PR #44, opened 2017-12-08
Strategy Scheduler.

MichaelRawson:
This PR implements a scheduler within Vampire to execute strategy sets with. The current `PortfolioMode` is then re-implemented as a degenerate case where the scheduler simply returns schedules in sequential order.

This should likely not be merged without scrutiny - there's a couple of architectural problems at the very least that should be fixed first - see FIXME in `PortfolioMode.hpp`. Plus bugs, probably. Seems to work alright on a few problems though.

### DIFF
--- a/Lib/Sys/Multiprocessing.cpp
+++ b/Lib/Sys/Multiprocessing.cpp
@@ -230,16 +230,17 @@ void Multiprocessing::killNoCheck(pid_t child, int signal)
   ::kill(child, signal);
 }
 
-void Multiprocessing::poll_child(pid_t child, bool &stopped, bool &exited, int &code)
+pid_t Multiprocessing::poll_children(bool &stopped, bool &exited, int &code)
 {
   CALL("Multiprocessing::poll_child");
 
   int status;
-  waitpid(child, &status, WNOHANG | WUNTRACED);
+  pid_t pid = waitpid(-1, &status, WUNTRACED);
   stopped = WIFSTOPPED(status);
   exited = WIFEXITED(status);
   if(exited)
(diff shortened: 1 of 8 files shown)


selig:
Let me look through the code.

Can this run with current schedules out of the box? If so, I suggest we run a StarExec job in debug mode across all of TPTP. It's a good initial test.

selig:
@quickbeam123 can you briefly scan through the changes made to Portfolio mode as you changed it recently?

MichaelRawson:
@selig : yes, seems to work fine so far. OK - will look into how to run a StarExec job for this.

Please do look at the changes to portfolio mode, I'm not sure what a lot of it does.

Lib/Sys/Multiprocessing.cpp
```diff
@@ -224,6 +224,24 @@ void Multiprocessing::kill(pid_t child, int signal)
...
+
+pid_t Multiprocessing::poll_children(bool &stopped, bool &exited, int &code)
+{
+  CALL("Multiprocessing::poll_child");
+
+  int status;
+  pid_t pid = waitpid(-1, &status, WUNTRACED);
+  stopped = WIFSTOPPED(status);
+  exited = WIFEXITED(status);
+  if(exited)
```
selig: Add {} please

MichaelRawson:
So a StarExec comparison has been done, and this branch proves a comparable number (actually slightly more) than baseline Vampire from recent `master`: 12,741 vs 12,727. Analysis of the job data also does not indicate any questionable proofs.

PR #48, opened 2018-04-26
Codatatypes

simonr89:
This pull request adds the features that correspond to the paper "Superposition with Datatypes and Codatatypes".

The new features are:

- Reasoning about codatatypes fixpoints (existence and uniqueness). Previously the treatment of codatatypes was known to be incomplete.
- Correct handling of the acyclicity property for mutually recursive datatypes.
- The injectivity and distinctness properties (for datatypes and codatatypes alike) can be axiomatized as before (-tar off) or the axioms can be replaced by an extension of the calculus (-tar full). A third option is to keep the axioms and only add some simplifying rules to the calculus (-tar simpl)
- The acyclicity property for datatypes can be axiomatized as before (-tac axiom) or a dedicated rule can be added to the calculus (-tac rule). It is also possible to completely ignore it (-tac off), at the expense for completeness.
- The uniqueness property for codatatype fixpoints is handled in the same way (-tau axiom/rule/off)


simonr89:
I'd like to finish things up before I leave Chalmers, so it would be good to do something with this request, either merge it or close it. It would also be helpful so that I know what to write in the manual.

MichaelRawson:
Hey @simonr89 - admittedly some time has passed! :-/ There are now considerable merge conflicts and I suspect Vampire has moved around under us. If you are still motivated to get this in, I am happy to work with you to make it happen, but I don't have the resources or knowledge to do it myself.

If not, let's admit defeat and close - it's still a monumental achievement! If we close it I'll keep the branch and add a note to `HACKING` about interesting branches for people to browse in future - we shouldn't lose the accumulated knowledge.

PR #57, opened 2018-07-11
FREQUENCY should be the default SymbolPrecedence; it's much better th…

quickbeam123:
…an ARITY

### DIFF
--- a/Shell/Options.cpp
+++ b/Shell/Options.cpp
@@ -1679,7 +1679,7 @@ void Options::Options::init()
     _termOrdering.description="The term ordering used by Vampire to orient equations and order literals";
     _termOrdering.tag(OptionTag::SATURATION);
     _lookup.insert(&_termOrdering);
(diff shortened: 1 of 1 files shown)


easychair:
1) This is an interesting statement, what evidence is it based on? I do not
remember making any experiments on this option.
2) Any change in the default value of options should be accompanied by
changing in the portfolio mode.

Generally, any changes in default values should be based on experiments,
not on statements.

Furthermore, if there are new values of some parameters, they should be
communicated to Spider and tested.

Andrei

quickbeam123:
My evidence were two 300s runs of vampire in early 2017 with

-av off -sa discount -awr 10 -lcm predicate

on the 17280 first-order problems of TPTP library, version 6.4.0.

One run was doing "-sp arity" and solved 9277 problems,
the other was doing "-sp frequency" and solved 9457 problems.

It would be nice if this could be confirmed by spider. It seems it does not know about "frequency" yet.

By the way, if we decide to change the defaults, another candidate would be "-sac on" which also tends to perform better on average than the default ("-sac off").

PR #79, opened 2019-07-30
revert z3 interface to 4.8.5

quicquid:
This reverts the z3 interface to 4.8.5 release.


easychair:
Vampire is broken, does not compile with libz3.a for 4.8.5.

A

PR #83, opened 2019-08-20
Skip-small-neg

quickbeam123:
Implementing Ahmed's proposal to skip the smaller sides of negative equations for rewriting by superposition.

The change (across TPTP, single 10s discount run) brings close to no improvement, but is a reasonable one, so should probably be merged when issues with CASC are resolved.


easychair:
> Implementing Ahmed's proposal to skip the smaller sides of negative equations for rewriting by superposition.
>
> The change (across TPTP, single 10s discount run) brings close to no improvement, but is a reasonable one, so should probably be merged when issues with CASC are resolved.

Yes, we should do it and yes, the net effect will probably be small or
none, even if there was no post-check, as we discussed on Skype.

I think I will start post-CASC runs with the new Vampire in the next few days.

A

PR #88, opened 2019-09-01
boolean bindings in newcnf need to by applied also to the already cre…

### LINKED ISSUE #62: Newcnf forgets (or fails to propage) Boolean bindings
The satisfiable input:

? [X0 : 'USort()'] : ! [X1 : $o] : ? [X2 : 'USort()'] : ((X1 | X0 = X2) & (~X1 | X0 != X2))

gets clausified under newcnf as the unsatisfiable:

cnf(u6,hypothesis,sK0 != sK1(X1)).
cnf(u5,hypothesis,sK0 = sK1(X1)).

It seems the problem comes from the last step in which:

----------------- OUTPUT -----------------
GC(1) {T} X0 = X2 | X1 --> $false | X2 --> sK1(X1) | X0 --> sK0
GC(1) {F} X0 = X2 | X1 --> $true | X2 --> sK1(X1) | X0 --> sK0
----------------- OUTPUT -----------------

are turned into clauses, and the bindings for X1 do not get applied to those for X2, which is becoming a skolem with a dependence on X1.

quickbeam123:
…ated biding. Fixes #62

### DIFF
--- a/Shell/NewCNF.cpp
+++ b/Shell/NewCNF.cpp
@@ -142,7 +142,7 @@ void NewCNF::process(Literal* literal, Occurrences &occurrences) {
   LOG2("process(Literal*)", literal->toString());
   LOG2("occurrences.size", occurrences.size());
 
-//  ASS_REP(!literal->shared(), literal->toString());
+  // shared literals don't contain fool stuff
   if (literal->shared()) {
     return;
   }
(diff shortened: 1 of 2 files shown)


Shell/NewCNF.cpp
```diff
@@ -427,6 +427,40 @@ void NewCNF::process(BinaryFormula* g, Occurrences &occurrences)
   }
 }
 
+void NewCNF::BindingStore::pushAndRememberWhileApplying(Binding b, BindingList* &lst)
+{
+  CALL("NewCNF::pushAndRememberWhileApplying");
+
+  // turn b into a singleton substitution
+  static Substitution subst;
```
selig: Shouldn't we reset subst here?

Shell/NewCNF.cpp
```diff
@@ -427,6 +427,40 @@ void NewCNF::process(BinaryFormula* g, Occurrences &occurrences)
...
+  subst.bind(b.first,b.second);
+
+  // to go through the bindings from the end, put them on a stack...
+  static Stack<BindingList*> st(5);
```
selig: Shouldn't we reset st here?

PR #93, opened 2019-09-17
Add max_age option and implementation.

MichaelRawson:
Hi folks,

I recently hacked Vampire to print all clauses generated by one inference step from a given set. The hack is not very useful in general. However, I noticed that there is a maximum-weight setting, but not a maximum-age setting for some reason - possibly there is one, but I'm not aware of any reason why it shouldn't work. Since it's a minimal change we might consider merging this to implement that option. What do you think?

Obviously this renders the prover incomplete in some cases, but presumably this is the same as the maximum-weight setting.

### DIFF
--- a/Saturation/SaturationAlgorithm.cpp
+++ b/Saturation/SaturationAlgorithm.cpp
@@ -161,8 +161,8 @@ SaturationAlgorithm::SaturationAlgorithm(Problem& prb, const Options& opt)
     _extensionality = 0;
   }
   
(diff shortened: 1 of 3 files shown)


easychair:
Max-weight is never used and is essentially useless by itelf, max-age would
be even more useless. If I remember correctly, max-weight is used by LRS -
LRS changes max-weight from time to time, this is why it is still there.

Andrei

PR #95, opened 2019-10-14
implement interactive clause selection

gleiss:
Adds an option manual_cs (with default off) to
perform clause selection interactively.
If manual_cs is set to on, then whenever a
clause has to be selected, the user is asked
to enter the id of a clause currently in the
set of passive clauses. The corresponding
clause then gets activated.
While this mechanism can be used on
its own, it is designed to be used in
conjunction with the Saturation Visualization.

### DIFF
--- a/Saturation/ManCSPassiveClauseContainer.cpp
+++ b/Saturation/ManCSPassiveClauseContainer.cpp
@@ -1,0 +1,84 @@
+
+#include <iostream>
+#include "ManCSPassiveClauseContainer.hpp"
+#include "Lib/VirtualIterator.hpp"
+
+namespace Saturation
+{
+using namespace Lib;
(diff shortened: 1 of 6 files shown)


Saturation/ManCSPassiveClauseContainer.cpp
```diff
@@ -0,0 +1,84 @@
...
+	removedEvent.fire(cl);
+	ASS(cl->store()!=Clause::PASSIVE);
+}
+
+Clause* ManCSPassiveClauseContainer::popSelected()
+{
+  	ASS(!clauses.empty());
+
+	// print existing clauses
+	std::cout << "Pick a clause from: ";
```
selig: Can you remove debugging output before we merge, or hide behind a debug flag

PR #100, opened 2019-11-29
Use subsets of induction term occurrences instead of all occurrences

hzzv:
When instantiating the induction schemas, try to use all possible subsets of occurrences of the induction term as a variable (and keep all other occurrences constant, as they are in the input literal). However, only do this if there are <=10 occurrences (somewhat arbitrary cutoff; if there are >10, there would be too many possibilities to try).

This change adds two options:
--induction_term_subset - turns on the behavior (default is off).
--max_induction_term_subset_size - maximum number of occurrences of the induction term to use as a variable (default is 3).

### DIFF
--- a/Inferences/Induction.cpp
+++ b/Inferences/Induction.cpp
@@ -87,6 +87,7 @@ Literal* LiteralSubsetReplacement::transformSubset() {
   // Increment _iteration, since it either is 0, or was already used.
   _iteration++;
   static unsigned maxSubsetSize = env.options->maxInductionTermSubsetSize();
+  // Note: __builtin_popcount() is a GCC built-in function.
   int setBits = __builtin_popcount(_iteration);
   // Skip this iteration if not all bits are set, but more than maxSubset are set.
   while ((_iteration <= _maxIterations) &&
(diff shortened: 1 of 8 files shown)


Inferences/Induction.cpp
```diff
@@ -674,4 +721,13 @@ bool InductionClauseIterator::notDone(Literal* lit, Term* term)
   return true;
 }
 
+Term* InductionClauseIterator::getPlaceholderForTerm(Term* t) {
+  CALL("InductionClauseIterator::getPlaceholderForTerm");
+
+  unsigned termType = env.signature->getFunction(t->functor())->fnType()->result();
+  unsigned placeholderConstNumber = env.signature->addFreshFunction(0, "placeholder");
```
selig: I'm a little concerned about creating a large number of fresh placeholder constants. I think it's not possible to have a single placeholder but is there a way to reuse. My understanding is that these objects are just used locally within the induction process? Could we have a finite number and keep track of how many we have used so far in this particular induction call?

Inferences/Induction.cpp
```diff
@@ -63,11 +63,46 @@ TermList TermReplacement::transformSubterm(TermList trm)
...
+  }
+  return trm;
+}
+
+Literal* LiteralSubsetReplacement::transformSubset() {
+  CALL("LiteralSubsetReplacement::transformSubset");
+  // Increment _iteration, since it either is 0, or was already used.
+  _iteration++;
+  static unsigned maxSubsetSize = env.options->maxInductionTermSubsetSize();
+  int setBits = __builtin_popcount(_iteration);
```
selig: Add a comment about compiler support - I don't think it's currently an issue but will help somebody debug in the future if it becomes an issue.

Inferences/Induction.cpp
hzzv: You're right. The placeholder constant is actually never used in a formula in the saturation set, so we can create just one per each type. I changed this in the code.

Inferences/Induction.cpp
hzzv: Good point, I added a comment.

hzzv:
In addition to reacting to your comments I also renamed the comman line options, so they correspond to the name "generalized induction".

PR #102, opened 2020-01-07
Fix implementation related to (internally added) theory axioms

gleiss:


Kernel/Unit.cpp
```diff
@@ -127,8 +127,9 @@ unsigned Unit::getPriority() const
   if(!env.clausePriorities) return 1;
 
   // Put these at the end of the priority list
+  // TODO: it is unclear why the current code checks here for EQUALITY_PROXY_AXIOM1 but not for EQUALITY_PROXY_AXIOM2.
```
selig: Because this code was just focussing on things without 'parents' ... the code is unused anyway as env.clausePriorities is only set in an experimental (unused) part of LTB.

PR #108, opened 2020-03-31
Updates for Question Answering

selig:
This pull request has been created to view the changes to question answering in this old branch. It might get merged if those changes look okay.

### DIFF
--- a/Debug/RuntimeStatistics.hpp
+++ b/Debug/RuntimeStatistics.hpp
@@ -24,7 +24,7 @@ collected nor output.
 #if VDEBUG
 #define RUNTIME_STATS 1
 #else
-#define RUNTIME_STATS 1
+#define RUNTIME_STATS 0
 #endif
 #endif
 
--- a/Debug/RuntimeStatistics.hpp
+++ b/Debug/RuntimeStatistics.hpp
@@ -24,7 +24,7 @@ collected nor output.
 #if VDEBUG
 #define RUNTIME_STATS 1
 #else
-#define RUNTIME_STATS 0
+#define RUNTIME_STATS 1
 #endif
 #endif
 
--- a/Saturation/SaturationAlgorithm.cpp
+++ b/Saturation/SaturationAlgorithm.cpp
@@ -922,7 +922,8 @@ void SaturationAlgorithm::addUnprocessedClause(Clause* cl)
     return;
   }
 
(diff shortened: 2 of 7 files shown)


quickbeam123:
Is this fixing Uwa at the same time?

selig:
No, the part fix got included in my rebase by mistake. I’ll finish fixing that separately. Giles

Sent from my iPhone

On 31 Mar 2020, at 18:43, Martin Suda <notifications@github.com> wrote:

﻿

Is this fixing Uwa at the same time?

logicReasoner:
What is the current status of this functionality? 
Additionally, can Vampire be made to return all (more than one found answer) when using the `-qa` option? Currently it seems to return just one (the first found?) answer.
Thanks in advance!

selig:
Great question. I should finish it. This branch incorporates some experimental things related to question answering (explored in this paper https://easychair.org/publications/paper/5L6q) and my plan was to tidy them up. This will be easier with a 'customer' prodding me...

This branch introduces a new  option "question_count" (bad name, I'll change that) that limits the number of answers. By default it will continue returning answers whilst it can still find them (different from master where it terminates on the first answer). But note that the way that Vampire works it will currently return a lot of duplicates - the plan is to filter these out. This branch also fails to notice when there are an identifiably finite set of answers e.g. it will always say that there may be more. Again, something that could be fixed in some cases but is not yet.

Do you have any question answering related feature requests?

logicReasoner:
@selig Your plan sounds great! 
As a final addition maybe something like the following option implemented by the E prover could be useful:
```
--conjectures-are-questions
    Treat all conjectures as questions to be answered. This is a wart
    necessary because CASC-J6 has categories requiring answers, but does not
    yet support the 'question' type for formulas.
```

selig:
@logicReasoner This is actually already an undocumented feature... kind of. Vampire doesn't distinguish between the conjecture and question role. In question answering mode any existential  conjecture or question will be treated as a question. I can't remember if we have support for multiple questions (we don't support multiple conjectures).

MichaelRawson:
This is quite a long-standing merge! With respect to @hzzv's changes in #442, does this still make sense? If so, how much effort is it to get this merged, @selig?

MichaelRawson:
Closing for lack of activity, feel free to re-open if not in fact dead.

PR #114, opened 2020-04-16
Subsumption demodulation

JakobR:
This PR adds forward and backward subsumption demodulation.

By default, both rules are disabled (the options are `-fsd on` and `-bsd on`).

I've tried to not touch existing Vampire code too much. One exception is that I've refactored the class `MLMatcher` (without changing its existing functionality -- the matching for subsumption demodulation is in `MLMatcherSD`).


Inferences/SubsumptionDemodulationHelper.cpp
```diff
@@ -0,0 +1,278 @@
...
+  // Copy given literals so we can sort them
+  vvector<Literal*> c1(lits1, lits1+n1);
+  vvector<Literal*> c2(lits2, lits2+n2);
+
+  // These will contain literals from c1/c2 with equal occurrences removed
+  vvector<Literal*> v1;
+  vvector<Literal*> v2;
+
+  // The equality tests below only make sense for shared literals
+  std::for_each(c1.begin(), c1.end(), [](Literal* lit) { ASS(lit->shared()); });
```
selig, line 192: Will this for_each run even in release mode? Perhaps it should be inside a VDEBUG

PR #127, opened 2020-07-08
User weight function

joe-hauns:
Adds an options to let the user supply the weight function for KBO, or to make vampire to randomly assign weights. 
```
$ vampire -explain function_weights
--function_weights (-fw)
        [experimental]
        Path to a file that defines weights for KBO for function symbols, or 'random'.

        If 'random' is used the weights will be assigned randomly.

        If the option is a file path, each line in the file is expected to contain
        a function name, followed by the functions arity, and a positive integer,
        that specifies symbols weight.

        Additionally there are special values that can be specified:
        - `$default    <number>` specifies the default symbol weight, that is used
        for all symbols not present in the file (if not specified 0 is used)
        - `$introduced <number>` specifies the weight used for symbols introduced
        during preprocessing or proof search
        - `$var        <number>` specifies the weight used for variables
        - `$int        <number>` specifies the weight used for integer constants
        - `$rat        <number>` specifies the weight used for rational constants
        - `$real       <number>` specifies the weight used for real constants


        ===== example ============
        $add 2 2
        $mul 2 7
        f    1 2
        $default 2
        $var     2
        ===== end of example =====

        If this option is empty all weights default to 1.

        default: <empty>
```
Further an option to display the weights, as well as the symbol precedence is added:
```
$ vampire -explain show_ordering
--show_ordering
        Display the used simplification ordering's parameters.
        default: off
``` 
Example output:
```
% Function precedences, smallest symbols first (line format: `<name> <arity>`)
% ===== begin of function precedences =====
% a 0
% -1 0
% $sum 2
% 1 0
% 2 0
% 0 0
% $uminus 1
% ===== end of function precedences =====
%
% Predicate precedences, smallest symbols first (line format `<name> <arity>`)
% ===== begin of predicate precedences =====
% = 2
% p 1
% $less 2
% ===== end of predicate precedences =====
%
% Predicate levels (line format: `<name> <arity> <level>`)
% ===== begin of predicate levels =====
% = 2 0
% p 1 1
% $less 2 1
% ===== end of predicate precedences =====
%
% Weights of function (line format: `<name> <arity> <weight>`)
% ===== begin of function weights =====
% $uminus 1 16910
% 2 0 54519
% 1 0 54519
% 0 0 54519
% -1 0 54519
% $sum 2 56290
% a 0 60855
% $introduced 59958
% $var 38110
% $real 54622
% $rat 49157
% $int 54519
% ===== end of function weights =====
....
```
In addition to setting the weights of symbols the admissibility of the weight function wrt. the precedence is checked and a user-error is thrown in case of a violation. This behaviour can be changed by setting the option ` --kbo_admissibility_check warning`. 
Example output:
```
WARNING: Your KBO is probably not well-founded. Reason: weight of constants (i.e. a) must be greater or equal to the variable weight (1)
...
```
Further unit tests for KBO are added.


JakobR:
I was actually going to open a similar PR soon :)

Differences to this branch (as far as I can tell):
- Options to give weights on command line (I've been meaning to add a similar option for precedence but haven't done it yet)
- Option `show_ordering` to print ordering parameters
- Stored weight in `Signature::Symbol`
- Use custom weights also for term/literal weight computation in `TermSharing`
- No checks for inadmissible weight functions yet
- No unit tests

I've pushed it here in case you want to take a look: https://github.com/vprover/vampire/tree/customizable-ordering

joe-hauns:
Sorry for the long delay. I was on vacation for the last two weeks. 

Concerning the differences:

> Option `show_ordering` to print ordering parameters
what was ur use case for the `show_ordering`? To find out how exactly vampire assigns the precedence (automatically), or for debugging purposes?

> Stored weight in `Signature::Symbol`
I didn't really think of this option, but thinking about it now I'd say KBO is the better place, since 
1. we don't use any unnecessary memory if LPO is used instead of KBO
2. all weights are within a continous block of memory which makes computing KBO more cache-efficient

> Use custom weights also for term/literal weight computation in TermSharing
I wasn't aware that there is some weight computation going on there as well. So that might be source of bugs. What are these weights used for?

> No unit tests
Do you mabe have any special cases in mind we should add as test cases?

quickbeam123:
Vampire uses `weights` stored with terms and thus literals at various places in indexing and also for computing clause weight which is then used for clause selection. These are not the same weights as the symbol weights used by KBO. Experimenting with both could be interesting, but we should definitely not mix them. (Not sure whether agreeing on a terminology which does not confuse the two and reflecting it in the identifiers in the code could be useful...)

easychair:
> Vampire uses weights stored with terms and thus literals at various places in indexing and also for computing clause weight which is then used for clause selection. These are not the same weights as the symbol weights used by KBO. Experimenting with both could be interesting, but we should definitely not mix them. (Not sure whether agreeing on a terminology which does not confuse the two and reflecting it in the identifiers in the code could be useful...)

You probably mean that they are not the same as TERM/LITERAL weights
computed by KBO. If we start using any KBO with weights different from
the default, we should definitely not mix them.

> Use custom weights also for term/literal weight computation in TermSharing
> I wasn't aware that there is some weight computation going on there as well. So that might be source of bugs. What are these weights used for?

I understand that

(a) term weights are constant
(b) terms weight are computed from subterm weight

This applies to both kinds of weights.

This means that TermSharing is The Right Place for computing these weights.

A

joe-hauns:
( updated the `vampire -explain ...` texts in the first message of the pull request to match the latest commit )

JakobR:
> what was ur use case for the `show_ordering`? To find out how exactly vampire assigns the precedence (automatically), or for debugging purposes?

The idea was to be able to see the exact ordering that Vampire uses. Ideally in a format that can be used as input for the ordering options, so we can use the same ordering in subsequent Vampire runs or as a starting point for custom weights/precedences.


> I didn't really think of this option, but thinking about it now I'd say KBO is the better place, since
>     1. we don't use any unnecessary memory if LPO is used instead of KBO
>     2. all weights are within a continous block of memory which makes computing KBO more cache-efficient

Ok, makes sense. Especially if we want to keep them separate from the weights for clause selection.


> 
> > No unit tests
> 
> Do you mabe have any special cases in mind we should add as test cases?

Nothing specific (yet), I just thought it should be mentioned :)

Kernel/KBO.cpp
```diff
@@ -291,19 +295,267 @@ void KBO::State::traverse(Term* t1, Term* t2)
...
+
+  if ( _funcWeights._specialWeights._numReal < varWght
+    || _funcWeights._specialWeights._numInt  < varWght
+    || _funcWeights._specialWeights._numRat  < varWght
+      ) {
+    throw UserErrorException("weight of (number) constant symbols must be >= variable weight (", varWght, ").");
+  }
+
+  if (varWght <= 0) {
+    throw new UserErrorException("variable weight must be greater than zero");
```
JakobR, line 542: Should be `throw` instead of `throw new`

Kernel/Ordering.cpp
```diff
@@ -637,26 +667,64 @@ static void loadPermutationFromString(DArray<unsigned>& p, const vstring& str) {
   }
 }
 
+bool isPremutation(const DArray<int>& xs) {
```
JakobR, line 670: typo :)

joe-hauns:
As Jakob noticed, in the current state of vampire choosing different weights for predicates does not make any sense, since we're comparing predicates by precedence and levels only. 

My suggestion would be to remove the command-line option but leave the implementation there  (disabled by a macro), in case it will become useful at some point.

What do u say?

quickbeam123:
> As Jakob noticed, in the current state of vampire choosing different weights for predicates does not make any sense, since we're comparing predicates by precedence and levels only.
> 
> My suggestion would be to remove the command-line option but leave the implementation there (disabled by a macro), in case it will become useful at some point.
> 
> What do u say?

Sounds like a sensible idea, I support it!

PR #141, opened 2020-09-07
Fix math induction (use correct substitution; use BR only when needed)

hzzv:
This commit changes math induction to use produceClauses(...) (similarly to struct induction).
To this end, I changed the parameters of produceClauses(...) to include the substitution, which is used in binary resolution of the CNF of the induction hypothesis with the premise. Note that the substitution is the same for all the applications of BR within one application of math induction (y -> induction term); and it is always identity for struct induction (for struct induction, the substituted variable only has one occurrence in the induction hypothesis, and the literal containing it is removed in BR).

A sample SMT2 input to check that the math induction works (it proves that 2*(1 + ... + n) = n*(n+1)):
;; Definition: triangle(n) = 1 + ... + n
(declare-fun triangle (Int) Int)
(assert (forall ((x Int)) (= (triangle 0) 0)))
(assert (forall ((x Int)) (=> (< 0 (+ x 1)) (= (triangle (+ x 1)) (+ (triangle x) (+ x 1))))))
;; Conjecture: 2*triangle(n) = n*(n+1)
(assert (not (forall ((x Int)) (=> (not (< x 0)) (= (* 2 (triangle x)) (* x (+ x 1)))))))

Run it using the following command (it should find the proof within 30-45 seconds):
vampire -ind math --input_syntax smtlib2 triangle_formula.smt2

### DIFF
--- a/Inferences/Induction.cpp
+++ b/Inferences/Induction.cpp
@@ -223,7 +223,6 @@ void InductionClauseIterator::process(Clause* premise, Literal* lit)
       Set<Term*>::Iterator citer2(ta_terms);
       while(citer2.hasNext()){
         Term* t = citer2.next();
-        //cout << "PERFORM INDUCTION on " << t->toString() << endl;
         static bool one = env.options->structInduction() == Options::StructuralInductionKind::ONE ||
                           env.options->structInduction() == Options::StructuralInductionKind::ALL; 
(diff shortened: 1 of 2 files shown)


selig:
Excellent, this may explain why it wasn't working as I'd hoped originally!

Btw, adding '-sos theory -sstl 1 ' turns the proof into a 1 second thing rather than waiting 30 seconds :)

PR #144, opened 2020-09-30
fixes theory instantiation tautology deletion

joe-hauns:
Fixes a subtle bug that potentially has big impacts on the performance of theory instantiation.

When the generating inference rule `TheoryInstAndGen` is called in `SaturationAlgorithm::activate`, it shall set the flag `redundant` in order to signal whether the premise clause can be deleted from the search space. This does never happen right now. 
The issue is that the flag redundant is set in the mapping function `InstanceFn::operator()` in the iterator returned by `TheoryInstAndSimpl::generateClauses`. But this function will only be called when `next`/`hasNext` is called on the returned iterator. This means the flag will only be set after line 1137 in `SaturationAlgorithm.cpp`, but is read in line 1116. Therefore the deletion of the redundant clause will never be performed. 

I tested this hypothesis by adding an assertion `ASS(!redundant)`  before line 1116, and ran vampire over 56000 smtlib arithmetic examples (`-t 0 -sched casc -thi full -uwa one_side_interpreted`, debug build), and the assertion is never triggerd. After applying the fix that is in this pull request the assertion is triggered. (Of course this assertion was removed before making the pull request).

### DIFF
--- a/Inferences/TheoryInstAndSimp.cpp
+++ b/Inferences/TheoryInstAndSimp.cpp
@@ -915,6 +915,7 @@ ClauseIterator TheoryInstAndSimp::generateClauses(Clause* premise,bool& premiseR
     // measure time of the overall processing
     auto it4 = getTimeCountedIterator(it3,TC_THEORY_INST_SIMP);
 
+    // persistent iterator is needed for the flag `premiseRedundant` to be set before returning from the function
     return getPersistentIterator(it4);
   }
 }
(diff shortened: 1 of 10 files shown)


joe-hauns:
As we discovered in regression test tautology deletions do not strictly help us in proof search. The problem is that some tautologies might be helpful in proof search. 

One example for a failed proof when tautology deletion is performed is the following.
```
$vampire -sas z3 -thi all  UFNIA/2019-Zohar-ic/combined/int_check_ne_bvurem0_rtl.smt2
```

The problem is that this clause is deleted:
```
0 = $remainder_e(X9,1) | $less(X9,0) <- (25, 28)
```
But in the proof in master we need the following descendant of the clause:
```
 0 = $remainder_e(x0,1) <- (25, 28, ~88) 
```
Therefore the proof fails. 

The essence of the problem is that we would delete something that can be solved by z3, but we don't (yet) have rules or axioms to really make this clause redundant.

Therefore as discussed with @quickbeam123 I added the option `--theory_instantiation_tautology_deletion`/`-thitd` to enable/disable the tautolgy deletion, in order to preserve the old behaviour by default.

quickbeam123 (approved):
All looking good!

PR #149, opened 2020-10-26
Silence warnings for default builds.

MichaelRawson:
Currently there are some warnings in both release and debug builds, which can make it hard to see the bugs you just created. This PR silences warnings from release and debug builds with the default configuration. In principle there should be no change in runtime behaviour that wasn't already a bug.

The major change is changing the definition of `ASSERTION_VIOLATION` in release mode from no-op to either:
1. `__builtin_unreachable` on compilers that support it (GCC, Clang, ICC). This should allow for further optimisation.
2. `while(true) {}` otherwise. This is undefined behaviour and conforming compilers can in principle optimise this into whatever is fastest based on context. Open to suggestions for better ideas.

@joe-hauns - there's some history of you changing `Assertion.hpp` but it doesn't seem to work out?

### DIFF
--- a/Debug/Assertion.hpp
+++ b/Debug/Assertion.hpp
@@ -225,6 +225,15 @@ private:
 
 #else // ! VDEBUG
 
+/* this point in the code is statically unreachable */
+#ifdef __GNU_C__
+// builtin if the GNU C dialect is availble: GCC, Clang, ICC
+#define __UNREACHABLE { __builtin_unreachable(); }
+#else // !__GNU_C__
+// otherwise, infinite loop - UB post-C++11 and should be optimised out
+#define __UNREACHABLE while(true) {}
+#endif
+
 #define DEBUG_CODE(X)
 
 #define __IGNORE_WUNUSED(...) __PUSH_DIAGNOSTICS("GCC diagnostic ignored \"-Wreturn-type\"", __VA_ARGS__)
@@ -248,7 +257,7 @@ private:
 #define ASS_ALLOC_TYPE(PTR,TYPE)
 #define ASS_METHOD(OBJ,METHOD)
 
-#define ASSERTION_VIOLATION 
+#define ASSERTION_VIOLATION __UNREACHABLE
 #define ASSERTION_VIOLATION_REP(Val) ASSERTION_VIOLATION
 #define ASSERTION_VIOLATION_REP2(Val1,Val2)  ASSERTION_VIOLATION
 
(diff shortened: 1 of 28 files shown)


quickbeam123:
What were the warnings in DEBUG you experienced? We always strived not to have any and I thought we were doing pretty well on this!

MichaelRawson:
@quickbeam123 - not many! Should probably be clearer, it's mostly the release build. There's some in `lglib.c` pertaining to useless asserts and checking an assignment - fixed in the PR.

joe-hauns:
In debug builds eliminating the warnings did work out. At least for `clang-1100.0.33.17` and some rather recent versions of gcc. For release builds the only warnings that were left (for me at least) are the ones that originate from ASSERTION_VIOLATION, and my current work-around was to have a local version of `Assertion.hpp` where 
`#define ASSERTION_VIOLATION`
is replaced by 
`#define ASSERTION_VIOLATION throw;` 
in order to get rid of the warnings in release builds. 
So basically the reason why I it didn't work out is that I didn't know about `__builtin_unreachable` and the UB trick with `while(true){}`. So I think this sounds like a great solution to me!

selig:
These changes have the potential to change the behaviour of Vampire in release mode if we should have an assertion violation in debug mode e.g. where an unreachable (in debug mode) return after an assertion violation is removed.

It's a grey area because the program is buggy so we don't really know what's going to ho happen but  perhaps we should have some release-mode behaviour for an assertion violation?

MichaelRawson:
@selig - yep. My thinking is that if you reach `ASSERTION_VIOLATION` in release mode, UB is not too bad, particularly if you can get a performance win from it. If we want it to do something more reasonable, perhaps we can `__builtin_trap` (definitely crash but similar performance) or throw an exception. I'd argue doing nothing is the worst of both worlds though, execution continues buggily but doesn't benefit from the compiler knowing that the scenario shouldn't be possible.

joe-hauns:
I think we technically have undefined behaviour in some cases already, namely in fuctions like this 
```
int foo() {
  if (condition()) {
     return bar();
  } else {
    ASSERTION_VIOLATION
  }
}
```
where we just don't return anything. And as you say, when we have defined behaviour, we probably don't have the behaviour we expect, since the person putting the `ASSERTION_VIOLATION` somewhere doesn't expect the exectution to continue after it. So I think we basically have some kind of UB for `ASSERTION_VIOLATION`s atm anyways.

But maybe `__builtin_trap` is a good alternative as well since we don't change behaviour between release and debug at all.

quickbeam123:
Please try to build with Z3 as well.

MichaelRawson:
@quickbeam123 - fixed as discussed, clean compile with Z3 enabled as well. :-)

PR #152, opened 2020-10-27
Polymorphic vampire

ibnyusuf:
Rank-1 polymorphism and HOL. Testing still needs to be carried out before the merge can take place. Please use comments to report any bugs or outstanding tasks.


ibnyusuf:
**Outstanding task:**

Review all places in the code where we iterate over the signature and add a check to ensure that we are not iterating over type constructors.

MichaelRawson:
Another from LIA.

```
Condition sort == Term::boolSort() in file Parse/SMTLIB2.cpp, line 1113 was violated, as:
sort == 6777 Aborted by signal SIGSEGV on 20190429-UltimateAutomizerSvcomp2019/jain_7_true-unreach-call_true-no-overflow_false-termination.i_11.smt2
% Version: Vampire 4.5.1 (commit 146698fd on 2020-10-29 08:48:07 +0000)
% Termination reason: Unknown
% Termination phase: Parsing

% Memory used [KB]: 383
% Time elapsed: 0.0000 s
----  Runtime statistics ----
-----------------------------
Version : Vampire 4.5.1 (commit 146698fd on 2020-10-29 08:48:07 +0000)
Control points passed: 39616
last control point:
              -Environment::out (39616)
main
 vampireMode()
  doProving()
   getPreprocessedProblem
    UIHelper::getInputProblem
     SMTLIB2::parse(istream&)
      SMTLIB2::parse(LExpr*)
       SMTLIB2::readBenchmark
        SMTLIB2::readAssert
         SMTLIB2::parseTermOrFormula
          SMTLIB2::parseAsBuiltinTermSymbol
           SMTLIB2::ParseResult::asTerm
            Term::toString
             System::handleSignal
```

ibnyusuf:
Okay I just started to run the branch on 10k random problems of smt-lib, and got the following violations:
Condition in file /home/joe.sch/vampire/Kernel/Signature.cpp, line 165 violated (3062 times)
Condition sort == Term::boolSort() in file /home/joe.sch/vampire/Parse/SMTLIB2.cpp, line 1077 was violated, as: (3 times)
Condition sort == Term::boolSort() in file /home/joe.sch/vampire/Parse/SMTLIB2.cpp, line 1113 was violated, as: (868 times)
examples:
1)
./vampire-polymorphic_vampire-Debug --mode portfolio -sched smtcomp --time_limit 20 --input_syntax smtlib2 UF::20170428-Barrett::cdt-cade2015::nada::gandl::stern_brocot_tree::x2015_09_10_16_52_13_375_1100659.smt_in.smt2
./vampire-polymorphic_vampire-Debug --mode portfolio -sched smtcomp --time_limit 20 --input_syntax smtlib2 UFLIA::grasshopper::uninstantiated::union_check_heap_access_38_21.smt2
./vampire-polymorphic_vampire-Debug --mode portfolio -sched smtcomp --time_limit 20 --input_syntax smtlib2 AUFLIA::20170829-Rodin::smt3696567545984296619.smt2
2)
./vampire-polymorphic_vampire-Debug --mode portfolio -sched smtcomp --time_limit 20 --input_syntax smtlib2 UFDT::20170428-Barrett::rec-fun-ijcar2016::wo::isa_nun_bench_v4_noenc::nunchaku2154182.nun.smt2
./vampire-polymorphic_vampire-Debug --mode portfolio -sched smtcomp --time_limit 20 --input_syntax smtlib2 LIA::psyco::134.smt2
./vampire-polymorphic_vampire-Debug --mode portfolio -sched smtcomp --time_limit 20 --input_syntax smtlib2 LIA::psyco::066.smt2
3)
./vampire-polymorphic_vampire-Debug --mode portfolio -sched smtcomp --time_limit 20 --input_syntax smtlib2 UFLIA::boogie::Chunker11-AdditiveExpose_Chunker..cctor.smt2
./vampire-polymorphic_vampire-Debug --mode portfolio -sched smtcomp --time_limit 20 --input_syntax smtlib2 NRA::keymaera::intersection-example-onelane.proof-node9774.smt2
./vampire-polymorphic_vampire-Debug --mode portfolio -sched smtcomp --time_limit 20 --input_syntax smtlib2 UFLIA::boogie::PureReceiverMightBeCommitted_T..ctor.smt2

ibnyusuf:
% lrs+11_5_av=off:cond=on:lma=on:nm=64:newcnf=on:nwc=1:stl=30:updr=off_22 on x2015_09_10_16_52_13_375_1100659.smt_in
% fmb+10_1_av=off:fmbes=contour:fmbsr=1.5:ile=on:updr=off_28 on x2015_09_10_16_52_13_375_1100659.smt_in
Condition in file ./Kernel/Term.hpp, line 421 violated:
_args[0]._info.shared
----- stack dump -----
Version : Vampire 4.5.1 (commit 11c52c69 on 2020-10-29 11:03:21 +0000)
Control points passed: 6192070
last control point:
                -Stack::push (6192070)
main
 PortfolioMode::perform
  PortfolioMode::searchForProof
   PortfolioMode::performStrategy
    PortfolioMode::runSchedule
     ScheduleExecutor::run
      ScheduleExecutor::spawn
       PortfolioSliceExecutor::runSlice
        PortfolioMode::runSlice
         PortfolioMode::runSlice(Option&)
          ProvingHelper::runVampire
           ProvingHelper::runVampireSaturationImpl
            MainLoop::run
             FiniteModelBuilder::init
              DefinitionIntroduction::hasNext
               DefinitionIntroduction::process
----- end of stack dump -----

ibnyusuf:
% WARNING: value z3 for option sas not known
% dis+1011_8_afp=10000:afq=1.2:amm=sco:anc=none:bce=on:gs=on:gsem=off:ile=on:lma=on:nm=16:newcnf=on:nwc=2.5:sas=z3:sos=all:sac=on:sp=occurrence:updr=off_2 on 066
% WARNING: value z3 for option sas not known
% dis+11_3_afr=on:afp=4000:afq=2.0:amm=sco:anc=none:bce=on:cond=on:fsr=off:ile=on:lwlo=on:nm=64:newcnf=on:nwc=1:sas=z3:tha=off:thf=on_2 on 066
% SZS status Satisfiable for 066
% # SZS output start Saturation.
tff(u620,hypothesis,
    'W_S1_V1').

tff(u619,hypothesis,
    ((~(![X16 : $int, X20 : $int, X15 : $int, X32 : $int, X3 : $int, X36 : $int, X7 : $o, X27 : $int, X31 : $int, X19 : $int, X23 : $int, X10 : $int, X2 : $int, X35 : $int, X6 : $int, X26 : $o, X30 : $int, X18 : $o, X22 : $int, X9 : $int, X13 : $int, X34 : $int, X5 : $int, X25 : $o, X29 : $int, X17 : $int, X12 : $int, X0 : $int, X4 : $o, X24 : $int, X28 : $int] : ((~$less(X28,X27) | ~sP194(X34,X28,X13,X28) | ~sP193(X36,X35,X31,X30,X27,X26,X24,X23,X22,$true,X19,X18,X16,X13,X28,X10,X9,X7,X6,X4,X2,X0) | ~sP192(X35,X31,X28,X28) | ~sP191(X28,X23,X16,X28) | ~sP190(X36,X28,X27,X26,X19,X18,X16,X10,X9,X7,X0) |

ibnyusuf:
**Resolved**
**Condition env.signature->functionArity(symbol) == arity in file /home/joe.sch/vampire/Shell/FOOLElimination.cpp, line 883 was violated, as:**

ibnyusuf:
**Resolved**
```
group 3: 80 occurences
		Condition in file /home/joe.sch/vampire/Kernel/Clause.cpp, line 509 violated:
```

ibnyusuf:
Assertion violation in FMB

`./vampire_dbg_polymorphic_vampire_4978 --decode fmb+10_1_av=off:fde=unused:ile=on:irw=on:lcm=predicate:lma=on:nm=16:nwc=1.7:sos=all:sp=reverse_arity_2 --input_syntax smtlib2 ../Downloads/SMTLIB_problems/smt9008971064966552750.smt2 --show_preprocessing off > output.vamp `

ibnyusuf:
**Outstanding Task**

Do not let NewCNF run on polymorphic problems.

`../vampire/vampire_dbg_polymorphic_vampire_4982 --decode dis+10_16_awrs=decay:awrsf=256:afr=on:afp=40000:afq=2.0:amm=off:bsr=on:cond=on:er=known:gsp=input_only:irw=on:lma=on:nm=6:newcnf=on:nwc=3:sd=4:ss=axioms:s2a=on:sp=frequency:updr=off_8 Problems/SWW/SWW475_5.p > output.vamp`

ibnyusuf:
**Outstanding Task**

Make sure that Vampire avoid all parts of the code not updated to handle polymorphism when faced with a polymorphic problem. For example induction, theories etc.

ibnyusuf:
**Points for reviews to note:**

1. All code exclusively relating to higher-order logic can be ignored when reviewing. It requires checking that such code is truly not reachable for first-order problems. The following is a (non-exhaustive) list of files that should never been involved in first-order reasoning:

```
         Inferences/ArgCong.o\
         Inferences/NegativeExt.o\
         Inferences/Narrow.o\
         Inferences/SubVarSup.o\
         Inferences/Injectivity.o\
         Inferences/CombinatorDemodISE.o\
         Inferences/CombinatorNormalisationISE.o\
         Inferences/CNFOnTheFly.o\
         Inferences/CasesSimp.o\
         Inferences/BoolSimp.o\
         Inferences/PrimitiveInstantiation.o\
         Inferences/Choice.o\
         Inferences/ElimLeibniz.o\
         Inferences/BoolEqToDiseq.o\
         Shell/LambdaElimination.o\
         Kernel/ApplicativeHelper.o\
        Kernel/SKIKBO.o\
```

quickbeam123:
There seems to be the following issue with many(most/all) of the Kanig benchmarks:

```
./vampire_z3_dbg_static_polymorphic_vampire_5033 AUFDTLIRA/20200306-Kanig/spark2014bench/evoting__evoting.ads_107_59_index_check___00.smt2 --input_syntax smtlib2
Condition in file /nfs/sudamar2/projects/vampire/Shell/Preprocess.cpp, line 233 violated:
!prb.hasPolymorphicSym()
----- stack dump -----
Version : Vampire 4.5.1 (commit 75b62285 on 2020-11-06 16:06:41 +0100)
Control points passed: 990187
last control point:
     -Problem::hasPolymorphicSym (990187)
main
 vampireMode()
  doProving()
   getPreprocessedProblem
    Preprocess::preprocess
----- end of stack dump -----
```

FMB/FiniteModelBuilder.cpp
```diff
@@ -789,11 +794,13 @@ void FiniteModelBuilder::init()
   _distinctSortConstantCount.ensure(_sortedSignature->distinctSorts);
   _fminbound.ensure(env.signature->functions());
   for(unsigned f=0;f<env.signature->functions();f++){
+    if(env.signature->isTypeConOrSup(f)){ continue; }
     if(del_f[f]) continue;
 
     if(env.signature->functionArity(f)==0){ 
-      unsigned vsrt = env.signature->getFunction(f)->fnType()->result();
-      if(vsrt != Sorts::SRT_BOOL){
```
quickbeam123, line 796: I see `vsrt != Sorts::SRT_BOOL` on the left vs `SortHelper::isBoolSort(vsrtT)` on the right. Could this be inadvertently flipping the logic of the condition?

ibnyusuf: It would indeed appear to be an inverting of the logic. I have now fixed.

ibnyusuf:
A review is required of 

`TPTPPrinter::ensureNecesarySorts()`

At the moment it is commented out on the branch. Is it important? If so, it can be resurrected.

ibnyusuf:
```
void TPTPPrinter::ensureHeadersPrinted(Unit* u)
{
  CALL("TPTPPrinter::ensureHeadersPrinted");

  if(_headersPrinted) {
    return;
  }
  
  //ensureNecesarySorts();
```

Can you review how we output sorts in poly-vamp. The difference is that we now output _all_ sorts even if they are not used in a problem.

ibnyusuf:
Also, in `Shell::Preprocess`, I have added the following code:

```
  if(env.options->choiceReasoning()){
    env.signature->addChoiceOperator(env.signature->getChoice());
  }
```
I don't think this is the right place to extend the signature, so feel free to move it to somewhere more reasonable.

Indexing/CodeTree.hpp
```diff
@@ -135,7 +135,7 @@ class CodeTree
     ILStruct* previous;
 
     unsigned isVarEqLit:1;
-    unsigned varEqLitSort:31;
```
quickbeam123, line 138: This changes the memory footprint from 32 bit (for compilers able to compile bitfields) to 1+64, which is probably at least 80 bits (on 64 bits platforms) due to alignment. 

In the future, for occasions where nothing but sorts' (not)identity is relevant for the use, we could instead term's id:

```
  /** For shared terms, this is a unique id used for deterministic comparison */
  unsigned _id;
```

Especially, after sorts get their own TermSharing instance, as planned, these will be small unsigneds uniquely numbering the sorts! (I.e. no need for `SortHelper::sortNum` as used, e.g., in FMB.)

ibnyusuf: Sounds like the way to go!

Indexing/TermSharing.cpp
```diff
@@ -259,6 +327,8 @@ Literal* TermSharing::insertVariableEquality(Literal* t,unsigned sort)
   if (s == t) {
     t->markShared();
     t->setId(_totalLiterals);
+    //TODO AYB why weight 3? Should the sort be taken into account?
+    //Look at Kernel/LiteralByMatchability
     t->setWeight(3);
```
quickbeam123, line 332: Did you later figure this out?

I'd say "X=Y" has weight 3 simply because of two the variables and one "=". 

Did you experiment, in the HOL setting, with additional weight for the sorts?

ibnyusuf: I have not yet done so. As the comment suggestions, this is likely to cause issues with polymorphic problems in `Kernel::LiteralByMatchability`. Consider the following code:

```
    static unsigned computeRating(Literal* lit)
    {
      return lit->weight() - lit->getDistinctVars();
    }
```

If `lit->weight()` is always three for two variable equalities, this function is going to work as expected. However, such an issue should not affect monomorphic problems.

quickbeam123: Is it because `lit->getDistinctVars();` will potentially return more than 2 (including variables in the sort)? 

In any case, I understand this is an open issue relevant only for the poly problems.

ibnyusuf:
Sort out the functions `TermList::freeVariables()` and `TermList::freeVars()`. The second should essentially be unnecessary.

vampire.cpp
```diff
@@ -670,9 +678,9 @@ void axiomSelectionMode()
 
   ScopedPtr<Problem> prb(UIHelper::getInputProblem(*env.options));
 
-  if (prb->hasFOOL()) {
+  /*if (prb->hasFOOL()) {
     FOOLElimination().apply(*prb);
-  }
+  }*/
```
quickbeam123, line 683: I never used axiomSelectionMode, but FOOLElimination something to avoid at the moment?

UnitTests/tUnificationWithAbstraction.cpp
```diff
@@ -74,16 +75,16 @@ TermList int_plus(TermList n1, TermList n2)
 {
   return binary(Theory::INT_PLUS,n1,n2);
 }
-Literal* equals(TermList t1, TermList t2)
+Literal* equals(TermList t1, SortType t2)
```
quickbeam123, line 78: The `SortType` here looks like a mistake

Test/SyntaxSugar.hpp
```diff
@@ -82,78 +84,78 @@
...
-                                                                                                                        \
-    operator TermList() { return _self; }                                                                               \
-    operator TermWrapper() { return _self; }                                                                            \
-    unsigned functor() { return _self.toTerm().term()->functor(); }                                                     \
-  };                                                                                                                    \
+#define __DECLARE_CONST(name, sort)                                                                           \
+  class __ ## name ## __CLASS {                                                                               \
+    TermWrapper _self;                                                                                        \
+    public:                                                                                                   \
+    __ ## name ## __CLASS(const char* name, TermList s)                                                       \
```
quickbeam123, line 121: Maybe SortType instead?

UnitTests/tUnificationWithAbstraction.cpp
ibnyusuf: Agreed, I have reverted to `TermList`

Shell/Statistics.hpp
```diff
@@ -188,6 +200,11 @@ class Statistics
   unsigned taNegativeInjectivitySimplifications;
   unsigned taAcyclicityGeneratedDisequalities;
 
+  //to be moved to the property object once that 
+  //is controlled by environment
+  bool higherOrder;
+  bool polymorphic;
```
quickbeam123, line 206: Didn't we resolve the problem with Property already?

Shell/Skolem.cpp
```diff
@@ -393,23 +400,55 @@ Formula* Skolem::skolemise (Formula* f)
...
       }
+      ASS(termArgs.size() == argSorts.size());
+
+      VList* vl = VList::empty();
+      for(int i = args.size() -1; i >= 0 ; i--){
+        VList::push(args[i].var(), vl);
+      }
 
-      Formula::VarList::Iterator vs(f->vars());
+      for(unsigned i = 0; i < termArgs.size() && !_appify; i++){
```
quickbeam123, line 424: Wow, isn't this simply the same as 

```
if (_appify) {
  for ...
```

?

ibnyusuf: It is, I prefer the `&& !_appify` because of its conciseness, but I can imagine that maybe it is a bit confusing.

Shell/Statistics.hpp
ibnyusuf: No, not yet.

Shell/Skolem.cpp
quickbeam123: I won't say you should fix it. But would vote against using such idioms for code that you expect others to read. 

Sometimes _concise_ is not the same thing as _easy to digest while skimming through_.

Shell/Naming.cpp
```diff
@@ -1106,42 +1108,67 @@ bool Naming::canBeInDefinition(Formula* f, Where where) {
...
-  if (env.colorUsed) {
+  unsigned length = VList::length(freeVars);
+  /*if (env.colorUsed) { //TODO what are colors all about?
     Color fc = f->getColor();
     if (fc != COLOR_TRANSPARENT) {
       predSym->addColor(fc);
     }
     if (f->getSkip()) {
       predSym->markSkip();
     }
```
quickbeam123, line 1122: This looks like we are breaking so color-handling code? 

Could you revive if in the non-appify branch?

Shell/Property.cpp
```diff
@@ -141,9 +149,9 @@ void Property::add(UnitList* units)
   }
 
   // information about sorts is read from the environment, not from the problem
-  if (env.sorts->hasSort()) {
+  /*if (env.sorts->hasSort()) {
```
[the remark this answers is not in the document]
ibnyusuf: This was commented out from when the `Sort` class was completely removed. Now that it has been partially reintroduced, I have uncommented the code.

Kernel/Unit.hpp
```diff
@@ -75,6 +75,9 @@ class Unit
   /** see isPureTheoryDescendant in Inference.cpp */
   bool isPureTheoryDescendant() const { return _inference.isPureTheoryDescendant(); }
   /** see isTheoryAxiom in Inference.cpp */
+  bool isCombAxiomsDescendant() const { return _inference.isCombAxiomsDescendant(); }
+  bool isProxyAxiomsDescendant() const { return _inference.isProxyAxiomsDescendant(); }
+  bool isHolAxiomsDescendant() const { return _inference.isHolAxiomsDescendant(); }
   bool isTheoryAxiom() const { return _inference.isTheoryAxiom(); }
```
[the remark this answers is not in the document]
ibnyusuf: Fixed

Kernel/desktop.ini
```diff
@@ -0,0 +1,3 @@
+[LocalizedFileNames]
```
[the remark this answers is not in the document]
ibnyusuf: Seems to have been added by accident.

Shell/Naming.cpp
ibnyusuf: This has been revived.

Shell/FOOLElimination.cpp
```diff
@@ -374,9 +393,9 @@ void FOOLElimination::process(TermList ts, Context context, TermList& termResult
   if (context == FORMULA_CONTEXT) {
     return;
   }
-
   // preprocessing of the term does not affect the sort
-  ASS_EQ(sort, SortHelper::getResultSort(termResult, _varSorts));
+  //ASS_EQ(sort, SortHelper::getResultSort(termResult, _varSorts));
+  //TODO assert that it is a variant
```
quickbeam123, line 398: Can we do it easily, these days?

Shell/FOOLElimination.cpp
```diff
@@ -496,14 +539,16 @@ void FOOLElimination::process(Term* term, Context context, TermList& termResult,
         process(*term->nthArgument(1), context, elseBranch, elseBranchFormula);
 
         // the sort of the term is the sort of the then branch
-        unsigned resultSort = 0;
+        TermList resultSort;
         if (context == TERM_CONTEXT) {
           resultSort = SortHelper::getResultSort(thenBranch, _varSorts);
           ASS_EQ(resultSort, SortHelper::getResultSort(elseBranch, _varSorts));
         }
```
quickbeam123, line 547: This used to be initialized to = 0 (default sort), whenever (context != TERM_CONTEXT);
now we leave unitialized `resultSort` in that case. Is it OK to do so for some reason?

ibnyusuf: I think the old version was defensive programming since `SortHelper::getResultSort(thenBranch, _varSorts);` should return a result in all cases. 

I have reintroduced the defense and the line is now:

`TermList resultSort = Term::defaultSort();`

Shell/FOOLElimination.cpp
ibnyusuf: It doesn't seem so. The closest we have is `MatchingUtils::isVariant`, but this function works on literals, not on TermLists. It seems a bit overkill to re-engineer the function for the sake of a single assertion.

quickbeam123: OK!

InstGen/ModelPrinter.cpp
```diff
@@ -326,13 +329,16 @@ void ModelPrinter::analyzeEqualityAndPopulateDomain()
 
     ALWAYS(ecElIt.hasNext());
     unsigned firstFunc = ecElIt.next();
+    //TODO is the below correct? AYB
```
quickbeam123, line 332: Any update on this? Or should the todo carry over to after merge?

quickbeam123: There is also one additional "TODO" slightly above.

Saturation/SaturationAlgorithm.cpp
```diff
@@ -524,10 +547,20 @@ void SaturationAlgorithm::onClauseReduction(Clause* cl, Clause* replacement,
...
+
   if (env.options->showReductions()) {
     env.beginOutput();
     env.out() << "[SA] " << (forward ? "forward" : "backward") << " reduce: " << cl->toString() << endl;
-    if(replacement){ env.out() << "     replaced by " << replacement->toString() << endl; }
+    ClauseStack::Iterator rit(repStack);
+    while(rit.hasNext()){
+      Clause* replacement = rit.next();
+      if(replacement){ env.out() << "      replaced by " << replacement->toString() << endl; }
+    }
```
quickbeam123, line 563: Is the use case for multiple replacements higher-order logic only?

Saturation/SaturationAlgorithm.cpp
```diff
@@ -541,6 +574,7 @@ void SaturationAlgorithm::onClauseReduction(Clause* cl, Clause* replacement,
   }
 
   if (replacement) {
+    //TODO fix this
     onParenthood(replacement, cl);
```
quickbeam123, line 578: Is that thing that should be fixed the fact, that only the first replacement will have "onParenthood" called?

Did you decide it's OK never to change this?

Saturation/SaturationAlgorithm.cpp
```diff
@@ -959,13 +1011,37 @@ bool SaturationAlgorithm::forwardSimplify(Clause* cl)
...
+      ClauseIterator results = se->perform(cl);
+ 
+      if (results.hasNext()) {
+        while(results.hasNext()){
+          Clause* simpedCl = results.next();
+          ASS(simpedCl != cl);
+          repStack.push(simpedCl);
+          addNewClause(simpedCl);
+        }
+        onClauseReduction(cl, pvi(ClauseStack::Iterator(repStack)), 0);
```
quickbeam123, line 1039: You put the replacements on a stack, turn that into an iterator, then pass it to onClauseReduction and then turn the iterator back into a stack. 

Did you consider passing `const& ClauseStack` directly instead?

quickbeam123: Or simply (Clause*)[] and a count?

InstGen/ModelPrinter.cpp
ibnyusuf: The confusion comes from an ignorance of what exactly `IntUnionFind`. Previously it took the set of all function symbols and did stuff (which I don't understand). It now takes a set which includes functional symbols and type constructors. Is it sufficiently to filter out the type constructors on line 332, or could there be other issues?

Saturation/SaturationAlgorithm.cpp
ibnyusuf: Correct. We currently have higher-order immediate simplifcation inferences that return multiple conclusions. 

It could also be used for `InjectivityGIE` in `TermAlgebraReasoning` which has been inaccurately (in comparison to the paper) coded as a generating inference. However, this hasn't been done yet.

Parse/TPTP.cpp
```diff
@@ -2385,10 +2885,37 @@ void TPTP::term()
...
+          ASSERTION_VIOLATION;
+      }
+      return;
+    }
+
     case T_STRING:
     case T_INT:
     case T_REAL:
     case T_RAT: {
+      if(env.statistics->polymorphic || env.statistics->higherOrder){
```
quickbeam123, line 2916: I still have more than 100 files to review. Won't you have a look at this?

What's the main problem with parsing polymorphism along with theories?

(Thinking of a problem with a general theory of lists and a conjecture about a list of ints and their, say, sum.

Definitely not about making e.g. $sum polymorhic for int/rat/real.)

Saturation/SaturationAlgorithm.cpp
ibnyusuf: That is correct. Assuming we never run symbol elimination along with higher-order we should be safe (looking at onParenthood). I have made the comment more explanatory.

Parse/TPTP.cpp
```diff
@@ -2775,7 +3328,7 @@ void TPTP::endFormula()
   case FORALL:
   case EXISTS:
     f = _formulas.pop();
-    _formulas.push(new QuantifiedFormula((Connective)con,_varLists.pop(),_sortLists.pop(),f));
+    _formulas.push(new QuantifiedFormula((Connective)con,_varLists.pop(),_sortLists.pop(),f)); //TODO this is broken currently.
     _states.push(END_FORMULA);
```
quickbeam123, line 3332: Sounds scary. Could you explain more? Is it just a weird case that is broken?

Parse/TPTP.cpp
ibnyusuf: The issue is that the theory reasoning has not been updated in any way to deal with truly polymorphic problems (e.g. it may make use of `NonVariableIterator`).

Until that has been dealt with, it seems safest to raise a parse error for all problems that combine polymorphism with theories (some may work, but we have no guarantees)

Parse/TPTP.cpp
ibnyusuf: That is a very old comment that I unfortunately left lying around. It all appears to work at the moment.

Saturation/SaturationAlgorithm.cpp
ibnyusuf: That is quite strange!

I think the reason for using an iterator is that it is easier to set up single element iterators than single element stacks, but I will attempt to change this.

quickbeam123: Note that a single element (Clause*)[] is just the address of the pointer holding the clause in the current context. 

Though this can only be used, of course, if the caller lives at all times when the reference is needed. But I think this is the case here!

Parse/TPTP.cpp
quickbeam123: Great!

InstGen/ModelPrinter.cpp
quickbeam123: This is not helping me much to resolve this.

How about we try running Instgen on some satisfiable problems with proof printing on to see whether it does something reasonable?

Saturation/SaturationAlgorithm.cpp
quickbeam123: Thanks!

Indexing/ClauseCodeTree.cpp
```diff
@@ -646,17 +646,17 @@ inline bool ClauseCodeTree::ClauseMatcher::canEnterLiteral(CodeOp* op)
...
       matchIndex--;
       MatchInfo* mi=ils->getMatch(matchIndex);
       unsigned liIntex = mi->liIndex;
       Literal* lit = (*query)[lInfos[liIntex].litIndex];
       ASS(lit->isEquality());
-      unsigned argSort = SortHelper::getEqualityArgumentSort(lit);
-      if(idxVarSort!=argSort) {
-	ils->deleteMatch(matchIndex); //decreases ils->matchCnt
+      TermList argSort = SortHelper::getEqualityArgumentSort(lit); 
+      if(idxVarSort!=argSort) {//TODO check that this is what we want. Perhaps require unification
```
quickbeam123, line 658: I forgot, can we resolved this with current knowledge?

ibnyusuf: Unfortunately not. I don't understand CodeTrees, so I am not sure what this function does, or what ramifications changing away from syntactic equality check may have. The good news is that the only place we use ClauseCodeTrees is in demodulation, so having a too strict test here will at worst result in some lost simplifications.

quickbeam123: OK, then let's leave a TODO

Kernel/Signature.cpp
```diff
@@ -213,16 +234,32 @@ Signature::Signature ():
...
+ * from Term::defaultSort before the Signature has been consstucted. hence
+ * the function below
+ */
+void Signature::addEquality()
+{
+  CALL("Signature::addEquality");
+  // initialize equality
+  addInterpretedPredicate(Theory::EQUAL, OperatorType::getPredicateType(2), "=");
+  ASS_EQ(predicateName(0), "="); //equality must have number 0
+  getPredicate(0)->markSkip(); //TODO is this safe? Doesn't look like it!
```
quickbeam123, line 260: Hi, do you recall what was the issue with this one?

Kernel/MLMatcher.cpp
```diff
@@ -506,6 +501,8 @@ void MLMatcher::Impl::initMatchingData(Literal** baseLits0, unsigned baseLen, Cl
       zeroAlts++;
       singleAlts++;
     } else if(currAltCnt==1 && !(resolvedLit && resolvedLit->couldBeInstanceOf(s_baseLits[i], true)) ) {
+      //TODO the above requires updating to polymorphism
+      //for now, a minor source of incompleteness
```
quickbeam123, line 505: I have forgotten... Do we have the mechanism in place, not to report SAT on polymorphic inputs which we finitely saturate?

Kernel/Inference.hpp
```diff
@@ -54,6 +54,52 @@ inline std::underlying_type<UnitInputType>::type toNumber(UnitInputType t) { ret
...
+ *    be working on type arguments. Please view TermIterators.hpp for a list of
+ *    iterators available.
+ *  - TermSubstitutionTrees, do NOT carry out any type checking. Thus, in the case
+ *    where either the search or query term is a variable, a type check needs to
+ *    be carried out by the inference code. This check may be a unification check or 
+ *    a matching check depending on whether the inference is using unification or
+ *    matching. Please view Superposition.cpp for an example of a unification check
+ *    and ForwardDemodulation for an example of a matching check.
+ */
+
```
quickbeam123, line 102: Thank you, Ahmed, for writing this up! Looks very useful!

ibnyusuf: Thanks, I hope it saves others some of the difficulties I went through!

Kernel/MLMatcher.cpp
ibnyusuf: No we don't. I think that the current thing to do would be to fix the above, which hopefully I can do with Jakob's help. If not, I will edit the reporting of SAT.

ibnyusuf: Looking through the code, It seems that I already dealt with the above issue.

`couldBeInstanceOf` makes a call to `Literal::headersMatch` which in `master` contains a syntactic equality sort check making it unsuitable. However, I have removed this check from the polymorphic version and pushed it into the places in the code that call `Literal::headersMatch`. 

If you agree with this, I think the comment can be deleted and this issue resolved.

quickbeam123: Sure, go for it!

PR #156, opened 2020-10-28
Simple CI to get started.

MichaelRawson:
This PR creates a GitHub action that can be _manually_ run to check that a branch compiles. More advanced things to come if this works well.

### DIFF
--- a/.github/workflows/debug-build.yml
+++ b/.github/workflows/debug-build.yml
@@ -7,12 +7,10 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - name: Create Build Directory
-      run: cmake -E make_directory ${{runner.workspace}}/build
+      run: mkdir ${{runner.workspace}}/build
     - name: Configure Build
-      shell: bash
       working-directory: ${{runner.workspace}}/build
       run: cmake $GITHUB_WORKSPACE -DCMAKE_BUILD_TYPE=Debug
     - name: Build
       working-directory: ${{runner.workspace}}/build
-      shell: bash
-      run: make -j8
+      run: make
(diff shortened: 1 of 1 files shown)


MichaelRawson:
@selig figured out I could test this by forking Vampire, then testing on my own copy. It (somewhat surprisingly IMO) builds fine, example run here: https://github.com/MichaelRawson/vampire-ci/actions/runs/341351848

I doubt we're likely to run into usage limits: https://docs.github.com/en/free-pro-team@latest/actions/reference/usage-limits-billing-and-administration

What else would you like implemented to make this useful for people? Currently you have to push a button to run the build, and the build is a CMake/debug configuration without Z3. Do we want build artifacts/different build configuration/automatic builds on e.g. pull requests?

selig:
Having it run automatically on PR so that it impacts whether the PR is mergeable makes sense.

Can we check whether the build completes without warnings? Adding unit tests would be good (I think you're already thinking about this based on your question to Joe).

Adding the built system as an artefact seems like a good bonus, but are there limits that we might hit?

MichaelRawson:
You can check the log for warnings - not sure about whether you can report them in e.g. the pull request.

Artefacts shouldn't be a problem - apparently they're deleted after a configurable period (max. 90 days) anyway.

quickbeam123:
What is the logic here? That once this is in master, github enables "a new button"/something in the web gui that will allow us to do the CI?

MichaelRawson:
Yes - it magically appears under the "actions" tab. It can be automatically run but I haven't done it yet so we can decide if we like it/add other stuff before we start doing it automatically.

quickbeam123:
Then let's have some fun with it ;)

MichaelRawson:
@quickbeam123 - lol, merging the Z3-submodule branch before this one was brave! Hold on tight...

PR #160, opened 2020-10-30
fix warnings from static analysis, possible bugs in some cases

MichaelRawson:
I ran the `clang-tidy` tool over Vampire. It discovered some false-positives, some fishy code, and some that look like bugs. This patches all of these: see comments in-line. The tree now compiles cleanly under `clang-tidy`, except for third-party code (minisat).

### DIFF
--- a/DP/SimpleCongruenceClosure.hpp
+++ b/DP/SimpleCongruenceClosure.hpp
@@ -100,11 +100,11 @@ private:
   /** Equality between two constants */
   struct CEq
   {
-    CEq() : c1(0), c2(0) {}
+    CEq() : c1(0), c2(0), foOrigin(false), foPremise(nullptr) {}
     CEq(unsigned c1, unsigned c2, Literal* lit)
      : c1(c1), c2(c2), foOrigin(true), foPremise(lit) {}
     CEq(unsigned c1, unsigned c2)
-     : c1(c1), c2(c2), foOrigin(false) {}
+     : c1(c1), c2(c2), foOrigin(false), foPremise(nullptr) {}
 
     bool isInvalid() const { ASS_EQ(c1==0, c2==0); return c1==0; }
     vstring toString() const;
--- a/FMB/FiniteModelBuilder.cpp
+++ b/FMB/FiniteModelBuilder.cpp
@@ -2588,8 +2588,6 @@ bool FiniteModelBuilder::SmtBasedDSAE::increaseModelSizes(DArray<unsigned>& newS
       }
       _smtSolver.add(sum == _context.int_val(_lastWeight));
 
-      result = _smtSolver.check();
-
       if (_smtSolver.check() == z3::check_result::sat) {
         loadSizesFromSmt(newSortSizes);
         // cout << "\nFound "; output_sizes(_distinctSortSizes); cout << endl;
--- a/Indexing/CodeTree.cpp
+++ b/Indexing/CodeTree.cpp
@@ -836,7 +836,7 @@ void CodeTree::compileTerm(Term* trm, CodeStack& code, CompileContext& cctx, boo
 	  sti.right();
 	}
 	else {
-	  code.push(CodeOp::getTermOp(CHECK_FUN, s.term()->functor()));
+	  code.push(CodeOp::getTermOp(CHECK_FUN, t->functor()));
 	}
       }
     }
@@ -1325,7 +1325,6 @@ bool CodeTree::RemovingMatcher::next()
       if(!backtrack()) {
 	return false;
       }
-      shouldBacktrack=false;
     }
     else {
       //the SEARCH_STRUCT operation does not appear in CodeBlocks
--- a/Indexing/CodeTree.cpp
+++ b/Indexing/CodeTree.cpp
@@ -1323,8 +1323,10 @@ bool CodeTree::RemovingMatcher::next()
     }
     if(shouldBacktrack) {
       if(!backtrack()) {
-	return false;
+        return false;
       }
+      // dead store, left here in case it should have been a static?
+      // shouldBacktrack = false;
     }
     else {
       //the SEARCH_STRUCT operation does not appear in CodeBlocks
--- a/Indexing/SubstitutionTree.cpp
+++ b/Indexing/SubstitutionTree.cpp
@@ -259,9 +259,10 @@ start:
     BindingMap::Iterator svit(svBindings);
     BinaryHeap<Binding, BindingComparator> remainingBindings;
     while (svit.hasNext()) {
-      Binding b;
-      svit.next(b.var, b.term);
-      remainingBindings.insert(b);
+      unsigned var;
+      TermList term;
+      svit.next(var, term);
+      remainingBindings.insert(Binding(var, term));
     }
     while (!remainingBindings.isEmpty()) {
       Binding b=remainingBindings.pop();
--- a/Indexing/SubstitutionTree.hpp
+++ b/Indexing/SubstitutionTree.hpp
@@ -652,8 +652,6 @@ public:
     TermList term;
     /** Create new binding */
     Binding(int v,TermList t) : var(v), term(t) {}
-    /** Create uninitialised binding */
-    Binding() {}
 
     struct Comparator
     {
--- a/Inferences/Induction.cpp
+++ b/Inferences/Induction.cpp
@@ -630,6 +630,7 @@ void InductionClauseIterator::performStructInductionThree(Clause* premise, Liter
       while(vtit.hasNext()){
         smallers = new FormulaList(new AtomicFormula(Literal::create1(sty,true,TermList(vtit.next(),false))),smallers);
       }
+      ASS(smallers);
       Formula* ax = Formula::quantify(new BinaryFormula(Connective::IMP,smaller_coni, 
                      (FormulaList::length(smallers) > 1) ? new JunctionFormula(Connective::AND,smallers)
                                             : smallers->head()
--- a/Kernel/RobSubstitution.hpp
+++ b/Kernel/RobSubstitution.hpp
@@ -148,7 +148,7 @@ public:
   struct TermSpec
   {
     /** Create a new TermSpec struct */
-    TermSpec() {}
+    TermSpec() : index(0) {}
     /** Create a new TermSpec struct */
     TermSpec(TermList term, int index) : term(term), index(index) {}
     /** Create a new TermSpec struct from a VarSpec*/
--- a/Kernel/SpassLiteralSelector.cpp
+++ b/Kernel/SpassLiteralSelector.cpp
@@ -76,8 +76,8 @@ void SpassLiteralSelector::doSelection(Clause* c, unsigned eligible)
     }
   }
 
-  LiteralList* sel = LiteralList::empty();
-
+  // sel always initialised below
+  LiteralList* sel;
   if(singleSel) {
     LiteralList::destroy(maximals);
     sel = LiteralList::empty();
--- a/Kernel/TermIterators.hpp
+++ b/Kernel/TermIterators.hpp
@@ -283,7 +283,7 @@ public:
     CALL("NonVariableIterator::NonVariableIterator");
     _stack.push(term);
     if (!includeSelf) {
-      next();
+      NonVariableIterator::next();
     }
   }
   // NonVariableIterator(TermList ts);
--- a/Lib/Metaiterators.hpp
+++ b/Lib/Metaiterators.hpp
@@ -369,7 +369,7 @@ public:
   DECL_ELEMENT_TYPE(ELEMENT_TYPE(Inner));
 
   FilteredIterator(Inner inn, Functor func)
-  : _func(func), _inn(inn), _nextStored(false) {}
+  : _func(func), _inn(inn), _next(), _nextStored(false) {}
   bool hasNext()
   {
     if(_nextStored) {
--- a/Lib/MultiCounter.hpp
+++ b/Lib/MultiCounter.hpp
@@ -56,6 +56,7 @@ public:
     if (v >= _top) { // not enough capacity
       expandToFit(v);
     }
+    ASS(_counts);
     _counts[v]++;
   } // MultiCount::inc
 
@@ -68,6 +69,7 @@ public:
     if (v >= _top) { // not enough capacity
       expandToFit(v);
     }
+    ASS(_counts);
     _counts[v]--;
   } // MultiCount::dec
 
@@ -80,6 +82,7 @@ public:
     if (v >= _top) { // not enough capacity
       expandToFit(v);
     }
+    ASS(_counts);
     _counts[v] = c;
   } // MultiCounter::set
 
@@ -121,5 +124,3 @@ private:
 }
 
 #endif // __MultiCounter__
-
-
--- a/Lib/SmartPtr.hpp
+++ b/Lib/SmartPtr.hpp
@@ -63,6 +63,7 @@ public:
   SmartPtr(const SmartPtr& ptr) : _obj(ptr._obj), _refCnt(ptr._refCnt)
   {
     if(_obj && _refCnt) {
+      ASS(_refCnt->_val > 0);
       (_refCnt->_val)++;
     }
   }
--- a/Lib/Timer.cpp
+++ b/Lib/Timer.cpp
@@ -223,21 +223,13 @@ void Lib::Timer::syncClock()
 {
   if(s_initGuarantedMiliseconds==-1) {
     //we're unable to sync clock as we weren't able to obtain number of ticks in the beginning
-    bool reportedProblem = false;
-    if(!reportedProblem) {
-      reportedProblem = true;
-      cerr << "cannot syncronize clock as times() initially returned -1" << endl;
-    }
+    cerr << "cannot synchronize clock as times() initially returned -1" << endl;
     return;
   }
   int newMilliseconds = guaranteedMilliseconds();
   if(newMilliseconds==-1) {
     //we're unable to sync clock as we cannot get the current time
-    bool reportedProblem = false;
-    if(!reportedProblem) {
-      reportedProblem = true;
-      cerr << "could not syncronize clock as times() returned -1" << endl;
-    }
+    cerr << "could not synchronize clock as times() returned -1" << endl;
     return;
   }
 
--- a/Lib/Timer.cpp
+++ b/Lib/Timer.cpp
@@ -221,15 +221,22 @@ void Lib::Timer::deinitializeTimer()
 
 void Lib::Timer::syncClock()
 {
+  static bool reportedProblem = false;
   if(s_initGuarantedMiliseconds==-1) {
     //we're unable to sync clock as we weren't able to obtain number of ticks in the beginning
-    cerr << "cannot synchronize clock as times() initially returned -1" << endl;
+    if(!reportedProblem) {
+      reportedProblem = true;
+      cerr << "cannot syncronize clock as times() initially returned -1" << endl;
+    }
     return;
   }
   int newMilliseconds = guaranteedMilliseconds();
   if(newMilliseconds==-1) {
     //we're unable to sync clock as we cannot get the current time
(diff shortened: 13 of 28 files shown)


DP/SimpleCongruenceClosure.hpp
```diff
@@ -100,11 +100,11 @@ class SimpleCongruenceClosure : public DecisionProcedure
   /** Equality between two constants */
```
MichaelRawson, line 100: Some constructors do not initialise some members - not a problem _if everything is initialised before use_, but let's make it deterministic and cause a null-pointer exception if members are not subsequently initialised later.

Indexing/CodeTree.cpp
```diff
@@ -836,7 +836,7 @@ void CodeTree::compileTerm(Term* trm, CodeStack& code, CompileContext& cctx, boo
 	  sti.right();
 	}
 	else {
-	  code.push(CodeOp::getTermOp(CHECK_FUN, s.term()->functor()));
```
selig, line 839: Although I'm not sure whether CodeTree is currently in use in normal operation, we should check these changes. What led to these suggestions?

FMB/FiniteModelBuilder.cpp
```diff
@@ -2588,8 +2588,6 @@ bool FiniteModelBuilder::SmtBasedDSAE::increaseModelSizes(DArray<unsigned>& newS
       }
       _smtSolver.add(sum == _context.int_val(_lastWeight));
```
MichaelRawson, line 2590: @selig - I believe this is your domain. Seems like you check Z3 twice and throw away the first result?

Indexing/CodeTree.cpp
MichaelRawson: Hang on, I'll comment each file! This one is because `t` is set to `s.term()` (pure getter) earlier, but never used.

Lib/Timer.cpp
```diff
@@ -223,21 +223,13 @@ void Lib::Timer::syncClock()
 {
   if(s_initGuarantedMiliseconds==-1) {
     //we're unable to sync clock as we weren't able to obtain number of ticks in the beginning
-    bool reportedProblem = false;
-    if(!reportedProblem) {
-      reportedProblem = true;
-      cerr << "cannot syncronize clock as times() initially returned -1" << endl;
-    }
+    cerr << "cannot synchronize clock as times() initially returned -1" << endl;
```
selig, line 226: I think this code is stopping is from printing the error more than once but, for this to work, we should make reportedProblem static.

Lib/Timer.cpp
```diff
@@ -223,21 +223,13 @@ void Lib::Timer::syncClock()
...
   }
   int newMilliseconds = guaranteedMilliseconds();
   if(newMilliseconds==-1) {
     //we're unable to sync clock as we cannot get the current time
-    bool reportedProblem = false;
-    if(!reportedProblem) {
-      reportedProblem = true;
-      cerr << "could not syncronize clock as times() returned -1" << endl;
-    }
+    cerr << "could not synchronize clock as times() returned -1" << endl;
```
selig, line 232: Same here

SAT/Preprocess.cpp
```diff
@@ -221,11 +221,16 @@ void Preprocess::propagateUnits(SATClauseIterator clauses,
...
+#ifdef VDEBUG
+    // set to true when kept is assigned something
+    bool kept_assigned = false;
+#endif
     for(unsigned i=0;i<clen;i++) {
       SATLiteral lit=(*cl)[i];
       bool posUnit;
       if(!unitBindings.find(lit.var(), posUnit)) {
 	kept=lit;
+	kept_assigned = true;
```
selig, line 233: doesn't this also need to be in a VDEBUG?

Indexing/SubstitutionTree.cpp
```diff
@@ -259,9 +259,10 @@ void SubstitutionTree::insert(Node** pnode,BindingMap& svBindings,LeafData ld)
     BindingMap::Iterator svit(svBindings);
     BinaryHeap<Binding, BindingComparator> remainingBindings;
     while (svit.hasNext()) {
```
MichaelRawson, line 261: There's a default constructor for Binding (now killed) that leaves things unitialised. We can actually get rid of it altogether in this case.

Indexing/SubstitutionTree.hpp
```diff
@@ -652,8 +652,6 @@ class SubstitutionTree
     TermList term;
     /** Create new binding */
     Binding(int v,TermList t) : var(v), term(t) {}
```
MichaelRawson, line 654: remove empty Binding constructor

Inferences/Induction.cpp
```diff
@@ -630,6 +630,7 @@ void InductionClauseIterator::performStructInductionThree(Clause* premise, Liter
       while(vtit.hasNext()){
         smallers = new FormulaList(new AtomicFormula(Literal::create1(sty,true,TermList(vtit.next(),false))),smallers);
       }
```
MichaelRawson, line 632: actually this might be dubious, let me get back to this one

Kernel/SpassLiteralSelector.cpp
```diff
@@ -76,8 +76,8 @@ void SpassLiteralSelector::doSelection(Clause* c, unsigned eligible)
     }
   }
```
MichaelRawson, line 78: unused init, always initialised in the block below

Kernel/TermIterators.hpp
```diff
@@ -283,7 +283,7 @@ class NonVariableIterator
     CALL("NonVariableIterator::NonVariableIterator");
     _stack.push(term);
     if (!includeSelf) {
```
MichaelRawson, line 285: In constructors, virtual calls aren't virtual. Make this explicit.

SAT/RestartStrategy.hpp
```diff
@@ -118,7 +118,7 @@ class MinisatRestartStrategy : public RestartStrategy {
    * According to "Armin Biere, PicoSAT Essentials" paper.
    */
   MinisatRestartStrategy(size_t initCnt = 100, float increase=1.1f)
```
MichaelRawson, line 120: In constructors, virtual calls aren't virtual. Make this explicit.

FMB/FiniteModelBuilder.cpp
selig: Yeah, probably leftover from debugging ! I don't think it will be a performance hit as z3 will cache the result as nothing is added but deleting the line makes sense as it's not doing anything!

DP/SimpleCongruenceClosure.hpp
selig: All of the constructor fixes here look sensible.

Kernel/RobSubstitution.hpp
```diff
@@ -148,7 +148,7 @@ class RobSubstitution
   struct TermSpec
   {
     /** Create a new TermSpec struct */
```
MichaelRawson, line 150: Would we prefer to set this to some invalid value? I believe this 0 is actually a reasonable value which might not cause a crash.

Indexing/CodeTree.cpp
```diff
@@ -1325,7 +1325,6 @@ bool CodeTree::RemovingMatcher::next()
       if(!backtrack()) {
 	return false;
       }
-      shouldBacktrack=false;
```
selig, line 1328: What about this - is it just never read?

Inferences/Induction.cpp
selig: Adding in some assertions where you "think" they should be true (but with a comment on why) is reasonable defensive programming - if we hit the assertion and think it's okay then we relax or remove it later.

Lib/MultiCounter.hpp
```diff
@@ -56,6 +56,7 @@ class MultiCounter
     if (v >= _top) { // not enough capacity
       expandToFit(v);
     }
```
MichaelRawson, line 58: Not actually required, obviously. In some convoluted cases the static analyser gets confused - I think this is worth the extra lines, it's not much cost in debug and zero cost in release.

Lib/Timer.cpp
MichaelRawson: Ah, that makes more sense. I thought it was a refactor-gone-wrong. I'll static it.

SAT/Preprocess.cpp
MichaelRawson: Yes it does!

SAT/Preprocess.cpp
```diff
@@ -245,6 +250,7 @@ void Preprocess::propagateUnits(SATClauseIterator clauses,
     unsigned newLen=clen-removedLitIndexes.length();
 
     if(newLen==1) {
```
MichaelRawson, line 252: The static analyser can find a lengthy code path where `kept` isn't initialised here. I think in practice it's OK, but this serves as documentation and also prevents a disaster if I'm wrong.

SAT/VariableSelector.cpp
```diff
@@ -177,7 +177,11 @@ for(const TermList* t = _stack.pop(); !t->isEmpty(); t = t->next()){
       }
       RSTAT_MCTR_INC("nscore",level_sum);
       // if we want average, we set divid=true earlier
-      if(divide){ level_sum = (level_sum/count);}//This division will truncate
```
MichaelRawson, line 180: Divide-by-zero. I don't think this ever happens, but good to trap it.

Shell/FOOLElimination.cpp
```diff
@@ -496,11 +496,11 @@ void FOOLElimination::process(Term* term, Context context, TermList& termResult,
...
 
         TermList thenBranch;
-        Formula* thenBranchFormula;
+        Formula* thenBranchFormula {};
         process(*term->nthArgument(0), context, thenBranch, thenBranchFormula);
 
         TermList elseBranch;
-        Formula* elseBranchFormula;
+        Formula* elseBranchFormula {};
         process(*term->nthArgument(1), context, elseBranch, elseBranchFormula);
```
MichaelRawson, line 504: Null out pointers that "should" be initialised by subsequent `process` calls. Little to no cost in release (either an XOR or hopefully dead-store elimination), traps if something disastrous happens.

Shell/NewCNF.cpp
```diff
@@ -1086,12 +1086,9 @@ void NewCNF::processBoolterm(TermList ts, Occurrences &occurrences)
     case Term::SF_ITE: {
       Formula* condition = sd->getCondition();
```
MichaelRawson, line 1088: Analyser is apparently confused by this for-loop in particular. As it's a bit pointless anyway it can go.

Shell/Options.cpp
```diff
@@ -2587,7 +2587,7 @@ bool Options::RatioOptionValue::readRatio(const char* val, char separator)
       return false;
     }
     char copy[128];
```
MichaelRawson, line 2589: Safety. Already checked by preconditions but worth having as can help compiler tell when e.g. buffer is too small. Not in a hot loop.

Shell/SimplifyFalseTrue.cpp
```diff
@@ -371,12 +371,12 @@ TermList SimplifyFalseTrue::simplify(TermList ts)
...
-        bool isFalse[2];
-        for (BRANCH branch : {THEN, ELSE }) {
-          branches[branch] = simplify(*term->nthArgument(branch));
-        }
+        TermList branches[2] = {
+          simplify(*term->nthArgument(THEN)),
+          simplify(*term->nthArgument(ELSE)),
+        };
+        bool isTrue[2] {};
+        bool isFalse[2] {};
```
MichaelRawson, line 380: Same change as in NewCNF.

Shell/SimplifyProver.cpp
```diff
@@ -325,7 +325,7 @@ void SimplifyProver::unbindVar(const vstring& varName)
   CALL("SimplifyProver::unbindVar");
 
   IntList* bindings = 0;
-  _variables.find(varName,bindings);
+  ALWAYS(_variables.find(varName,bindings));
   IntList* tl = bindings->tail();
```
MichaelRawson, line 329: This must always succeed, otherwise deref null.

Shell/SimplifyProver.cpp
```diff
@@ -1501,15 +1501,15 @@ void SimplifyProver::undoLet()
     case K_FORMULA:
       {
 	Lib::List<Formula*>* binding;
-	_formulaLet.find(symb,binding);
+	ALWAYS(_formulaLet.find(symb,binding));
 	_formulaLet.replace(symb,binding->tail());
```
MichaelRawson, line 1505: This must always succeed, otherwise deref null.

Shell/SimplifyProver.cpp
```diff
@@ -1501,15 +1501,15 @@ void SimplifyProver::undoLet()
...
 	delete binding;
       }
       return;
     case K_TERM:
       {
 	Lib::List<TermList>* binding;
-	_termLet.find(symb,binding);
+	ALWAYS(_termLet.find(symb,binding));
 	_termLet.replace(symb,binding->tail());
```
MichaelRawson, line 1513: This must always succeed, otherwise deref null.

Shell/TheoryFinder.cpp
```diff
@@ -164,10 +164,10 @@ class TheoryFinder::Backtrack
 public:
   /** code pointer */
   unsigned cp;
-  /** object on which the instruction should be executed */
-  const void* obj;
   /** position in the object stack */
   unsigned objPos;
+  /** object on which the instruction should be executed */
+  const void* obj;
 }; // TheoryFinder::Backtrack
```
MichaelRawson, line 171: Aligment. The previous ordering of fields generates 8 bytes of padding, this ordering has no padding at all. Nothing seems to depend upon the precise order.

Shell/TheoryFinder.cpp
```diff
@@ -418,6 +424,7 @@ bool TheoryFinder::matchCode(const void* obj,
 #endif
     unsigned l = code[cp+1];
     // bit field of choices for this literal
+    ASS(clength_assigned);
     unsigned choice = (1u << clength) - 1;
```
MichaelRawson, line 428: Same as with `kept_assigned` above - really has to be initialised otherwise something bad will happen.

Indexing/CodeTree.cpp
MichaelRawson: Yeah, dead-store elimination caught it.

MichaelRawson: Ah - unless this is another should-have-been-a-static case @selig ?

Lib/SmartPtr.hpp
```diff
@@ -63,6 +63,7 @@ class SmartPtr {
   SmartPtr(const SmartPtr& ptr) : _obj(ptr._obj), _refCnt(ptr._refCnt)
   {
     if(_obj && _refCnt) {
```
MichaelRawson, line 65: "should" always be true - otherwise the memory has already been deleted. Also a classic thread-safety bug (need an atomic refcount), might canary later!

Lib/Metaiterators.hpp
```diff
@@ -369,7 +369,7 @@ class FilteredIterator
   DECL_ELEMENT_TYPE(ELEMENT_TYPE(Inner));
 
   FilteredIterator(Inner inn, Functor func)
-  : _func(func), _inn(inn), _nextStored(false) {}
+  : _func(func), _inn(inn), _next(), _nextStored(false) {}
   bool hasNext()
```
MichaelRawson, line 373: This one looks innocent, but is actually pernicious. ELEMENT_TYPE is both POD (usually raw pointers) or user-defined types in Vampire. In the POD case this change just adds the same zero-initialise as elsewhere.

If it's a user-defined type, this change has no effect. However, I notice that this means any use of the filter class will incur whatever the default constructor does, then throw it away later. We probably don't want (or intend?) to do this, but these are the current semantics of `FilteredIterator` so I don't change it here.

Inferences/Induction.cpp
MichaelRawson: Actually this assert has to be true or we will crash - in the ternary operator below, either `smallers` has some length more than one (and is therefore non-null), or we call `->head()` - which will crash if `smallers` is `nullptr`.

MichaelRawson:
Thanks for the review @selig, I fixed your suggestions. Can you look at the comments on `RobSubstitution.hpp`, `CodeTree.cpp` and `Metaiterators.hpp` before merging?

Indexing/CodeTree.cpp
selig: Bleh, it's been many years since I've looked at CodeTree. Can we make it a comment-out instead of a delete in case it helps somebody in the future to know that it was there in the past!

MichaelRawson: Done!

Indexing/CodeTree.cpp
JakobR: @selig I think only forward demodulation currently uses code trees (the `DEMODULATION_LHS_SUBST_TREE` index)

Kernel/RobSubstitution.hpp
quickbeam123: Could we live without this 0-param constructor altogether?

Lib/Metaiterators.hpp
quickbeam123: In other word, the new version is just explicitly saying what was implicitly happening anyway, but we know suspect, it's not such a great idea for non PODs, right?

MichaelRawson: Exactly that!

Kernel/RobSubstitution.hpp
MichaelRawson: Seems not - it's used in a `DHMap<VarSpec, TermSpec>` member which then requires the default constructor - unless you can think of some way around it?

SAT/VariableSelector.cpp
quickbeam123: I don't think we are using the in-house SAT solver anymore (not in the casc strategies...), so this code is not essential...

MichaelRawson: Great - do we intend to retire the in-house solver? If so I can add it to my "unused code" issue.

Shell/Options.cpp
quickbeam123: When you do a final pass of this, could you please factor out "128" to a "const int COPY_SIZE" or something similar?

Shell/SimplifyFalseTrue.cpp
quickbeam123: Is 

`bool isTrue[2] {};`

the same as 

`bool isTrue[2] {false, false};`

?

MichaelRawson: I think so. https://en.cppreference.com/w/cpp/language/zero_initialization says "If T is array type, each element is zero-initialized."

Kernel/RobSubstitution.hpp
quickbeam123: Oh, I see. Then let's not waste too much effort here atm.

SAT/VariableSelector.cpp
quickbeam123: Please, add it to the issue. Let's see whether Giles does not have any plans with it. I am happy with it going.

Shell/SimplifyFalseTrue.cpp
quickbeam123: I need to get used to see-ing {} as an initializer. Not sure if we should insist on the more explicit bool isTrue[2] {false, false};

MichaelRawson: I agree it's a bit of an odd syntax choice! I'm OK with whichever - `T t{};` where `T` is non-POD is quite helpful though to avoid `T t();` (function declaration) or `T t = T();` (unnecessary copy).

MichaelRawson:
@quickbeam123 - most recent commit factors out `COPY_SIZE`. Good to go?

PR #161, opened 2020-11-02
CMake changes: Z3 in submodule, some reorganisation.

MichaelRawson:
As discussed with @selig, this implements Z3-as-a-submodule. You still have to build Z3 in the submodule yourself (because different systems might need different handling), but now we hard-code the path in our CMake to the submodule. This "heavily encourages" use of a fixed-in-time, source-build Z3.

Also:
- misc. CMake reshuffles/fixes
- remove existing `z3` header directory, not used as far as we know

Minisat will be left alone as it's not very large and no longer changes. It may be patched in future to perform better for Vampire's purposes.

I'm not sure if (all of) this should be merged. From my point of view it's an improvement: a bug in some random version of Z3 on the build machine cannot cause bugs in Vampire, only bugs in the version we stick to. The build system is also simpler and less fragile for finding Z3. People might reasonably object that if they already have a build of Z3 somewhere else, they can't easily use that for building Vampire (although I think symlinking the build directory works).


quickbeam123:
```
Martins-MacBook-Pro-2:build mbassms6$ git submodule update --init
Submodule 'z3' (https://github.com/Z3Prover/z3) registered for path '../z3'
fatal: destination path '/Users/mbassms6/GitHub/vampire/z3' already exists and is not an empty directory.
fatal: clone of 'https://github.com/Z3Prover/z3' into submodule path '/Users/mbassms6/GitHub/vampire/z3' failed
Failed to clone 'z3'. Retry scheduled
fatal: destination path '/Users/mbassms6/GitHub/vampire/z3' already exists and is not an empty directory.
fatal: clone of 'https://github.com/Z3Prover/z3' into submodule path '/Users/mbassms6/GitHub/vampire/z3' failed
Failed to clone 'z3' a second time, aborting
```
../z3 is actually my z3 installation. Can I use it as is without having to clean a clone again?

MichaelRawson:
@quickbeam123  I think yes, provided it matches the same commit. The submodule will be populated so git should do its thing, you only need to `git submodule update` if you don't have the submodule contents.

If not, what about:
- move the `z3` directory somewhere else
- `git restore` it so it's just the submodule handle thingy
- `git submodule update --init` - does a clone but no build
- move contents of directory back again - should just be build files?

Or something like that?

quickbeam123:
Worked like a charm! Thanks!

PR #168, opened 2020-11-11
Integer arithmetic

joe-hauns:
This PR contains the following additions
# Command line options
- the options `-gve` can now be performed either `cautious` or `force`. When `force` is enabled the rule is used as a simplification rule, even if it violates the simpl ordering. in `cautious` mode the rule generates a new clause instead of simplifiying if the ordering would be violated.
- `[ -asg | --arithmetic_subterm_generalization ] [ off | cautious | force ]`, enables various arithmetic suberm generalization rules described in the attached paper.
- new modes for arithmetic evaluation `[ -ev | --evaluation ] [ simple | force | cautious ]` 
  - `simple` is our old evalution strategy
  - `force `, `cautious` the new evaluation strategy described in the attached paper
- `[-pum | --push_unary_minus ] [on | off]` enables disables the corresponding rule from the paper.
- `[-canc | --cancellation ] [ off | force | cautious ]` enables the corresponding rule in cautious or force mode
[paper_arith_simpl.pdf](https://github.com/vprover/vampire/files/5523609/paper_arith_simpl.pdf)
 
# Code
- macro `DBG` `DBGE` for printing debug output
- partial refactor of iterators (only `DECL_RETURN_TYPE` for closures was removed yet, and a new interface has been added. This will be cleaned up after this PR)
- `Kernel/Polynomial.hpp` data-structures `PolyNf` for terms in the normal form described in the paper
- `Lib/Perfect.hpp`, a generic class for perfect sharing of structures
- `Lib/Coproduct.hpp` a generic tagged union
- `Lib/Option.hpp` a generic datatype for representing a either a value or nothing
- `Lib/TypeList.hpp`some template-metaprogramming utilities used for `Lib/Coproduct.hpp`
- `Inferences/LfpRule.hpp` a rule that is the least fixed point of another rule
- `Kernel/BottomUpEvaluation.hpp` a template function to bottom-up evaluate an arbitrary structure that is traversable similar to a term. It can be used to bottom-up evaluate `TermList` as well as `PolyNf`
- some additional functions for datastructures in `Lib`
- Various things for making it easy to write nice unit tests. see `UnitTests/tEqualityResolution.cpp` , `UnitTests/tSyntaxSugar.cpp`, and `UnitTests/tGaussianElimination.cpp` for examples
- various minor additions 

(Sorry for the massive PR, for the future there will be more smaller ones. It got somehow out of hands ..)


selig:
I'm looking through. One comment is that this PR contains some orthogonal improvements to debugging code. Maybe this should be a separate PR but I think we can live with it here. But more generally, perhaps we could document some of this separately. Perhaps a top-level Debug helper file that gives hints on how to use the inbuilt debugging stuff.

Shell/Options.cpp
```diff
@@ -1987,6 +2024,20 @@ void Options::init()
...
     _introducedSymbolPrecedence.tag(OptionTag::SATURATION);
 
+    _evaluationMode = ChoiceOptionValue<EvaluationMode>("evaluation","ev",
+                                                        EvaluationMode::SIMPLE,
+                                                        {"simple","force","cautious"});
+    _evaluationMode.description=
+    "Choses the algorithm used to simplify interpreted integer, rational, and real terms. \
+                                 \
+    - simple: will only evaluate expressions built from interpreted constants only.\
+    - polynomial: will evaluate abstract expressions to a weak polynomial normal form. This is more powerful but may fail in some cases where the resulting polynomial is not strictly smaller than the initial one wrt. the simplification ordering. \
```
quickbeam123, line 2034: The description seems to be out of sync with the three values `{"simple","force","cautious"}`.

joe-hauns: Fixed all the option descriptions!

Api/Problem.hpp
```diff
@@ -361,7 +361,7 @@ class Problem
     friend class Problem;
     void validate() const;
 
-    void prepareOptionsReader(OptionsReader& rdr);
+    // void prepareOptionsReader(OptionsReader& rdr);
```
quickbeam123, line 364: In Api, this looks pretty inconsequential, but why comment it out?

joe-hauns: I think we did already talk about that, but in order to have i there publicly as well:
That was dead code, and instead of deleting it straight away i just commented it out.

Issue #157, opened 2020-10-28
Removing unused files, dead code.


MichaelRawson:
Some files in Vampire's tree appear to no longer be used. While this is not a problem in itself, it adds to the noise for new contributors and some dead code is still being compiled or even linked (see e.g. PR #153  to remove Lingeling sources).

After some discussion with others, I believe some of the following could be removed:

- [ ] `Saturation/Limits` are empty files.
- [ ] `Shell/SimplifyProver` isn't used anywhere - used to be support for the Simplify theorem prover's language
- [ ] `BFNT` and `BFNTMainLoop` is an old form of finite model building.
- [ ] Some of the scripts in `scripts` seem like they might no longer serve a purpose.
- [ ] `Kernel/MatchTag` is switched off. Seems to have been the result of an experiment that was never finished.

Credit to @ibnyusuf and @selig for identifying some of these. What I'd like from the team is help to identify more unused files or dead code, as well as rescuing files that need to be kept.

Eventually things that are still on the list will be removed by PR, although one of the many virtues of source control is the ability to reanimate code from beyond the grave. Thank you!

quickbeam123:
What about the whole API folder? 

There is also some mysterious looking ProofSimplifier in Shell.

MichaelRawson:
Very possibly @quickbeam123! I'm going to chat with Giles Monday about any politics I need to be aware of, then start deleting things in order of least-likely-to-be-wanted first.

ibnyusuf:
Inferences/InterpretedSimplifier is completed commented out. Perhaps it can be removed.

quickbeam123:
Some values in TimeCounterUnit most likely don't occur in the code anymore... Like TC_BDD_CLAUSIFICATION, etc. Could be removed as well!

quickbeam123:
There also some code, authored by Ioan, generally guarded by `#if GNUMP` and thus not compiled currently, which could perhaps be removed.

quickbeam123:
Got another one: CTFwSubsAndRes.*
- maybe useful if at some point want to revive code trees for subsumption, but currently not even compiled
- by the way IndexManager holds a reference to CodeTreeSubsumptionIndex via FW_SUBSUMPTION_CODE_TREE, but only CTFwSubsAndRes would ever request this index. 

So, transitively, CodeTreeSubsumptionIndex is also unused and maybe some other indices are formally alive but unused because of how IndexManager works.

MichaelRawson:
[...]

PR #170, opened 2020-11-11
remove InterpretedSimplifier, long dead

MichaelRawson:
See issue #157 - `InterpretedSimplifier` was commented out circa 2011 and not revived since.


selig:
joe-hauns will we use any of this?

MichaelRawson:
@joe-hauns ^^

joe-hauns:
I had a look at it at some point, and made some notes what's going on there, and I think it's not worth the effort trying to revive it.
The plans I have are somewhat similar to but more general to what's going on in there. So I think it's better to reimplement it with the more general thing in mind.

PR #174, opened 2020-11-12
Build Z3 in CI as well.

MichaelRawson:
Not ready yet. Opening PR to test.


MichaelRawson:
This is ready to go now.

Github Actions allows caching the results of build steps, so this PR builds the Z3 submodule, but also caches the result.

There is a slight hitch: caches are kept for 7 days, then if they're not used in that time they are discarded. Caches can only be accessed by children of the branch the cache was created from. Therefore, this also adds a scheduled build of `master` every day at UTC 00:00, along with the manual and PR builds. Perhaps this is not such a bad thing!

PR #176, opened 2020-11-12
add experimental LTO support to the CMake build

MichaelRawson:
Link-time optimisation ought to make a difference to Vampire's overall performance, but this is an unknown quantity and needs tweaking and testing. This adds a flag to the CMake build to allow experimentation with LTO. The build seems to work, but takes around a minute to link.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,9 +30,19 @@ endif()
 # statically link libraries
 option(BUILD_SHARED_LIBS "Build an executable dynamically linking to libraries instead of a statically-linked one (static not working on Mac)" ON)
 if (NOT BUILD_SHARED_LIBS)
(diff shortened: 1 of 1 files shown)


MichaelRawson:
Having looked a bit harder I think we want this. The build takes a while, but it doesn't seem to break anything: if we only use LTO for competitions and test the build well I don't think this matters.

Performance is encouraging: the executable becomes slightly larger (probably cross-compilation-unit inlining?) but you get an improvement in performance of ~5% in terms of time-to-proof. If people are interested what this means in "real" terms I can schedule a StarExec job.

In principle we can also LTO the Z3 link, but the Z3 LTO build is quite fussy and it doesn't seem to make that much difference.

PR #179, opened 2020-11-12
stop suppressing compiler warnings about exceptions

MichaelRawson:
`Debug/Assertion` has some code to suppress warnings about throwing exceptions where you shouldn't throw exceptions. This is hiding bugs: for example, in C++11 destructors are implicitly `noexcept` so failing an assertion doesn't throw the exception, but calls `std::terminate`.

This removes the code to suppress warnings. Cue warnings! In most cases these related to testing assertions in destructors, which as mentioned does not have the intended effect and so are just removed. If reviewers feel them critical we can discuss the best path to replacing them with something functional.

### DIFF
--- a/Debug/Assertion.hpp
+++ b/Debug/Assertion.hpp
@@ -25,25 +25,6 @@
 #ifndef __Assertion__
 #define __Assertion__
 
-#define __PUSH_DIAGNOSTICS(diag, ...) \
-    _Pragma("GCC diagnostic push") \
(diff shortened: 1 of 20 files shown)


quickbeam123:
As discussed on slack, what about leaving ASS's in destructors (and the ability use them there), but making them just do what they usually do (print stack trace, report-spider-fail, etc) and then abort rather than throwing?

PR #180, opened 2020-11-16
Unit tests

joe-hauns:
# Changes
- unit tests are now integrated with ctest. 
- CI workflow includes running unit tests now.
- old broken tests were removed

# How to manually run the unit tests?

Just run `ctest` in your build directory after building vampire with cmake:
```
mkdir cmake-build 
cd cmake-build
cmake -G "Unix Makefiles" ..
make -j 8

ctest --output-on-failure
```


MichaelRawson:
Looks good to me wrt build/system unit tests. :-)

A bit of a niggle is that the CMake build now has to re-build the same objects again, but slightly differently for unit tests because they don't have the same configuration. This means the from-scratch build takes longer. Also personally I'd like unit tests to fail if the build configuration breaks the test, but maybe people disagree?

I managed to make the build treat all objects equally and re-use object files. The work is in a branch called `michael-unit-test-cmake-fixes`: take a look and if you agree, merge it into here.

joe-hauns:
I had a look, and your CMakeLists looks much tidier! I find it also nice that Z3 is included in the tests iff it is included in the main build! So go ahead and merge it in here! 

The only thing I found nicer in what I wrote was the globbing for files tests cases using`UnitTests/t*.cpp` instead of listing all the tests explicitly in the `CMakeLists.txt`. Is there any technical reason why u favor the explicit listing?

MichaelRawson:
I don't think I have permission to touch your branch, but you should be able to `git merge` and then push?

Just to match the rest of the file; there's pros and cons to globbing. The pro is you don't have to list everything! The con is that if you add another test you need to remember to re-run CMake in order for the file to be picked up. This can be unintuitive: imagine I `git pull` to get the latest shiny unit tests, then run `make` in my already-configured build tree. If they are explicitly listed, CMake will re-run automatically because the `CMakeLists.txt` file changed. If not, the unit test just won't run for me until I remember to `cmake` again.

joe-hauns:
Merged and pushed it! 
I think uniformity is enough of an argument for not globbing.

MichaelRawson (approved):

PR #182, opened 2020-11-17
stop suppressing compiler warnings about exceptions, redux

MichaelRawson:
As discussed for #179, when an assertion is violated, this calls `System::terminateImmediately` and some other cleanup functions, then terminates, instead of raising an exception which may not be permitted in all contexts.


selig:
@quickbeam123 anything stopping us on this?

quickbeam123:
Just my promise that I will look at all other PRs after Ahmed's one is merged :)

PR #188, opened 2020-11-24
remove code commented out or not built

MichaelRawson:
See #157. Some files are commented out (permanently) in the build system: this removes them. Bonus: remove some redundant includes.

EDIT: also remove things guarded by `GNUMP` as it doesn't compile.


MichaelRawson:
@quickbeam123 - the commit sequence is designed to be cherry-picked from if people want some of these. :-)

selig:
I think we should look through these in a little detail.

For example, the files in the CTFwSubsAndRes commit were removed with the plan to "definitely" put them back when we fix code trees.

MichaelRawson:
@selig - agreed, hence the file-per-commit. CTFw has been out since 2016 so I figured we don't miss it, but by all means let me know if I can reanimate something.

MichaelRawson:
I think people prefer these removed one-by-one, and this needs rebasing anyway. Let's can it and do them individually.

PR #189, opened 2020-11-24
Remove "summary" from file headers

hzzv:
This PR removes the "summary" paragraph from the header of all files that contained it.


selig:
Thanks!


 @quickbeam123 please don't merge yet as I need to add a couple of files.

selig:
@hzzv  Could you post the script you used for this on Slack so that it can be applied to pull requests still using the old header.

MichaelRawson:
@selig - while you're at it, is there a reason for the blank line at the top of (all) source files?

selig:
@MichaelRawson I'm guessing this is a side-effect of the script I wrote to add the headers.

hzzv:
@selig Which Slack channel should I post the script to?

PR #190, opened 2020-11-24
remove unused files from Test

MichaelRawson:
See #157. Some of the files in `Test/` are apparently no longer reachable for tests. This removes them. @joe-hauns - if you're happy with these (I preserved things I think you want) perhaps we can also remove `vtest`?


joe-hauns:
Looks all sound! 

Concerning `vtest`: As it seems as the verdict of last meeting that we use `cmake` as our official build system and `make` only for workflows that are not yet ported/not easily portable to `cmake`, I think we can get rid of vtest, and only support unit testing via `ctest`! This will make a bunch of things in `UnitTesting.cpp` redundant, and lets us simplify defining unit tests and so on. 
I'll have a look at that tomorrow morning if noone objects to that!

joe-hauns (approved):

selig:
Sounds good to me... I don't think anybody uses vtest properly at the moment so replacing it with something else (and then encouraging people to use that) seems fine.

PR #191, opened 2020-11-25
Removing vtest

joe-hauns:
Resolves the bullet `vtest` in #157 .
- `vtest.cpp` removed, more minimalist main function implemented in `UnitTesting.cpp`
- `Test/UnitTesting.*` cleaned up
- `cmake` will build tests only in `Debug` mode, since assertions (our primary means of testing) are being compiled away in `Release` builds. If we want unit tests in debug build we could refactor the test cases to use some other macro than `ASS` that is still present in `Release` mode.


CMakeLists.txt
```diff
@@ -698,38 +705,40 @@ set(VAMPIRE_CASC_SOURCES
...
-    UnitTests/tRatioKeeper.cpp
-    UnitTests/tTwoVampires.cpp
-    UnitTests/tOptionConstraints.cpp
-    UnitTests/tSCCAnalyzer.cpp
-    UnitTests/tDHMultiset.cpp
-    UnitTests/tList.cpp
-    UnitTests/tStack.cpp
-)
-source_group(unit_tests FILES ${UNIT_TESTS})
+if(COMPILE_TESTS)
```
MichaelRawson, line 708: I don't think this section wants guarding - my understanding is that `source_group` is for IDE support.

MichaelRawson (changes requested):
Looks good - see comment inline but otherwise great. :+1:

joe-hauns:
Good shout! I did remove the guard.

CMakeLists.txt
joe-hauns: Good shout. Did remove the guard!

MichaelRawson (approved):

MichaelRawson:
@joe-hauns - as per discussion, perhaps you can add an option to temporarily disable unit tests so that people can hack on Vampire, iterate quickly and then turn them on again to finish? Something like:

```cmake
option(DISABLE_UNIT_TESTS "disable building unit tests _temporarily_" OFF)
if(DISABLE_UNIT_TESTS)
  set(COMPILE_TESTS OFF)
endif()
```

joe-hauns:
What would be aim of that? To bring down compile times?

MichaelRawson:
Yeah, see Ahmed's comments in Slack. Suppose you're implementing some feature and painstakingly debugging some file. Changing a single source file requires a re-compile of the source file (fine), re-linking Vampire (fine), and also re-linking each unit test (unavoidable, but can take a while). You'd like to temporarily disable unit tests, finish your feature and then turn them back on again.

On the other hand, we don't want to disable them by default or they'll never be used.

I think a happy medium is having an opt-out flag: this way you can turn it off temporarily, but there's some friction to not use the unit tests, and if you forget to run them the CI will and catch whatever bug you just introduced. What do you think?

joe-hauns:
Yeah I already noticed the long link times myself. I'm currently bundeling the unit tests into one executable (while keeping every test case as one cmake test). I hope this will bring down compile times. If not I'll add an option as suggested as well!

CMakeLists.txt
```diff
@@ -828,30 +828,40 @@ endif()
 # UNIT TESTING 
 ################################################################
```
MichaelRawson, line 830: I think this loop only exists to set `UNIT_ID_STR` and `UNIT_ID` and setup CTest. Could these be set from `__FILE__` in the source file instead? That way the whole unit testing build is just another target executable and a small loop to setup CTest.

MichaelRawson (commented):
Looking good, `vtest` lives again! I wonder if the build can be simplified even further by using the `__FILE__` builtin macro and trimming the extension instead of defining `UNIT_ID`. That way we might not even need the `t` prefix on files?

If not, no bother, we can always work out how to improve it later. :+1:

CMakeLists.txt
joe-hauns: I looked into that, and the issue is that `__FILE__` is a string literal, and we need an identifier specifying the test name, if we want to be able to write 
```
TEST_FUN(ident) { /*...*/}
```
If we'd accept to have 
```
TEST_FUN(ident, { /* ... */ })
```
instead we might be able to do with `__FILE__` only. But it requires patching old tests, and fiddling around with macro expansion .. I don't know if it's worth it.

MichaelRawson: Makes sense to me - never mind, can always be simplified later if we figure out how to do it! Good changes.

PR #195, opened 2020-11-27
remove TWL solver and its dependencies

MichaelRawson:
See #157. `TWLSolver` is reachable, but isn't the default and wasn't used in the most recent strategy schedule. It also allows removing and simplifying a lot of code. Additionally, the SAT solving interfaces can be simplified in future as there is no longer any requirement to interface to `TWL` (only Minisat and Z3 at present).


MichaelRawson:
Also removed the apparently-unwanted SAT mode here as it's somewhat related.

PR #196, opened 2020-11-30
Remove Global.cpp

MichaelRawson:
`Global.cpp` is a bit scary. This PR removes it and most of the scariness.

The toplevel file `Global.cpp` contains some global declarations that induce an ordering on static initialisation. C++ guarantees objects with static storage duration (`static` class members, globals) in the same compilation unit are initialised in the order they appear, but does __not__ guarantee anything about the order of initialisation for objects __across__ compilation units.

I note that `Global.cpp` does implement this initialisation correctly. However, it is surprising for newcomers, somewhat fragile and scatters class members far from their parents. It turns out that this constrained initialisation is not actually required and therefore `Global.cpp` can meet its demise.

There are two commits: the first is trivial, the second less so.

### Commit 1
`Global.cpp` contains some static class declarations that are POD (and therefore _constant initialised_, no ordering concerns here), and some that are simply not used any more. The PODs are moved into their respective class files. The unused fields are removed altogether, along with the code that was no longer using them (notably `Lib::XML`). This change should be completely safe.

### Commit 2
This is a bit more interesting and hazardous. There are a few remaining items in `Global.cpp`:
1. The `s_terminationHandlers` field from `Lib::System`. This is a prioritised list of termination handlers that are called before the process terminates to clean up resources, currently `Semaphore`s and `SyncPipe`s.
1. The global `env` object.
1. A struct `__Lib_System_Init_Invoker` whose sole purpose is to be immediately declared, causing its constructor to run. The constructor calls `Lib::System::onInitialization()`, so that effectively this ensures that `onInitialization` is run after the previous 2 objects are initialised.

`s_terminationHandlers` has a twin for "initialisation handlers". However, initialisation handlers are implemented as function-static variable (not class-static) so that it acts as a singleton, initialised on the first call to the function. `s_terminationHandlers` is not implemented this way because it would be destructed before the `System::onTermination` function uses it, registered with `atexit`. However, this can be solved: C++ guarantees that as long as an object is initialised _before_ a call to `atexit(f)`, it will be destructed _after_ the eventual call to `f` on termination. `s_terminationHandlers` becomes a method in `System` hiding a static variable. Question: should `System::onTermination` be private?

`env` is declared here so that it starts after the creation of `s_terminationHandlers` and before `onInitialization`. `s_terminationHandlers` have now been dealt with, and `onInitialization` does not require the existence of `env` (and is removed anyway). `env` no longer needs to be here and is moved to its natural home, `Environment`.

`onInitialized` runs callbacks registered previously in the startup sequence. However, nothing actually registers these, so the whole machinery can just go.

Initial testing suggests that the changes have not introduced any obvious bugs. While the change is not risk-free, I feel this one is worth it in the long run.


quickbeam123:
Hi Michael, do you still remember what Commit 2 actually did? The description above seems to describe the situation before the the commit, but is not so clear about what steps have been taken... Could we briefly discuss?

MichaelRawson:
@quickbeam123 - commit 2 removes the remaining non-trivial stuff in Globals, but some of these changes definitely warrant discussion! I'm free all day tomorrow if you want to ping me on skype or something, or we can discuss here - IIRC it's:
1. make `s_terminationHandlers` behave like the initialisation handlers - this is safe (I think) because of C++'s guarantees about `atexit` and object destruction
2. no more order dependencies for `env` so that can move to `Environment`
3. nothing actually registers initialisation handlers

PR #197, opened 2020-12-01
C++14

MichaelRawson:
This makes the necessary changes to compile with the C++14 standard. No obvious damage is done: the compile takes around the same time on my machine (GCC and Clang), unit tests pass and running a few random problems doesn't crash Vampire.

Pros:
* all the work the standards people do to make our lives easier and our binaries faster
* working toward more recent standards like C++17 or even '20?!
* I get new toys for the CAPS project, notably `std::shared_timed_mutex`
* modern tooling can work better on modern standards
* it's possible that the standard library now provides things we implement ourselves so Vampire can shrink a little

Cons:
* may inadvertently introduce obscure bugs
* people with older build machines may find it difficult to build Vampire.
  * partial rebuttal: cross-compiling is possible

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,8 +4,8 @@
 cmake_minimum_required (VERSION 3.12.0)
 project(Vampire)
 
-# require the compiler to use C++11
-set(CMAKE_CXX_STANDARD 11)
+# require the compiler to use C++14
+set(CMAKE_CXX_STANDARD 14)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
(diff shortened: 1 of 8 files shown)


JakobR:
Looks good to me. Maybe we should add the `std::` prefix to existing uses of `make_unique` and `exchange`?

MichaelRawson:
@JakobR - good idea - somehow the existing calls to `Lib::make_unique` got resolved to the right thing (?!). Everything's `std::` prefixed now. :-)

JakobR:
Yes, no idea how that still compiled, but looks good now :-)

JakobR (approved):

PR #202, opened 2020-12-14
Fingerprint indices experiment.

MichaelRawson:
This branch implements the _fingerprint indexing_ technique described in [_Fingerprint Indexing for Paramodulation and Rewriting_](http://wwwlehre.dhbw-stuttgart.de/~sschulz/PAPERS/schulz_fp-index.pdf). Fingerprint indexing is a simple imperfect indexing technique for first-order unification and matching.

It's possible that Vampire could benefit from this technique, alongside or in place of existing index structures. A quick hack is promising: replacing a single substitution-tree index with a fingerprint index gives a slight performance win, even though the fingerprint index is not well-tuned.

This should not be merged and is here to gauge interest and solicit feedback from more experienced team members. There are at least some bugs. TODO:
- [x] implement instances and generalisations as well as unification
- [ ] implement literal indexing
- [ ] guard against, or implement support for, theories/HOL
- [ ] test on problems
- [ ] cleanup, format, test, document etc.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -311,6 +311,7 @@ set(VAMPIRE_INDEXING_SOURCES
     Indexing/CodeTree.cpp
     Indexing/CodeTreeInterfaces.cpp
 #    Indexing/FormulaIndex.cpp
+    Indexing/FingerprintIndex.cpp
     Indexing/GroundingIndex.cpp
     Indexing/Index.cpp
     Indexing/IndexManager.cpp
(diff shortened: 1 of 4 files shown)


MichaelRawson:
This needs some love and should probably be part of investigating non-perfect indexing more generally. Closing, branch can persist if we want to revive it.

ibnyusuf:
In case we ever want to try this again, I have made some imporvements as part of the delayed unification work. I have:

- Added a `LiteralFingerprintIndex` and implemented `getUnifications` for it (but not `getGeneralisations` or `getInstantiations`)
- Fixed a subtle bug which meant that in some cases, terms were put in the wrong bucket and not returned by a query when they should have been.

For fingerprint indices to be truely useful, some further imporvements are required:

- `TermFingerprintIndex` needs to be modified to support the indexing of variables. Currently it does not support this.
- Further code sharing between `TermFingerprintIndex` and `LiteralFingerprintIndex` should be achieved via the use of templates. 
- The retrieval of literal instantiations and generalisations needs to be implemented.

Issue #198, opened 2020-12-01
Proposal: Automatic detection of input syntax


gleiss:
## Motivation
Vampire supports input written in the TPTP- and the SMT-LIB2-syntax. Currently, Vampire interprets its input by default using TPTP, which can be changed to SMT-LIB2 with the option `--input_syntax smtlib2`.

While the current behaviour is nice if you prefer TPTP, it is not optimal for users prefering SMT-LIB2, due to the following reasons:
- it is cumbersome for a user to type `--input_syntax smtlib2` all the time.
- the user needs to know that it's possible to change the input syntax to SMT-LIB2 and that it is possible to do so using `--input_syntax smtlib2`. Note that this might in particular be a problem for new (industrial) users of Vampire.

If you are using TPTP, you might ask why others don't do the same. There are several reasons for that:
- SMT-LIB2 is the de facto standard in industrial applications of theorem provers, so industrial users might already be familiar with SMT-LIB2, but not with TPTP.
- Some features of Vampire are only accessible using SMT-LIB2 syntax, for instance native term algebras or recursive function definitions, so if you need these features, you have to use SMT-LIB.
- (arguable reasons for why SMT-LIB2 is a better input format than TPTP in some context ;) )

## Proposal
Here is a proposal to improve the current situation using an automatic detection of the input syntax (including some tips from @quickbeam123):
- change the values for `--input_syntax` to `auto`, `tptp`, and `smtlib2`, with default value `auto`
- with the `auto` option value, Vampire uses the file extension of the input file to decide which input syntax is used. If the ending is one of `smt`, `smt2`, `smtlib`, `smtlib2`, then Vampire uses SMT-LIB2 syntax, otherwise it uses TPTP syntax.
- with the option value `tptp` resp. `smtlib2`, Vampire uses TPTP resp. SMT-LIB2 syntax (independently from the file extension)

Additionally, for the `auto` option value, Vampire could output a message, which informs the user about the detected language, e.g. `Assuming SMT-LIB2 input syntax, since file extension is "smt2"`, or `Assuming TPTP input syntax, since file extension is not one of "smt", "smt2, "smtlib", "smtlib2"`. We should only print such a message in the modes which are allowed to be chatty.

Some thoughts on the proposed solution:
- The solution tries to cover the most relevant cases for SMT-LIB2 users, but at the same time tries to preserve the current behaviour as much as possible.
[...]

PR #204, opened 2020-12-17
Initial Attempt at Auto-detecting Input langauge

selig:
In issue [198](https://github.com/vprover/vampire/issues/198) Bernhard made the very sensible suggestion to have an auto-detect mode for the input syntax.

The proposal was to do this solely based on filename. This will probably cover 99% of the use cases as it is a defacto standard in SMT to use smt or smt2 as file extensions. 

However, there was the counterproposal to look inside the file to see how we should parse it.

This pull request is an initial attempt to combine the two. We first check the filename to work out which way to parse the file initially but then fallback to the other parser (we now only have two) if this fails.

This PR is not ready to merge as I haven't tested on more than a few examples. 

It currently (perhaps naively) assumes that all parser errors are being raised as ParseExceptions and this needs to be checked.

The code in UIHelper could probably be refactored a little. I welcome initial thoughts.

### DIFF
--- a/Shell/Options.cpp
+++ b/Shell/Options.cpp
@@ -343,9 +343,9 @@ void Options::init()
 
     _inputSyntax= ChoiceOptionValue<InputSyntax>("input_syntax","",
                                                  //in case we compile vampire with bpa, then the default input syntax is smtlib
-                                                 InputSyntax::TPTP,
+                                                 InputSyntax::AUTO,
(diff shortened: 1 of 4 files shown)


selig:
Hi @quickbeam123 I think this is a good compromise on the auto-detection of input language. What do you think?

quickbeam123:
Sorry for the long silence! I like the idea! I just tested it and maybe there is a small problem still (see slack).

PR #206, opened 2020-12-18
remove generalizationExists, never called

MichaelRawson:
See #157. `TermIndexingStructure::generalizationExists` is never called so can be removed safely, along with all implementations.


quickbeam123:
Not sure I agree. An index provides a certain interface for logical operations on objects stored therein. The fact we at the moment don't use a certain feature of a library (here the index lib) does not mean it couldn't be useful in the future.

MichaelRawson:
Fair enough, this is why we have review! I'm not sure any of the current implementations do anything that isn't materially `getGeneralizations().hasNext()` so maybe it could be moved into the base class for optional overriding, but that's another change. Closing.

PR #209, opened 2021-01-25
Add integer induction with different base cases

hzzv:
Overview of the changes:
- Main goal is that integer induction uses not only 0 as the base case, but also any ground term G, if there is an active clause G<=inductionTerm or G<inductionTerm (for induction with inductive step +1), or G>=inductionTerm or G>inductionTerm (for induction with inductive step -1).
- In order to do this, integer induction can be applied either (a) when the premise which comes into the Induction inference class is the induction literal (switched on by -mik one), or (b) when the premise is a unit clause which is ground integer comparison (switched on by -mik two).
In the former case, if the induction literal is ~L[indTerm], then we fetch all comparisons ~indTerm<b1, b2<indTerm, ~b3<indTerm and indTerm<b4, and for each of them instantiate the induction hypothesis with using b1 or b2 or b3 or b4 as the base case (with induction step +1 in case of b1 and b2, and -1 in case of b3, b4).
In the latter case, let's say the comparison is ~term2<term1 (meaning term1<=term2). Then, we fetch all active literals which are ground and contain term2, and for each of them instantiate an induction hypothesis using term1 as the base case and +1 as the inductive step. Similarly, we fetch all active literals which are ground and contain term1, and for each of them instantiate an induction hypothesis using term2 as the base case and -1 as the inductive step. Analogously for comparison term1<term2.
- Whether the induction was triggered by the induction literal Li or the integer comparison Lc (containing the base case b), the resulting clausified induction hypothesis is resolved against both Li and Lc. The conclusion of induction hypothesis is: !y. (comp(y, b)  -->  Li[y]), where 'comp' is the same comparison as used in Lc. (E.g., if Lc is b<indTerm, then the conclusion of IH is !y.(b<y  -->  Li[y]).)
- While struct induction uses notDone(...) to reduce redundancy, integer induction needs a different method, since redundancy depends not only on the combination of induction literal/induction term, but also on the base case, direction of the induction step (+1/-1) and whether the base case was default (0) or came from an integer comparison. To this end, I added new method notDoneInt(...).
- To be able to fetch corresponding integer comparisons for a given induction term, and corresponding induction literals for a given induction term, I added 2 new indexes (one LiteralIndex, one TermIndex), which index only literals/terms suitable for induction. The indexes themselves only check whether clauses/literals/terms are suitable for induction, not whether induction is on (that is done in the Induction class).
- I also added a new class InductionHelper into a separate file in the Inferences/ directory. Its methods (e.g., checking whether a literal can be used as induction literal) are used both in Induction.cpp and in the Indexes.
- Additionally, integer induction cannot be applied on indTerm in literal L in case that the top-most functor of L is <, and indTerm is  one of <'s arguments. (This would mean we're trying to prove indTerm < someGroundTerm (or similar) by induction on indTerm.)


Indexing/TermIndex.cpp
```diff
@@ -427,33 +457,33 @@ void HeuristicInstantiationIndex::handleClause(Clause* c, bool adding)
         /*cout << "lhs is " + lit->nthArgument(0)->toString() << endl;
         cout << "arg " + leftArgs[i].toString() << endl;*/
         cout << "inserting " + lambdaTerm->toString() << endl;
-        cout << "inserting " + combTerm.toString() << endl;          
+        cout << "inserting " + combTerm.toString() << endl;
```
selig, line 460: leftover debug printing?

selig: But this isn't your code... I don't think it should be here though.

selig:
Do you have any sample test files that you've been using to test that it works properly? E.g. small files that the options in this PR should easily be able to solve.

I'd like to start adding small test files with PRs to help us do regression testing.

Indexing/TermIndex.cpp
hzzv: Should I remove it completely or just comment it out? (There's also similar debug printing in lines 475-478.)

selig: I think the least disruptive thing is to comment these print statements out. Thanks for dealing with this here - it must have been missed in a previous PR.

hzzv: Done.

hzzv:
> Do you have any sample test files that you've been using to test that it works properly? E.g. small files that the options in this PR should easily be able to solve.

Here are some small problems:

https://github.com/vprover/inductive_benchmarks/blob/master/benchmarks/int/val/smt2/declared_axgeq3_conjgeq3_val3_inter.smt2 solved using `-ind int`

https://github.com/vprover/inductive_benchmarks/blob/master/benchmarks/int/power/smt2/1_all_1.smt2 solved using `--forced_options "drc=off:iik=two:ind=int:sos=theory:sstl=1:to=lpo"`

https://github.com/vprover/inductive_benchmarks/blob/master/benchmarks/int/power/smt2/0_all_0.smt2 solved using `--forced_options "drc=off:iik=two:ind=int:to=lpo"`

https://github.com/vprover/inductive_benchmarks/blob/master/benchmarks/int/sum/smt2/y_geq_0.smt2 solved using `--forced_options "ind=int:sa=discount:sos=theory:sstl=1:to=lpo"`

PR #211, opened 2021-02-22
SMT-LIB 2.6 new keywords

mezpusz:
This pull request extends our current SMT-LIB parsing capabilities with some useful keywords based on [this documentation](http://smtlib.cs.uiowa.edu/papers/smt-lib-reference-v2.6-r2017-07-18.pdf):
- A `match` term is a case distinction of n cases on a term algebra variable. Each case is a pair `(p t)` where `p` (the pattern) is either a constructor term, a variable, or an underscore. The term `p` can declare new variables, so we create a new variable scope for each such pair before parsing and add the new variables. E.g.: `(match x (nil y) ((cons u v) v))` contains two cases `nil` and `(cons u v)`, where the second branch declares `u` and `v`. Since the standard does not require the variables to be fresh, just to be distinct, we create a new scope for each case, just like in case of quantifier blocks.
The case distinction must be exhaustive and non-overlapping (this latter is in fact not required by the standard). The exhaustiveness is ensured by either listing all constructor terms or by adding at most one 'else' branch where `p` is a variable or an underscore. An underscore means that the variable is not used in the block and can be ignored. In accord with the standard, we disallow nested constructor terms, e.g. a pattern `(cons u (cons v w))` is not allowed, one needs to first add `(cons u u0)` to an outer case distinction and then make a case distinction on `u0` which contains `(cons v w)`.
If there is an else branch with an explicit variable `x`, we create the missing n ctor terms and substitute each for `x` into n copies of the else body.
- A `define-fun-rec` block is the same as a `define-fun` block except we declare the function before parsing its body, therefore the function can call itself.
- A `declare-datatype` is a singular datatype declaration. E.g.:
`(declare-datatype list ((nil) (cons (head nat) (tail (list nat)))))`
The first argument is the name of the datatype, the second is the list of its ctors. The keyword `par` is reserved for parametric declarations, if the ctor list contains an atom `par`, we report error (since these are not supported).

The `match` blocks are converted to special terms `$match(matched_term,p1,t1,...,pn,tn)` which are then eliminated in either FOOLElimination or NewCNF:
- In FOOLElimination, we declare a new function of arity m for each match block with m free variables, then replace the match with the function and the concrete free variables as its arguments. We also add a clause for each case. So a match occurrence:
`L[$match(matched_term,p1,t1,...,pn,tn)] v C`
becomes
`L[mG(x1,...,xm)] v C, matched_term != p1 v mG(x1,...,xm) = t1, ..., matched_term != pn v mG(x1,...,xm) = tn`
- In NewCNF, we parse any subformula/subterm similarly for matches as for ITEs TODO

### DIFF
--- a/Kernel/Term.cpp
+++ b/Kernel/Term.cpp
@@ -1394,9 +1394,17 @@ bool Term::isBoolean() const {
         return false;
       case SF_ITE:
       case SF_LET:
-      case SF_LET_TUPLE:
-      case SF_MATCH: {
+      case SF_LET_TUPLE: {
         const TermList *ts = term->nthArgument(0);
+        if (!ts->isTerm()) {
+          return false;
+        } else {
+          term = ts->term();
+          break;
+        }
+      }
+      case SF_MATCH: {
+        const TermList *ts = term->nthArgument(2);
         if (!ts->isTerm()) {
           return false;
         } else {
--- a/Kernel/Term.cpp
+++ b/Kernel/Term.cpp
@@ -1130,11 +1130,11 @@ Term* Term::createTuple(Term* tupleTerm) {
   return s;
 }
 
-Term* Term::createMatch(TermList sort, TermList matchedSort, unsigned int arity, TermList* elements) {
+Term *Term::createMatch(TermList sort, TermList matchedSort, unsigned int arity, TermList *elements) {
   CALL("Term::createMatch");
-  Term* s = new(arity, sizeof(SpecialTermData)) Term;
+  Term *s = new (arity, sizeof(SpecialTermData)) Term;
   s->makeSymbol(SF_MATCH, arity);
-  TermList* ss = s->args();
+  TermList *ss = s->args();
   s->getSpecialData()->_matchData.sort = sort;
   s->getSpecialData()->_matchData.matchedSort = matchedSort;
 
--- a/Kernel/Term.cpp
+++ b/Kernel/Term.cpp
@@ -564,6 +564,10 @@ vstring Term::headToString() const
         varList += "]";        
         return "(^" + varList + " : (" + lambdaExp.toString() + "))";
       }
+      case Term::SF_MATCH: {
+        // we simply let the arguments be written out
+        return "$match(";
+      }
       default:
         ASSERTION_VIOLATION;
     }
@@ -1126,6 +1130,24 @@ Term* Term::createTuple(Term* tupleTerm) {
   return s;
 }
 
+Term* Term::createMatch(TermList sort, TermList matchedSort, unsigned int arity, TermList* elements) {
+  CALL("Term::createMatch");
+  Term* s = new(arity, sizeof(SpecialTermData)) Term;
+  s->makeSymbol(SF_MATCH, arity);
+  TermList* ss = s->args();
+  s->getSpecialData()->_matchData.sort = sort;
+  s->getSpecialData()->_matchData.matchedSort = matchedSort;
+
+  for (unsigned i = 0; i < arity; i++) {
+    ASS(!elements[i].isEmpty());
+    *ss = elements[i];
+    ss = ss->next();
+  }
+  ASS(ss->isEmpty());
+
+  return s;
+}
+
 /** Create a new complex term, copy from @b t its function symbol and arity.
  *  Initialize its arguments by a dummy special variable.
  */
@@ -1372,7 +1394,8 @@ bool Term::isBoolean() const {
         return false;
       case SF_ITE:
       case SF_LET:
-      case SF_LET_TUPLE: {
+      case SF_LET_TUPLE:
+      case SF_MATCH: {
         const TermList *ts = term->nthArgument(0);
         if (!ts->isTerm()) {
           return false;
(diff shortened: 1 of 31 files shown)


selig:
I'm assuming that this implementation doesn't contradict the SMT-LIB 2.6 standard wrt these keywords. I'm interested in this comment "In order to align with CVC4" - I infer that CVC4 don't fully conform to the standard?

mezpusz:
> I'm assuming that this implementation doesn't contradict the SMT-LIB 2.6 standard wrt these keywords. I'm interested in this comment "In order to align with CVC4" - I infer that CVC4 don't fully conform to the standard?

I used this documentation when making the changes:
http://smtlib.cs.uiowa.edu/papers/smt-lib-reference-v2.6-r2017-07-18.pdf
Actually, looking at the documenation once again, the comment about CVC4 is misleading since the standard does not allow nested constructors anyways, so I will rewrite that.

The pull request itself is not yet ready, I converted it to draft to signify that. But discussion about some questions like what SMT-LIB standard we want to align with would be good.

Parse/SMTLIB2.cpp
```diff
@@ -160,6 +162,19 @@ void SMTLIB2::readBenchmark(LExprList* bench)
       continue;
     }
 
+    // TODO(mhajdu): is it possible that codatatypes are
```
mezpusz, line 165: I couldn't find 'declare-codatatype(s)' in the 2.6 documentation but the plural keyword is parsed [here](https://github.com/vprover/vampire/pull/211/files#diff-0f0f75c3f4c1c2f5fc5b277a47a0393152567d1a4f14eb32e99564fd598c2dcfR189). So where does this come from?

simonr89: This is not an official SMTLIB keyword, but co-inductive data types are a fairly natural extension. The syntax of the declaration would be exactly similar, only the semantics would change (in particular w.r.t. circular terms). CVC4 supports them, and there was some work for Vampire but it never made it to master, so you can safely ignore this keyword.

mezpusz: Thanks for the clarification. Then, I won't parse the 'co' counterpart now but leave the last argument `codatatype` of `readDeclareDatatype` there to be consistent with the plural `readDeclareDatatypes`.

quickbeam123: Wasn't codatatypes a pull request at some ancient time?

simonr89: Yes there is a 3 year old pull request that provides proper handling of codatatypes, plus some minor improvements for datatypes too (e.g., proper handling of cyclic terms in nested datatypes). The code has diverged a bit since then obviously, but if anybody is interested in reviewing that pull request, I could find some time to help.

selig:
Is there a set of input problems that can be used to test these new features? E.g. that could be used for regression testing to make sure we don't break things in the future.

mezpusz:
> Is there a set of input problems that can be used to test these new features? E.g. that could be used for regression testing to make sure we don't break things in the future.

We have the [inductive benchmark set](https://github.com/vprover/inductive_benchmarks) of our own which already contains a lot of these function definitions under the `benchmarks/dty/*/smt2_rec/` folders.

I also tested it by hand-converting the UFDTLIA set of SMT-LIB which only contained declare-funs + axioms before. This has a much greater variety of function definitions. I only have this locally but we could create our own fork. Whether the SMT-LIB community will eventually want to have the converted ones is not known to me.

Also, the [TIP set](https://github.com/tip-org/benchmarks) contains many problems all in this format but unfortunately the datatypes in them are mostly parametric which we don't yet support in the SMTLIB parser. A question is whether we will have support for parametric types in the near future.

selig:
That's great, as I will discuss this afternoon, my aim is to try and include a small number of test problems in checks/Problems to be included in our CI tests. So we'll want to select a few representative problems.

Let's discuss parametric types this afternoon (will you attend the seminar?) Could you point me to an interesting file in the TIP benchmarks that we could use for our dicussion?

mezpusz:
Yes, that would be great. I'll also attend the seminar. I think this problem containing parametric lists is sufficient for us:
https://github.com/tip-org/benchmarks/blob/master/benchmarks/isaplanner/prop_01.smt2
Or here's another one that has multiple type parameters in some blocks:
https://github.com/tip-org/benchmarks/blob/master/benchmarks/tip2015/list_assoc.smt2

Parse/SMTLIB2.cpp
mezpusz: I removed the `codatatype` argument for now.

Kernel/Term.cpp
```diff
@@ -1381,6 +1403,15 @@ bool Term::isBoolean() const {
           break;
         }
       }
+      case SF_MATCH: {
+        const TermList *ts = term->nthArgument(2);
```
quickbeam123, line 1407: Couldn't this crash for a match over the empty type?

mezpusz: As per the 2.6 standard and [this](https://github.com/vprover/vampire/blob/mhajdu-smtlib2-new-keywords/Parse/SMTLIB2.cpp#L1018) and [this](https://github.com/vprover/vampire/blob/mhajdu-smtlib2-new-keywords/Parse/SMTLIB2.cpp#L1108) line, declaring an empty type shouldn't be possible and we also add missing ctors when some are covered with a variable pattern.

quickbeam123: Thank you for the clarification!

PR #212, opened 2021-03-04
using z3-4.8.10

joe-hauns:
The current version of z3, `z3-4.8.9` yields segmentation faults, when datatypes are used with push/pop semantics. These features are necessary to implement theory instantiation with generalisation on datatypes though therefore I propose to update to the newer z3 version. 

Checks ran to verify that the new z3 version does not mess up proof search
- ran `-mode portofolio -sched smtcomp -t 120` over 10k randomly selected smtlib problems 
    => no assertion violations, no performance difference
- ran `-mode portofolio -sched smtcomp -t 120` over 50k randomly selected smtlib problems in Release mode
     => minor performance gain -10/+20
     => some segmentation faults from the finite model builds (should be unrelated to this changes since they r present in master as well)


quickbeam123:
We still have issues in FMB? I thought I fixed them. Could you please share some of the master runs that fail there?

joe-hauns:
Yes. I'm digging into the issue atm. 
I think the issue is an out of bounds access to `cnames` in `FiniteModelMultiSorted::toString`.
The configuration to reproduce it is:
```
vampire --decode fmb+10_1_av=off:fde=unused:ile=on:irw=on:lcm=predicate:lma=on:nm=16:nwc=1.7:sos=all:sp=reverse_arity_2 --input_syntax smtlib2 $smtlib/AUFLIA/20170829-Rodi
n/smt8675261522418160957.smt2
```
In release mode there's a segfault, while in debug mode vampire immediately terminates with exit status 4. I'm not sure what this status means though.

joe-hauns:
*compiling with `clang-1100.0.33.17`

quickbeam123:
I am finding it hard to see what are the commits actually changing... (I know this most likely simply because I don't know how submodules work on git(hub), but could you please quickly explain?)

Isn't there a textual info somewhere in our repo stating which version of the submodule we prefer building against? Confusingly, I don't see this in .gitmodules

joe-hauns:
Well the thing is that I did not (manually) change any files. I just `cd`ed into the `z3` directory, checked out the commit that is tagged with `z3-4.8.10`, `cd`ed back to the vampire dir, and then commited the `z3` directory.

So all the changes are performed by git somewhere in its hidden file structure I think .. But tbh I don't really know how git submodules work, but just followed stackoverflow.

PR #213, opened 2021-03-24
Fix multicore proof printing

ibnyusuf:
This pull requests attempts to solve an issue related to proof printing when running with multiple cores. It is possible that whilst a child process is printing a proof, another child interrupts the proof with other info (as only further proofs are locked out, not all print statements).

The attempted solution is to print the proof to a file and then, once all child processes have been killed, print the file to `env.out()` and delete the file. There is also the option to leave the proof in the file.

### DIFF
--- a/CASC/PortfolioMode.cpp
+++ b/CASC/PortfolioMode.cpp
@@ -559,6 +559,11 @@ void PortfolioMode::runSlice(Options& strategyOpt)
     } else {
       UIHelper::outputResult(output);
     }
+    if (env.options->printProofToFile() && outputAllowed()) {
+      env.beginOutput();
+      addCommentSignForSZS(env.out()) << "Proof written to " << fname << endl;
+      env.endOutput();
+    }
   }
   else{
     /*
--- a/CASC/PortfolioMode.cpp
+++ b/CASC/PortfolioMode.cpp
@@ -387,13 +387,13 @@ bool PortfolioMode::runSchedule(Schedule& schedule)
     vstring fname = env.options->problemName() + "-vampire.proof";    
     BYPASSING_ALLOCATOR; 
     
-    ScopedPtr<ifstream> input(new ifstream(fname.c_str()));
+    ifstream input(fname.c_str());
 
-    bool openSucceeded = !input->fail();
+    bool openSucceeded = !input.fail();
 
     if (openSucceeded) {
       env.beginOutput();
-      env.out() << input->rdbuf();
+      env.out() << input.rdbuf();
       env.endOutput();
     }
 
--- a/CASC/PortfolioMode.cpp
+++ b/CASC/PortfolioMode.cpp
@@ -398,6 +398,12 @@ bool PortfolioMode::runSchedule(Schedule& schedule)
       env.beginOutput();
       env.out() << input.rdbuf();
       env.endOutput();
+    } else {
+      if (outputAllowed()) {
+        env.beginOutput();
+        addCommentSignForSZS(env.out()) << "Failed to restore proof from tempfile " << outputFileName << endl;
+        env.endOutput();
+      }
     }
 
     //If for some reason, the proof could not be opened
--- a/CASC/PortfolioMode.cpp
+++ b/CASC/PortfolioMode.cpp
@@ -30,6 +30,8 @@
 #include "Shell/TheoryFinder.hpp"
 
 #include <unistd.h>
+#include <fstream>
+#include <stdio.h>
 
 #include "Saturation/ProvingHelper.hpp"
 
@@ -378,7 +380,34 @@ bool PortfolioMode::runSchedule(Schedule& schedule)
   PortfolioSliceExecutor executor(this);
   ScheduleExecutor sched(&policy, &executor);
 
-  return sched.run(schedule);
+  bool result = sched.run(schedule);
+
+  //All children have been killed. Now safe to print proof
+  if(result && !env.options->printProofToFile()){
+    vstring fname = env.options->problemName() + "-vampire.proof";    
+    BYPASSING_ALLOCATOR; 
+    
+    istream* input=new ifstream(fname.c_str());
+
+    bool openSucceeded = !input->fail();
+
+    if (openSucceeded) {
+      env.beginOutput();
+      env.out() << input->rdbuf();
+      env.endOutput();
+    }
+
+    delete static_cast<ifstream*>(input);
+    input=0;
+
+    //If for some reason, the proof could not be opened
+    //we don't delete the proof file
+    if(openSucceeded){
+      remove(fname.c_str()); 
+    }
+  }
+
+  return result;
 }
 
 /**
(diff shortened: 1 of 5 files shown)


CASC/PortfolioMode.cpp
```diff
@@ -378,7 +380,34 @@ bool PortfolioMode::runSchedule(Schedule& schedule)
...
+
+    bool openSucceeded = !input->fail();
+
+    if (openSucceeded) {
+      env.beginOutput();
+      env.out() << input->rdbuf();
+      env.endOutput();
+    }
+
+    delete static_cast<ifstream*>(input);
```
quickbeam123, line 400: Is the static cast really necessary? I would expect polymorphism to take care of this. 

Also, can't you declare `input` as `ifstream*` in the first place?

Or even: Can't you declare `input` as `ifstream` and rely on RAII for the cleanup?

CASC/PortfolioMode.cpp
```diff
@@ -378,7 +380,34 @@ bool PortfolioMode::runSchedule(Schedule& schedule)
   PortfolioSliceExecutor executor(this);
   ScheduleExecutor sched(&policy, &executor);
 
-  return sched.run(schedule);
+  bool result = sched.run(schedule);
+
+  //All children have been killed. Now safe to print proof
+  if(result && !env.options->printProofToFile()){
+    vstring fname = env.options->problemName() + "-vampire.proof";
```
quickbeam123, line 387: Just an idea: this looks like the proof is to be written to a place where the benchmark lies. This could often be a read only location...

Maybe we should force the user to specify the output location whenever `printProofToFile` is true (making the option a string directly). And maybe we should use the temp folder or at least the current vampire's directory (and a name which contains the process's pid) for when it it false (and we still need a file, like this).

CASC/PortfolioMode.cpp
ibnyusuf: I copied this code from `UIHelper::getIputProblem`

```
  istream* input;
  if (inputFile=="") {
    input=&cin;
  } else {
    // CAREFUL: this might not be enough if the ifstream (re)allocates while being operated
    BYPASSING_ALLOCATOR; 
    
    input=new ifstream(inputFile.c_str());
    if (input->fail()) {
      USER_ERROR("Cannot open problem file: "+inputFile);
    }
  }
```
But on consideration, I think the existing code is suboptimal. I have refactored to declare as `ifstream` and use RAll

CASC/PortfolioMode.cpp
ibnyusuf: Update: the user is now given the option to specify the output location. If they leave it blank, the output location is the directory from which the run command happens. Likewise, this is the directory used when `printProofToFile` is false.

quickbeam123:
There is still an issue of multiple vampires (running from the same directory) solving the same problem (or, more generally, solving different problems but of the same filename, not necessarily the same path).

I call for a general discussion whether we should not strive to solve this as well..

ibnyusuf:
Is the problem that someone might try to simultaneously run two Vampires in portfolio mode from the same location on different problems which happen to have the same name? I think that such an occurrence is likely to be rare enough that we don't need to handle it, but it would certainly be good to have other opinions as well.

quickbeam123:
I can easily imagine running two versions of vampire at the same time on the same benchmark set in parallel. With a network file system, this does not even need to be initiated from the same machine ...

quickbeam123:
(Ahmed, you should "try hard to break it" and only be happy when you fail. Not "to try hard to defend as sufficient what you currently have" ;))

quickbeam123:
On a more constructive note, I think we could solve this robustly with [`tmpfile`](https://en.cppreference.com/w/cpp/io/c/tmpfile) First open it in master, then the first successful child writes the proof to it, finally master rewinds and prints the content. (and the file gets even deleted automatically)

ibnyusuf:
Sounds like a good idea. Does that mean we remove the option to save to file or we only make use of the tmpfile when the save to file option has not been selected?

MichaelRawson:
Firstly, nice work!

One thing @selig suggested at some point in the past was a "deduction log" similar to [what SAT solvers do](https://www.cs.utexas.edu/~marijn/publications/APPA.pdf) - this way we don't have to keep stuff around in memory for proof printing and can resurrect it from file later. This is of course a much bigger change and needs some careful design, but thought it might be worth mentioning.

selig:
Yep, there are orthogonal issues here. The most pressing one is that Vampire seems to keep getting race conditions with proof printing  - the current mechanism for proof printing to stdout is messy. The other, broader, issue is that we're doing quite a bit of work at runtime to be able to print a proof e.g. we can't really delete any clauses. So perhaps we can do what SAT people do. Let's discuss that separately.

For this PR... I'm not sure we completely address the race condition thing. We remove races on stdout which is the one that Geoff was complaining about but we preserve our previous mechanism for deciding who prints proofs. I note the comment about printing multiple proofs. Maybe we want to allow each proof attempt to print its proof and then quit and then just ensure we print at least one (and possibly tidy up within the master thread afterward).

Martin's suggestion also avoids the race on stdout and if that's our main aim here then it does seem the simplest

quickbeam123:
> Sounds like a good idea. Does that mean we remove the option to save to file or we only make use of the tmpfile when the save to file option has not been selected?

I think the save to file option is useful. 

I am just not sure whether we shouldn't force the user to specify the actual filename rather than just a folder!
(That would make it clear it's their responsibility not to have name clashes. As it stands, what if I run vampire on the whole smtlib outputting the proofs to ./proofs... Will there be a clash or not? I don't know.)

ibnyusuf:
> > Sounds like a good idea. Does that mean we remove the option to save to file or we only make use of the tmpfile when the save to file option has not been selected?
> 
> I think the save to file option is useful.
> 
> I am just not sure whether we shouldn't force the user to specify the actual filename rather than just a folder!
> (That would make it clear it's their responsibility not to have name clashes. As it stands, what if I run vampire on the whole smtlib outputting the proofs to ./proofs... Will there be a clash or not? I don't know.)

One option is to keep incrementing an integer and appending it to the of the autogenerated name until the name is unique for the output folder. However, I am not sure this is very helpful. If a  user opens ./proofs and finds proofs very_hard_problem0.proof and very_hard_problem1.proof she/he won't know which proof was from which Vampire.

Perhaps we could allow users to specify file names with auto-generated names the fallback (with a warning).

quickbeam123:
I am fine if the interface stays simple on the vampire side, forcing the user to use shell tools and solve the name problem themselves. But if you have a use case for something fancier yourself, feel free to go ahead!

joe-hauns:
I'd also agree that it's better to let the user specify the file names, as it's less behaviour to document (i.e. how exactly the output files will be named), hence also less unexpected things that could go wrong from the user side.

Concerning the fallback idea you mentioned, do you mean fallback if no output file is specified? From the user side I'd expect that there's no output file if I don't explicitly specify one.. which would mean that the temporary proof file will b deleted.

PR #214, opened 2021-03-30
Fix forward subsumption resolution

JakobR:
The current version may miss some inferences.

### DIFF
--- a/Inferences/ForwardSubsumptionAndResolution.cpp
+++ b/Inferences/ForwardSubsumptionAndResolution.cpp
@@ -39,7 +39,6 @@
 
 #include "ForwardSubsumptionAndResolution.hpp"
 
-extern bool reporting;
 
 namespace Inferences
 {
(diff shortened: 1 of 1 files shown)


quickbeam123:
Jakob, you are now the leading expert on this part of the codebase so I would trust your update.

Maybe, before we do the merge, do you think this could benefit from some testing?

JakobR:
Yes, we should definitely do some checks first. (Let's put this PR on hold until we're done with them.)

selig:
Hi @JakobR do you think this is something we should aim to get in before competitions or keep on hold until afterwards?

JakobR:
Since it's "just" a completeness issue (and apparently didn't cause noticable trouble so far), I would prefer to keep it on hold for now.

quickbeam123:
Hi @JakobR, could this be now revived and merged? I am just starting a quick test!

quickbeam123:
This looks like a clear improvement! 

Especially:
`./vampire -sa discount -awr 10 -fsd on Problems/SYO/SYO679-1.p`
becomes much easier thanks to this!

PR #217, opened 2021-04-07
Polymorphic vampire refactoring sorts

ibnyusuf:
A significant refactoring of how sorts are dealt with:

- This pull request introduces a new child class of `Term`, `AtomicSort` to represent sort terms.
- It splits the signature into three, predicate symbols, function symbols and type constructors
- It introduces a new set in `TermSharing` called `_sorts` to store instances of `AtomicSort`

These changes generally make code clearer and force a programmer to think whether the term they are dealing with is actually a sort and potentially treat it differently if it is. This can lead to efficiency gains since sorts are untyped and therefore some type checks can be omitted (see `TermSharing` for an example of this).

(Still needs to be tested)


quickbeam123:
Terminology question: is the "Atomic" in "AtomicSort" indicative of there also being other, non-atomic sorts?

ibnyusuf:
Well we have quantified sorts (otherwise known as types) stored in the `OperatorType` class. Another motivation to add the "Atomic" is that there is already a Sort class in Lib.

ibnyusuf:
Ready for review. I have tested across TPTP with time limit of 150s and mode --casc and these are the results

Proved unsat, theorem, contra axioms
Master: 12408
Refactored sorts: 12398

selig:
Given the desire for stability prior to CASC let's wait until after the competition submission to merge this one. Making the code nicer is definitely a good thing.

CASC/PortfolioMode.cpp
```diff
@@ -134,8 +134,11 @@ bool PortfolioMode::searchForProof()
     //we normalize now so that we don't have to do it in every child Vampire
     ScopedLet<Statistics::ExecutionPhase> phaseLet(env.statistics->phase,Statistics::NORMALIZATION);
     Normalisation().normalise(*_prb);
-
-    TheoryFinder(_prb->units(),property).search();
+    
+    //TheoryFinder can not yet find polymorphic theory axioms
```
quickbeam123, line 138: Not being able to find something it was not meant to find does not seem like a reason not to call.

Does TheoryFinder cope with polymorphic input?

ibnyusuf:
Comment from @joe-hauns 

> aaah okay i wasn't aware that that's gonna b eliminated! I'd say it's best to keep some sort of future-proof wrapper of whatever kind, in order to make it clear in which places we rely on s->functor() is used as an id.
> I'm thinking like something like s->monomorphicId()  which has an assertion in there in order to make sure we only call it with a monomorphic sort.

CASC/PortfolioMode.cpp
ibnyusuf: No it doesn't, it crashes. I should probably have worded that comment differently.

ibnyusuf:
Branch requires retesting to ensure that merging in Integer Arithmetic branch hasn't created any issues.

quickbeam123:
@ibnyusuf is the conflict in `TPTP.cpp` a problematic one?

ibnyusuf:
> @ibnyusuf, is the conflict in `TPTP.cpp` a problematic one?

No, trivial, though I am wondering how the conflict came about?

I am still in the process of reviewing the places in the code we call `args()` to see if we would do better to call `termArgs()`. I'll incorporate this fix into those I carry out as part of the review.

quickbeam123:
Excellent! In the meantime, I slowly started going through the diff.

PR #224, opened 2021-04-22
Fix unexpected interaction between uwa and HOL

ibnyusuf:
fixing bug whereby running with `uwa on` had unexpected outcomes for higher-order problems

### DIFF
--- a/Indexing/SubstitutionTree.cpp
+++ b/Indexing/SubstitutionTree.cpp
@@ -949,14 +949,20 @@ bool SubstitutionTree::UnificationsIterator::associate(TermList query, TermList 
 {
   CALL("SubstitutionTree::UnificationsIterator::associate");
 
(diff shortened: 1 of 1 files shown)


selig:
@ibnyusuf I'm guessing this is necessary before CASC?

ibnyusuf:
> @ibnyusuf I'm guessing this is necessary before CASC?

@selig it is not critical, but would be nice to have the fix in prior to CASC.

selig:
So the unexpected outcomes are completeness related rather than soundness? Or just efficiency in general? I should probably look at the code/fix but...

ibnyusuf:
The bug is a soundness issue, but only if someone runs with ` -uwa on` with a higher-order problem which shouldn't happen. Still, we can't control what users do, so it would be nice to sort this out to prevent unexpected behavior.

PR #230, opened 2021-05-26
Ucore stuff

selig:
This implements a unsat-core mode for SMT-LIB. It can be triggered either by setting the output mode to ucore or by including

(get-unsat-core)

immediately after (check-sat) in the problem. Note that we still ignore

(set-option :produce-unsat-cores true)

as we produce models anyway.

The current commit adds the functionality but @quickbeam123 it might be interesting to have a little think about whether we could actively minimise our unsat cores... nothing springs to mind immediately (we should already do minimisation on AVATAR unsat core).


Kernel/InferenceStore.cpp
```diff
@@ -1043,6 +1043,60 @@ InferenceStore::ProofPrinter* InferenceStore::createProofPrinter(ostream& out)
   return 0;
 }
 
+struct StringHash{
+
+  static bool equals(vstring s1, vstring s2)
+  {
+    cout << s1 << " eq  " << s2 << endl;
```
quickbeam123, line 1050: A leftover debug code?

Kernel/InferenceStore.cpp
```diff
@@ -1043,6 +1043,60 @@ InferenceStore::ProofPrinter* InferenceStore::createProofPrinter(ostream& out)
...
+    if(u->number() <= Unit::getLastParsingNumber()){
+      if(!u->isClause()){
+        vstring label =  u->getFormula()->getLabel();
+        if(!printed.contains(label)){
+          out << label << endl; 
+          printed.insert(label);
+        }
+      }
+      else{
+        out << u->toString() << endl;
```
quickbeam123, line 1086: Input formulas are printed only once, input clauses (they probably don't exists in smtlib, right?) as many times as encountered?

selig:
Thanks @quickbeam123 I've tidied up the code. In particular, I've decided that ucore mode will just work with SMT :named labels - if we want  to extend  it to other kinds of labels later we can... so no clause printing and an extra check to make sure we only print non-default labels. As per the competition rules.

selig:
Of course, what we could do is carry on searching for proofs whilst remembering the shortest unsat core found so far... but that's engineering I'd prefer to avoid. I think we just enter to demonstrate that it's a feature we support but don't assume that we're going to win any medals!

PR #232, opened 2021-06-02
Fixing Vampire poly parsing

ibnyusuf:
Fixing parser bug whereby Vampire wasn't parsing `$i` and `$o` when these appeared as type arguments.

### DIFF
--- a/Parse/TPTP.cpp
+++ b/Parse/TPTP.cpp
@@ -2901,10 +2901,9 @@ void TPTP::term()
     
     case T_INTEGER_TYPE:
     case T_REAL_TYPE:
-    case T_RATIONAL_TYPE: {
-      //USER_ERROR("Polymorphic Vampire is currently not compatible with theory reasoning");
-      //the code below is in preparation for 
-      //when theorey reasoning is updated to deal with polymorphism
+    case T_RATIONAL_TYPE: 
+    case T_BOOL_TYPE:
+    case T_DEFAULT_TYPE: {
       resetToks();
       switch (tok.tag) {
         case T_INTEGER_TYPE:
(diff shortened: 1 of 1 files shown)


quickbeam123:
Hi Ahmed, could you please provide a 3 sentence "executive summary" of the change? (Not so much what you achieved--which is already stated--but how). Did you do any testing of the new version? Cheers!

ibnyusuf:
> Hi Ahmed, could you please provide a 3 sentence "executive summary" of the change? (Not so much what you achieved--which is already stated--but how). Did you do any testing of the new version? Cheers!

Sure, the pull request modifies the `term()` function in the parser which reads terms, to be able to handle the tokens `T_BOOL_TYPE` and `T_DEFAULT_TYPE` which it was previously unable to do. In TF1, these can appear as (type) arguments of a function symbol.

By allowing `T_BOOL_TYPE` as an argument, function symbols can at times act as predicate symbols (consider the symbol `id :: !>[X : $tType] : X -> X ` and the term `f($o, $true)`). Therefore, the `formulaInfix()` parser function has been modified to deal with this case by calling `endTermAsFormula` function if needed.

With regards testing, none of the TF1 problems in the existing TPTP library contain `$o` or `$i` as arguments. I have tested the fix on a problem sent to me Geoff which I attach here. I have also tested on 5 randomly selected problems from TF1 to ensure that nothing else has been broken by this fix.
[0001_Set_Monad_00009_2.txt](https://github.com/vprover/vampire/files/6585914/0001_Set_Monad_00009_2.txt)

quickbeam123:
Thank you, this helps! I vaguely remember there were some TPTP (TF1?) problems where a similar issue was arising with $int / $rat / $real or some other "built-in" sort. Would this be covered by your fix as well?

ibnyusuf:
> Thank you, this helps! I vaguely remember there were some TPTP (TF1?) problems where a similar issue was arising with $int / $rat / $real or some other "built-in" sort. Would this be covered by your fix as well?

You have a good power of recollection! That was the case, but has already been fixed (prior to merging polymorphism into master).

quickbeam123:
Perfect!

PR #233, opened 2021-06-02
extend extra schedules

selig:
Our current method for dealing with new options when not creating new schedules is to add them at the 'end' using PortfolioMode::getExtraSchedules. This does two things. Firstly, it uses a set of extra_opts to extend the given schedule. Secondly, it extends the time limit of the given schedule. This second feature allows portfolio mode to carry on trying with longer and longer time limits without giving up.

The way we incorporate the extra_opts is to copy the entire schedule with each extra_opt extended. Note that this is expensive! In this PR in the worst-case extra_opts will have 18 new options to try. So a '5-minute' schedule would increase to 90 minutes. In reality, it won't increase this much as many strategies are incomplete (terminating early).

If there are some options here that should just be added to the main SMT-COMP schedules let me know. If there are some things that you think should be added/changed add a comment.

Note that in SMT-COMP we will run through the extra options before the fallback as not very strong. BUT this does mean that we have to process all of the extra strategies before we get to the slightly more diverse fallback... it's all a trade-off.


MichaelRawson:
@selig this will probably compile if you rebase it onto master now, some CI shenanigans have gone through in the last few days.

quickbeam123:
@selig, would you agree this not a real PR? And when the competitions are over we close it into oblivion?

PR #234, opened 2021-06-04
bust cache: maybe this fixes Z3/GLIBCXX?

MichaelRawson:
CI recently failed with a linker problem. Try this to see if rebuilding Z3 (and therefore relinking it) helps.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -23,7 +23,7 @@ jobs:
       uses: actions/cache@v2
       with:
         path: ${{runner.workspace}}/vampire/z3/build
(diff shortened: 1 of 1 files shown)


MichaelRawson:
Blech, OK, this seemed to fix it. Working hypothesis: old build machine had a different glibc(xx), which we used to build Z3, which we then cached (because otherwise CI takes 12 minutes, rather than 4). When Github's CI machines get a software upgrade, we get linker failures.

Two possible solutions:
- remember this happens occasionally and bump the version (as here) when it dies
- don't cache Z3: build times for PRs are longer (~12 minutes, currently ~4), but we don't need to run CI every night to keep the cache warm

Let me know if we want the latter, that's a slightly different change.

PR #236, opened 2021-06-08
support large proofs in LaTeX

MichaelRawson:
Vampire's LaTeX output does not support large proofs well, such as those found by AVATAR. If the inference is too wide or long, it overflows the page. There is not an easy fix for this with the usual document model.

This branch changes the output routine to produce one page per inference, shrunk to size. Since the output is only intended for human consumption (?) this is not a breaking change. See attached files showing LaTeX output from `PUZ010-1`.

[before.pdf](https://github.com/vprover/vampire/files/6617080/before.pdf)
[after.pdf](https://github.com/vprover/vampire/files/6617144/after.pdf)

### DIFF
--- a/Shell/LaTeX.cpp
+++ b/Shell/LaTeX.cpp
@@ -67,13 +67,13 @@ using namespace Kernel;
 
 vstring LaTeX::header()
 {
-    vstring res =  "\\documentclass[fleqn]{article}\n"
-    "\\usepackage{fullpage,latexsym}\n"
-
-    "\\newenvironment{VampireProof}{%\n"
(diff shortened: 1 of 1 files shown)


easychair:
LaTeX has a \nopagebreak command that can be used for a single inference.
The result is that some pages will still have several inferences, which is
more convenient for humans.

A

MichaelRawson:
Hey Andrei,

I tend to agree. I'll look into `\nopagebreak` and see if I can get it to play well with [`standalone`](https://ctan.org/pkg/standalone).

PR #238, opened 2021-06-21
Option cleanup

quickbeam123:
Removing some unused options and cleaning the Options module a bit


ibnyusuf:
Look reasonable. If we can remove these from `Options.*` and Vampire still compiles it is safe to assume that these options are not functional.

ibnyusuf (approved):

PR #247, opened 2021-07-19
Reorganising HOL options

ibnyusuf:
- Added a new tag `HIGHER_ORDER` to the option tags and reclassified all higher-order options with this tag
- Reorganised the Options.cpp file to bring the various higher-order options together


quickbeam123:
I know the warning message (and now also the tag) are already there, but could you possibly find a more discriminating name for `superposition`?

ibnyusuf:
Do you mean something that makes it stand out as being linked to HOL? Whatabout `superposition_hol`?

quickbeam123:
Yes. `superposition_hol` would sure work!

ibnyusuf:
Renamed

PR #253, opened 2021-07-22
disable C++11 thread-safe statics, unused

MichaelRawson:
C++11 mandates that concurrent initialisation of local `static` variables is thread-safe. Vampire generally assumes that we work in a single-threaded context (don't I know it...), but thread-safe initialisation adds [some overhead](https://godbolt.org/z/vx9fd541K) regardless of whether we use it or not. Disable this via compiler flag.

No measurable performance improvement, but every little helps. Binary size is down 30KB on my machine.

Should in principle be safe since the behaviour is identical in a single-threaded context. Scary global change to code generation, however...

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,6 +9,13 @@ set(CMAKE_CXX_STANDARD 14)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
 
+# configure compilers
+if(CMAKE_CXX_COMPILER_ID STREQUAL GNU OR CMAKE_CXX_COMPILER_ID MATCHES Clang$)
+  add_compile_options(-Wall)
+  # we don't use multithreading in (mainline!) Vampire
+  add_compile_options(-fno-threadsafe-statics)
+endif()
+
 # compile command database
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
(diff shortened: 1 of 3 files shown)


quickbeam123:
Interesting! Especially the "Binary size is down 30KB on my machine."

I guess, I will give it a try in Makefile too ;)

MichaelRawson:
@quickbeam123 - apologies for the lack of build-system parity. ;-) Also planning a follow-up RTTI PR if you like this one which is more impactful...

selig:
@quickbeam123 should we add the Makefile change to this PR?

quickbeam123:
That's what I was planning on doing!

MichaelRawson:
@quickbeam123 - if you get chance, yes please. Otherwise in a day or two I will get around to it. :-)

quickbeam123:
So, a bit confusingly, I got the opposite result wrt size of the executable:

via Makefile
```
11485444 with -fno-threadsafe-statics
11448892 without

```
via cmake
```
9266656 with -fno-threadsafe-statics
9266304 without
```

using
```
g++ --version
g++ (MacPorts gcc11 11.1.0_2) 11.1.0

```

MichaelRawson:
Weird. Maybe something to do with Linux/Mac ABI differences? It's possible that this flag _increases_ instructions generated on some platforms, but I'd be surprised. Thanks for working out the Makefile, anyway.

PR #256, opened 2021-08-02
Use of `virtual`, `override` and `final` keywords.

MichaelRawson:
Vampire uses a fair amount of dynamic dispatch via class hierarchies.

First change: use the C++11 `override` keyword wherever possible. This allows the compiler to detect errors when e.g. a non-virtual function is overridden. I think this is fairly uncontroversial and does not change machine code.

Second change: use `final` instead of `override` wherever possible. The `final` keyword indicates that a method (or whole class) may not be overridden by subclasses. I think the _semantic_ uses for this in Vampire are fairly rare, but it allows the compiler to "devirtualise" calls in some cases, which may improve performance. I added a note to `HACKING.md` to indicate that this may be changed in most cases if we need to override a method. This might be more controversial and does impact generated machine code (although hopefully positively).

### DIFF
--- a/CASC/PortfolioMode.hpp
+++ b/CASC/PortfolioMode.hpp
@@ -42,15 +42,15 @@ class PortfolioMode;
 class PortfolioProcessPriorityPolicy : public ProcessPriorityPolicy
 {
 public:
-  float staticPriority(vstring sliceCode) override;
-  float dynamicPriority(pid_t pid) override;
+  float staticPriority(vstring sliceCode) final;
+  float dynamicPriority(pid_t pid) final;
 };
 
 class PortfolioSliceExecutor : public SliceExecutor
 {
 public:
   PortfolioSliceExecutor(PortfolioMode *mode);
-  void runSlice(vstring sliceCode, int remainingTime) override;
+  void runSlice(vstring sliceCode, int remainingTime) final;
 
 private:
   PortfolioMode *_mode;
--- a/DP/ShortConflictMetaDP.hpp
+++ b/DP/ShortConflictMetaDP.hpp
@@ -43,20 +43,20 @@ public:
   ShortConflictMetaDP(DecisionProcedure* inner, SAT2FO& sat2fo, SATSolver& solver)
   : _inner(inner), _sat2fo(sat2fo), _solver(solver) {}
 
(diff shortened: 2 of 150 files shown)


selig:
This sounds generally sensible. How did you do the refactor, was it semi-automatic? 

This feels like a 'style thing' that we may want to advise/enforce in some way, otherwise it's not going to apply to new code.

Obviously the longer we wait the more patching up to the PR will be required.

quickbeam123:
I am a bit hesitant about taking this one in (although I believe it would be best to have all overrides marked by `override`), simply because of how many files it impacts (while being only a change of syntax). 

My policy so far was: 
1) to add `override` when I was writing or modifying a method for some other reason,
2) realize I am getting a warning if this is used inconsistently within a translation unit, and
3) adapt the whole translation unit to use `override` everywhere, so that the warning goes away.

On the other hand, since Michael already did the work, ...

MichaelRawson:
@selig  - `override` everywhere is done automatically via `clang-tidy`, which uses a per-compilation-unit model. `final` is more tricky since it can involve more than one compilation unit - for this I `s/override/final/`, fix all the compiler errors, then read the diff to check I didn't do anything stupid. 

OK, so I leave the cost/benefit analysis up to @quickbeam123 - for @selig it is possible to add a (subset of) [`clang-tidy` checks](https://clang.llvm.org/extra/clang-tidy/checks/list.html) to CI, but unless we are willing to fail a build on lints this relies on people reading build logs before merging. Does anyone do this?!

If we want to merge, let me know and I'll fix conflicts.

quickbeam123:
One more remark about `final`.

My gut feeling of using it almost everywhere possible, to solicit an optimization, goes a bit against the spirit of some form of a generally understood comment to the effect of "this class's method has been designed such that it shouldn't make sense to override it". (Not that the need to use such comment occurs too often, I know ...)

Anyway, I am not sure I support this form eager finalisation.

MichaelRawson:
I agree that the `final` change is dubious! I was hoping that given e.g. the stacks of iterators we use this might make a big difference, but either it doesn't or the compiler was already smart enough to devirtualise in some cases. If we like `override` but not `final` let's cherry-pick.

MichaelRawson:
Merge conflicts, and we're not sure that this a good thing.

PR #259, opened 2021-08-04
Induction with recursive functions

mezpusz:
The new inferences can be found in this paper:
https://easychair.org/publications/preprint/tLl5

A new induction inference `GeneralInduction` has been added, which implements:
- the original induction rule `sik one` where for one induction term `t` of datatype sort and literal `~L[t]` we create an induction formula
```(L[b1] & ... & !x1,...,xn. (L[xi1] & ... & L[xik]) -> L[r(x1,...,xn)) & ...) -> !x. L[x]```
where `b1,...` are the base constructors of the term algebra of `t` and `r(x1,...,xn), ...` are the recursive constructors with `xi1,...,xik, ...` of the same term algebra sort.
- a new induction `-sik rec_ref` based on recursive functions where the above rule is generalized so that multiple induction terms of term algebra sorts can be possible within one induction formula and not just term algebra constructors can be present but any term with the restriction that the cases based on the induction terms must obey a well-founded lexicographic order with the subterm subrelation.
- multi-clause induction (see paper). This further generalizes the above by adding "side literals" `L1,...,Ln` apart from the "main literal" L to the induction formulas, that is, each base case, step case conclusion and step case hypothesis is of the form:
```(L1[.] & ... & Ln[.]) -> L```
where [.] denoted the holes for the induction terms.
- generalizations over:
  - term occurrences: given a term with n occurrences in the main and side literals, we may induct on any combination of these occurrences. In particular, the induction rule `rec_def` excludes some of these possibilities, inducting on the so-called active occurrences always (see paper).
  - complex terms: this just treats complex (non-constant) terms as regular induction terms. There is a flag which disallows inductin on a complex term if it has only one occurrence. This makes sense in most cases but otherwise there are counterexamples where one occurrence is inducted upon.

Function definition rewriting:
- this rule uses equality-based function definition axioms in the orientation they were given in the input, disregarding Vampire's ordering
- due to term sharing, when an equality is changed, its orientation might also change, hence the weird flag in `Term` that saves if this happened. After going through the preprocessing, these function definition axioms remain unchanged and therefore they are not present in the saturation loop as given clauses.

Induction hypothesis rewriting:
- the main idea here is to create new "lemmas" for induction (when multiple induction steps are needed for a proof) which would be impossible due to the ordering constraints
- we do this lazily, that is, we don't exactly know which literals are induction hypothesis are which are conclusion, we just match them based on what "induction skolems" -- which currently uniquely map the induction case -- are marked for their clauses. When we see a disequality and an equality where the induction skolems of the first are a superset of that of the second, we rewrite the first with the second. We unmark the used induction skolems in the resulting clause so that the same induction hypothesis (or any clause stemming from it) is not used twice.
- the resulting clause is given to induction without returning it directly to the saturation loop.

Unit tests:
- some modifications are cherry picked from the branch `inequality-resolution` in order to test inferences using indices

### CODE REVIEW

File: Shell/InductionSchemeGenerator.hpp [hunk 1]
```diff
@@ -0,0 +1,243 @@
...
+  }
+
+  uint64_t val() {
+    ASS(_finished);
+    return _iter;
+  }
+
+  uint64_t num_bits() const {
+    ASS(_finished);
+    return __builtin_ctz(_max);
```
hzzv, line 86: Maybe add here ASS(_max);? Documentation for __builtin_ctz(x) says that the result is undefined if x==0. https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html

Shell/InductionSchemeGenerator.cpp
```diff
@@ -0,0 +1,572 @@
...
+      }
+    }
+    Stack<TermList>::Iterator tvit(ta_vars);
+    while (tvit.hasNext()) {
+      recursiveCalls.emplace_back();
+      recursiveCalls.back().bind(0,tvit.next());
+    }
+    step.bind(0, TermList(Term::create(con->functor(),(unsigned)argTerms.size(), argTerms.begin())));
+    scheme.addCase(std::move(recursiveCalls), std::move(step));
+  }
```
hzzv, line 567: IIUC, while the scheme contains the 'term' (from arguments of generateStructural()) as an induction term, the cases only depend on the sort. However, the cases are always generated anew even if they were already generated for a different term of the same sort. Could this be optimized (eg. by caching the cases for each sort)?

Inferences/GeneralInduction.cpp
```diff
@@ -0,0 +1,517 @@
...
+    res.emplace_back(SLQueryResult(literal, premise), vset<pair<Literal*,Clause*>>());
+    while (itSides.hasNext()) {
+      auto qr = itSides.next();
+      if (InductionHelper::isInductionClause(qr.clause) && sideLitCondition(literal, premise, qr.literal, qr.clause)) {
+        res.back().second.emplace(qr.literal, qr.clause);
+      }
+    }
+  }
+  // pair current literal as side literal with possible main literals
+  // this results in only one side literal (the current)
```
hzzv, line 504: What about the inference where the current literal is one of multiple side literals? Is there a reason why we don't perform it?

Kernel/Inference.hpp
```diff
@@ -957,6 +965,23 @@ class Inference
...
 
+  DHSet<unsigned>* inductionInfo() const { return _inductionInfo; }
+  void addToInductionInfo(unsigned e) {
+    if (!_inductionInfo) {
+      _inductionInfo = new DHSet<unsigned>();
+    }
+    _inductionInfo->insert(e);
+  }
+  void removeFromInductionInfo(unsigned e) {
+    if (_inductionInfo) {
```
hzzv, line 976: Is it expected that this function will be called also when _inductionInfo was never allocated? If not, I think it would be good to use ASS(_inductionInfo) to catch potential bugs when the caller is expecting that _inductionInfo is present but in reality it is not.

Inferences/GeneralInduction.cpp
```diff
@@ -0,0 +1,517 @@
...
+  // Resolve all induction clauses with the main and side literals
+  auto resSubst = ResultSubstitution::fromSubstitution(&subst, 0, 1);
+  ClauseStack::Iterator cit(hyp_clauses);
+  while(cit.hasNext()){
+    Clause* c = cit.next();
+    for (const auto& v : hypVars) {
+      c->inference().addToInductionInfo(rvs.get(v));
+    }
+    auto qr = mainQuery;
+    qr.substitution = resSubst;
```
hzzv, line 356: Can we create 'qr' only once? It seems to be the same one for each iteration of this while loop.

Shell/InductionSchemeGenerator.hpp
mezpusz: Good catch! I added the assertion.

Shell/InductionSchemeGenerator.cpp
mezpusz: Good point. However, `InductionScheme`s are non-copyable because of the `DHMap` inside `Substitution`, so I ended up creating an intermediate container specific for these terms.

Inferences/GeneralInduction.cpp
mezpusz: The reason is we need another term index query to get other side literals for the main literal which would be as if the main literal triggered the induction. However, we could try it, I guess in some cases it would help, in others it would give more/bigger clauses without any benefit.

Kernel/Inference.hpp
mezpusz: Yes, it should be only performed on a subset of values that are queried from the very same inference, so the check makes sense.

Inferences/GeneralInduction.cpp
mezpusz: I made `mainQuery` and `sideLitQrPairs` non-references, so we copy the former and move the latter (since it is used only once) from outside and then we alter these before the binary resolution parts. I also perform the `Literal::complementaryLiteral` and `subst->apply` there since after that we don't use them anyways. Do you think this is reasonable?

Indexing/TermIndex.hpp
```diff
@@ -80,6 +80,37 @@ class SuperpositionLHSIndex
...
+public:
+  CLASS_NAME(ICSubtermIndex);
+  USE_ALLOCATOR(ICSubtermIndex);
+
+  ICSubtermIndex(TermIndexingStructure* is)
+    : TermIndex(is) {}
+
+protected:
+  void handleClause(Clause* c, bool adding);
+};
```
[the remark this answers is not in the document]
mezpusz: I changed their names to more proper non-abbreviations.

Shell/InductionSchemeGenerator.cpp
hzzv: Ok!

Inferences/GeneralInduction.cpp
hzzv: Looks good to me.

Indexing/TermIndex.hpp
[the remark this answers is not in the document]
hzzv: Thanks!

Inferences/GeneralInduction.cpp
hzzv: My point here is that whether we perform the inference then depends on the order of selection, so in some sense we are compromising completeness.
If it is not too much work, could you add that inference and see if it helps or hurts the performance?

Inferences/GeneralInduction.cpp
```diff
@@ -346,25 +346,27 @@ void GeneralInduction::generateClauses(
 
   // Resolve all induction clauses with the main and side literals
   auto resSubst = ResultSubstitution::fromSubstitution(&subst, 0, 1);
+  mainQuery.substitution = resSubst;
+  // Be aware that we change mainLit and sideLitQrPairs here irreversibly
```
hzzv, line 350: This comment reminded me - are there any pointer/reference arguments in any of the methods you introduce that are not supposed to be changed? If yes, it would be good to mark them const.

Inferences/GeneralInduction.cpp
mezpusz: Done.

Indexing/TermIndex.cpp
```diff
@@ -129,6 +188,10 @@ void DemodulationSubtermIndexImpl<combinatorySupSupport>::handleClause(Clause* c
 
   static DHSet<TermList> inserted;
 
+  if (c->inference().rule() == InferenceRule::IH_REWRITING) {
```
mezpusz, line 191: Remove this once no such clauses pop up in saturation where they shouldn't.

UnitTests/tInductionPreprocessor.cpp
```diff
@@ -0,0 +1,359 @@
...
+  checkCases(*schemes.begin(), {
+    { { { 0, r(x2).toTerm() }, { 1, r(r(x3)).toTerm() } }, { { { 0, f(x2,r(r(x3))).toTerm() }, { 1, x3.toTerm() } },
+                                                             { { 0, x2.toTerm() }, { 1, r(r(x3)).toTerm() } } } },
+    { { { 0, b.toTerm() }, { 1, x4.toTerm() } },           { } },
+    { { { 0, x5.toTerm() }, { 1, b.toTerm() } },           { } },
+    { { { 0, x6.toTerm() }, { 1, r(b).toTerm() } },        { } },
+  });
+
+  schemes.clear();
+  auto t2 = f(sk1,sk1);
```
hzzv, line 326: Maybe just use f(sk1,sk1) in the line below (similarly to above)?

UnitTests/tInductionPreprocessor.cpp
```diff
@@ -0,0 +1,359 @@
...
+    { { { 0, b1.toTerm() }, { 1, x5.toTerm() } },           { } },
+    { { { 0, b2.toTerm() }, { 1, x6.toTerm() } },           { } },
+    { { { 0, r2(x7,x8).toTerm() }, { 1, x9.toTerm() } },    { } },
+    { { { 0, x10.toTerm() }, { 1, b.toTerm() } },           { } },
+  });
+
+  // no scheme is generated if no Skolems are present or the term is non-ground
+  schemes.clear();
+  ft.requestInductionScheme(f(sk1,x).toTerm().term(), schemes);
+  ft.requestInductionScheme(f(b,sk1).toTerm().term(), schemes);
```
hzzv, line 354: Isn't this ground and isn't sk1 a skolem constant?

UnitTests/tInductionSchemeGenerator.cpp
```diff
@@ -0,0 +1,221 @@
...
+  sides.insert(make_pair(lit1, clause({ lit1 })));
+
+  gen.generate(qrMain, sides, res);
+
+  checkResult(res, {
+    { { sk2, { { lit2, 1 } } }, { sk3, { { lit2, 1 } } } },
+  });
+}
+
+TEST_FUN(test_04) {
```
hzzv, line 188: Please add a comment here.

UnitTests/tInductionSchemeGenerator.cpp
```diff
@@ -0,0 +1,221 @@
...
+  DECL_SKOLEM_CONST(sk2,s)                                                                 \
+  DECL_SKOLEM_CONST(sk3,s)                                                                 \
+  DECL_CONST(b, s)                                                                         \
+  DECL_FUNC(r, {s}, s)                                                                     \
+  DECL_TERM_ALGEBRA(s, {b, r})                                                             \
+  DECL_FUNC(f, {s, s}, s)                                                                  \
+  DECL_FUNC(g, {s, s}, s)                                                                  \
+  DECL_PRED(p, {s, s})
+
+void checkResult(vvector<pair<InductionScheme, OccurrenceMap>> res,
```
hzzv, line 32: Maybe better name this "checkOccurrenceMap" or something like that, since it checks only the occ. map and not the scheme?

UnitTests/tGeneralInduction.cpp
```diff
@@ -0,0 +1,463 @@
...
+      .indices({ index() })
+      .input( clause({ ~p(g(sK1)) }))
+      .expected({
+        clause({ ~p(g(b)), p(g(x)) }),
+        clause({ ~p(g(b)), ~p(g(r(x))) }),
+      })
+    )
+
+// multi-clause does not work due to clauses
+// not having complex terms in common
```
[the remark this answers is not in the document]
mezpusz: This is simply a choice I made based on experimental observations. I think we mentioned it in some early version of the paper. We can measure what happens without this filter and add a flag if it gives some additional results.

UnitTests/tGeneralInduction.cpp
```diff
@@ -0,0 +1,463 @@
...
+        clause({ g(sK3) == f(b,sK3), g(sK3) == f(r(x4),sK3) }),
+        clause({ g(sK3) == f(b,sK3), ~p(f(g(sK3),f(sK3,r(x4)))) }),
+      })
+    )
+
+// side premise triggers multi-clause
+TEST_GENERATION_INDUCTION(test_11,
+    Generation::TestCase()
+      .context({ clause({ ~p(f(sK1,sK2)) }),
+                 clause({ p(sK2) }) })
```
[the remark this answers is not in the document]
mezpusz: Makes sense. Also, here I had some problems matching the clauses and if we change the order of formula 1 and 2, the test won't work. This change somewhat makes it more resilient to this error as well.

UnitTests/tInductionHypothesisRewriting.cpp
```diff
@@ -0,0 +1,158 @@
...
+      .indices({ lhsIndex(), subtermIndex() })
+      .input( fromInduction(clause({  sk2 != f(f(sk2,sk2), sk1) })) )
+      .expected({
+              clause({  sk1 != f(f(sk2,sk2), sk1) }),
+              clause({  sk2 != f(f(sk1,sk1), sk1) }),
+              clause({  sk2 != f(f(sk2,sk2), sk2) }),
+      })
+    )
+
+// induction skolems are used only once
```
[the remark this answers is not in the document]
mezpusz: The "used again" below means the resulting clause is used again. I will clarify this.

UnitTests/tInductionPreprocessor.cpp
```diff
@@ -0,0 +1,359 @@
...
+                   { clause({ q(r1(x,y),r(z)), ~q(y,r(z)), g(b) == b, q(z,b) }),0, false } })
+
+  ASS(!env.signature->getFnDefHandler()->hasInductionTemplate(f.functor(), true));
+  ASS(!env.signature->getFnDefHandler()->hasInductionTemplate(g.functor(), true));
+  ASS(!env.signature->getFnDefHandler()->hasInductionTemplate(h.functor(), true));
+  ASS(!env.signature->getFnDefHandler()->hasInductionTemplate(p.functor(), false));
+  ASS(!env.signature->getFnDefHandler()->hasInductionTemplate(q.functor(), false));
+}
+
+// not useful functions
```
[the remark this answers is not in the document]
mezpusz: It means they are not recursive. I'll clarify this as well.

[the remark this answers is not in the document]
mezpusz: Or they are recursive but their arguments don't change. This is kind of confused in the code as well. If there are no inductive arguments, it can simply mean there are no recursive calls or that none of them change in any recursive call. I'll add comments to clarify.

UnitTests/tInductionHypothesisRewriting.cpp
```diff
@@ -0,0 +1,158 @@
...
+        clause({ f(sk1,sk2) != f(sk2,sk3) }),
+        clause({ f(sk1,sk4) != f(sk4,sk3) })
+      })
+    )
+
+// cases where nothing happens
+TEST_GENERATION(test_03,
+    Generation::TestCase()
+      .context({
+        fromInduction(clause({ sk1 != sk2 })), // due to same polarity
```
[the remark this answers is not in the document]
mezpusz: Added to the second test (if that's what you meant).

PR #262, opened 2021-08-09
SMTLIB problem export

joe-hauns:
Adds the options:
- `--export_avatar <filename>` 
- `--export_thi <filename>`

These options will cause vampire to create an smtlib file that contains equivalent commands to the z3 API functions that are being called in `Z3Interfacing`.

### DIFF
--- a/SAT/Z3Interfacing.cpp
+++ b/SAT/Z3Interfacing.cpp
@@ -121,16 +121,17 @@ Z3Interfacing::Z3Interfacing(SAT2FO& s2f, bool showZ3, bool unsatCore, vstring c
   _assumptions(),
   _showZ3(showZ3),
   _unsatCore(unsatCore),
-  _out(exportSmtlib == "" ? Option<std::ofstream>()
-                          : Option<std::ofstream>(std::ofstream(exportSmtlib.c_str())) )
+  _out()
 {
   CALL("Z3Interfacing::Z3Interfacing");
   BYPASSING_ALLOCATOR
-  _solver.reset();
-
+  _out = exportSmtlib == "" ? Option<std::ofstream>()
+                            : Option<std::ofstream>(std::ofstream(exportSmtlib.c_str())) ;
   if (_out.isSome() && _out.unwrap().fail()) {
     throw UserErrorException("Failed to open file: ", exportSmtlib);
   }
+
+  _solver.reset();
   outputln("(check-sat)");
   outputln("(get-model)");
   outputln("(reset)");
--- a/SAT/Z3Interfacing.cpp
+++ b/SAT/Z3Interfacing.cpp
@@ -42,7 +42,7 @@
 namespace Lib {
 using SortId = TermList;
 
-template<> 
+template<>
 struct BottomUpChildIter<z3::expr>
 {
   unsigned _idx;
@@ -69,13 +69,13 @@ struct BottomUpChildIter<z3::expr>
 namespace SAT
 {
 
-using namespace Shell;  
-using namespace Lib;  
+using namespace Shell;
+using namespace Lib;
 
 //using namespace z3;
 
-Z3Interfacing::Z3Interfacing(const Shell::Options& opts,SAT2FO& s2f, bool unsatCore):
-  Z3Interfacing(s2f, opts.showZ3(), /* unsatCore */ unsatCore)
+Z3Interfacing::Z3Interfacing(const Shell::Options& opts, SAT2FO& s2f, bool unsatCore, vstring const& exportSmtlib):
+  Z3Interfacing(s2f, opts.showZ3(), /* unsatCore */ unsatCore, exportSmtlib)
 { }
 
 const char* errToString(Z3_error_code code)
@@ -98,7 +98,7 @@ const char* errToString(Z3_error_code code)
   }
 }
 
-void handleZ3Error(Z3_context ctxt, Z3_error_code code) 
+void handleZ3Error(Z3_context ctxt, Z3_error_code code)
 {
   DEBUG(errToString(code))
   throw z3::exception(errToString(code));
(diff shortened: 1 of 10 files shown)


SAT/Z3Interfacing.cpp
```diff
@@ -98,50 +98,63 @@ const char* errToString(Z3_error_code code)
...
+            BYPASSING_ALLOCATOR;
+            _solver.check();
           ),
-         _solver.get_model())), 
+         _solver.get_model())),
   _assumptions(),
   _showZ3(showZ3),
-  _unsatCore(unsatCore)
+  _unsatCore(unsatCore),
+  _out(exportSmtlib == "" ? Option<std::ofstream>()
```
selig, line 124: Interesting. This seems like it needs to bypass the allocator. We have a bypass allocator call in the body, so does that happen before the initialisation? I'm not sure of the current allocator bypass checks.

SAT/Z3Interfacing.cpp
```diff
@@ -98,50 +98,63 @@ const char* errToString(Z3_error_code code)
...
   BYPASSING_ALLOCATOR
   _solver.reset();
 
-  z3::set_param("rewriter.expand_store_eq", "true");
+  if (_out.isSome() && _out.unwrap().fail()) {
+    throw UserErrorException("Failed to open file: ", exportSmtlib);
+  }
+  outputln("(check-sat)");
+  outputln("(get-model)");
+  outputln("(reset)");
```
selig, line 136: Vampire itself doesn't handle reset...  although I started to add in a branch a long time ago. But I'm guessing the aim is to hand these SMT-LIB files to other solvers!

SAT/Z3Interfacing.cpp
```diff
@@ -98,50 +98,63 @@ const char* errToString(Z3_error_code code)
...
 
   z3::params p(_context);
-  p.set("model.compact", false); // keeps z3 from compressing its model. ~50% of the runtime of get_model is spent doing that otherwise
+  auto setOption = [&](auto k, auto v){
+    p.set(k,v);
+    outputln(";- z3 parameter: ", k, "=", v);
+  };
+  setOption("rewriter.expand_store_eq", true);
```
selig, line 143: Can you add a comment about what this option is doing?

SAT/Z3Interfacing.cpp
```diff
@@ -294,46 +320,46 @@ SATSolver::Status Z3Interfacing::solveUnderAssumptions(const SATLiteralStack& as
...
-SATSolver::VarAssignment Z3Interfacing::getAssignment(unsigned var) 
+SATSolver::VarAssignment Z3Interfacing::getAssignment(unsigned var)
 {
   CALL("Z3Interfacing::getAssignment");
   BYPASSING_ALLOCATOR;
 
   ASS_EQ(_status,SATISFIABLE);
   bool named = isNamedExpr(var);
   z3::expr rep = named ? getNameExpr(var) : getRepresentation(SATLiteral(var,1)).expr;
+  outputln("(get-value (", rep, "))");
```
selig, line 331: Does the z3  documentation say that get-value and model.eval do the same thing? It would be interesting to check. We've had issues in the past with eval being a bit lazy... and the model completion option (not there for get-value) is necessary.

SAT/Z3Interfacing.cpp
joe-hauns: Oh good spot. I just checked and it only create instances of `Z3Interfacing` in contexts where we call `BYPASSING_ALLOCATOR` anyways. I will fix this in order to be future proof anyway.

SAT/Z3Interfacing.cpp
joe-hauns: Yes so the idea is to export our z3 calls in an smtlib file in order to b able to file issues for the z3 people easily.

SAT/Z3Interfacing.cpp
joe-hauns: Tbh I don't have any idea. the option was present before. I just kept it. But I'm happy to remove it as it would simplify the code ;) 
Git blame says @quickbeam123 has added the option in the first place.
`z3 -pd` explains
```
    expand_store_eq (bool) reduce (store ...) = (store ...) with a common base into selects (default: false)
```
I reckon we want to keep that.. (?)

SAT/Z3Interfacing.cpp
```diff
@@ -1145,9 +1154,44 @@ Z3Interfacing::~Z3Interfacing()
...
+      outputln("(declare-fun ", name, " () Bool)");
+      return _context.bool_const(name.c_str());
+  });
 }
 
 
+z3::expr Z3Interfacing::getNamingConstantFor(TermList toName, z3::sort sort)
+{
+  return _termIndexedConstants.getOrInit(toName, [&](){
+      auto name = "n" + toName.toString();
```
selig, line 1179: We never actually looked at whether this single letter prefix was enough to heuristically avoid clashes, which could lead to unsoundness (I think)... the same for the new prefix for constants below. There's obviously a way we could do it

SAT/Z3Interfacing.cpp
selig: I think it could help with performance things as well as debugging. I'd also like to use this interface to collect some problems and submit them to the interactive part of SMT-LIB

SAT/Z3Interfacing.cpp
selig: Sorry, I assumed it was new as I didn't see it removed from elsewhere. I  can see in another thread that this option was suggested by Nikolaj so let's definitely keep it ;)

SAT/Z3Interfacing.cpp
joe-hauns: No it doesn't. I just used that as I thought it would be the intuitive equivalent. 
This is what the documentation says:
```

/**
   \brief Evaluate the AST node \c t in the given model.
   Return \c true if succeeded, and store the result in \c v.

   If \c model_completion is \c true, then Z3 will assign an interpretation for any constant or function that does
   not have an interpretation in \c m. These constants and functions were essentially don't cares.

   If \c model_completion is \c false, then Z3 will not assign interpretations to constants for functions that do
   not have interpretations in \c m. Evaluation behaves as the identify function in this case.

   The evaluation may fail for the following reasons:

   - \c t contains a quantifier.

   - the model \c m is partial, that is, it doesn't have a complete interpretation for uninterpreted functions.
   That is, the option \ccode{MODEL_PARTIAL=true} was used.

   - \c t is type incorrect.

   - \c Z3_interrupt was invoked during evaluation.

   def_API('Z3_model_eval', BOOL, (_in(CONTEXT), _in(MODEL), _in(AST), _in(BOOL), _out(AST)))
*/
Z3_bool_opt Z3_API Z3_model_eval(Z3_context c, Z3_model m, Z3_ast t, bool model_completion, Z3_ast * v);
```
But I think there might be a discrebancy between `evaluate` and `get-value` indeed. Or at least this would explain an issue I just ran into while looking for a bug.

joe-hauns: Okay I dug into the z3 code, and found out that there is only one difference between `get-value` and `model.evaluate(...)`. That is, in the case of `get-value`, the model is being compressed before it is evaluated. 
So I think we can keep it as is. 
This also explains the discrepancy I ran into.

PR #266, opened 2021-08-17
Fix no-Z3 build and tests

mezpusz:
Build was failing without Z3 due to test for `TheoryInstAndSimp` is not behind `VZ3` flag.

Also, empty tests fail due to `vtest run <test>` returns -1 on missing tests.
This is currently fixed by returning 0, alternatively we could exclude those tests (tZ3Interfacing.cpp, tTheoryInstAndSimp.cpp) from CMake or add empty tests in else branch of VZ3.


selig:
Looks okay to me. 

@joe-hauns can you comment on "empty tests fail due to vtest run <test> returns -1 on missing tests" - you most recently touched vtest. What do you think?

Test/UnitTesting.cpp
```diff
@@ -47,7 +47,7 @@ UnitTesting& UnitTesting::instance()
 bool UnitTesting::runTest(vstring const& unitId, vstring const& testCase) 
 {
   auto unit = findUnit(unitId);
-  if (unit == nullptr) return false;
+  if (unit == nullptr) return true;
```
joe-hauns, line 50: I'm not sure whether that's the right thing to do. How can i reproduce the issue that u wanna solve with that?
Only adding the `#if VZ3` does the job for compiling and testing everything without `z3` on my machine.

I think the `return false;` should stay so we get a test failure if we try to run a non-existing test case.

joe-hauns: ah sorry just found what the issue you meant was. I'll have a quick look and give a new comment then.

joe-hauns: I'd propose to let the `return false;`s stay `return false;`s, and skipping the z3 tests in the makefile instead. I made these changes and pushed them.

mezpusz: Thanks for the change, I agree that it's a better solution. There is probably no way anyway to make sure no one adds new tests that would fail without Z3.

PR #267, opened 2021-08-17
changing the z3 option model.compact to true.

joe-hauns:
This gives worse performance in some cases, but also gets rid of some weird cases where `z3` does not evaluate boolean values correctly when it found a model, which causes `avatar` to fail.


selig:
@quicquid apparently you might run some tests on this. If not please merge as seems fine to me.

quickbeam123:
No worries, I will take care of the testing...

PR #269, opened 2021-08-26
Common interface for re-using introduced symbols.

MichaelRawson:
Do not merge, opening early so people can discuss.

There's several places in Vampire where we introduce new symbols that "represent" some formula (or term?). Examples include Skolemnization (regular CNF, NewCNF, induction (via NewCNF), CNF-on-the-fly), or naming subformulae (aforementioned CNF, FOOL, probably others).

It can be worthwhile to re-use symbols/terms for formulae, up to some re-use criterion on formulae. I have some data for a previous hack if anyone's interested, but this does happen frequently and it can improve proof search. CNF-on-the-fly already does this (thanks @ibnyusuf  for pointing this out).

I'm also interested in this for multi-threaded Vampire, where I'm trying to share AVATAR's solver between threads. If two different proof attempts separately introduce "the same" symbol, it would be nice if they could be treated as such by AVATAR.

So far I've implemented an interface (`Shell::NameReuse`) for re-using symbols, with two implementations so far:
- do not reuse symbols (default)
- re-use symbols if they represent exactly the same formula

I've added an option to select how to re-use Skolem symbols, and patched `Shell::Skolem` and `Shell::NewCNF` accordingly.
In general I've tried to make changes such that we only pay for what we use: the default behaviour of not re-using symbols should be (almost) free.

Some examples of the new behaviour:

```
fof(test, axiom, ?[X, Y]: p(X, Y)).
fof(test, axiom, ?[X, Y]: p(X, Y)).

% default behaviour
% vampire --mode clausify -updr off -skr none
cnf(u3,axiom,p(sK0,sK1)).
cnf(u4,axiom,p(sK2,sK3)).

% "exact" reuse behaviour
% vampire --mode clausify -updr off -skr exact
cnf(u6,axiom,p(sK0,sK1)).
cnf(u10,axiom,p(sK0,sK1)).
```
-----
```
fof(test, axiom, ![X]: ((?[Y]: p(X, Y)) & (?[Y]: p(X, Y)))).

% vampire --mode clausify -updr off -skr none
cnf(u4,axiom,p(X0,sK1(X0))).
cnf(u3,axiom,p(X0,sK0(X0))).

% vampire --mode clausify -updr off -skr exact
cnf(u10,axiom,p(X0,sK0(X0))).
cnf(u9,axiom,p(X0,sK0(X0))).
```
-----

Still left to do:
- [x] discuss the proposed interface and see if we like it
- [x] integrate this everywhere we want it: subformula naming especially
- [x] discuss where we don't
- [x] discuss options and how fine-grained we want these
- [x] come up with new and exciting re-use criteria: CNF-on-the-fly already re-uses _instances_, could consider e.g. associative/commutative operators, reuse-by-CNF-of-formula
- [x] fix variable-order problem
- [x] add statistics
- [x] run some experiments
- [x] decide on whether to make this default (probably not: quadratic behaviour)
- [ ] review implementation

### DIFF
--- a/Shell/NameReuse.cpp
+++ b/Shell/NameReuse.cpp
@@ -62,7 +62,7 @@ bool NameReuse::get(const vstring &key, unsigned &symbol)
 void NameReuse::put(vstring key, unsigned symbol)
 {
   CALL("NameReuse::put");
-  // std::cout << "put: " << env.signature->functionName(symbol) << " for " << key << std::endl;
+  //std::cout << "put: " << symbol << " for " << key << std::endl;
   _map.insert(key, symbol);
 }
 
--- a/Shell/NameReuse.cpp
+++ b/Shell/NameReuse.cpp
@@ -22,42 +22,31 @@
 
 namespace Shell {
 
-static NameReuse *make_policy(Options::NameReuse option)
-{
-  CALL("NameReuse::make_policy");
-  switch (option) {
-    case Options::NameReuse::NONE:
-      return new NoNameReuse();
-    case Options::NameReuse::EXACT:
-      return new ExactNameReuse();
-  }
-}
-
 NameReuse *NameReuse::skolemInstance()
 {
   CALL("NameReuse::skolemInstance");
-  static NameReuse *instance = make_policy(env.options->skolemReuse());
+  static NameReuse *instance = new NameReuse();
   return instance;
 }
 
 NameReuse *NameReuse::definitionInstance()
 {
   CALL("NameReuse::definitionInstance");
-  static NameReuse *instance = make_policy(env.options->definitionReuse());
+  static NameReuse *instance = new NameReuse();
   return instance;
 }
 
-Formula *ExactNameReuse::normalise(Formula *f)
+Formula *NameReuse::normalise(Formula *f)
 {
-  CALL("ExactNameReuse::normalise");
+  CALL("NameReuse::normalise");
   //std::cout << "normalise: " << f->toString() << std::endl;
   Rectify rectify;
   return rectify.rectify(f);
 }
 
-bool ExactNameReuse::get(Formula *normalised, unsigned &symbol)
+bool NameReuse::get(Formula *normalised, unsigned &symbol)
 {
-  CALL("ExactNameReuse::get");
+  CALL("NameReuse::get");
   //std::cout << "get: " << normalised->toString() << std::endl;
   return _map.find(normalised->toString(), symbol);
   /*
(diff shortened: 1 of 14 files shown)


Shell/Skolem.cpp
```diff
@@ -469,34 +483,69 @@ Formula* Skolem::skolemise (Formula* f)
...
+          if(VList::isNonEmpty(remainingVars)) {
+            reuse_formula = new QuantifiedFormula(
+              Connective::EXISTS,
+              remainingVars,
+              remainingSorts,
+              SubstHelper::apply(reuse_formula->qarg(), _subst)
+            );
+          }
+        }
+
```
quickbeam123, line 548: What is this code useful for?

Shell/NameReuse.hpp
```diff
@@ -0,0 +1,60 @@
...
+  // singletons for...
+  // skolems
+  static NameReuse *skolemInstance();
+  // definitions
+  static NameReuse *definitionInstance();
+
+  // convert `f` to a string in some way to use as a key: saves recomputing it later
+  vstring key(Formula *f);
+
+  // try and reuse a symbol for `normalised`
```
quickbeam123, line 43: `normalised` what?

Shell/NameReuse.cpp
```diff
@@ -0,0 +1,75 @@
...
+{
+  CALL("NameReuse::put");
+  //std::cout << "put: " << symbol << " for " << key << std::endl;
+  _map.insert(key, symbol);
+}
+
+Lib::Stack<unsigned> NameReuse::freeVariablesInKeyOrder(Formula *f)
+{
+  CALL("NameReuse::freeVariablesInKeyOrder");
+  return Lib::Stack<unsigned>::fromIterator(FormulaVarIterator(f));
```
quickbeam123, line 72: FormulaVarIterator -> Stack -> BottomFirstIterator (at use site). Is that necessary?

Shell/Naming.cpp
```diff
@@ -1146,26 +1166,37 @@ Literal* Naming::getDefinitionLiteral(Formula* f, VList* freeVars) {
...
-    predSym->setType(OperatorType::getPredicateType(arity - typeArgArity, termVarSorts.begin(), typeArgArity));
+      predSym->setType(OperatorType::getPredicateType(arity - typeArgArity, termVarSorts.begin(), typeArgArity));
+    }
     return Literal::create(pred, arity, true, false, allVars.begin());
   } else {
-    unsigned fun = env.signature->addNameFunction(typeVars.size());
-    TermList sort = AtomicSort::arrowSort(termVarSorts, AtomicSort::boolSort());
-    Signature::Symbol* sym = env.signature->getFunction(fun);
-    sym->setType(OperatorType::getConstantsType(sort, typeArgArity)); 
+    unsigned fun = reused_symbol;
```
quickbeam123, line 1191: Kudos for treating the HO case as well!

PR #278, opened 2021-09-29
When outputting unsat core, visit each Unit only once

zhangwen0411:
The current implementation of `InferenceStore::outputUnsatCore` can visit a Unit more than once during the traversal. This can cause outputting an unsat core to take a very long time.


selig:
Thanks! A silly mistake of mine. 

This one's obviously correct and helpful so I'll merge straight away. I need to take a careful look at the other one and I'm in the middle of  moving  house so might not get merged until next week.

PR #281, opened 2021-10-06
reorganise CI sanity checks

MichaelRawson:
The `checks/sanity` script got a little cluttered. Clean it up, reorganise the file structure to be less confusing and add some comments.


MichaelRawson:
I think @joe-hauns is on holiday until November, @quickbeam123 - but it can obviously wait, just FYI.

PR #283, opened 2021-10-06
fix cygwin compilation (without timer support)

quicquid:
Cygwin's gcc/glibc does not define some constants and functions without explicitly enabling them as a feature via -D_BSD_SOURCE. See also https://man7.org/linux/man-pages/man7/feature_test_macros.7.html .


quickbeam123:
I guess this is a step in the right direction, but might not fix the whole cigwin compile issue, right?

PR #296, opened 2021-11-02
remove unused material in `SAT::Preprocess`

MichaelRawson:
See #157. There's a number of unused methods and general gubbins in `SAT::Preprocess` that was useful at some point, but apparently no longer - probably from the days of `TWLSolver`. Remove these methods, remove a load of includes and fix formatting on a few lines.

There is now a sole method `SATClause *removeDuplicateLiterals(SATClause *)` - maybe this should move into `SATClause`?


quickbeam123:
Yes, please move `removeDuplicateLiterals` (as a static method) to `SATClause`! Afterwards, remove `SAT::Preprocess` completely. 

It might not be so annoying to have a few utility functions laying around (currently) unused, but getting rid of the whole annoying name clash on `Preprocess` will more than justify this PR!

MichaelRawson:
Done! It might be cleaner to have this be a non-static method now, but let's not introduce bugs unnecessarily.

PR #300, opened 2021-11-24
support (grounded) polymorphic functions for Z3

MichaelRawson:
This allows Vampire's Z3 translation to reason about fully-applied polymorphic functions by encoding them into a family of monomorphic functions. For example, if we have `f: !>[A: $tType]: A > A` and we create the terms `f($int, 1)` and `f($real, 2.0)` then Z3 cannot deal with `f` natively, but it is quite capable of reasoning about `f$int(1)` and `f$real(2.0)` where `f$int: $int > $int` and `f$real: $real > $real`.

Mostly I think the changes to achieve this are natural. There are two non-obvious points:
- Introduce a new `BottomUpChildIter` type, `SortlessTermList`: when building up such terms for Z3, we don't want to include sort terms in arguments as we'd get e.g. `f$int($int, 1)`. This seems like the cleanest way of doing it.
- The Z3 interface has a type called `FuncOrPredId`, which is what it sounds like. I add a (nullable) `Term *` to this structure, which is merely a Vampire term including the sort arguments that we need to distinguish e.g. `f$int` from `f$real`, such as `f($int, 1)`. We can supply this term when we need to construct such a term! I admit this is a bit weird, but it avoids an allocation and changes the least code.

Z3 seems quite happy with ad-hoc polymorphism, so I don't actually bother changing the name of the function. It will then have two separate functions called `f` with different types. Everyone OK with this?

This now proves the problem @ibnyusuf provided called `prob_07476_368335__18231272_1.p` relatively easily with only `-sas z3`. Naturally we should test thoroughly to make sure we're not accidentally unsound/out-of-spec.


SAT/Z3Interfacing.hpp
```diff
@@ -141,19 +141,55 @@ class Z3Interfacing : public PrimitiveProofRecordingSATSolver
...
+      if(!l.forSorts)
+        return true;
+
+      // compare sort arguments
+      for(int i = 0; i < l.forSorts->numTypeArguments(); i++)
+        // sorts are perfectly shared
+        if(!l.forSorts->nthArgument(i)->sameContent(r.forSorts->nthArgument(i)))
+          return false;
+
+      return true;
```
joe-hauns, line 179: Why can't we check that by just checking whether `l.forSorts == r.forSorts` could it be that two ids get different `forSorts`s that represent the same type arguments?

joe-hauns (commented):
Looks all nice and tidy! Just one minor comment/question concerning `operator==`.

SAT/Z3Interfacing.hpp
MichaelRawson: I believe so! If for instance we are trying to insert `f($int, 1)` and `f($int, 2)` then they will have the same type arguments but a different `forSorts`.

joe-hauns: Aaah right! I see! All good then! I was just confused for a sec.

SAT/Z3Interfacing.hpp
```diff
@@ -141,19 +141,55 @@ class Z3Interfacing : public PrimitiveProofRecordingSATSolver
...
+    Term *forSorts;
 
     friend struct std::hash<FuncOrPredId> ;
     friend bool operator==(FuncOrPredId const& l, FuncOrPredId const& r)
-    { return l.id == r.id && l.isPredicate == r.isPredicate; }
+    {
+      if(l.id != r.id || l.isPredicate != r.isPredicate)
+        return false;
+      if(!l.forSorts)
+        return true;
```
ibnyusuf, line 171: If `forSorts` is always set, it should be possible to remove this check since in the case where `numTypeArguments() == 0` the for loop below will never be entered.

SAT/Z3Interfacing.hpp
```diff
@@ -141,19 +141,55 @@ class Z3Interfacing : public PrimitiveProofRecordingSATSolver
...
-    static FuncOrPredId function(FuncId id) { return FuncOrPredId ( id, false ); }
-    static FuncOrPredId predicate(PredId id) { return FuncOrPredId ( id, true ); }
+    explicit FuncOrPredId(unsigned id, bool isPredicate, Term *forSorts = nullptr) : id(id), isPredicate(isPredicate), forSorts(forSorts) {}
+    explicit FuncOrPredId(Term* term) :
+      FuncOrPredId(
+        term->functor(),
+        term->isLiteral(),
+        term->numTypeArguments() == 0
+          ? nullptr
+          : term
```
ibnyusuf, line 151: Another option, would be to always set `forSorts` in both the mono and the poly case. This would lead to some minor code simplifications below.

SAT/Z3Interfacing.hpp
```diff
@@ -141,19 +141,55 @@ class Z3Interfacing : public PrimitiveProofRecordingSATSolver
...
-    { return out << (self.isPredicate ? "pred " : "func " )
-      << (self.isPredicate ? env.signature->getPredicate(self.id)->name() : env.signature->getFunction(self.id)->name());
+    {
+      out << (self.isPredicate ? "pred " : "func ");
+      out << (
+        self.isPredicate
+          ? env.signature->getPredicate(self.id)->name()
+          : env.signature->getFunction(self.id)->name()
+      );
+      if(self.forSorts)
```
ibnyusuf, line 189: If we always set `forSorts` this check becomes redundant

Kernel/BottomUpEvaluation/SortlessTermList.hpp
```diff
@@ -0,0 +1,53 @@
...
+
+namespace Lib {
+
+// just wrap a TermList for use in BottomUpChildIter
+struct SortlessTermList { TermList inner; };
+
+// identical to BottomUpChildIter<TermList>, except no sort arguments are returned
+// useful for e.g. Z3 expressions where we handle these as part of function definitions
+template<>
+struct BottomUpChildIter<SortlessTermList>
```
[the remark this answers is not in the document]
MichaelRawson: I don't think so - @joe-hauns, any objections?

SAT/Z3Interfacing.hpp
MichaelRawson: Sadly there are some places where we don't (?) have access to a concrete term. I guess in principle we could create a "dummy" term or review all of these places if you are very keen?

Kernel/Term.cpp
```diff
@@ -332,17 +332,22 @@ bool Term::isApplication() const {
   return !isSort() && !isLiteral() && env.signature->isAppFun(_functor);    
 }
 
+unsigned Term::numTypeArguments() const {
+  CALL("Term::numTypeArguments");
+
```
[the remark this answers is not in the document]
MichaelRawson: Good idea. Btw - I lifted this implementation from `Term::termArgs` directly below, which includes the `isSpecial` check. However, the implementation from `Term::hasTermArgs` doesn't. Any potential problem here?

Kernel/BottomUpEvaluation/SortlessTermList.hpp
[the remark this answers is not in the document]
joe-hauns: No objections. I did not know about the type arguments when I wrote `BottomUpChildIter<TermList>`. Or I think that they were not yet in master at that time. So `Kernel/BottomUpEvaluation/SortlessTermList.hpp` behaves how I'd expected `Kernel/BottomUpEvaluation/TermList.hpp` to behave!

SAT/Z3Interfacing.hpp
ibnyusuf: Sorry, I don't follow what you mean by "concrete term". We are passing a term into the constructor, can we not just use that?

MichaelRawson: The term is defaulted to `nullptr` in the first constructor (i.e. not this one) because at the call site we have access to a symbol, but not a term using that symbol. So `forSorts` might be null later regardless of what we do here - but you're right, for this particular bit of code we can just use the term.

Kernel/Term.cpp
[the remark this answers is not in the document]
ibnyusuf: I can see no reason for the absence of the `isSuper()` check in `hasTermArgs`. I am about to push a fix that replaces the function `hasTermArgs` with one called `numTermArguments`. It can obviously be used in a context requiring a Boolean as well, but has nicer symmetry  with `numTypeArguments`.

ibnyusuf:
I have tested this PR on a random sample of 98 TF1 problems coming from Petar's Sledgehammer benchmark set and there were no issues.

MichaelRawson:
OK, so there was a bug. The problem was that if you had a 'phantom' type variable (I think I am using this word correctly, @ibnyusuf ?) we could end up having merging symbols that should be distinct in Z3.

For example, consider the propositional problem

```
p | q
~p | q
p | ~q
~p | ~q
```

Unsatisfiable, right? Now let's sprinkle some types all over it:

```
p: !>[A: $tType]: $o
q: $o
a: $tType

p(a) | q
~p(a) | q
p(a) | ~q
~p(a) | ~q
```

A little weird, but still unsat. But now if we introduce another type:

```
p: !>[A: $tType]: $o
q: $o
a: $tType
b: $tType

p(a) | q
~p(a) | q
p(a) | ~q
~p(b) | ~q
```

this should be satisfiable because the last clause doesn't forbid `p(a), q`, (and normal Vampire is OK with this) but `-sas z3` says unsatisfiable because it doesn't distinguish between two `p` functions with the same type after substitution.

I've fixed this for now by adding a tag for each instantiated type variable to the Z3 name. I'm not sure this is the best solution, but I'm reasonably confident it's correct. Z3 does deal with e.g. compound symbol names correctly, e.g. `p(f(a))`.

quickbeam123:
Well discovered, @MichaelRawson ! (Already running new performance check.)

As a side note: Since you are now adding these tags, isn't the comment

`// Z3 seems OK with having overloaded functions`

superfluous / incorrect?

MichaelRawson:
> isn't the comment superfluous / incorrect?

Correct! Removed.

PR #301, opened 2021-11-25
Fix KBOForEPR polymorphism bug

ibnyusuf:
KBOForEPR assumes that all arguments to a predicate symbol are either variables or constants. We were not checking this properly for polymorphic predicate symbols previously. This PR should fix the bug.

### DIFF
--- a/Shell/Property.cpp
+++ b/Shell/Property.cpp
@@ -723,10 +723,7 @@ void Property::scan(TermList ts,bool unit,bool goal)
       _hasLogicalProxy = true;
     }
 
-    // TODO change this to use Term::numOfTypeArgs() once
-    // Micahel's PR is merged
-    OperatorType* type = func->fnType();
-    if(!t->isApplication() && type->typeArgsArity()){
+    if(!t->isApplication() && t->numTypeArguments() > 0){
       _hasPolymorphicSym = true;
     }
 
--- a/Shell/Property.cpp
+++ b/Shell/Property.cpp
@@ -685,8 +685,9 @@ void Property::scan(TermList ts,bool unit,bool goal)
     }
   } else {
     if(t->isSort()){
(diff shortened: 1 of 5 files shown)


Shell/Property.cpp
```diff
@@ -687,6 +684,12 @@ void Property::scan(TermList ts,bool unit,bool goal)
         break;
     }
   } else {
+    if(t->isSort()){
+      if(t->arity() > _maxTypeConArity)
+      _maxTypeConArity = t->arity();
```
MichaelRawson, line 689: Indentation here is confusing. Assume L689 wants indenting by one and then we `return` in any event if `t->isSort()`.

MichaelRawson (approved):

MichaelRawson:
Nice! Theoretical question: if we have only fully-applied type constructors (like in the Z3 PR), could we still use KBOForEPR in that case?

Shell/Property.cpp
ibnyusuf: Thanks, resolved

ibnyusuf:
> Nice! Theoretical question: if we have only fully-applied type constructors (like in the Z3 PR), could we still use KBOForEPR in that case?

No. In the current implementation of KBOforEPR, we make certain assumptions that are no longer valid in the presence of function symbols or type constructors with arity > 0.

For example:

```
Comparison arComp=Int::compare(l1->arity(),l2->arity());
    if(arComp!=Lib::EQUAL) {
      //since on the ground level each literal argument must be a constant,
      //and all symbols are of weight 1, the literal with higher arity is
      //heavier and therefore greater
      return fromComparison(arComp);
    }
```

As soon as we allow symbols of arity > 0 the weight of the ground instance of a variable can be arbitrarily large, and the above optimisation is no longer valid.

PR #306, opened 2021-12-02
since tryParseTPTP tries to only throw TPTP::ParseErrorException, ...

quickbeam123:
..., that's what we should be catching to retry with tryParseSMTLIB2

### DIFF
--- a/Shell/UIHelper.cpp
+++ b/Shell/UIHelper.cpp
@@ -218,6 +218,24 @@ UnitList* UIHelper::tryParseSMTLIB2(const Options& opts,istream* input,SMTLIBLog
           return parser.getFormulas();
 }
 
+// Call this function to report a parsing attempt has failed and to reset the input
+template<typename T>
+void resetParsing(T exception, vstring inputFile, istream*& input){
+           env.beginOutput();
+           addCommentSignForSZS(env.out());
+           env.out() << "Failed with\n";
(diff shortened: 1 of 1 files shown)


quickbeam123:
Until now we had:

```
./vampire ../../sledgehammer-eval/problem-examples/Core_DOM/SMT2/0023_Core_DOM_Functions/prob_01627_079330__15026148.smt_in
Parsing Error on line 1
Bad character at position 0
```

without a retry (we go for TPTP first, because of the non-standard file extension `smt_in`).

selig:
Great catch and I agree with your change but I noticed we're also missing exceptions from the SMT parser (from the lexer and parser, which are the ones we'd expect to hit if we have a TPTP file).

My proposed solution works around the fact that the various exceptions involved here are a bit of a mess!

quickbeam123:
Since we are at this, would it be too bad to just catch ANY exception (instead of a potentially open eded list and a templated call)? 

Alternatively, a common ancestor (in the class hierarchy) of all the parsing exceptions?

selig:
I think with ANY exception we still need the templated call. I tried to look for a suitable common ancestor and it didn't work but that might be my C++ fu failing. Just using Lib::Exception ended up with the wrong cry function being called. It's important we show the user the output of the failing exception when skipping it.

quickbeam123:
I see. A general exception cannot not cry! :)

PR #309, opened 2022-01-06
Print result before property is deleted

ibnyusuf:
This PR resolves the issue wherein the `env.property` object is deleted prior to proof printing when running in Vampire mode. 

However, it does not reolve the somewhat odd ownership of the `property` object. Unlike the `options`, `signature` etc. objects, the environment only mantains a pointer to the property object and [does not own it](https://github.com/vprover/vampire/blob/dd79494d88a52a3c92bd32f5fd5c3cd7fcdc28a9/Shell/Property.cpp#L92). 

Is this something that we look to change?

### DIFF
--- a/CASC/PortfolioMode.cpp
+++ b/CASC/PortfolioMode.cpp
@@ -136,7 +136,7 @@ bool PortfolioMode::searchForProof()
     Normalisation().normalise(*_prb);
     
     //TheoryFinder cannot cope with polymorphic input
-    if(!env.statistics->polymorphic){
+    if(!env.property->hasPolymorphicSym()){
       TheoryFinder(_prb->units(),property).search();
     }
   }
@@ -236,7 +236,7 @@ void PortfolioMode::getExtraSchedules(Property& prop, Schedule& old, Schedule& e
    //extra_opts.push("etr=on");         // equational_tautology_removal
    extra_opts.push("av=on:atotf=0.5");     // turn AVATAR off
 
-   if(!env.statistics->higherOrder){
+   if(!prop.higherOrder()){
      //these options are not currently HOL compatible
      extra_opts.push("bsd=on:fsd=on"); // subsumption demodulation
      extra_opts.push("to=lpo");           // lpo
--- a/Indexing/IndexManager.cpp
+++ b/Indexing/IndexManager.cpp
@@ -145,7 +145,7 @@ Index* IndexManager::create(IndexType t)
   bool isGenerating;
   static bool const useConstraints = env.options->unificationWithAbstraction()!=Options::UnificationWithAbstraction::OFF;
   static bool const extByAbs = (env.options->functionExtensionality() == Options::FunctionExtensionality::ABSTRACTION) &&
-                    env.statistics->higherOrder;
+                    env.property->higherOrder();
                     
   switch(t) {
   case GENERATING_SUBST_TREE:
--- a/Indexing/LiteralSubstitutionTree.cpp
+++ b/Indexing/LiteralSubstitutionTree.cpp
@@ -35,7 +35,7 @@ LiteralSubstitutionTree::LiteralSubstitutionTree(bool useC)
   //However, there is no need to guard aginst it, as equalityProxy removes all
   //equality literals. The flag below is only used during the unification of 
   //equality literals.
-  _polymorphic = env.statistics->polymorphic || env.statistics->higherOrder;
+  _polymorphic = env.property->hasPolymorphicSym() || env.property->higherOrder();
 }
 
 void LiteralSubstitutionTree::insert(Literal* lit, Clause* cls)
--- a/Indexing/TermSharing.cpp
+++ b/Indexing/TermSharing.cpp
@@ -82,8 +82,8 @@ void TermSharing::setPoly()
 
   //combinatory superposiiton can introduce polymorphism into a 
   //monomorphic problem
-  _poly = env.statistics->higherOrder ||
-          env.statistics->polymorphic ||
+  _poly = env.property->higherOrder() ||
+          env.property->hasPolymorphicSym() ||
           env.options->equalityProxy() != Options::EqualityProxy::OFF ||
           env.options->saturationAlgorithm() == Options::SaturationAlgorithm::INST_GEN;
 }
--- a/Inferences/EqualityFactoring.cpp
+++ b/Inferences/EqualityFactoring.cpp
@@ -116,7 +116,7 @@ struct EqualityFactoring::ResultFn
     ASS_NEQ(sLit, fLit);
 
     static Options::FunctionExtensionality ext = env.options->functionExtensionality();
-    bool use_ho_handler = (ext == Options::FunctionExtensionality::ABSTRACTION) && env.statistics->higherOrder;
+    bool use_ho_handler = (ext == Options::FunctionExtensionality::ABSTRACTION) && env.property->higherOrder();
 
     if(use_ho_handler){
       TermList sLHSreplaced = sLHS;
--- a/Inferences/EqualityResolution.cpp
+++ b/Inferences/EqualityResolution.cpp
@@ -78,7 +78,7 @@ struct EqualityResolution::ResultFn
     static Options::FunctionExtensionality ext = env.options->functionExtensionality();
     bool use_uwa_handler = uwa != Options::UnificationWithAbstraction::OFF;
     bool use_ho_handler = (ext == Options::FunctionExtensionality::ABSTRACTION) &&
-                          env.statistics->higherOrder;
+                          env.property->higherOrder();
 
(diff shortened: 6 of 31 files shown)


quickbeam123:
Could you please remind me what are the symptoms of the current situation in master (i.e., of "env.property object is deleted prior to proof printing")? Do we get a crash under some conditions? (Is it during higher-order / polymorphic proof printing?)

I slightly dislike that the contract of `doProving` has now been extended to "... and exactly in vampire mode we also print the result". I guess adding a bool argument (print/notPrint) to the function instead, to make this explicit, could be better design. (But vampire is full of these out-of-the-blue hacks already, so this might be a matter of taste.)

Concerning ownership, I guess the original idea was that it's the `Problem` who owns `Property` (cf. `prb->getProperty()` in `doProving`), but later someone needed it more globally and added:
`3e73ab1fe1 (Andrei Voronkov      2011-07-23 20:24:13 +0000   92)   env.property = this;`
to the `Property::Property` constructor. Quite an ugly solution, given how deeply it is hardwired.

For a fix, how about we consider whether `Problem` itself should not be part of `env`? (It could even be owned by it!)
@MichaelRawson, didn't you (through your work on the multi-threading) come to a better understanding of what the ideal solution of all the global-ish objects in vampire should look like? (I kind of like how z3 up front admits something global will be need, calls it a context and starts passing it around...)

quickbeam123:
OK, this is interesting, good to know! 

Then maybe one more question. As the working logic behind property is that it gets updated along the preprocessing pipeline; so for example, if we were to somehow eliminate the only polymorphic predicate with some preprocessing transformation, it would be nice if the updated property object somehow reflected that. Is that something we could already support now? 

Also, wouldn't it make sense for you to add to this PR all the changes that would migrate the odd information from statistics to property to kind of justify and capitalize on the change?

ibnyusuf:
> Then maybe one more question. As the working logic behind property is that it gets updated along the preprocessing pipeline; so for example, if we were to somehow eliminate the only polymorphic predicate with some preprocessing transformation, it would be nice if the updated property object somehow reflected that. Is that something we could already support now?

Good question. On further analysis, we store information about polymorphism in both statistics and in the property. The inforamtion in property is updated during the course of preprocessing whilst that in statistics is not. As part of this PR I should clean up the mess. Adding an on hold flag to this PR now.

> Also, wouldn't it make sense for you to add to this PR all the changes that would migrate the odd information from statistics to property to kind of justify and capitalize on the change?

Good point, I will do that

MichaelRawson:
The solution for the specific problem describes looks fine. I might also prefer a flag passed in, so that I can tell roughly what stuff does from the call site!

More generally I think @quickbeam123 is correct, we could use a "context" object that we pass around instead that holds this stuff, and maybe a separate "problem" object also. There are a _lot_ of advantages to this approach, not least the multithreading case. I doubt this is happening any time soon given how much effort it was just to change `env` to a pointer...but I don't personally see any issue with `env` owning `env.property` at least.

Incidentally

`env.property = this;`

is one of my least favourite parts of the Vampire codebase, I'd be very pleased if that could disappear! :-)

ibnyusuf:
This is ready to review again.

I have now:

-  transferred ownership of `Property` to `Environment`
-  moved `higherOrder` and `polymorphism` flags to the `Property` object from the `Statistics` class

Some questions:

- When creating a copy of the `Problem` object we no longer create a copy off the `Property` object, merely copy its pointer (both `Problem` objects contain pointers to `env.property`). Is this safe to do?
- In `Statistics` we have a `hasTypes`flag? Should this also be moved to the Property object?

MichaelRawson (approved):
Nice job, solid change that makes sense.

MichaelRawson:
Sorry, missed the update.

1. I think so...but we should be careful that we never mutate the shared `Property` unless we mean to. There again, I'm not really sure why we copy `Problem` objects in the first place.
2. I would say so, it's not really a statistic.

ibnyusuf:
> Sorry, missed the update.
> 
>     1. I think so...but we should be careful that we never mutate the shared `Property` unless we mean to. There again, I'm not really sure why we copy `Problem` objects in the first place.

The only place in the code that I can find where we copy Problem objects is when we run instance generation along side saturation. In this case both loops have their own copy of the problem. However, this appears to occur after preprocessing, so we should not be modifyingthe problem properties at all.

quickbeam123:
These changes all look reasonable, but I think we should take the opportunity and resolve the "bigger picture" as well.

For me there are two main questions:
1) Who actually owns `Property`? 
2) And is it really for everybody to access via `env.property`?

Regarding 1), it feels like
`env` always has one as 
`property = new Property;` in `Environment::Environment()` 
and also cleans it, thanks to 
`delete property;` in `Environment::~Environment()`,
but is keen on having the one (any) `Problem` creates for itself: 
`env.property = _property;` in `Problem::refreshProperty()`.
This seems to work, as long as we only have one `Problem`. (The copy in InstGen, as Ahmed figured out, seems at the moment harmless.) But I find is quite ugly.

Regarding 2), there seem to be many places in the code where we simply do:
`env.property->something()`,
but when we have a problem at hand, we prefer
`prb.something()`.
The wrappers in `Problem` talk to its `_propery`, but there is also the opportunity for various places in the code to tell the `Problem` that its `_propery` is not up to date and any of its "getters" will trigger refreshProperty() and compute it anew.

The question is, are we sure all the places talking directly to `env.property->something()` are getting the most up to date results (thanks to some complex code invariants) (or don't care about the most up to date results for some even more complex unstated reasons), or have they just been accessing property directly through `env` instead of talking to a `Problem`, because a programer was lazy (and `Problem` was somewhere far far away)?

To me it seems that while global things are bad design, we are not going to get rid off env any time soon. However, global `env.property` whose meaning is something like "I am the last version of `Problem._property` of the last `Problem` who got refreshed" is totally awful. And the solution (for now) should be, I think, a global `env.problem` (as opposed to `env.property`)  who's `_property` nobody should access directly. 

Now, does this make sense to others too? @ibnyusuf , do you think it would be feasible?

MichaelRawson:
I think all that makes sense. Global things are a little inflexible but not inherently buggy, and if we wanted we could transition to passing `env` (`ctx`?) around mechanically at some point. I've time to help next week if anyone wants it, give me a ping.

ibnyusuf:
I agree that having a single `Problem` object within the environment that cannot be copied is the cleanest most future proof solution. 

It should be feasible, but will require a good amount of niggly refactoring. I don't have much availability this week to carry it out, so @MichaelRawson if you have some time please feel free to go ahead. If not, I can make an attempt next week.

quickbeam123:
Don't be afraid to say it would be too much of digging. We already have an improvement in this PR.

As a matter of fact, I just noticed a line preprocessing `env.signature->hasTermAlgebras()`. The path to perfection is going to be longer than expect, for sure...

MichaelRawson:
Yep - can I suggest we merge this (something better than nothing) @quickbeam123?

I put something on my Vampire backlog to fix this. Having properties randomly (maybe even defensively?) refreshed is awful. I will at some point get around to doing this, but (as Ahmed correctly points out) the refactoring really is niggly and the probability of introducing bugs is quite high, so I will tread carefully.

ibnyusuf:
One thing that I don't follow regarding the current design is why we delete and recreate the property object at each refresh? Surely we could just update the data in the object?

MichaelRawson:
> One thing that I don't follow regarding the current design is why we delete and recreate the property object at each refresh? Surely we could just update the data in the object?

Yeah, I have no idea. Some kind of immutable design I could understand, but given the attention to performance elsewhere it does seem incongruous.

quickbeam123:
> Having properties randomly (maybe even defensively?) refreshed is awful.
They are not refreshed defensively. Only lazily. The logic in Problem makes sure this is only done if the problem was modified (and there is interest in knowing the property after). So the solution in problem is actually quite nice (if not an overkill) and the current design encourages ignoring the niceness and potentially falling prey to out of date Properties...

quickbeam123:
> why we delete and recreate the property object at each refresh?

Not sure either, but I guess the author thought that maintaining the path from a defaulted state to an up-to-date one is better maintained when the defaulted one comes (only) from the constructor. 

Note that the constructor there is not public and the only way to get a `Property*` is via `Property* Property::scan(UnitList* units)`

selig:
My two-cents on why it is refreshed not updated is that (i) it's harder to make mistakes, and (ii) it could actually be cheaper.

To update we need to 'remove' an object before adding its replacement. It's common to have the old and new objects at the same time but not always. Needing to match remove/add introduces a  potential for making mistakes.

We often refresh property after making big changes to the problem. If we had to remove/add everything then we do twice  as much work as just forgetting the old thing and adding the new things. Of course, if we start refreshing the property frequently for small changes then this argument goes away.

PR #311, opened 2022-01-13
Fix LPO returning INCOMPARABLE for some comparable cases

mezpusz:
This PR contains some fixes for the LPO ordering implementation. The original "bug" was that term 1. should have been GREATER than term 2. by the subterm property but it was in fact INCOMPARABLE with precedence f < g < c:

1. g(f(x,y),c)
2. c

This was caused by two issues:
1. the function `LPO::lpo` failed to meet its specification, namely to return the correct result whenever its first argument is greater or equal than its second, it never returned EQUAL when the two terms were equal and it didn't return anything in one branch.
2. in the function `LPO::alpha`, the above function `LPO::lpo` was called with arguments switched, so it sometimes returned INCOMPARABLE when it was actually GREATER.

Additionally, some base unit tests were added including the original "bug".

Regression tests with `-t 10 --forced_options to=lpo`, format "solved (unique)":
| Benchmark set | master | this branch | total |
| -------------------- | --------- | --------------- | ----- |
| TPTP (`--mode casc`) | 11754 (40) | 11958 (244) | 23291 |
| SMT-LIB (`--mode smtcomp`*) | 55843 (128) | 56062 (347) | 93519 |

* single-core, time limit is overriden

### DIFF
--- a/Kernel/LPO.cpp
+++ b/Kernel/LPO.cpp
@@ -10,8 +10,9 @@
 /**
  * @file LPO.cpp
  * Implements class LPO for instances of the lexicographic path
- * ordering
- *
+ * ordering based on Bernd Loechner's thesis "Advances in
+ * Equational Theorem Proving - Architecture, Algorithms, and
+ * Redundancy Avoidance" Section 4.2
  */
 
 #include "Debug/Tracer.hpp"
@@ -100,7 +101,8 @@ Ordering::Result LPO::clpo(Term* t1, TermList tl2) const
     ASSERTION_VIOLATION;
     // shouldn't happen because symbol precedence is assumed to be
     // total, but if it is not then the following call is correct
-    return cAA(t1, t2, t1->args(), t2->args(), t1->arity(), t2->arity());
+    //
+    // return cAA(t1, t2, t1->args(), t2->args(), t1->arity(), t2->arity());
   }
 }
 
@@ -242,7 +244,9 @@ Ordering::Result LPO::lexMAE(Term* s, Term* t, TermList* sl, TermList* tl, unsig
       ASSERTION_VIOLATION;
     }
   }
-  return GREATER;
+  // reached only when the terms are equal but this is checked already
+  // at the start of LPO::lpo, which is the only caller of this function
+  ASSERTION_VIOLATION;
 }
 
 // greater if s is greater than every term in tl
--- a/Kernel/LPO.cpp
+++ b/Kernel/LPO.cpp
@@ -176,11 +176,11 @@ Ordering::Result LPO::alpha(TermList* sl, unsigned arity, Term *t) const
   ASS(t->shared());
 
   for (unsigned i = 0; i < arity; i++) {
-    switch (lpo(t, *(sl - i))) {
+    switch (lpo(*(sl - i), TermList(t))) {
     case EQUAL:
-    case LESS:
-      return GREATER;
     case GREATER:
+      return GREATER;
+    case LESS:
     case INCOMPARABLE:
       break;
     default:
@@ -196,30 +196,26 @@ Ordering::Result LPO::lpo(TermList tl1, TermList tl2) const
 {
   CALL("LPO::lpo(TermList, TermList)");
 
+  if(tl1==tl2) {
+    return EQUAL;
+  }
   if(tl1.isOrdinaryVar()) {
     return INCOMPARABLE;
   }
   ASS(tl1.isTerm());
-  return lpo(tl1.term(), tl2);
-}
-
-// used when the first term is not a variable
-Ordering::Result LPO::lpo(Term* t1, TermList tl2) const
-{
-  CALL("LPO::lpo(Term*, TermList)");
-
+  Term* t1 = tl1.term();
   ASS(t1->shared());
 
   if(tl2.isOrdinaryVar()) {
     return t1->containsSubterm(tl2) ? GREATER : INCOMPARABLE;
   }
-  
+
   ASS(tl2.isTerm());
   Term* t2=tl2.term();
 
   switch (compareFunctionPrecedences(t1->functor(), t2->functor())) {
   case EQUAL:
-    lexMAE(t1, t2, t1->args(), t2->args(), t1->arity());
+    return lexMAE(t1, t2, t1->args(), t2->args(), t1->arity());
   case GREATER:
     return majo(t1, t2->args(), t2->arity());
   default:
@@ -257,7 +253,7 @@ Ordering::Result LPO::majo(Term* s, TermList* tl, unsigned arity) const
   ASS(s->shared());
 
   for (unsigned i = 0; i < arity; i++) {
-    switch(lpo(s, *(tl - i))) {
+    switch(lpo(TermList(s), *(tl - i))) {
     case GREATER:
       break;
     case EQUAL:
--- a/Kernel/LPO.hpp
+++ b/Kernel/LPO.hpp
@@ -39,6 +39,14 @@ public:
   LPO(Problem& prb, const Options& opt) :
     PrecedenceOrdering(prb, opt)
   {}
+  LPO(Problem& prb, const Options& opt, DArray<int> funcPrec) :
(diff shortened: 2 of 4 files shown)


Kernel/LPO.cpp
```diff
@@ -196,30 +196,26 @@ Ordering::Result LPO::lpo(TermList tl1, TermList tl2) const
...
     return t1->containsSubterm(tl2) ? GREATER : INCOMPARABLE;
   }
-  
+
   ASS(tl2.isTerm());
   Term* t2=tl2.term();
 
   switch (compareFunctionPrecedences(t1->functor(), t2->functor())) {
   case EQUAL:
-    lexMAE(t1, t2, t1->args(), t2->args(), t1->arity());
```
MichaelRawson, line 222: This is horrifying! We should probably use some compiler machinery for warning about unused return values, but that's a separate PR.

Kernel/LPO.cpp
```diff
@@ -176,11 +176,11 @@ Ordering::Result LPO::alpha(TermList* sl, unsigned arity, Term *t) const
   ASS(t->shared());
 
   for (unsigned i = 0; i < arity; i++) {
-    switch (lpo(t, *(sl - i))) {
```
MichaelRawson, line 179: Isn't this OK? The arguments are switched, I agree, but on closer inspection the logic below is also switched to fix it. I like the `LPO::lpo(Term *, TermList)` function for when we know that the LHS is a compound term as we can skip a check in some cases. Maybe just the `return lexMAE` fix below was enough to fix the bug?

Kernel/LPO.cpp
```diff
@@ -257,7 +253,7 @@ Ordering::Result LPO::majo(Term* s, TermList* tl, unsigned arity) const
   ASS(s->shared());
 
   for (unsigned i = 0; i < arity; i++) {
-    switch(lpo(s, *(tl - i))) {
```
MichaelRawson, line 260: Same comment about the original `LPO::lpo` override here, but more importantly isn't the logic below wrong? We'd only ever go around the loop once - I think the `GREATER` case should be `continue`, not `break`.

MichaelRawson: `lexMAE` looks wrong from that perspective as well.

MichaelRawson (changes requested):
OK, looks like there's some more issues that should be fixed. Some minor comments:
- maybe a comment at the top of the file would be useful to say that this comes from 'things to know when implementing LPO', and maybe which version of the algorithm we use - R5?
- L103 is unreachable after `ASSERTION_VIOLATION`, we shouldn't even codegen that

Kernel/LPO.hpp
```diff
@@ -39,7 +39,10 @@ class LPO
   LPO(Problem& prb, const Options& opt) :
     PrecedenceOrdering(prb, opt)
   {}
-  virtual ~LPO() {}
+  LPO(const DArray<int>& funcPrec, const DArray<int>& predPrec, const DArray<int>& predLevels, bool reverseLCM) :
```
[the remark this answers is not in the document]
mezpusz: What do you mean? I changed this for the unit testing actually.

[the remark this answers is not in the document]
MichaelRawson: That's what I meant - it looks similar to the KBO machinery, which I think is also for testing.

Kernel/LPO.cpp
mezpusz: Agreed, it should be quite trivial to catch these.

Kernel/LPO.cpp
mezpusz: Suppose some i exists such that s[i] > t. The issue is that in this case `lpo(t, s[i])` may still return `INCOMPARABLE` in some cases instead of `LESS` whereas with `lpo(s[i], t)` we should always get `GREATER` (assuming `lpo` abides by its specification). Does that make sense?

MichaelRawson: Ahah, OK, that makes sense! Looks like we were being too clever then.

Kernel/LPO.cpp
mezpusz: I also thought it was an issue but the `break` is for the `switch` confusingly... One issue though with `lexMAE` is that it should return `EQUAL` not `GREATER` after the loop by the original specification, but since we check for equality in `lpo` it shouldn't ever happen.

Kernel/LPO.cpp
```diff
@@ -196,30 +196,26 @@ Ordering::Result LPO::lpo(TermList tl1, TermList tl2) const
 {
   CALL("LPO::lpo(TermList, TermList)");
 
+  if(tl1==tl2) {
```
mezpusz, line 199: This check is not in the original specification but the original did not use term sharing either... Does this make sense here? Otherwise we have to return `EQUAL` in `lexMAE` as I wrote in the other thread.

Kernel/LPO.hpp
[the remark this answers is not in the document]
mezpusz: Yes, as far as I know, we either use options to get the precedences we want or have this ctor... either way works but I wanted to align with `tKBO.cpp`.

Kernel/LPO.cpp
mezpusz: Btw, does it make something less efficient that we create a `TermList` in two more cases in your opinion? Apart from the obvious creating one more object and some additional checks? Since we use `lpo` once with `(TermList, Term*)` and once with `(Term*, TermList)` either we implement it twice or have this version, but I would gladly hear that there is a third option.

Kernel/LPO.cpp
MichaelRawson: I think so! It should be a fairly fast check and it saves us recursing through terms uselessly. `return GREATER` at the end of `lexMAE` is confusing and looks like a bug though - could we safely make it `ASSERTION_VIOLATION` instead, with a comment explaining things? We know statically that `lexMAE` is only called from `lpo` after the equality check. Should also be a slight performance win as the compiler will know that it doesn't have to generate code for the loop exit (or loop condition, come to think of it).

Kernel/LPO.cpp
MichaelRawson: No, creating a `TermList` from a `Term *` is basically free I expect. It should be the size of a machine register with automatic storage, and the constructor just copies the pointer into a union. The compiler should completely elide it. The extra checks are very slightly less efficient but I expect some amount of compiler help/branch prediction with that. Overall what we have is fine, and I'll take correctness over (possibly, very slightly) a performance win and some maintenance burden.

MichaelRawson (approved):

MichaelRawson:
OK, looks good! @quickbeam123 and @selig - when you get around to it I think this is ready to go and has a really positive impact on LPO performance as a bonus. @mezpusz is running some tests (let us know when you're done!) to make sure the review changes didn't break anything, but I'd be surprised.

mezpusz:
> OK, looks good! @quickbeam123 and @selig - when you get around to it I think this is ready to go and has a really positive impact on LPO performance as a bonus. @mezpusz is running some tests (let us know when you're done!) to make sure the review changes didn't break anything, but I'd be surprised.

Tests are done, TPTP: 11962, SMT-LIB: 56055, diffs are 15 and 61 between the two tests, respectively.

MichaelRawson:
OK - inclined to think that's mostly experimental noise, but maybe one of the other two can confirm. Were there any extra crashes? If not then I think we're good. :-)

mezpusz:
> OK - inclined to think that's mostly experimental noise, but maybe one of the other two can confirm. Were there any extra crashes? If not then I think we're good. :-)

There was one segfault in SMT-LIB in the latest test, namely UFNIA::lahiri-cav09-storm-queries::usbsamp_example_cegar_2_2_3.smt2 but it does not reproduce and the logfile is empty... Maybe part of the random noise?

quickbeam123:
We will be doing more testing with Filip, so will come back if we get into problems.

Thank you, guys!

PR #312, opened 2022-01-14
no longer need to offset arguments for Z3/polymorphic equality proxy

MichaelRawson:
Now that we fixed `Z3Interfacing` for grounded polymorphic functions in #300, we no longer need to work around polymorphic equality proxy, and the workaround is now a bug. Remove it.


MichaelRawson:
@quickbeam123 - yep, on the current master try this input:
```
tff(type, type, f: $int > $int).
tff(ax,axiom, f(0) = 1).
```

and `-ep R -sas z3` and you should get a crash from reading past the end of the array [https://github.com/vprover/vampire/blob/master/SAT/Z3Interfacing.cpp#L918](here).

PR #313, opened 2022-01-17
Fix sym counter

ibnyusuf:
The symbol counter contained a number of hidden bugs which Spider brought to notice.

Leading amongst these was the use of an iterator over all subterms for non-shared terms which could result in an exponential number of calls to` count(..)` function.

### DIFF
--- a/Shell/SymCounter.cpp
+++ b/Shell/SymCounter.cpp
@@ -298,7 +298,6 @@ void SymCounter::count(Term* term, int polarity, int add)
     NonVariableIterator nvi(term);
     while (nvi.hasNext()) {
       Term *t = nvi.next().term();
-      ASS(t != term);
       int fun = t->functor();
       if(!t->isSort()){      
         ASS_REP(_noOfFuns > fun, t->toString());
--- a/Shell/SymCounter.cpp
+++ b/Shell/SymCounter.cpp
@@ -166,7 +166,7 @@ void SymCounter::count (Formula* f,int polarity,int add)
 
     case BOOL_TERM: {
       TermList ts = f->getBooleanTerm();
-      if (!ts.isTerm()) return;
+      if (!ts.isTerm()) return;     
       count (ts.term(), polarity, add);
       return;
     }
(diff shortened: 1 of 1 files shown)


Shell/SymCounter.cpp
```diff
@@ -267,18 +275,25 @@ void SymCounter::count(Term* term, int polarity, int add)
...
+        if(arg->isTerm()){
+          count(arg->term(), 1, add);
+        }
       }
     }
   } else {
     int fun = term->functor();
-    ASS_REP(_noOfFuns > fun, term->toString());
-    _funs[fun].add(add);
+    if(!term->isSort()){
```
MichaelRawson, line 290: We never have predicates in term position...right?

Shell/SymCounter.cpp
```diff
@@ -166,7 +166,7 @@ void SymCounter::count (Formula* f,int polarity,int add)
 
     case BOOL_TERM: {
       TermList ts = f->getBooleanTerm();
-      if (!ts.isTerm()) return;
+      if (!ts.isTerm()) return;
```
MichaelRawson, line 169: various bits of whitespace - delete to minimise diff

Shell/SymCounter.cpp
```diff
@@ -196,9 +196,16 @@ void SymCounter::count(Literal* l,int polarity,int add)
...
   if (!l->shared()) {
-    NonVariableIterator nvi(l);
-    while (nvi.hasNext()) {
-      count(nvi.next().term(), 1, add);
+    for(TermList* arg=l->args(); arg->isNonEmpty(); arg=arg->next()) {
+      if(arg->isTerm()){
+        count(arg->term(), 1, add);
+      }
+    }
+    if(l->isTwoVarEquality()){
```
MichaelRawson, line 204: Why is this special-cased exactly? I'm sure it makes sense but I should probably figure it out.

MichaelRawson (approved):
Looks generally sensible, just a few comments. :-)

Shell/SymCounter.cpp
ibnyusuf: Correct. In FOOL and HOL, the predicates in term positions are treated as special terms, not predicates.

Shell/SymCounter.cpp
ibnyusuf: In the two variable equality case, we cannot work out the sort of the variables, so have to store it as a separate field in the literal. In the polymorphic case, this field may be a complex term and therefore needs scanning.

Shell/SymCounter.cpp
ibnyusuf: Done

Shell/SymCounter.cpp
MichaelRawson: Aha, OK. I will keep an eye out for other such instances then.

PR #315, opened 2022-01-24
warn if fail to use Ordering::Result

MichaelRawson:
In #311, we had a bug in which a computed `Ordering::Result` was not used as intended. Introduce a `VWARN_UNUSED` macro (GCC, Clang) to warn about unused values and apply it to `Ordering::Result` so that it warns about similar code (and would have produced warnings for the recent bug).

Also open to suggestions about what else we could apply this to. Container types?

### DIFF
--- a/Kernel/LPO.hpp
+++ b/Kernel/LPO.hpp
@@ -45,19 +45,19 @@ public:
   ~LPO() override = default;
 
   using PrecedenceOrdering::compare;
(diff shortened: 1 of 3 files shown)


quickbeam123:
Interesting! So this can be used also to types not just functions?

MichaelRawson:
Oops, looks like types are clang-only on closer inspection. But yes!

PR #319, opened 2022-02-01
fixes found by clang-tidy

MichaelRawson:
I ran `clang-tidy` again, and both Vampire and the tool have changed so I found some more (potential) problems. Mostly these are just opportunities for adding assertions, but there's one or two places where something is dubious. I will add comments here to help reviewers.

### DIFF
--- a/CASC/PortfolioMode.cpp
+++ b/CASC/PortfolioMode.cpp
@@ -582,7 +582,7 @@ void PortfolioMode::runSlice(Options& strategyOpt)
   if(outputResult) { // this get only true for the first child to find a proof
     ASS(!resultValue);
 
-    if (outputAllowed() && (Lib::env.options && Lib::env.options->multicore() != 1)) {
+    if (outputAllowed() && env.options->multicore() != 1) {
       env.beginOutput();
       addCommentSignForSZS(env.out()) << "First to succeed." << endl;
       env.endOutput();
--- a/Kernel/KBO.cpp
+++ b/Kernel/KBO.cpp
@@ -549,7 +549,7 @@ void KBO::checkAdmissibility(HandleError handle) const
   for (FunctionSymbol i = 0; i < nFunctions; i++) {
     auto sort = env.signature->getFunction(i)->fnType()->result();
     /* register min function */
-    auto maxFn = maximalFunctions.getOrInit(std::move(sort), [&](){ return i; } );
+    auto maxFn = maximalFunctions.getOrInit(sort, [&](){ return i; } );
     if (compareFunctionPrecedences(maxFn, i) == LESS) {
       maximalFunctions.replace(sort, i);
     }
--- a/Kernel/PolynomialNormalizer.cpp
+++ b/Kernel/PolynomialNormalizer.cpp
@@ -192,6 +192,7 @@ Option<NormalizationResult> normalizeSpecializedFractional(Interpretation i, Nor
 {
   switch (i) {
     case NumTraits::divI:
+      ASS(ts != nullptr);
       return normalizeDiv<NumTraits>(ts[0], ts[1]);
     default:
       return Option<NormalizationResult>();
@@ -286,10 +287,13 @@ NormalizationResult normalizeNumSort(TermList t, NormalizationResult* ts)
     if (fn.isInterpreted()) {
       switch(fn.interpretation()) {
         case NumTraits::mulI:
+          ASS(ts != nullptr);
           return normalizeMul<NumTraits>(ts[0], ts[1]);
         case NumTraits::addI:
+          ASS(ts != nullptr);
           return normalizeAdd<NumTraits>(ts[0], ts[1]);
         case NumTraits::minusI:
+          ASS(ts != nullptr);
           return normalizeMinus<NumTraits>(ts[0]);
         default:
         {
--- a/Kernel/RobSubstitution.hpp
+++ b/Kernel/RobSubstitution.hpp
@@ -46,7 +46,7 @@ public:
   CLASS_NAME(RobSubstitution);
   USE_ALLOCATOR(RobSubstitution);
   
-  RobSubstitution() : _nextUnboundAvailable(0) {} //,_nextAuxAvailable(0) {}
+  RobSubstitution() : _funcSubtermMap(nullptr), _nextUnboundAvailable(0) {}
 
   SubstIterator matches(Literal* base, int baseIndex,
 	  Literal* instance, int instanceIndex, bool complementary);
--- a/Kernel/SubstHelper.hpp
+++ b/Kernel/SubstHelper.hpp
@@ -80,12 +80,9 @@ public:
   static Literal* apply(Literal* lit, Applicator& applicator)
   {
     CALL("SubstHelper::apply(Literal*...)");
-    TermList sort;
-    if(lit->isTwoVarEquality()){
-      sort = lit->twoVarEqSort();
-    }
     Literal* subbedLit = static_cast<Literal*>(apply(static_cast<Term*>(lit),applicator));
     if(subbedLit->isTwoVarEquality()){ //either nothing's changed or variant
+      TermList sort = lit->twoVarEqSort();
       TermList newSort = apply(sort, applicator);
       if((sort != newSort)){
         subbedLit = Literal::createEquality(subbedLit->polarity(), *subbedLit->nthArgument(0), *subbedLit->nthArgument(1), newSort);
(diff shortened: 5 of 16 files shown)


CASC/PortfolioMode.cpp
```diff
@@ -582,7 +582,7 @@ void PortfolioMode::runSlice(Options& strategyOpt)
   if(outputResult) { // this get only true for the first child to find a proof
     ASS(!resultValue);
 
-    if (outputAllowed() && (Lib::env.options && Lib::env.options->multicore() != 1)) {
```
MichaelRawson, line 585: We use `env.options` above and below here, so we know it isn't `nullptr`. `Lib::env` and `env` are the same, I think.

Kernel/KBO.cpp
```diff
@@ -549,7 +549,7 @@ void KBO::checkAdmissibility(HandleError handle) const
   for (FunctionSymbol i = 0; i < nFunctions; i++) {
     auto sort = env.signature->getFunction(i)->fnType()->result();
     /* register min function */
-    auto maxFn = maximalFunctions.getOrInit(std::move(sort), [&](){ return i; } );
```
MichaelRawson, line 552: `sort` has type `TermList`, and that's trivially-copyable, so moving it is not required and we use it again below.

Kernel/PolynomialNormalizer.cpp
```diff
@@ -192,6 +192,7 @@ Option<NormalizationResult> normalizeSpecializedFractional(Interpretation i, Nor
 {
   switch (i) {
     case NumTraits::divI:
```
MichaelRawson, line 194: In these cases `ts` might be `nullptr` in a convoluted way if there are no sub-terms. Assert that it is not when we expect subterms.

Kernel/RobSubstitution.hpp
```diff
@@ -46,7 +46,7 @@ class RobSubstitution
   CLASS_NAME(RobSubstitution);
   USE_ALLOCATOR(RobSubstitution);
   
-  RobSubstitution() : _nextUnboundAvailable(0) {} //,_nextAuxAvailable(0) {}
+  RobSubstitution() : _funcSubtermMap(nullptr), _nextUnboundAvailable(0) {}
```
MichaelRawson, line 49: POD types are not automatically initialised in constructors. This will cause a crash more readily if `_funcSubtermMap` is not set before use somehow.

Kernel/SubstHelper.hpp
```diff
@@ -80,12 +80,9 @@ class SubstHelper
   static Literal* apply(Literal* lit, Applicator& applicator)
   {
     CALL("SubstHelper::apply(Literal*...)");
```
MichaelRawson, line 82: We only actually use `sort` in the case when `subbedLit` is still a two-variable equality. Move it into there, save some compute and avoid the uninitialised mess.

Kernel/TermIterators.hpp
```diff
@@ -469,7 +469,7 @@ class FirstOrderSubtermIt
     } 
     _stack.push(term);
     if (!includeSelf) {
```
MichaelRawson, line 471: Dynamic dispatch is disabled in constructors. Make this explicit.

Lib/Coproduct.hpp
```diff
@@ -275,8 +275,8 @@ class Coproduct<A, As...>
   using Self = Coproduct<A, As...>;
   using Content = decltype(_content);
 
-  /** unsafe default constructor. tag as well as content will be uninit */
-  Coproduct() {}
+  /** unsafe default constructor, content will be uninit */
+  Coproduct() : _tag(std::numeric_limits<unsigned>::max()) {}
```
MichaelRawson, line 279: Try and make this as likely as possible to blow up if we don't initialise it correctly.

Lib/Coproduct.hpp
```diff
@@ -234,7 +234,7 @@ namespace CoproductImpl {
       using Ts = TL::List<As...>;                                                                             \
       auto unwrap = Unwrap<acc, Ts>{};                                                                        \
       if (acc == idx) {                                                                                       \
-        ::new (&self) TL::Get<acc,Ts>(unwrap(MOVE(value)));                                                   \
+        ::new (&self._head) TL::Get<acc,Ts>(unwrap(MOVE(value)));                                             \
```
MichaelRawson, line 237: So I'm a little unsure what the language spec says on this, but the linter is confused and so am I. I think (same for `Option` as well) that we want to assign to one specific field of the union, not the whole thing - even though because of the memory layout that's the same thing. Did I get the right field?

Lib/List.hpp
```diff
@@ -649,6 +649,7 @@ class List
 	return;
       }
 
+      ASS_NEQ(_prev,0);
```
MichaelRawson, line 652: In this case we know that `_prev` must not be `nullptr`, but it's not a trivial invariant in some cases.

Lib/Option.hpp
```diff
@@ -33,7 +33,7 @@ template<class T>
 struct MaybeUninit {
   union Value { 
     T init; int uninint[0]; 
-     Value() {};
+    Value() : uninint{} {};
```
MichaelRawson, line 36: Make it explicit which field we expect to be active while uninitialised.

Lib/Option.hpp
```diff
@@ -45,7 +45,7 @@ struct MaybeUninit {
   { return MV(_elem.init); }                                                                                  \
                                                                                                               \
   void init(T REF content)                                                                                    \
-  { ::new(&_elem)T(MV(content)); }                                                                            \
+  { ::new(&_elem.init)T(MV(content)); }                                                                       \
```
MichaelRawson, line 48: Same as with `Coproduct` - I think we want to assign to the union field, not the whole thing.

SAT/Z3Interfacing.cpp
```diff
@@ -436,8 +436,10 @@ struct EvaluateInModel
     } else if (expr.is_app()) {
       auto f = expr.decl();
       auto vfunc = self._fromZ3.get(f);
```
MichaelRawson, line 438: Again, `evaluatedArgs` might be `nullptr` in some cases. Assert that it isn't when we need them.

Test/SimplificationTester.hpp
```diff
@@ -41,7 +41,7 @@ class Success
   Option<ClausePattern> _expected;
 
 public:
-  Success() {}
+  Success() : _input(nullptr) {}
```
MichaelRawson, line 44: Make this blow up noisily if we mess up later.

UnitTests/tBottomUpEvaluation.cpp
```diff
@@ -58,24 +58,26 @@ TEST_FUN(example_02__compute_size) {
...
   DECL_DEFAULT_VARS
   DECL_SORT(s)
-  DECL_CONST(a, s)
   DECL_FUNC(f, {s}, s)
   DECL_FUNC(g, {s,s}, s)
 
   /* defines how to evaluate bottom up. 
    * computes the size of the term (number of function & variable symbols) */
   struct Eval {
-    TermList replacement;
```
MichaelRawson, line 68: Never used.

UnitTests/tGaussianElimination.cpp
```diff
@@ -54,7 +54,7 @@ class GveSimplTester : public Test::Simplification::SimplificationTester
     static GaussianVariableElimination gve = GaussianVariableElimination();
 
     /* applies gve and evaluation until they're not applicable anymore */
-    Kernel::Clause* last = simpl(in);
+    Kernel::Clause* last = nullptr;
```
MichaelRawson, line 57: Dead-store eliminated because of the do-while loop below.

Lib/Coproduct.hpp
joe-hauns: Yes absolutely! This was definitely not intended!

UnitTests/tOption.cpp
```diff
@@ -126,7 +126,7 @@ class LeaqueChecker
     _moved = o._moved;
     return *this; 
   }
-  LeaqueChecker()  { instances++; }
+  LeaqueChecker() : _moved(false) { instances++; }
```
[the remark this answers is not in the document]
joe-hauns: Right! I think it probably worked because we run tests in debug mode only, where arbitrary initialization happend consistently (if I rember correctly)!

joe-hauns (approved):
Went over all the changes and looks all great!

PR #320, opened 2022-02-19
Nice options

quickbeam123:
Various cleanups around the Options module. 
Mostly more stringent option constrains so that dependencies are explicit.


Shell/Options.cpp
```diff
@@ -133,6 +133,10 @@ void Options::init()
     _lookup.insert(&_mode);
     _mode.addHardConstraint(If(equal(Mode::CONSEQUENCE_ELIMINATION)).then(_splitting.is(notEqual(true))));
 
+    auto UsingCascTechnology = [this] {
```
selig, line 136: Is this CASC technology or Portfolio technology? I think the latter.  Perhaps add a comment next to the mode enum saying that this may need updating if a new portfolio-based mode is added.

Shell/Options.cpp
```diff
@@ -1019,418 +1068,423 @@ void Options::init()
...
+    "\n"
+    "\nThe different option values define the behaviour of which theory literals to select."
+    "\n- all    : hmmm.. what could that mean?!"
+    "\n- neg_eq : only negative equalities"
+    "\n- strong : interpreted predicates, but no positive equalites"
+    "\n- overlap: all literals that contain variables that are also contained in a strong literal"
+    "\n- new    : deprecated"
+    "\n- full   : deprecated"
+    "";
+    _theoryInstAndSimp.tag(OptionTag::INFERENCES);
```
selig, line 1092: Add problem  constraints for hasTheories?

Shell/Options.cpp
```diff
@@ -1019,418 +1068,423 @@ void Options::init()
...
+      "- one_side_constant: only if one of s or t is an interpreted constant (e.g. a number)\n"
+      "- all: always apply\n"
+      "- ground: only if both s and t are ground\n"
+      "See Unification with Abstraction and Theory Instantiation in Saturation-Based Reasoning for further details.";
+    _unificationWithAbstraction.tag(OptionTag::INFERENCES);
+    _lookup.insert(&_unificationWithAbstraction);
+
+    _useACeval = BoolOptionValue("use_ac_eval","uace",true);
+    _useACeval.description="Evaluate associative and commutative operators e.g. + and *.";
+    _useACeval.tag(OptionTag::INFERENCES);
```
selig, line 1140: Add problem constraint for  hasTheories?

selig:
This is some nice tidying.

If you care about problem constraints then there are quite a few that could be added related to theories.

The higher-order options are currently free of constraints. Perhaps this is a good reminder to do this? I'm guessing you're not looking at the higher-order part of  Vampire right now but it would be really useful to Spider this at some point.

Shell/Options.cpp
```diff
@@ -1019,418 +1068,423 @@ void Options::init()
...
+    { return ChoiceOptionValue<ArithmeticSimplificationMode>(l,s,d, {"force", "cautious", "off", }); };
+    _cancellation = choiceArithmeticSimplificationMode(
+       "cancellation", "canc",
+       ArithmeticSimplificationMode::OFF);
+    _cancellation.description = "Enables the rule cancellation around additions as described in the paper Making Theory Reasoning Simpler ( https://easychair.org/publications/preprint/K2hb ). \
+                                In some rare cases the conclusion may be not strictly simpler than the hypothesis. With `force` we ignore these cases, violating the ordering and just simplifying \
+                                anyways. With `cautious` we will generate a new clause instead of simplifying in these cases.";
+    _lookup.insert(&_cancellation);
+    _cancellation.tag(OptionTag::INFERENCES);
+    _cancellation.setExperimental();
```
joe-hauns, line 1158: I think we could remove these setExperimental calls as the rules are in master, and in a paper, right?

joe-hauns: I said these, but i actually just meant this one.

Shell/Options.cpp
joe-hauns: Yeah I think that's a shout. For this one as well as for 
- _inequalityNormalization
- _unificationWithAbstraction
- _cancellation
- _highSchool
- _gaussianVariableElimination
- _pushUnaryMinus
- _arithmeticSubtermGeneralizations

joe-hauns (approved):
Looks all sweet to me as well! 
Noted the other options that could potentially be equipped with hasTheories, and besides that I don't have anything to note :)

quickbeam123:
Thanks for the feedback. This was definitely not meant to be finished when submitted. More as a first of several iterations. Awareness raising ;)

Feel free to add more reasonable constraints and flags. Or we can have another PR later...

Organising some such "ontology" around the HO options would be great! @ibnyusuf , could you find some time...

What I discovered for myself through this work is that we are actually (with the current capabilities of the Options' "language", thanks to @selig 's effort from some time ago) already quite close to automatically generating a full "parameter configuration space" (PCS) file for some third-party tools such as ParamILS or SMAC. Not that we would immediately want to do just that, but being able to "push-button"-communicate a PCS even to our own tools is an advantage, since the options slowly change during the developments and new spidering efforts normally (I assume) consist of some painful manual alignments after each dormant phase.

quickbeam123:
"Ontology" is probably a bit elevated term for what I had in mind. Basically, it's just about adding
`addProblemConstraint`, `reliesOn` (and perhaps `addHardConstraint`/`addConstraint`) whenever they make sense.

The first will give you ability to make spider-like tools (and the user as well) happy, by telling them that changing a HO option for a first-order problem won't help. (Maybe you will need to add a `OptionProblemConstraint` descendant for "isHO".) 

The second you already use, I see, but maybe there are some other implicit incompatibilities not yet specified? Like does it make sense to build a HO strategy that uses equality_proxy? Or what about HO-inferences and instgen as the `sa`, etc. Sometimes we already bail out from running such combinations, but the user error is elsewhere and not "known" to the Options module.

Shell/Options.cpp
```diff
@@ -3250,16 +3335,28 @@ bool Options::checkGlobalOptionConstraints(bool fail_early)
...
    CALL("Options::checkProblemOptionConstraints");
 
   bool result = true;
 
   VirtualIterator<AbstractOptionValue*> options = _lookup.values();
   while(options.hasNext()){
-    result = options.next()->checkProblemConstraints(prop) && result; 
+    AbstractOptionValue* opt = options.next();
+
+    if ((before_preprocessing && opt->getTag() != OptionTag::PREPROCESSING) ||
```
MichaelRawson, line 3354:
```
bool tagIsPreprocessing = opt->getTag() == OptionTag::Preprocessing;
if(before_processing != tagIsPreprocessing)
```
? Just a matter of taste.

Shell/Options.cpp
```diff
@@ -133,6 +133,15 @@ void Options::init()
     _lookup.insert(&_mode);
     _mode.addHardConstraint(If(equal(Mode::CONSEQUENCE_ELIMINATION)).then(_splitting.is(notEqual(true))));
```
MichaelRawson, line 134: Is it worth specialising these `equal(true)` calls to something like `_splitting.isTrue()`?

MichaelRawson (approved):
Looks good. The only slight consideration is compile times for `Shell::Option`, but they're not obscene yet so it'll be OK.

Shell/Options.cpp
quickbeam123: It took me a while to get what you meant, but I agree your solution is easier to read in the end! ;)

Shell/Options.cpp
quickbeam123: It could be, but I am too lazy to do it at the moment. Looking at `is`

```
template<typename T>
Options::AbstractWrappedConstraintUP Options::OptionValue<T>::is(OptionValueConstraintUP<T> c)
{
    return AbstractWrappedConstraintUP(new WrappedConstraint<T>(*this,std::move(c)));
}
```

what should be the specialization? ;)

quickbeam123:
Thank you, @ibnyusuf - looks like a nice addition!

Issue #322, opened 2022-02-22
The symbol $sum is not used with a single sort


stanmoon:
I am trying to write simple proofs using sorts but cannot figure it out how to use equality for constants and integer functions.
Below a simple example:

`fof(simple_conj,conjecture,(
! [X: $int] : $sum(X,0) = X
)).`

Vampire output is as follows:

`The symbol $sum is not used with a single sort`

I'm using Vampire 4.5.1 (commit 57a6f78c on 2020-07-15 11:59:04 +0200)

Thanks!

selig:
Try

`tff(simple_conj,conjecture,( ! [X: $int] : $sum(X,0) = X )).`

In fof mode Vampire will treat `$sum` as an uninterpreted function.

The error message does need improving!

selig:
Correction. In fof mode, Vampire will treat `0` as an uninterpreted constant. `$sum` will be treated as interpreted.  The complaint is that `$sum` is being called on an uninterpreted input.

I have created a pull request (in progress) for a better error message. The improvement is currently meant to be

> The interpreted function symbol $sum is not used with a single sort.
> Argument 0 is $int and argument 1 is $i
> Check that you are using tff if you want numbers to be interpreted

Please say if you think this is still confusing.

stanmoon:
Thank you, that explains it very well!

PR #323, opened 2022-02-23
Improving the error message when an interpreted symbol is used incorr…

selig:
Small improvement in error messages in response to Issue #322.

TODO:  is _isFof always set and used properly here?


selig:
Actually, this doesn't work yet as we step through arguments backward so it should print

> Argument 0 is $int and argument 1 is $i

but actually prints

> Argument 0 is $i and argumet 1 is $int

The most intuitive thing from a user perspective would be to step through arguments left-to-right but that's going to be annoying so perhaps I should just fix the way indices are reported i.e.

> Argument 1 is $i and argumet 0 is $int

quickbeam123:
Sorry, I already pulled. But feel free to continue with the branch and create another PR later.

PR #326, opened 2022-02-25
remove references to EGSubst

MichaelRawson:
Back in the dawn of time, there was [apparently](https://link.springer.com/chapter/10.1007%2F978-3-642-04617-9_55) several unification algorithms in Vampire. Remove references to an old one.


MichaelRawson:
For GitHub, ignore. Related to #157.

easychair:
They were used for experiments on comparison of unification algorithms, as
published. Indeed, one is enough: simple original Robinson's, it is
slightly better than the rest of them, on the average, and also simpler.

A

PR #328, opened 2022-03-04
Smaller refactors in Induction code

mezpusz:
- use TA ctor's `recursive()` instead of manually computing whether a ctor has recursive argument
- use `List` helper functions to avoid `0`s and `new`s
- use `JunctionFormula::generalJunction` where otherwise we branch on the length of the list argument anyways
- replace `0`s to `nullptr`s

### DIFF
--- a/Inferences/Induction.cpp
+++ b/Inferences/Induction.cpp
@@ -274,11 +274,11 @@ void Induction::detach() {
   CALL("Induction::detach");
 
   if (InductionHelper::isIntInductionOn()) {
-    _comparisonIndex = 0;
+    _comparisonIndex = nullptr;
     _salg->getIndexManager()->release(UNIT_INT_COMPARISON_INDEX);
   }
   if (InductionHelper::isIntInductionTwoOn()) {
-    _inductionTermIndex = 0;
+    _inductionTermIndex = nullptr;
     _salg->getIndexManager()->release(INDUCTION_TERM_INDEX);
   }
   GeneratingInferenceEngine::detach();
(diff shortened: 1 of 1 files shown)


MichaelRawson:
Looks great! Assuming that you haven't broken anything since you're working on this yourself. Thank you so much for the code-quality improvements, it really does make a difference. :-)

While you're at it - if there are any remaining '0's used to mean empty-list, maybe they could be `nullptr`?

mezpusz:
> Looks great! Assuming that you haven't broken anything since you're working on this yourself. Thank you so much for the code-quality improvements, it really does make a difference. :-)
> 
> While you're at it - if there are any remaining '0's used to mean empty-list, maybe they could be `nullptr`?

How about `List::empty()`? Here and there the scheme `ASS(list);` is also used to check for non-emptiness, what about those? To me, they look informative enough but we can go for `ASS_NE(list, List::empty())` if we want. :)

MichaelRawson:
Even better! Then if we ever change how `List` is implemented it won't break. `ASS(list)` is probably fine, we can make the non-empty list truthy if we ever need to change that.

MichaelRawson (approved):

quickbeam123 (approved):
I am happy to push the button. It's just seems you still wanted to change

`nullptr` to `List::empty()`

?

mezpusz:
> I am happy to push the button. It's just seems you still wanted to change
> 
> `nullptr` to `List::empty()`
> 
> ?

Hm, I think I changed those (actually, only `FormulaList::empty()`s were involved) but let me know if I missed one in Induction.cpp.

quickbeam123:
Sorry, I thought there were some others. My bad!

PR #330, opened 2022-03-07
Nice skolemization

quickbeam123:
Just some comments and a (performance) fix for the repeated normalization taking too long on a large quantifier blocks under skolem reuse (skr). The fix is different in Skolem (store just one symbol and reuse all -- relying on consecutivity of allocation (which could probably fail in HOL!)) and in NewCnf (where I just skip skr for blocks larger than 5 variables).

### DIFF
--- a/Shell/NewCNF.cpp
+++ b/Shell/NewCNF.cpp
@@ -993,7 +993,7 @@ Term* NewCNF::createSkolemTerm(unsigned var, VarSet* free, Formula *reuse_formul
   ASS(domainSorts.isEmpty());
   ASS(fnArgs.isEmpty());
 
-  NameReuse *name_reuse = env.options->skolemReuse()
+  NameReuse *name_reuse = (env.options->skolemReuse() && reuse_formula)
     ? NameReuse::skolemInstance()
     : nullptr;
   unsigned reused_symbol = 0;
@@ -1113,7 +1113,7 @@ void NewCNF::skolemise(QuantifiedFormula* g, BindingList*& bindings, BindingList
       Formula *reuse_formula = nullptr;
       VList *remainingVars = nullptr;
       SList *remainingSorts = nullptr;
-      if(env.options->skolemReuse()) {
+      if(env.options->skolemReuse() && VList::length(g->vars()) <= 5) { // give up on skolemReuse for long quantifier blocks
         BindingList::Iterator bit(bindings);
         while (bit.hasNext()) {
           Binding b = bit.next();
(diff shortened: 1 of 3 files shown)


Shell/Skolem.cpp
```diff
@@ -415,13 +427,9 @@ Formula* Skolem::skolemise (Formula* f)
         dep = dep->getUnion(their_dep);
       }
 
-      /*
```
MichaelRawson, line 418: A great loss!

Shell/NewCNF.cpp
```diff
@@ -993,7 +993,7 @@ Term* NewCNF::createSkolemTerm(unsigned var, VarSet* free, Formula *reuse_formul
   ASS(domainSorts.isEmpty());
   ASS(fnArgs.isEmpty());
 
-  NameReuse *name_reuse = env.options->skolemReuse()
+  NameReuse *name_reuse = (env.options->skolemReuse() && reuse_formula)
```
MichaelRawson, line 996: Isn't `reuse_formula` only ever `nullptr` if `!env.options->skolemReuse()`? Think this might be a copy/paste error from the other fix.

Shell/NewCNF.cpp
```diff
@@ -1113,7 +1113,7 @@ void NewCNF::skolemise(QuantifiedFormula* g, BindingList*& bindings, BindingList
       Formula *reuse_formula = nullptr;
       VList *remainingVars = nullptr;
       SList *remainingSorts = nullptr;
-      if(env.options->skolemReuse()) {
+      if(env.options->skolemReuse() && VList::length(g->vars()) <= 5) { // give up on skolemReuse for long quantifier blocks
```
MichaelRawson, line 1116: If we were very keen on Spidering this could be a parameter...but let's not.

Shell/NewCNF.cpp
```diff
@@ -1146,7 +1146,7 @@ void NewCNF::skolemise(QuantifiedFormula* g, BindingList*& bindings, BindingList
         // ?[Y, Z]: F[X->sK0],
         // ?[Z]: F[X->sK0, Y->sK1],
         // but not F[X->sK0, Y->sK1, Z->sK2], since this doesn't need a Skolem term
-        if(env.options->skolemReuse()) {
+        if(env.options->skolemReuse() && reuse_formula) {
```
MichaelRawson, line 1149: See above.

Shell/Skolem.cpp
```diff
@@ -462,37 +469,43 @@ Formula* Skolem::skolemise (Formula* f)
       }
       SortHelper::normaliseArgSorts(typeVars, termVarSorts);
 
+      /*
```
MichaelRawson, line 472: This is sneaky, I love it.

MichaelRawson (approved):
Looks great! Thank you for removing an item from my to-do list. :-) Only some minor comments inline.

I was initially upset that if e.g. the second variable in a block could have been reused but the first could not, then we lose an opportunity, but then I realised that the first variable would have to not occur in the key formula for that to happen. At some point I will work out whether `Rectify` already removes such bindings, and if not fix it then.

Is there a reason the NewCNF fix is not the same?

Shell/NewCNF.cpp
quickbeam123: `(reuse_formula == nullptr)` with `env.options->skolemReuse()` is the bailout situation with `(VList::length(g->vars()) > 5)`

PR #331, opened 2022-03-07
Integer induction options changes + small induction bug fix

hzzv:
This PR includes:

1. Changing what -iik refers to. Previously this option controlled whether integer induction was triggered when the main literal was selected (value "one"), or when the comparison (bound) literal was selected ("two"), or in both cases ("all"). This wasn't very useful, and should be safe to change (the option is not referenced in any paper). After this change the option refers to the implementation of integer induction that is used - the value "one" being the one in Induction.cpp, the value "two" will refer to the implementation in GeneralInduction.cpp (once/if that gets merged).
2. Adding integer induction option --int_induction_strictness, which encodes types of literals and terms, on which is induction (dis)allowed. The default value 11 roughly corresponds to the condition that literals/terms had to satisfy before adding this option. For a detailed description of the strictness values see Inferences/InductionHelper.cpp or Shell/Options.cpp
3. Fixing a bug, where we used upward induction rule enums (e.g. InferenceRule::INT_INF_UP_INDUCTION_AXIOM) when a downward induction rule enum was correct (e.g. InferenceRule::INT_INF_DOWN_INDUCTION_AXIOM).
4. Changing integer induction schedules to reflect the changes in points 1 and 2 above. I also changed the sanity checks to use the portfolio mode for integer induction problems.
5. Adding tests for integer induction, also covering points 2 and 3 above. I also added the possibility to test post-conditions and pre-conditions in the TestCase class - this is useful for checking values of global variables (e.g. env.statistics->induction) pre- and post-run.

To verify the effect of these changes, I also ran an experiment with the {UF,""}{LIA,NIA} benchmarks and our inductive int benchmarks. Changes from points 1 and 2 are pretty much neutral w.r.t. the number of problems solved uniquely and at the same time using induction; they are slightly positive w.r.t the number of problems solved using induction, and slightly negative w.r.t. the number of problems solved overall. I think this is a reasonable result for changing the -iik option. Before the default was -iik one, and thus we did not apply induction if the main literal was selected before the comparison literal (i.e., when the comparison literal was not indexed before we selected the main literal, we gave up on induction on that main literal even if we later selected the comparison literal). This was confusing and not too helpful - if we turn on induction, then we want to apply it indiscriminately regardless of the order in which the premises were selected.
Additionally, I ran an experiment in the portfolio mode, comparing the previous induction schedule with the new one, only on 222 benchmarks that were uniquely solved using induction in the previous experiment. The new schedule is better (it solves 202 problems, compared to the old schedule which solves 182 problems, or to the casc schedule + induction, which solves 179 problems).


Shell/Options.cpp
```diff
@@ -1307,6 +1307,13 @@ void Options::init()
     _integerInductionInterval.reliesOn(Or(_induction.is(equal(Induction::INTEGER)),_induction.is(equal(Induction::BOTH))));
     _lookup.insert(&_integerInductionInterval);
 
+    _intInductionStrictness = UnsignedOptionValue("int_induction_strictness","intindst",11);
+    _intInductionStrictness.description = "Encodes term/literal combinations that are excluded from int induction. See InductionHelper.cpp for explanation.";
```
hzzv, line 1311: Should I also add the complete explanation of the values here?

Inferences/InductionHelper.cpp
```diff
@@ -139,15 +140,13 @@ bool InductionHelper::isIntInductionOn() {
 
 bool InductionHelper::isIntInductionOneOn() {
   CALL("InductionHelper::isIntInductionOneOn");
-  static bool one = env.options->intInduction() == Options::IntInductionKind::ONE ||
-                    env.options->intInduction() == Options::IntInductionKind::ALL;
+  static bool one = env.options->intInduction() == Options::IntInductionKind::ONE;
```
MichaelRawson, line 143: Could this just be a non-static `bool` in these cases since we're changing it? The corresponding load/eq instructions are likely to be faster as statics are surprisingly expensive.

Inferences/InductionHelper.cpp
```diff
@@ -209,22 +208,61 @@ bool InductionHelper::isInductionTermFunctor(unsigned f) {
...
+bool containsSkolem(Term* t) {
+  if (env.signature->getFunction(t->functor())->skolem()) return true; 
+  NonVariableIterator nvi(t);
+  while (nvi.hasNext()) {
+    if (env.signature->getFunction(nvi.next().term()->functor())->skolem()) return true;
+  }
+  return false;
+}
+
+bool termAndLiteralSatisfyStrictness(const TermList& tl, Literal* l, unsigned strictness) {
```
MichaelRawson, line 220: Could we introduce a "strictness" enumeration and/or make this a `switch` statement? It seems like there are only a finite number of permitted values with fixed semantics.

Inferences/InductionHelper.cpp
```diff
@@ -209,22 +208,61 @@ bool InductionHelper::isInductionTermFunctor(unsigned f) {
...
+      ((termst >= 1) && theory->isInterpretedConstant(f)) ||
+      ((termst == 2) && !containsSkolem(tl.term())))
   {
     return false;
   }
-  // Integer term tl from literal l cannot be used for induction if l is an integer
-  // comparison and tl is one of the comparison's arguments.
-  if (isIntegerComparisonLiteral(l) &&
-      ((tl == *l->nthArgument(0)) || (tl == *l->nthArgument(1)))) {
+  const bool isEquality = (theory->isInterpretedPredicate(l) && l->isEquality());
```
MichaelRawson, line 263: Isn't equality always an interpreted predicate? i.e. would `isEquality = l->isEquality()` suffice?

Inferences/InductionHelper.cpp
```diff
@@ -209,22 +208,61 @@ bool InductionHelper::isInductionTermFunctor(unsigned f) {
...
+  if ((isEquality && !termAndLiteralSatisfyStrictness(tl, l, eqst)) ||
```
MichaelRawson, line 264: `return !<condition>`, right?

Shell/Options.cpp
MichaelRawson: If you have the time/inclination that would be nice from an end-user perspective! Probably just copying in what you already have in a comment would be fine.

checks/sanity
```diff
@@ -56,11 +56,11 @@ check_smtcomp_status unsat --mode smtcomp ucore/test1.smt2
 check_exact_output ucore/test2.out --input_syntax smtlib2 -om ucore ucore/test2.smt2
 
 # Integer induction problems
-check_szs_status Unsatisfiable -ind int ind/int_invariant_infinite_geq3_val3.smt2
-check_szs_status Unsatisfiable -ind int ind/int_invariant_finite_a_to_b.smt2
-check_szs_status Unsatisfiable --forced_options "drc=off:ind=int:intindint=infinite:sos=theory:sstl=1:to=lpo" ind/int_power_0_all_0.smt2
-check_szs_status Unsatisfiable --forced_options "ind=int:sa=discount:sos=theory:sstl=1:to=lpo" ind/int_sum_y_geq_0.smt2
+check_szs_status Unsatisfiable --mode portfolio --schedule integer_induction --slowness 1 ind/int_invariant_infinite_geq3_val3.smt2
```
MichaelRawson, line 59: I think I would prefer invoking the exact strategy from this portfolio that solves the problem, otherwise this might break when we change the schedules. But I don't feel particularly strongly one way or another. :-)

MichaelRawson (approved):
Looks nice to me! Thank you for taking the time to do cleanup. Some minor comments inline.

Shell/Options.cpp
hzzv: Done.

Inferences/InductionHelper.cpp
hzzv: This is a good question and I'm not sure what's the best solution. This function only checks a part of the strictness condition - for either equality or comparison.
There are 5 permitted values for equalities, (the same) 5 for comparisons, and 3 for terms, so altogether it's 75. I think having 75 values for this option is not really practical.
Another possibility would be to keep the option as an unsigned value, but then convert its sub-values for equalities, comparisons and terms into enums. I'm not sure though whether this would be much better than what we have now (converting the sub-values into unsigneds).
The last possibility I see would be to split this option into 3 options, each one an enum, but that would (unnecessarily, seems to me) add 3 options when 1 is sufficient.
What do you think?

Inferences/InductionHelper.cpp
hzzv: Yes, thanks, done!

Inferences/InductionHelper.cpp
hzzv: Yes, done.

checks/sanity
hzzv: I'd actually prefer to use portfolio. The proofs that Vampire produces for these problems have up to 100 steps (it's not easy to find problems that require various types of integer induction and have short proofs), which means that if something unrelated changes in the proof search, Vampire might diverge and not find the proof. See Slack: https://vampireglobal.slack.com/archives/C014NQS2UNL/p1619758673000900
If we use the portfolio mode, there should always be a couple of schedules that solve each of the problems, so the sanity check won't break that easily.

Inferences/InductionHelper.cpp
```diff
@@ -209,22 +208,61 @@ bool InductionHelper::isInductionTermFunctor(unsigned f) {
          );
 }
 
+bool containsSkolem(Term* t) {
+  if (env.signature->getFunction(t->functor())->skolem()) return true;
```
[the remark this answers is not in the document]
hzzv: What about this?
Since induction is by default only applied to simple terms, let's have a special case for that (so we avoid creating the iterator). Otherwise (if t is complex) we use NonVariableIterator.

hzzv:
Thanks for such a quick review, Michael! I incorporated most of the comments, what do you think about the rest?

Inferences/InductionHelper.cpp
MichaelRawson: Thanks, pet peeve. :-) It's not a ridiculous overhead so they definitely have their place, but I think it's more than people expect given that it's a keyword. Post C++11 it has to be thread-safe as well. Have a play with https://godbolt.org/z/feTzv4oE3 if you're interested.

Inferences/InductionHelper.cpp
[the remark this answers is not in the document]
MichaelRawson: Ahah, OK - I see the logic now! My mistake. The original is probably fine then, but it looks like this is a further improvement.

Just while we're here, could the two helper functions `containsSkolem` and `termAndLiteralSatisfyStrictness` be `static`?

Inferences/InductionHelper.cpp
MichaelRawson: OK, I hoped that the 'strictness' semantics for equalities/comparisons/terms overlapped, but looking closer I don't think so. Then I think this makes sense and is an elegant solution: Martin might have some ideas about how to ensure that all invalid values for the option are prevented (e.g. `66`) as he's been improving the options machinery recently.

MichaelRawson:
Thanks for the changes, Petra! I think this is ready to go to Martin - @quickbeam123, there is a new option that takes a three-tuple of enumeration values 0-4, 0-4, 0-2, currently expressed as an unsigned integer constrained to be less than 443. However, if I understand correctly e.g. `66` is still an invalid value. Any ideas/opinions?

Inferences/InductionHelper.cpp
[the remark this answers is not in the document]
hzzv: Done.

quickbeam123:
@MichaelRawson , one way we do something similar are the split queue options. The option itself is a string, a the getter does the checking. E.g.

`Lib::vvector<float> Options::avatarSplitQueueCutoffs() const`

using

`Lib::vvector<A> parseCommaSeparatedList(vstring const& str) `

MichaelRawson:
Hey @hzzv - I took the liberty of changing the options into three options, each of which is an enumeration. I hope that's OK for your purposes, it seemed the most readable/maintainable/typesafe option to me. If you approve, let @quickbeam123 know and he'll merge it.

MichaelRawson:
So I broke a few things (mostly unit tests) and fixed them. Commit 8e85f72 fixes an existing logic error, which compiled because `finite` is a function in scope somewhere which always evaluates to `true` when used as a condition.

MichaelRawson:
OK, I broke it, sorry. It works on my machine! I've asked Joe if he has any idea why the CI is failing to build unit tests.

hzzv:
Thanks for adding the option enums @MichaelRawson! I can also try to fix it, but no sooner than over the weekend.

hzzv:
Sorry for the delay! Unfortunately I have no idea what might be (or was?) wrong :/ (It works locally for me as well.)
Was Joe able to help?

MichaelRawson:
No, sadly Joe is bamboozled too! Weird - I think it's something to do with the software setup on the CI machine, it apparently changed in between your last commit and my first. However, I've been adding tests back one by one (up to 18 so far) and seems OK, so maybe I'll find something that way. I'll make it work somehow and then hopefully this can be merged finally, sorry it's taking so long.

PR #334, opened 2022-03-21
normalise order of variables in quantifiers

MichaelRawson:
Currently Vampire does not sort the order of variables in quantifier blocks, e.g.

```
?[X, Y]: p(X, Y)
?[Y, X]: p(X, Y)
```

are treated differently, despite being equivalent. This small change sorts quantifier blocks in order of occurrence in the sub-formula if `--normalize on`. I doubt this makes much difference (if any at all) without symbol re-use, but with symbol re-use it's a slight win in principle. @quickbeam123 - if you get chance and think this is a good idea, take a look: might be interesting with respect to your ongoing randomisation program?

### DIFF
--- a/Shell/Flattening.cpp
+++ b/Shell/Flattening.cpp
@@ -21,6 +21,7 @@
 #include "Kernel/FormulaUnit.hpp"
 #include "Kernel/Problem.hpp"
 #include "Kernel/Unit.hpp"
+#include "Kernel/FormulaVarIterator.hpp"
 
 #include "Lib/Environment.hpp"
 #include "Shell/Options.hpp"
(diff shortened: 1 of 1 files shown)


quickbeam123:
A potentially interesting idea. However, some questions:
- it looks like only a merge of quantifier blocks possibly triggers the sorting, why not sort blocks that are not getting merged?
- actually, why should flattening be the right moment to do it?
(since it's conditioned under the normalization flag, why not do it during normalization? (I wouldn't mind giving up the additional opportunity of the merged-quantifier event))
- aren't we asking for some quadratic trouble with the FormulaVarIterator?

MichaelRawson:
Thanks for getting to this! Just an idea, thought I should get around to it.

> it looks like only a merge of quantifier blocks possibly triggers the sorting, why not sort blocks that are not getting merged?

We do! I removed the early returns from the original code to allow that.

> actually, why should flattening be the right moment to do it?

Good question - I'm not attached to flattening, it just went there because we already manipulate quantifier blocks and it gets invoked in a reasonable number of places. This probably needs a new flag of its own as it's not quite "normalisation", but I'm not sure what it should be or where it should be applied. Off by default because it's quite expensive.

> aren't we asking for some quadratic trouble with the FormulaVarIterator?

I'm not sure I understand the question. If you mean "is this quadratic in the size of the formula?"...then yes, but only if it's a weird shape. If on the other hand you mean "is there a better way of doing it than FormulaVarIterator"...then possibly, but I'm not seeing it. I didn't do the obvious "sort the quantifier block by how far away they are in the formula" algorithm because that's asymptotically even worse.

MichaelRawson:
Probably pointless now without name reuse, and shuffling should take care of the small amount of "wiggle room" left.

PR #336, opened 2022-03-23
Higher-order proof output fix

ibnyusuf:
Using > as a type constructor symbol name is not TPTP compliant.

### DIFF
--- a/Kernel/Signature.cpp
+++ b/Kernel/Signature.cpp
@@ -1171,7 +1171,7 @@ bool Signature::symbolNeedsQuoting(vstring name, bool interpreted, unsigned arit
   //also don't want them to be treated as interpreted symbols
   //hence the hack below, AYB
   if(name=="$int" || name=="$real" || name=="$rat" || 
-     name=="$i" || name=="$o" || name==">"){
+     name=="$i" || name=="$o"){
     return false;
   }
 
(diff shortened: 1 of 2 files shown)


ibnyusuf:
A potentially issue with renaming the arrow type constructor to `fun`, is that there may be a symbol named `fun` in the input. What is the standard means of dealing with this issue for internally created symbol names?

MichaelRawson:
http://www.tptp.org/TSTP/NewSymbolNames.html is the main standard I'm aware of...so maybe `sFfun` or `sGfun`? Once fixed, looks good to me. :-)

MichaelRawson (approved):

ibnyusuf:
One subtlety, this is not really a new symbol, but a new name for an input symbol (the arrow constructor > which occurs in higher-order type defs)

MichaelRawson:
> One subtlety, this is not really a new symbol

Hmm. Good point! Maybe we should ask Geoff what to do in this case, he usually has an opinion. We could also consider `'>'` which is valid TPTP I believe, but again could clash with `'>'` if somebody decides to use that.

selig:
Isn’t there a convention somewhere about using $ or $$ as a prefix? But yes, no harm in asking Geoff what he thinks as he makes the rules.

Sent from my iPhone

On 24 Mar 2022, at 04:11, Michael Rawson ***@***.***> wrote:

﻿

One subtlety, this is not really a new symbol

Hmm. Good point! Maybe we should ask Geoff what to do in this case, he usually has an opinion. We could also consider '>' which is valid TPTP I believe, but again could clash with '>' if somebody decides to use that.

PR #337, opened 2022-03-25
cleanup Term: unused stuff, comments, type-correctness, C++11 unions

MichaelRawson:
- remove some unused functions/includes/etc from `Term`
- move `ArgumentOrderVals` to `Term`: this allows using more descriptive types and enumeration values
- where possible, use non-POD types in unions since we're allowed to now
- add some comments about the hazardous bits

### DIFF
--- a/Indexing/ClauseVariantIndex.cpp
+++ b/Indexing/ClauseVariantIndex.cpp
@@ -350,9 +350,9 @@ struct HashingClauseVariantIndex::VariableIgnoringComparator {
       return Int::compare(t1->weight(),t2->weight());
     }
 
-    if(t1->vars()!=t2->vars()) {
+    if(t1->getVariableOccurrences()!=t2->getVariableOccurrences()) {
       // number of variable occurrences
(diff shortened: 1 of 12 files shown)


MichaelRawson:
@quickbeam123 - I've asked Ahmed about removing some of the stuff in `Term`. Squashing it further is on my to-do list but I think this can be merged.

quickbeam123:
I am ready, although! What would you think about static_asserting `ARCH_X64` and using `reserved` in `TermList[0]` for something which already is using up 32 bits in `Term`?

MichaelRawson:
I think that's a good idea! Thanks for the commit. If we are happy with 64-bit only we can look into removing `ARCH_X64` and asserting the behaviour we require in `Lib::Portability` or something.

JakobR:
I was worried when seeing the name `ARCH_X64` because my current laptop runs on `arm64` rather than `x64`. But it compiles/runs fine; it seems we only care about pointer size here. Maybe the name of `ARCH_X64` could be changed to reflect that?

PR #338, opened 2022-03-25
fixing bug with eqProxy quantification and adding warning

ibnyusuf:
Code changes are hopefully self explanatory.

In short, when quantifying formulas that may contain sort variables, we need to take care that the sort variables occur at the front of the quantification list.

### DIFF
--- a/Kernel/Formula.cpp
+++ b/Kernel/Formula.cpp
@@ -780,21 +780,15 @@ Formula* Formula::quantify(Formula* f)
   DHMap<unsigned,TermList> tMap;
   SortHelper::collectVariableSorts(f,tMap);
 
-  Set<unsigned> vars;
-  FormulaVarIterator fvit( f );
-  while(fvit.hasNext()) {
-    vars.insert(fvit.next());
-  }
-
   //we have to quantify the formula
   VList* varLst = VList::empty();
   VList::FIFO quantifiedVars(varLst);
 
-  Set<unsigned>::Iterator vit(vars);
-  while(vit.hasNext()) {
-    unsigned v = vit.next();
+  DHMap<unsigned,TermList>::Iterator tmit(tMap);
+  while(tmit.hasNext()) {
+    unsigned v; 
     TermList t;
-    ALWAYS(tMap.find(v,t));
+    tmit.next(v, t);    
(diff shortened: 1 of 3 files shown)


Shell/EqualityProxy.cpp
```diff
@@ -334,7 +334,13 @@ unsigned EqualityProxy::getProxyPredicate()
   Literal* proxyLit = Literal::create(newPred, 3, true, false, args.begin());
   Literal* eqLit = Literal::createEquality(true,var1,var2,sort);
   Formula* defForm = new BinaryFormula(IFF, new AtomicFormula(proxyLit), new AtomicFormula(eqLit));
-  Formula* quantDefForm = Formula::quantify(defForm);
+
+  VList* varLst = VList::empty();
+  VList::push(2, varLst);
+  VList::push(1, varLst);
+  VList::push(0, varLst); //order is important, the sort var must be first!
```
MichaelRawson, line 341: Reference issue here.

MichaelRawson (approved):
Looks reasonable. Could you file an issue to fix `Formula::quantify` so that we know about the general problem and reference it in the comment in `EqualityProxy`? Doesn't seem difficult to fix when we get time.

ibnyusuf:
The fix was straightforward enough, that I jsut went ahead and carried it out. `Formula::quantify` should now be poly safe. In the process, I also managed to remove some code duplication.

MichaelRawson (approved):
Nice job! Bonus points for fixing it already.

Kernel/Formula.cpp
```diff
@@ -776,6 +776,10 @@ Formula* Formula::createLet(unsigned predicate, VList* variables, Formula* body,
 
 Formula* Formula::quantify(Formula* f)
 {
+
+  DHMap<unsigned,TermList> tMap;
+  SortHelper::collectVariableSorts(f,tMap);
+
```
MichaelRawson, line 782: Minor optimisation: are they keys of `tMap` not also the members of `vars` below? i.e. could we reuse `tMap` and remove `vars` altogether?

ibnyusuf: Done

PR #342, opened 2022-03-30
improve hashing performance, safety

MichaelRawson:
Vampire has a few collections (and some other stuff, but mostly collections) that use hashing as part of their implementation, most commonly the single-hashing `Set` and `Map`. `DHSet` and `DHMap` use _double_ hashing, where a second hash function is used to compute an offset interval for open addressing. This second hash function does not have to have nice distribution characteristics or any of the usual desirable features of hash functions. All these collections have template parameters to set the hash function(s) used, which by default is `Lib::Hash` for single hashing and `FirstHashTypeInfo<T>::Type`, then `Lib::Hash` for double hashing. `Lib::Hash` is mostly 32-bit FNV-1a with a few quirks.

There are some shortcomings of the current implementation. The first is that it's surprising that the default _first_ hash (i.e. the one that matters) is not the same for e.g. `Set<T>` and `DHSet<T>`. Leading on from that, using FNV as the second hash for double-hashing collections is overkill: we don't need the nice distribution properties. It's also somewhat wasteful if `FirstHashTypeInfo<T>::Type = Lib::Hash`: currently if we get a collision in e.g. `DHSet<vstring>`, we recompute the same hash again and use it as an offset.

There's some implementation oddities too: strings are hashed backwards, for some reason I don't understand. `*Hash` classes inconsistently implement `equals`, which single-hashing collections use but double-hashing does not. There is a blanket implementation of `FirstHashTypeInfo<T> = Lib::Hash`, which is convenient but often surprising. `Lib::Hash` has a default overloaded hash function for any `T` which just hashes the bytes of `T`, which is enormously unsafe due to padding (there is a comment warning about this, in fairness, but how many people are thinking about struct alignment and hashing when they add a field to some struct?).

This PR fixes...most of this. Concretely:
1. Change the order of hashes for doubly-hashed collections
2. The small number of functions in `Hash.cpp` should be available for inlining: lift them to `Hash.hpp` and delete `Hash.cpp`
3. `PtrIdentityHash` is (almost?) never the right thing to do because of poor distribution characteristics: remove it.
4. There's a few magic numbers representing the 32-bit FNV offset basis and prime. Introduce constants for them.
5. Introduce an imperfect type trait `is_safely_hashable<T>` that approximates C++17 `has_unique_object_representation`. Use this to restrict the implementation of `Lib::Hash` for any `T`.
6. Un-break things where the above broke them. In many cases this was a (potential) bug. Justifications inline.
7. Remove some unused stuff, fiddle with templates a bit and comment liberally for people who come looking for a reason why they can't hash their shiny new object.


Forwards.hpp
```diff
@@ -74,11 +72,11 @@ typedef Stack<vstring> StringStack;
 
 typedef Map<vstring,unsigned,Hash> SymbolMap;
 
-template<typename T> struct FirstHashTypeInfo;
+template<typename T, typename=void> struct SecondaryHash;
```
MichaelRawson, line 75: `FirstHashTypeInfo` renamed to `SecondaryHash`

Inferences/ArithmeticSubtermGeneralization/AdditionGeneralizationImpl.cpp
```diff
@@ -24,7 +24,7 @@ namespace AdditionGeneralizationImpl {
 
 template<class NumTraits> class MonomSet;
 
-using GenMap = Map<Variable, AnyNumber<MonomSet>>;
+using GenMap = Map<Variable, AnyNumber<MonomSet>, StlHash>;
```
MichaelRawson, line 27: Anywhere where `StlHash` was inserted: this used to be resolving to the default hash functions, which might do all manner of crazy. Since for all these types there is some kind of `std::hash` implementation, I assume this was meant to use that.

Kernel/BottomUpEvaluation.hpp
```diff
@@ -44,10 +44,10 @@ namespace Memo {
   };
 
   /** a memoization realized as a hashmap */
-  template<class Arg, class Result>
+  template<class Arg, class Result, class Hash = Hash>
```
MichaelRawson, line 47: Defaults to `Lib::Hash` but needs to be `StlHash` in places.

Kernel/RobSubstitution.cpp
```diff
@@ -261,7 +261,7 @@ bool RobSubstitution::occurs(VarSpec vs, TermSpec ts)
       return false;
     }
   }
-  typedef DHSet<VarSpec, VarSpec::Hash1> EncounterStore;
+  typedef DHSet<VarSpec, VarSpec::Hash1, VarSpec::Hash2> EncounterStore;
```
MichaelRawson, line 264: So `VarSpec` has `Hash2`, but we didn't use it here...not sure why, but this seems like the intended behaviour.

Lib/Hash.hpp
```diff
@@ -41,47 +63,81 @@ struct HashUtils
...
+// TODO: is this reasonable?
+struct PairIdentityHash {
+  template<typename T>
+  static unsigned hash(T pp) {
+    return static_cast<unsigned>(pp.first^pp.second^(pp.first<<1));
   }
 };
 
-template<class T>
+// wrapper around std::hash
```
MichaelRawson, line 92: Some of the following custom hash structs had a `T` parameter that wasn't strictly necessary. I pushed it into the function as an overload instead which makes it a bit nicer to type...but it doesn't have to be that way if there's some reason not to.

Lib/Hash.hpp
```diff
@@ -116,156 +176,147 @@ class Hash
...
-      res = Hash::combineHashes(res, ElHash::hash(cont[i]));
+    for (int i = 0; i < size; i++) {
+      hash = (hash ^ val[i]) * FNV32_PRIME;
     }
-    return res;
+    return hash;
   }
 };
 
+// template specializations for SecondaryHash
```
MichaelRawson, line 250: Lots of these are made considerably cheaper to compute, but worse as hashes. I don't see a problem with this given how secondary hashes are used.

Lib/Hash.hpp
```diff
@@ -116,156 +176,147 @@ class Hash
...
-};
-
-struct GeneralPairSimpleHash {
-  template<typename T, typename U>
-  static unsigned hash(std::pair<T,U> pp) {
-    unsigned h1=FirstHashTypeInfo<T>::Type::hash(pp.first);
-    unsigned h2=FirstHashTypeInfo<U>::Type::hash(pp.second);
-    return static_cast<unsigned>(h1^h2^(h1<<1));
-  }
+struct SecondaryHash<T, typename std::enable_if<std::is_enum<T>::value>::type> {
```
MichaelRawson, line 255: Removing the blanket implementation for all `T` that forwards to `Lib::Hash` meant that enumerations were a compiler error. This implementation seems OK to me.

Lib/SharedSet.hpp
```diff
@@ -460,14 +460,8 @@ class SharedSet {
...
     CALL("SharedSet::hash(T*,size_t)");
-
-    unsigned res=1234567890; //arbitrary number, hopefully won't cause problems:)
-    const T* arre=arr+len;
-    while(arr!=arre) {
-      res=Hash::hash(*arr,res);
-      arr++;
-    }
-    return res;
+    static_assert(is_safely_hashable<T>::value, "T must be safely hashable");
```
MichaelRawson, line 463: This originally did something bonkers, but it seems it was only ever actually used with `unsigned`. Stick a `static_assert` here to warn future users of `SharedSet<T>`.

SAT/Z3Interfacing.hpp
```diff
@@ -197,10 +197,12 @@ class Z3Interfacing : public PrimitiveProofRecordingSATSolver
   Map<SortId, z3::sort> _sorts;
   struct Z3Hash {
     static unsigned hash(z3::func_decl const& c) { return c.hash(); }
+    static unsigned hash(z3::expr const& c) { return c.hash(); }
     static bool equals(z3::func_decl const& l, z3::func_decl const& r) { return z3::eq(l,r); }
+    static bool equals(z3::expr const& l, z3::expr const& r) { return z3::eq(l,r); }
```
MichaelRawson, line 202: Originally I think we were hashing the bytes of a `z3::expr` via `Lib::Hash`...this seems better.

joe-hauns (approved):
So from what I can tell it looks all like great improvements/tidying.
Buuuut i need to remark that I don't have a lot of knowledge about hashing. i.e. what good/bad hash functions are. 
So I'm trusting you about that part.

MichaelRawson:
Great, thank you @joe-hauns! I will run some ad-hoc perf to make sure nothing disastrous happened before the gatekeepers look at this. A "good" hash function for me is one that evenly distributes (statistically-speaking) inputs over hash table buckets. I'm fairly sure (but not certain - I will check an algorithms textbook) that the secondary hash for double hashing does not need to be so good.

MichaelRawson:
Performance seems unchanged. There were a few regressions involving `DHMap<unsigned, T, IdentityHash>`, which would now use the identity hash twice. These are now fixed, but they weren't catastrophic in the first place. @quickbeam123 this is ready to go when you get chance!

PR #346, opened 2022-04-05
Fix max sort sizes

selig:
This adds functionality to FMB to detect maximum sort sizes in particular circumstances. The idea is that if we have a clause like `X = a | X = b` then we know that the cardinality of the sort of X is at most 2.

These maximum sort sizes are not required for the soundness of finding models but (i) can improve efficiency when enumerating sort size vectors, and (ii) can allow us to return unsat when we know we are complete. We have one mechanism for detecting maximum sort sizes (only constants,  no positive equality). This PR adds a new method.

About 5 years ago we tried adding this generally and there were some bugs due to trying to capture all cases. We never reintroduced it. This time I focus just on the case where we have a clause of the form `{ X = t_i }` for some `X` and some terms `t_i`. In such a case the cardinality of the sort of `X` is bound by `i`. Of course, there are scenarios where it may be lower. There are also cases where we might miss a bound by being too simplistic.

We detect this sort bound during sort inference. When scanning clauses we look for clauses consisting only of positive equalities where every literal contains exactly one shared variable. There is some complication in mapping this bound back to the meta-sorts used in the finite model building process.

### DIFF
--- a/FMB/SortInference.cpp
+++ b/FMB/SortInference.cpp
@@ -239,6 +239,7 @@ void SortInference::doInference()
   IntUnionFind unionFind(count);
   ZIArray<unsigned> posEqualitiesOnPos;
   Stack<unsigned> varEqualityVampireSorts;
+  DHMap<unsigned,unsigned> vampireSortMax;
 
   ClauseIterator cit = pvi(ClauseList::Iterator(_clauses));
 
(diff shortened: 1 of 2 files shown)


FMB/SortInference.cpp
```diff
@@ -252,9 +253,30 @@ void SortInference::doInference()
    Array<Stack<unsigned>> varPositions(c->varCnt());
    ZIArray<unsigned> varsWithPosEq(c->varCnt());
    IntUnionFind localUF(c->varCnt()+1); // +1 to avoid it being 0.. last pos will not be used
+
+   // This set will be used to collect the variables that occur as part of an equality in every literal (implicitly becomes empty if there's a non-equality literal).
+   // We will use this information to detect maximum model sizes as any claues which is just {X = t_i} means that the sort of X is at most i
+   // We start by assuming that all variables are always in all equalities and then remove variables that are not
+   Set<unsigned> vars_always_in_equalities;
```
quickbeam123, line 260: Isn't it an overkill to use a set?

Yes, for the first literal, we don't know what the variable is. We could learn it there and check it's the same later on, couldn't we?

FMB/SortInference.cpp
```diff
@@ -252,9 +253,30 @@ void SortInference::doInference()
...
      if(l->isEquality()){
+
+       // Positive equality means we might  be in a { X = t_i } scenario
+       if(l->polarity()){
+         // For each variable in vars_always_in_equalities remove it if it isn't
+         Set<unsigned>::Iterator vaie_iter(vars_always_in_equalities);
+         while(vaie_iter.hasNext()){ 
+           unsigned var = vaie_iter.next();
+           if((l->nthArgument(0)->isVar() && l->nthArgument(0)->var() != var) ||
+              (l->nthArgument(1)->isVar() && l->nthArgument(1)->var() != var)){
```
quickbeam123, line 274: I think the code will still fail to work for a clause like:
`X = Y | X = c`
As one pass of the loop here will kick out X from vars_always_in_equalities and another will remove Y.

Please let me know if I missed something!

FMB/SortInference.cpp
selig: Probably, yes. In the first literal we might have two possibilities but we get one or zero possibilities by the second literal so we only need two variables, not a set. I'll refactor after discussing the next comment.

FMB/SortInference.cpp
selig: Yes, it should be && instead of || here, right? The first literal shouldn't kick anything out but then the second should kick out Y.

Refactoring to use two variables should make this more logical. I'll refactor and ask you to look again.

FMB/SortInference.cpp
quickbeam123: Great, sounds good!

FMB/SortInference.cpp
quickbeam123: There is probably many sensible ways of doing this ... 

The one I had in mind would probably check for the exact shape of the current literal: 1) positive two_var_eq, 2) positive one_var_eq_non_var, 3) rest and for the exact state: a) just starting, b) "these two vars possible candidates", c) "this single var possible"; and act accordingly.

PR #349, opened 2022-04-14
Do not create additional units when displaying theory axioms

mezpusz:
The motivation for this is mainly that when `--show_theory_axioms` (or `--show_everything`) is on, more units are added and the unit numbers slightly change, making it harder to track two diverging proofs.

Question: the original code avoided outputting a clause (calling `Clause::toString`) directly, is there any reason to do this? If so, I can live without merging this PR.

### DIFF
--- a/Shell/TheoryAxioms.cpp
+++ b/Shell/TheoryAxioms.cpp
@@ -55,14 +55,9 @@ void TheoryAxioms::addAndOutputTheoryUnit(Unit* unit, unsigned level)
   if(opt_level != Options::TheoryAxiomLevel::ON && level != CHEAP){ return; }
 
   if (env.options->showTheoryAxioms()) {
-    Unit* qunit = unit;
-    Formula* f = 0;
-    if(unit->isClause()){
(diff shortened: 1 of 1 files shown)


selig:
The reason to do that clauses don't contain the sorts of variables. By turning them into units we get valid TPTP we can parse using Vampire again. Given that lots of the interpreted functions are overloaded we can't infer the sorts of variables from the way they are used in the clause.

An alternative solution would be to always create these units but only print them under the guard. There is a small number so the cost would be very low.

mezpusz:
> The reason to do this (turn them into units) is that clauses don't contain the sorts of variables. By turning them into units they will become quantified formulas and we get valid TPTP that we can parse again using Vampire. Given that lots of the interpreted functions are overloaded we can't infer the sorts of variables from the way they are used in the clause.
> 
> An alternative solution would be to always create these units but only print them under the guard. There is a small number so the cost would be very low.

Thanks for the explanation, now it makes sense! I'll close the PR then, I don't think it's worth the effort to solve.

PR #356, opened 2022-04-26
Make LRS also respond to instruction limit 

quickbeam123:
What needs to be considered:
1) both time limit and instruction limit can be 0, which effectively stands from infinity (so it complicates the arithmetic with some if-then-elses)
2) when both limits are active, we plan to terminate when the earlier hits, so the smaller estimate is used.

(Similarly, do the same also for `avatar_turn_off_time_frac`.)

### DIFF
--- a/Lib/Timer.cpp
+++ b/Lib/Timer.cpp
@@ -33,7 +33,7 @@
 #include "Timer.hpp"
 
 #define DEBUG_TIMER_CHANGES 0
-#define MILLION (1 << 20)
+#define MEGA (1 << 20)
 
 // things that need to be signal-safe because they are used in timer_sigalrm_handler
 // in principle we also need is_lock_free() to avoid deadlock as well
@@ -54,7 +54,7 @@ int Timer::s_initGuarantedMiliseconds;
 
 unsigned Timer::elapsedMegaInstructions() {
 #ifdef __linux__
-  return (last_instruction_count_read >= 0) ? last_instruction_count_read/MILLION : 0;
+  return (last_instruction_count_read >= 0) ? last_instruction_count_read/MEGA : 0;
 #else
   return 0;
 #endif
@@ -131,7 +131,7 @@ timer_sigalrm_handler (int sig)
     // to get info about instructions burned even when not instruction limiting
     read(perf_fd, &last_instruction_count_read, sizeof(long long));
     
-    if (last_instruction_count_read >= MILLION*(long long)env.options->instructionLimit()) {
+    if (last_instruction_count_read >= MEGA*(long long)env.options->instructionLimit()) {
       Timer::setLimitEnforcement(false);
       if (protectingTimeout) {
         callLimitReachedLater = 2; // 2 for an instr limit
--- a/Lib/Timer.cpp
+++ b/Lib/Timer.cpp
@@ -33,7 +33,7 @@
 #include "Timer.hpp"
 
 #define DEBUG_TIMER_CHANGES 0
-#define MILLION 1000000
+#define MILLION (1 << 20)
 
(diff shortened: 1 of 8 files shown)


Lib/Timer.cpp
```diff
@@ -33,7 +33,7 @@
 #include "Timer.hpp"
 
 #define DEBUG_TIMER_CHANGES 0
-#define MILLION 1000000
+#define MILLION (1 << 20)
```
MichaelRawson, line 36: Huh? Maybe `INSTRUCTION_UNIT` or something?

Saturation/Splitter.cpp
```diff
@@ -703,9 +703,16 @@ void Splitter::init(SaturationAlgorithm* sa)
 #endif
 
   if (opts.splittingAvatimer() < 1.0) {
-    _stopSplittingAt = opts.splittingAvatimer() * opts.timeLimitInDeciseconds() * 100;
+    _stopSplittingAtTime = opts.splittingAvatimer() * opts.timeLimitInDeciseconds() * 100;
+    _stopSplittingAtInst = 0;
+#ifdef __linux__
```
MichaelRawson, line 708: If this kind of thing is going to proliferate, maybe we should introduce a new preprocessor define like `INSTRUCTION_COUNT_AVAILABLE`?

MichaelRawson (approved):
Cool! Nice generalisation, I wonder how reproducible it is. Just some nitpicks inline, the logic looks OK but I haven't tested it.

Lib/Timer.cpp
MichaelRawson: Ahah, OK - it's like megabytes or whatever. In which case either ignore me or something like `MEGA_INSTRUCTION`.

quickbeam123: I think it's just the MEGA from MEGA_INSTRUCTION; thanks for the suggestions!

Saturation/LRS.cpp
```diff
@@ -103,33 +103,59 @@ long long LRS::estimatedReachableCount()
...
+      && instrsBurned*100 < firstCheck*opt_instruction_limit
+  ) {
     goto finish;
   }
 
-  result = (processed*timeLeft)/timeSpent;
+  {
+    long long processed=env.statistics->activeClauses;
+
+    if (processed<=10) {
```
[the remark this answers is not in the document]
quickbeam123: It came from the original implementation, so I just copied it...

Saturation/LRS.cpp
```diff
@@ -103,33 +103,59 @@ long long LRS::estimatedReachableCount()
...
+    long long timeLeft; // (in milliseconds) 
+    if(_opt.simulatedTimeLimit()) {
+      timeLeft=_opt.simulatedTimeLimit()*100 - currTime;
+    } else {
+      timeLeft=opt_timeLimitDeci*100 - currTime;
+    }
+
+    long long instrsLeft = opt_instruction_limit - instrsBurned;
+
+    // note that result is -1 here already
```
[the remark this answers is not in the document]
quickbeam123: Except for the goto! (I know, it's evil...)

Saturation/Splitter.cpp
quickbeam123: A good point. I was just too lazy for this one. But will do it with the next!

quickbeam123:
I already tested it. It's a bliss to play with it! :)

Saturation/LRS.cpp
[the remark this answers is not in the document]
MichaelRawson: Ah! That is evil, good spot.

PR #357, opened 2022-04-26
remove ARCH_X*, add static assertions to check stuff we rely on

MichaelRawson:
See previous discussions about `Term`. Vampire is not especially portable, and more-or-less assumes 64-bit Intelesque platforms. We want to improve this, but in the meantime we also want to know if there's a portability hazard _before_ we get a weird bug report. Remove `ARCH_X64`, `ARCH_X32`, and statically assert that there are 8 bits in a `char` (not guaranteed!) and that pointers have 8 bytes. This is not all we assume, but it's a good start.


JakobR (approved):
Looks good to me!

PR #358, opened 2022-04-26
remove SINGLE_PARAM_HASH

MichaelRawson:
The doubly-hashing containers have a mechanism via `SINGLE_PARAM_HASH` where the hash function is optionally passed the current size of the hash table. This is cute and might in principle improve performance, but it's very seldom used: only in the unit test for `DHMap` and in `RobSubstitution`. I doubt anyone uses this, and I think it's very unlikely that it changes performance much.

### DIFF
--- a/Kernel/RobSubstitution.hpp
+++ b/Kernel/RobSubstitution.hpp
@@ -121,14 +121,14 @@ public:
     struct Hash1
     {
      static unsigned hash(VarSpec& o) {
-       return HashUtils::combine(Hash::hash(o.var), Hash::hash(o.index));
+       return HashUtils::combine(o.var, o.index);
      }
     };
     /** struct containing second hash function for DHMap object storing variable banks */
     struct Hash2
     {
       static unsigned hash(VarSpec& o) {
-        return o.var;
+        return HashUtils::combine(o.index, o.var);
       }
     };
   };


Kernel/RobSubstitution.cpp
```diff
@@ -1115,33 +1115,4 @@ ostream& operator<< (ostream& out, RobSubstitution::TermSpec ts )
...
-  else
-    return static_cast<unsigned>(res);
-}
-
-/**
- * Second hash function for DHMap. It just uses the hash function from Lib::Hash
- */
-unsigned RobSubstitution::VarSpec::Hash2::hash(VarSpec& o)
-{
-  return Lib::Hash::hash(reinterpret_cast<const unsigned char*>(&o), sizeof(VarSpec));
```
MichaelRawson, line 1143: Incidentally this was another compiler-changed-the-struct-padding bug waiting to happen.

quickbeam123:
Running some performance check on this ...

quickbeam123:
... your simplification does not seem to break anything, but makes vampire lose 1 TPTP problem in discount10 (50G instruction limit) on average. So, is it worth the cleanup?

MichaelRawson:
Is it easy for you to run another performance check? I got a measurable speedup (as measured by `perf`) on my machine by tweaking the hash function, but don't bother if it's a pain. And no, if we lose a problem it's not worth it, but I am a bit surprised!

quickbeam123:
No, I can easily rerun some more (ultimately, it just costs money for the electricity and a bit of my time). But it's important to get some intuitions about how much things matter in this. 

At the moment, however, something seems broken there...

PR #360, opened 2022-04-27
Add option to only collect unbound variables in SortHelper

mezpusz:
Proposes solution to changed behaviour of `Formula::quantify`, which now uses `SortHelper::collectVariableSorts` instead of `FormulaVarIterator`. The difference is that the former also collects bound variables, which can be now ignored by setting an argument to true.
Solution works similarly as in `FormulaVarIterator`.
Should fix unit test in #353 .

Does this affect any problems or tests? Are there any cases where we might want to ignore bound variables, similarly as in `Formula::quantify`?

### DIFF
--- a/Kernel/Formula.cpp
+++ b/Kernel/Formula.cpp
@@ -778,7 +778,7 @@ Formula* Formula::quantify(Formula* f)
 {
 
   DHMap<unsigned,TermList> tMap;
-  SortHelper::collectVariableSorts(f,tMap);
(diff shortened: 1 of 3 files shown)


quickbeam123:
Marton, did you run any tests on this? Does it fix your problem in the induction work?

quickbeam123:
I think, that's_ enough for the moment. Let me merge and test it the usual way :) Thank you for the work!

PR #361, opened 2022-04-28
update CI checkout to v3, try with GCC again

MichaelRawson:
There is a new version of the GitHub script that checks out the Vampire repository. Update it and see if anything breaks. Also at some point GCC broke the unit tests so we swapped to Clang, but let's see if that's still true...

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -19,6 +19,8 @@ jobs:
     - name: Configure Z3 Build
       working-directory: ${{runner.workspace}}/vampire/z3/build
       run: cmake .. -DCMAKE_BUILD_TYPE=Debug
+      env:
+        CXX: clang++
     - name: Z3 Build
       working-directory: ${{runner.workspace}}/vampire/z3/build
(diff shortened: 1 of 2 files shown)


MichaelRawson:
Unit tests still broken under GCC. Interesting but probably not critical for now.

quickbeam123:
What is the error in unit tests with gcc? Could we possibly also see it locally?

MichaelRawson:
https://github.com/vprover/vampire/runs/6208506355?check_suite_focus=true - we get some segmentation faults in the unit tests. I can't reproduce this on any of my machines (including an ancient GCC), and I don't remember Joe having any luck when we looked at this last time. Apparently the machine it runs on is Intel Xeon, so it's a little unusual - but probably so is the server you run benchmarks on at CIIRC, right?

On the plus side, while I was trying to reproduce it I found I forgot a `std::` prefix that Clang found somehow but GCC did not.

quickbeam123:
Hmm, maybe I should try vtest on one of our servers! A good point!

PR #362, opened 2022-05-03
reliesOn warns when options are getting explicitly set, not when they have a non-default value

quickbeam123:
the option constraint mechanism `reliesOn` has had, until now, the problem that it was checking whether the respective option has a nonDefault value as a precondition to triggering the check:

```
        void reliesOn(OptionValueConstraintUP<T> c){
            _constraints.push(If(getNotDefault()).then(std::move(c)));
        }
```

This was causing a problem, at least, for the combination of avatar (on by default), with `reliesOn(ProperSaturationAlgorithm())` (ProperSaturationAlgorithm means discount, otter or lrs) and fmb, where setting `-sa fmb -av off` would trigger a warning. 

(Actually, until recently, we were getting a warning for any `-sa fmb` strat which does not explicitly set av to off. This was also weird, because the user should not care about explicitly touching an irrelevant option. (The until recent implementation was using

`_splitting.addConstraint(If(equal(true)).then(ProperSaturationAlgorithm()));`

that I changed to the above `reliesOn` in https://github.com/vprover/vampire/commit/e9ecc9057d1ff50f8af1c2cf1909811399397f9f)

Since neither makes a lot of sense, the PR changes the meaning of `reliesOn` to
```
        void reliesOn(AbstractWrappedConstraintUP c){
            _constraints.push(If(hasBeenSet<T>()).then(unwrap<T>(c)));
        }
```
which resolves the issue for the above and also seems to follow the original intended semantics anyway. 

I kept 

```
        void reliesOnWhenNondef(AbstractWrappedConstraintUP c){
            _constraints.push(If(getNotDefault()).then(unwrap<T>(c)));
        }
```

and used it for the case of `_selection` and `_ageWeightRatio` which cannot be NOT SET via the `--decode` mechanism. The non-warning way to call an fmb strat via `--decode` needs to use the defaults: --decode fmb+10_1_...

### DIFF
--- a/Shell/Options.cpp
+++ b/Shell/Options.cpp
@@ -157,17 +157,17 @@ void Options::init()
          "struct_induction"});
     _schedule.description = "Schedule to be run by the portfolio mode. casc and smtcomp usually point to the most recent schedule in that category. Note that some old schedules may contain option values that are no longer supported - see ignore_missing.";
     _lookup.insert(&_schedule);
-    _schedule.reliesOnHard(UsingPortfolioTechnology());
+    _schedule.reliesOn(UsingPortfolioTechnology());
 
     _multicore = UnsignedOptionValue("cores","",1);
     _multicore.description = "When running in portfolio modes (including casc or smtcomp modes) specify the number of cores, set to 0 to use maximum";
     _lookup.insert(&_multicore);
-    _multicore.reliesOnHard(UsingPortfolioTechnology());
+    _multicore.reliesOn(UsingPortfolioTechnology());
 
     _slowness = FloatOptionValue("slowness","",1.3);
     _slowness.description = "The factor by which is multiplied the time limit of each configuration in casc/casc_sat/smtcomp/portfolio mode";
     _lookup.insert(&_slowness);
-    _slowness.reliesOn(UsingPortfolioTechnology());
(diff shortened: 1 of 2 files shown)


selig (approved):
Looks good. Definitely captures the original intent. 

One (non-critical) thing to consider is whether 'reliesOn' actually captures the right intent. reliesOn and reliesOnHard reflect two different scenarios:

- reliesOn means that setting this option will have no impact unless the captured condition on other option values is true
- reliesonHard means that setting this option (implicitly to a non-default value) requires that the captured condition on other option values is true

Can we think of better names?

MichaelRawson (approved):
Makes sense to me.

MichaelRawson:
>Can we think of better names?

`onlyUsefulWith` and `reliesOn` respectively? I guess that the current `reliesOn` produces a warning (not an error) so could also be `warnUnless` or something.

"relies" already implies that the constraint is hard to me, so it was surprising to find `reliesOnHard`.

MichaelRawson:
Interestingly CI seems to be hung running `checks/sanity` - I'll kill it and re-run manually, maybe GitHub goofed.

MichaelRawson:
Actually I can reproduce. Run a portfolio mode (e.g. `--mode casc Problems/PUZ/PUZ001+1.p`) and it seems to loop complaining about broken option constraints.

PR #366, opened 2022-05-11
remove old-style interpolants, orphaned code

MichaelRawson:
As agreed, remove the old style of interpolant generation. This allows us to remove references to the Yices SMT solver, the slightly-confusing `Shell::SMTFormula` and quite a few not-very-portable functions in `Lib::System`.


easychair:
Do we have a new style or do we just get rid of interpolation alltogether?

MichaelRawson:
Apparently there is a new style(s) `--show_interpolant off/new_heur/new_opt` - I would be a bit more careful before removing interpolation altogether, I know it's an advertised feature of Vampire. That said, I'm not especially convinced either style "out of the box" any more.

easychair:
This is right. I would keep interpolation. What I noticed is that the code
contains interpolant minimization. Do we have it in the new style?

Also, what kind of minimization is this? The only where we make large
theory formulas smaller or the one where we use "grey areas" and
optimization? Ideally, it would be good to have both. But I do not remember
- maybe the latter was done outside of Vampire. If the latter is the code
we remove, it would be good to have it for the new (remaining) interpolants.

Best, Andrei

MichaelRawson:
There is the `InterpolantMinimizerNew` that seems to get called, so I assume so. I think this is the "grey areas" work - I only came to this because I noticed some code that attempted to locate and then shell out to Yices, which wasn't in the best shape. I asked @quickbeam123 (sorry to drop you in it Martin!) and he said

> I remember I had yices working on my laptop in Vienna still, but it probably comes from the "Playing in the grey area of proofs." times. Our Splitting Proofs for Interpolation, if I remember correctly, did not rely on any yicesing.
I vote for discontinuing the connection. Whoever keeps working on interpolation will need to undust more than just this and that should be the perfect time to do things properly! ;)

I'm not against it remaining if it's desired - but if we want to do this it might be worth making it work with Z3 or adding an (optional?) dependency on the Yices API.

Lib/System.cpp
```diff
@@ -284,7 +274,7 @@ void System::terminateImmediately(int resultStatus)
   CALL("System::terminateImmediately");
 
   onTermination();
-  _exit(resultStatus);
+  std::quick_exit(resultStatus);
```
quickbeam123, line 277: I am getting 

```
Lib/System.cpp:278:8: error: reference to unresolved using declaration
  std::quick_exit(resultStatus);
       ^
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/c++/v1/cstdlib:155:1: note: using declaration annotated with 'using_if_exists' here
using ::quick_exit _LIBCPP_USING_IF_EXISTS;
^
Lib/System.cpp:278:19: error: redefinition of 'resultStatus'
  std::quick_exit(resultStatus);
                  ^
Lib/System.cpp:273:39: note: previous definition is here
void System::terminateImmediately(int resultStatus)
```

even after having included `<cstdlib>` Googling does not help much. This is probably an issue of

```
gcc --version
Apple clang version 13.1.6 (clang-1316.0.21.2.5)
Target: x86_64-apple-darwin21.4.0
```

Lib/System.hpp
```diff
@@ -69,10 +67,9 @@ class System {
 
   static bool fileExists(vstring fname);
 
+  // XXX
```
MichaelRawson, line 70: Oops. This needs to go, I'll do it before merge but after you finish any changes.

Lib/System.cpp
MichaelRawson: `gcc --version` returning `clang` is ...interesting, but probably not the issue. I obviously can't reproduce, does `::quick_exit` compile for you?

quickbeam123: Well, I meant the whole clang ecosystem ;)

My `cstdlib` contains:

```
#if !defined(_LIBCPP_CXX03_LANG) && defined(_LIBCPP_HAS_QUICK_EXIT)
using ::at_quick_exit _LIBCPP_USING_IF_EXISTS;
using ::quick_exit _LIBCPP_USING_IF_EXISTS;
#endif
```

`::quick_exit` does not help.

quickbeam123:
To the general discussion:
- the code that's being removed has been shown inferior to the new alternative (see https://link.springer.com/chapter/10.1007/978-3-319-63046-5_18)
- the new alternative still works and optimizes interpolants (just checked on https://github.com/vprover/vampire_benchmarks/tree/master/interpolation_examples) 
(we just no longer terminate, i.e. fail to find a local proof, on some of the examples, but that would also be an issue for the old interpolation code - it's probably a consequence of some general changes in vampire)

Maybe this PR should also consider removing the qualifier "New" from the modules that are left in Vampire (in particular `InterpolantMinimizerNew`). If we also decide to changed the options values of `show_interpolant`, we should update the interpolation_examples mentioned above as they hardcode the option values using the 
`vampire(option,show_interpolant,new_heur).` 
directive.

Lib/System.cpp
MichaelRawson: I'd expect `!defined(_LIBCPP_CXX03_LANG)` to be fine, so I guess it's failing on `_LIBCPP_HAS_QUICK_EXIT`...but I have no idea why, and I guess if we can't use the standards-compliant thing we should go back to using POSIX `_exit`.

MichaelRawson:
OK, moved remaining stuff from `Interpolants` to `InterpolantsNew`, then moved all the `New` things over the top of the old things. I haven't changed the option value so I don't have to change the benchmarks too, but if it's confusing to somebody we can change it later.

easychair:
The interpolant minimizer needs pseudo-boolean programming, I believe. This
is probably why Yices was used.

We should probably preserve the code that generates the pseudo-boolean
optimization problem. Who and how solves it (using an optimizer or a
sub-optimal SAT solving) can then be changed.

A

PR #368, opened 2022-05-16
Fix CI test cases

MichaelRawson:
Figured it out. The CI machinery will _merge_ the PR branch and master, which is usually fine (and intended) but in this case the merge went weirdly and somehow ended up inserting new test cases halfway through the file and with clashing names, which results in a linker error. This is very hard to debug because master looks OK and the branch looks OK, but the merged master and branch on the remote CI server is obviously not OK.

I'll keep this in mind for the future but I don't see a way to fix it properly, so for now just fix the linker errors and move along.

### DIFF
--- a/UnitTests/tInduction.cpp
+++ b/UnitTests/tInduction.cpp
@@ -143,11 +143,11 @@ private:
   };
 };
 
-#define TEST_GENERATION_INDUCTION(name, expr)                                                                  \
+#define TEST_GENERATION_INDUCTION(name, expr)                                                                 \
   TEST_FUN(name) {                                                                                            \
     GenerationTesterInduction tester;                                                                         \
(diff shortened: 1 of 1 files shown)


hzzv:
I think the issue is that this commit added a new mechanism for variable/skolem matching: https://github.com/vprover/vampire/commit/244e6e6910c88cd7c7f9931a6926e6526f9ce89a - am I right, @mezpusz?
I can try to fix it but probably not sooner than over the weekend.

mezpusz:
Yes, sorry, my new code probably broke it, one needs to use `skx` prefixed variables where a fresh Skolem is expected from now on. Should be fixed now.

MichaelRawson:
Hey, great! Thank you for the quick response, both of you. @quickbeam123 - should be ready now, the tests pass. ;-)

PR #369, opened 2022-05-17
disequation flattening

MichaelRawson:
Add a new inference to "flatten" disequations with the same top symbol, sometimes resulting in a smaller or more-splittable clause after simplification. I'm not aware of an existing option in Vampire to do this, but there could well be - `Injectivity` looked promising, but seems to do something HOLish and more complex. The inference can be expressed with congruence clauses injected with equality proxy, but equality proxy has its own performance characteristics. Suggestions for a better name welcome.

Naturally, this inference is not necessary: all proofs found with it can in principle be found without it. However, it's relatively cheap, simple, and it partially turns unit-equality problems into regular first-order problems that can use more of Vampire's capabilities (although you might question whether this is desirable).

I'm not sure whether this should be merged. I found quite a few problems within a few minutes of CPU time for which the eventual proof used the new inference. Of these, the unit-equality problems `GRP489-1` and `COL044-1` are of particular interest: `GRP489-1` is solved much more quickly with the new option, and `COL044-1` cannot be solved (in the default mode - the CASC portfolio gets it almost immediately by setting `-nwc 2`) without the new option, and the resulting proof constructs terms that are visibly smaller than the proof found by `-nwc 2`. If people are interested we can push this further.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -398,6 +398,7 @@ set(VAMPIRE_INFERENCE_SOURCES
     Inferences/BackwardSubsumptionResolution.cpp
     Inferences/BinaryResolution.cpp
     Inferences/Condensation.cpp
+    Inferences/DisequationFlattening.cpp
     Inferences/DistinctEqualitySimplifier.cpp
     Inferences/EqualityFactoring.cpp
     Inferences/EqualityResolution.cpp
@@ -434,6 +435,7 @@ set(VAMPIRE_INFERENCE_SOURCES
     Inferences/BackwardSubsumptionResolution.hpp
     Inferences/BinaryResolution.hpp
     Inferences/Condensation.hpp
+    Inferences/DisequationFlattening.hpp
     Inferences/DistinctEqualitySimplifier.hpp
     Inferences/EqualityFactoring.hpp
     Inferences/EqualityResolution.hpp
--- a/Inferences/DisequationFlattening.cpp
+++ b/Inferences/DisequationFlattening.cpp
@@ -19,8 +19,30 @@ using Kernel::Clause;
 
 Inferences::DisequationFlattening::~DisequationFlattening() {}
 
+bool Inferences::DisequationFlattening::eligibleForFlattening(Literal *l)
+{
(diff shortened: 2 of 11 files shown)


easychair:
You don't use it as a simplification rule, right?

A

MichaelRawson:
> You don't use it as a simplification rule, right?

No, indeed @easychair! I initially thought it could be, but it's hilariously incomplete. It's a generating inference that only flattens the top of disequation literals in a given clause.

MichaelRawson:
Fixed polymorphism and applied only to selected literal. Also implement Martin's suggestion where no inferences are allowed where they could be simulated by this inference followed by other inferences. Sadly mostly-negative effect (even on UEQ), closing - but feel free to play around if you think you can get something working.

ibnyusuf:
Was this closed because there wasn't sufficient beenfit?

MichaelRawson:
Yep, with a 30G instruction limit and `-sa discount` I got 5 losses and 2 wins on my machine for UEQ.

easychair:
I'd be really surprised if it helps in solving a problem that we
cannot solve without it.

Of course, we all know that with saturation small changes in the
search space can result in big changes in solving time,, so I am not
surprised that in some cases the time goes down from a lot to almost
zero.

Do you have problems that it solves that were probably not solved
otherwise at all? I can then try to check this problem in an old
Spider database.

I apologise for still being unable to allocate time to make the new
Spider run and I am trying to rearrange my time to make it possible.
What I would propose in any case is to have the new rule on a branch
and, when we have the new Spider, try to see if it can solve anything.

I can also see some cases when the new rule can be seen as a
simplifier, which again can be tested later.

Andrei

MichaelRawson:
> Do you have problems that it solves that were probably not solved otherwise at all?

No, hence my lack of hope for this - but if you get chance to play around with Spider it might be interesting. I think we have special schedules for UEQ (?). These strategies probably doesn't do much work for e.g. multi-literal clauses, AVATAR - whereas we do have these with this option, even in UEQ.

PR #371, opened 2022-05-17
Twee preprocessing

MichaelRawson:
As discussed, implement the Twee goal-transformation step, initially just for ground, compound terms in clauses derived from the conjecture. The technique is described in Twee's system description (Smallbone 2021), section 2, paragraph headed _Goal transformation_. The new flag is `-twee on`/`--twee on`.

Questions:
- Is the transformation applied at the right time in the preprocessing pipeline? I think so; the later the better, but before equality-proxy.
- Should properties be recomputed (immediately) after applying this transformation?
- Should the generated axioms be axioms or conjectures?
- Should we also introduce new names for constants? This seems useless prima facie but the symbol precedences might have an effect.
- Should we introduce functions for non-ground terms?
- Should we optionally include non-conjecture clauses?


ibnyusuf:
Interesting, have we carried out any testing?

MichaelRawson:
Nope - and in fact the few problems I tried it wasn't helpful. However, the results in the Twee paper are quite encouraging, and @quickbeam123 has some plans with it - I think for a start the introduced constants need to be engineered such that they are smaller than any other.

easychair:
We definitely should test such things. Can anybody describe in short
what it is or send a link to a comprehensible presentation?

A

easychair:
So it is a "tree automaton" transformation of ground subterms in the goal.

I do not have any intuition about it. It is probably not very useful for
non-UEQ problems.

A

Kernel/TermTransformer.hpp
```diff
@@ -47,18 +49,23 @@ class TermTransformer {
 };
 
 /**
- * TODO: WTF the name?
- *   --> rename to SumbtermEvaluator?
+ * Has similar philosophy to TermTransformer, but:
```
MichaelRawson, line 52: I assume this stuff is just incidental cleanup unrelated to Twee per se. But thanks for doing it! Might be worth figuring out if `Kernel::BottomUpEvaluation" can be unified with this somehow.

Kernel/Inference.cpp
```diff
@@ -715,6 +715,8 @@ vstring Kernel::ruleName(InferenceRule rule)
     return "definition unfolding";
   case InferenceRule::DEFINITION_FOLDING:
     return "definition folding";
+  case InferenceRule::FUNCTION_DEFINITION:
```
MichaelRawson, line 718: Nice, thanks!

Shell/Twee.cpp
```diff
@@ -0,0 +1,219 @@
...
+ * ----- 
+ * Twee’s frontend can optionally transform the problem to make the prover more goal-directed.
+ * The transformation is simple, but strange.
+ * For every function term f(...) appearing in the goal, we introduce a fresh constant symbol a and add the axiom f(...) = a
+ * For example, if the goal is f(g(a), b) = h(c) we add the axioms f(g(a), b) = sF0, g(a) = sF1, and h(c) = sF2.
+ * Simplification will rewrite the first axiom to f(sF1, b) = sF0 and the goal to sF0 = sF2.
+ * -----
+ * 
+ * We take an intepretation of this for our purposes, NB:
+ *  - any ground term, in a clause derived from the conjecture, is transformed in this way
```
MichaelRawson, line 24: Does this header need updating wrt the new functionality?

Shell/Twee.cpp
```diff
@@ -0,0 +1,219 @@
+/*
```
MichaelRawson, line 1: If this is now the "Twee goal transformation" should this file be `TweeGoalTransformation.cpp`?

Shell/Twee.cpp
```diff
@@ -0,0 +1,219 @@
...
+  private:
+    bool _groundOnly;
+
+    unsigned _typeArity;
+    TermStack _allVars; // including typeVars, which will come first, then termVars
+    TermStack _termVarSorts;
+
+    // a helper function to collect terms variables and their sorts
+    // all stored in the above private fields to be looked up by transformSubterm
+    void scanVars(Term* t) {
```
MichaelRawson, line 70: I take it this isn't some variation on `SortHelper::collect*`?

MichaelRawson (commented):
Looks good to me! And nice job making this work in the general case, it looks tricky. If performance testing shows that rewriting "in place" or handling the non-ground case is not worth it, it might be worth removing the extra complexity, but I think it does "something" so it's worth keeping.

Shell/Twee.cpp
```diff
@@ -0,0 +1,219 @@
...
+    }
+
+  protected:
+    TermList transformSubterm(TermList trm) override {
+      CALL("Definizator::transformSubterm");
+
+      // cout << "tf: " << trm.toString() << endl;
+      if (trm.isVar()) return trm;
+      Term* t = trm.term();
+      if (t->isSort() || t->arity() == 0 || (!t->ground() && _groundOnly)) return trm;
```
MichaelRawson, line 113: One more thing: did we try also replacing constants? I'm fairly sure it doesn't make sense to replace sorts but in principle we could for HOL or something.

Kernel/TermTransformer.hpp
quickbeam123: Exactly, according to the rule: if your PR touches something ugly, pay the price and make it nicer for rest of the codebase as well. ;)

A good question! @joe-hauns , you might know this. `TermTransformerTransformTransformed`, here being renamed to `BottomUpTermTransformer` is being used in, e.g., `InterpretedLiteralEvaluator`, which is now "your domain". What is it's relation to `BottomUpEvaluation` and why does "your" arithmetic code use both?

Shell/Twee.cpp
quickbeam123: It's did not look like an exact match of any. Although the polymorphic part is "inspired" by @ibnyusuf 's work in Skolem.

Shell/Twee.cpp
quickbeam123: I am fairly convinced that defining constants would only make it mildly slower with the default orderings... Defining sorts is not an option. Even the poly version of fol does not allow equalities between sorts to be stated!

Shell/Twee.cpp
quickbeam123: Damn, it did!

Shell/Twee.cpp
quickbeam123: Done!

quickbeam123:
Thank you, @MichaelRawson , for the check!

Concerning removing the extra complexity if it does not help (much), I agree in principle. But you didn't factor in the equation my emotional pain of letting go of something I spend an afternoon coding up!

Here are some results (with LRS):

problemsSTD_10Gi_rel_twee_6167_default.pkl 8769 0
problemsSTD_10Gi_rel_twee_6172_tgt-full.pkl 9225 456
problemsSTD_10Gi_rel_twee_6167_tgt-on.pkl 9265 496
problemsSTD_10Gi_rel_twee_6167ax_tgt-on.pkl 9284 515
problemsSTD_10Gi_rel_twee_6169_tgt-on.pkl 9291 522
problemsSTD_10Gi_rel_twee_6172_tgt-ground.pkl 9295 526

ax changed the status of the added definitions to axioms. 72 went bottom up. Full adds 37 robust uniques and (loses 35) in a different, shuffled experiment.

Kernel/TermTransformer.hpp
joe-hauns: So basically BUE (`BottomUpEvaluation`) does the same thing as the TTTT(`TermTransformerTransformTransformed`/`BottomUpTermTransformer`), but with a few performance optimizations. 
Say we have a term `f(t1...tn)`, and wanna bottom up apply the operation `φ` to it, then `TTTT` will compute `φ(t1) ...φ(tn)` first,  build the term `f(φ(t1)...φ(tn))`, and then compute `φ(f(φ(t1)...φ(tn)))`. 
BUE skips the step of creating `f(φ(t1)...φ(tn))`, and directly computes  `φ(f(φ(t1)...φ(tn)))`, which means we skip inserting some unnecessary intermediate terms in our term index.
Also BUE  not only works with terms but any sort of term-ish structure (i.e. DAGs). For example it's also used for z3 terms, and the arithmetically normalized terms `PolyNf`. 
Further BUE has some extras like never using dynamic dispatch, and cashing results in a hashmap, to not recompute results for the same subterms (which can be enabled using optional arguemnts).

The reason why both of them are used is that some of the old code I didn't write, but only updated uses TTTT. Everything I originally wrote uses BUE.

ibnyusuf: So via a little code refactoring, we could get rid of TTTT altogether and use only BUE?

selig: That sounds like a win

Is there an easy correctness argument that can be made for BUE being able to replace TTTT directly?

MichaelRawson:
> But you didn't factor in the equation my emotional pain of letting go of something I spent an afternoon coding up! ;)

Agreed! And it is neat. It can be reconsidered for removal in another ten years. ;-) Results look good though, I admit I'm kind of surprised.

MichaelRawson (commented):
Approve. Can't actually approve because it's my PR!

Kernel/TermTransformer.hpp
joe-hauns: So it works as follows:
In TTTT we have 
```
Signature: TermList evalute(TermList)
Call: evaluate(f(φ(t1)...φ(tn)))
```
so `evaluate` computes `φ`.
In BUE we have 
```
Signature: TermList evalute(TermList old, TermList* evalutedArgs)
Call: evaluate(f(t1, ... tn), { φ(t1)...φ(tn) })
```
This means we do not directly compute phi, but we get symbol + arity from `old` and the arguments from `evaluatedArgs`.

Concerning the hashmap: 
Whenever we `evaluate(old, ...)` returns a result `r`, we put `(old, r)` in a hashmap, and before bottom-up evaluating some subterm, we check whether the result is already present in the map.

I can look into doing this refactor at some point soon, but I'm pretty busy implementing thing for POPL atm.

joe-hauns: * and i agree that it would be good to have a separate PR for that

PR #372, opened 2022-05-23
delay reporting the error until we can check instruction limiting has been actually requested

quickbeam123:
@MichaelRawson, do you have access to a unix machine without "perf-ing" enabled? This new version should print an error there, but only once and only if `-i something_greater_zero` is requested. Cheers!


MichaelRawson (approved):
Yep, works. I will caution that some of the added stuff `std::strerror` and so on are also not technically signal-safe (but probably are), but IIRC we already weren't so we can't be _more_ unsafe...

PR #373, opened 2022-05-24
Remove unneeded backtracking option from unit tests other than induction

mezpusz:


joe-hauns (approved):
yess looks great! Thanks a lot for sorting that! 
Makes some of my merging way easier and disentangles things :)

PR #378, opened 2022-06-02
New precedence schemes and weight schemes for KBO

quickbeam123:


Kernel/KBO.cpp
```diff
@@ -346,20 +346,55 @@ struct FuncSigTraits {
...
+        [](auto _, auto prec) { return prec + 1; });
+    case Options::KboWeightGenerationScheme::INV_PRECEDENCE:
+      return KboWeightMap<SigTraits>::fromSomeUnsigned(precedenceExtractor,
+        [](auto max, auto prec) { return max-prec+1; });
+    case Options::KboWeightGenerationScheme::FREQUENCY:
+      return KboWeightMap<SigTraits>::fromSomeUnsigned(frequencyExtractor,
+        [](auto _, auto freq) { return freq > 0 ? freq : 1; });
+    case Options::KboWeightGenerationScheme::INV_FREQUENCY:
+      return KboWeightMap<SigTraits>::fromSomeUnsigned(frequencyExtractor,
+        [](auto max, auto freq) { return max > 0 ? max - freq + 1 : 1; });
```
joe-hauns, line 393: Really like this! U see so nicely what's going on!!

joe-hauns (commented):
I think that looks all very good! 
Very curious what results we get out of it!

Kernel/KBO.cpp
quickbeam123: You are the teacher of templated programming here!

PR #380, opened 2022-06-06
Allow loading schedule from a file

filipbartek:
Call `vampire --schedule file --schedule_file schedule.txt` to load the schedule from the input file "schedule.txt". Each line of "schedule.txt" is an encoded strategy.

# Known issues
- [x] If the input schedule file contains a line that is not a valid encoding of a strategy, Vampire fails in a dirty manner.

### DIFF
--- a/CASC/PortfolioMode.cpp
+++ b/CASC/PortfolioMode.cpp
@@ -178,6 +178,10 @@ bool PortfolioMode::performStrategy(Shell::Property* property)
     schedules.push(main);
   }
 
+  if (all_of(schedules.begin(), schedules.end(), [](const Schedule& schedule){ return schedule.isEmpty(); })) {
+    USER_ERROR("The schedule is empty.");
+  }
+
   int remainingTime = env.remainingTime()/100;
 
   while(remainingTime > 0) {
@@ -300,6 +304,9 @@ void PortfolioMode::getSchedules(Property& prop, Schedule& quick, Schedule& fall
   CALL("PortfolioMode::getSchedules");
 
   switch(env.options->schedule()) {
+  case Options::Schedule::FILE:
+    Schedules::getScheduleFromFile(env.options->scheduleFile(), quick);
+    break;
   case Options::Schedule::CASC_2019:
   case Options::Schedule::CASC:
     Schedules::getCasc2019Schedule(prop,quick,fallback);
--- a/CASC/Schedules.cpp
+++ b/CASC/Schedules.cpp
@@ -39,8 +39,8 @@ void Schedules::getScheduleFromFile(const vstring& filename, Schedule& quick)
   while (getline(schedule_file, line)) {
     // Allow structuring the schedule file with empty lines.
     // Allow documenting the schedule file with line comments.
-    // Interpret lines that start with '#' as comments.
-    if (line == "" or line[0] == '#') {
+    // Interpret lines that start with '%' as comments, following the TPTP convention.
+    if (line == "" or line[0] == '%') {
       continue;
     }
     Options opts;
--- a/CASC/Schedules.cpp
+++ b/CASC/Schedules.cpp
@@ -37,6 +37,12 @@ void Schedules::getScheduleFromFile(const vstring& filename, Schedule& quick)
   ASS(schedule_file.is_open());
   vstring line;
   while (getline(schedule_file, line)) {
+    // Allow structuring the schedule file with empty lines.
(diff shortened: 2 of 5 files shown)


CASC/PortfolioMode.cpp
```diff
@@ -178,6 +178,10 @@ bool PortfolioMode::performStrategy(Shell::Property* property)
     schedules.push(main);
   }
 
+  if (all_of(schedules.begin(), schedules.end(), [](const Schedule& schedule){ return schedule.isEmpty(); })) {
+    USER_ERROR("The schedule is empty.");
```
filipbartek, line 182: Otherwise we would end up in a hopeless infinite loop.

CASC/Schedules.cpp
```diff
@@ -16,10 +16,29 @@
...
+#include <fstream>
+
 using namespace Lib;
 using namespace Shell;
 using namespace CASC;
 
+void Schedules::getScheduleFromFile(const vstring& filename, Schedule& quick)
+{
+  if (filename == "") {
+    USER_ERROR("Schedule file was not set.");
```
filipbartek, line 28: I wonder if the default / invalid value can ever be "<empty>".

Shell/Options.cpp
```diff
@@ -155,10 +156,15 @@ void Options::init()
          "smtcomp",
          "smtcomp_2018",
          "struct_induction"});
-    _schedule.description = "Schedule to be run by the portfolio mode. casc and smtcomp usually point to the most recent schedule in that category. Note that some old schedules may contain option values that are no longer supported - see ignore_missing.";
+    _schedule.description = "Schedule to be run by the portfolio mode. casc and smtcomp usually point to the most recent schedule in that category. file loads the schedule from a file specified in --schedule_file. Note that some old schedules may contain option values that are no longer supported - see ignore_missing.";
     _lookup.insert(&_schedule);
     _schedule.reliesOn(UsingPortfolioTechnology());
 
+    _scheduleFile = StringOptionValue("schedule_file", "", "");
+    _scheduleFile.description = "Path to the input schedule file. Each line contains an encoded strategy. Disabled unless `--schedule file` is set.";
```
filipbartek, line 164: The backticks around `--schedule file` follow a convention used in some descriptions but not in the description of `_schedule`.

CASC/Schedules.cpp
MichaelRawson: I think this is fine. If somebody supplies an empty string somehow they probably don't have a file with an empty name.

MichaelRawson (approved):
Nice feature! Looks all OK to me. You can get slightly further validating input without much effort if you do something like
```
  while (getline(schedule_file, line)) {
    Options opts;
    try {
      opts.readFromEncodedOptions(line);
      opts.checkGlobalOptionConstraints();
    }
    catch (...) {
      USER_ERROR("bad strat: " + line);
    }
    quick.push(line);
  }
```
but it still doesn't completely validate it as you can have e.g. invalid options.

filipbartek:
> Nice feature! Looks all OK to me. You can get slightly further validating input without much effort if you do something like
> 
> ```
>   while (getline(schedule_file, line)) {
>     Options opts;
>     try {
>       opts.readFromEncodedOptions(line);
>       opts.checkGlobalOptionConstraints();
>     }
>     catch (...) {
>       USER_ERROR("bad strat: " + line);
>     }
>     quick.push(line);
>   }
> ```
Thanks for the suggestion, Michael! I've implemented it in 9f37455.

CASC/Schedules.cpp
```diff
@@ -16,10 +16,45 @@
...
+  if (schedule_file.fail()) {
+    USER_ERROR("Cannot open schedule file: " + filename);
+  }
+  ASS(schedule_file.is_open());
+  vstring line;
+  while (getline(schedule_file, line)) {
+    // Allow structuring the schedule file with empty lines.
+    // Allow documenting the schedule file with line comments.
+    // Interpret lines that start with '%' as comments, following the TPTP convention.
+    if (line == "" or line[0] == '%') {
```
filipbartek, line 43: > Comments extend from a % character to the end of the line
Source: https://www.tptp.org/TPTP/TR/TPTPTR.shtml

In SMT-LIB, comments start with a `;`.

filipbartek:
@quicquid I've resolved all Michael's comments. I believe this pull request is ready to be merged.

MichaelRawson:
Maybe @quickbeam123? But otherwise I agree!

PR #381, opened 2022-06-07
squish GCC warnings except one

MichaelRawson:
Haven't compiled with GCC in a while! Fix both debug and release warnings. In most cases a false positive, occasionally actually a bug but either dead-code or very unlikely. See comments inline.

### DIFF
--- a/Debug/Assertion.hpp
+++ b/Debug/Assertion.hpp
@@ -200,11 +200,9 @@ template <typename T>
 
 #define DEBUG_CODE(X) {}
 
-#define __IGNORE_WUNUSED(...) __PUSH_DIAGNOSTICS("GCC diagnostic ignored \"-Wreturn-type\"", __VA_ARGS__)
-
 #define ASS(Cond)  {}
-#define ALWAYS(Cond) (void) ( Cond );
-#define NEVER(Cond) (void) ( Cond );
+#define ALWAYS(Cond) if (!(Cond)) __UNREACHABLE;
+#define NEVER(Cond) if ((Cond)) __UNREACHABLE;
 
 #define ASS_REP(Cond, ReportedVal) {}
 #define ASS_REP2(Cond, ReportedVal, ReportedVal2) {}
--- a/FMB/DefinitionIntroduction.hpp
+++ b/FMB/DefinitionIntroduction.hpp
@@ -138,8 +138,7 @@ namespace FMB {
             // do not add definitions for constants
             if(ts->term()->arity()==0){args.push(*ts);}
             else{
-              Term* argT;
-              ALWAYS(_introduced.find(ts->term(),argT));
+              Term* argT = _introduced.get(ts->term());
               args.push(TermList(argT));
               updated = true; 
             }
--- a/Indexing/CodeTree.cpp
+++ b/Indexing/CodeTree.cpp
@@ -377,10 +377,8 @@ bool CodeTree::CodeOp::equalsForOpMatching(const CodeOp& o) const
     //during insertion into the code tree
     ASS_NEQ(instrSuffix(), SEARCH_STRUCT);
     return instrSuffix()==o.instrSuffix() && arg()==o.arg();
-#if VDEBUG
   default:
     ASSERTION_VIOLATION;
-#endif
   }
 }
 
--- a/Indexing/SubstitutionTree.cpp
+++ b/Indexing/SubstitutionTree.cpp
@@ -682,7 +682,10 @@ SubstitutionTree::UnificationsIterator::UnificationsIterator(SubstitutionTree* p
 svStack(32), literalRetrieval(query->isLiteral()),
   retrieveSubstitution(retrieveSubstitution), inLeaf(false),
 ldIterator(LDIterator::getEmpty()), nodeIterators(8), bdStack(8),
-clientBDRecording(false), tree(parent), useUWAConstraints(useC)
+clientBDRecording(false), useUWAConstraints(useC)
+#if VDEBUG
+  , tree(parent)
+#endif
 {
   CALL("SubstitutionTree::UnificationsIterator::UnificationsIterator");
 
--- a/Indexing/SubstitutionTree_FastGen.cpp
+++ b/Indexing/SubstitutionTree_FastGen.cpp
@@ -477,7 +477,7 @@ bool SubstitutionTree::FastGeneralizationsIterator::findNextLeaf()
   CALL("SubstitutionTree::FastGeneralizationsIterator::findNextLeaf");
 
   Node* curr;
-  bool sibilingsRemain;
+  bool sibilingsRemain = false;
   if(_inLeaf) {
     if(_alternatives.isEmpty()) {
       return false;
@@ -497,7 +497,7 @@ bool SubstitutionTree::FastGeneralizationsIterator::findNextLeaf()
   }
   for(;;) {
 main_loop_start:
-    unsigned currSpecVar;
+    unsigned currSpecVar = 0;
 
     if(curr) {
       if(sibilingsRemain) {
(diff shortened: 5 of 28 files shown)


Debug/Assertion.hpp
```diff
@@ -200,11 +200,9 @@ template <typename T>
 
 #define DEBUG_CODE(X) {}
 
-#define __IGNORE_WUNUSED(...) __PUSH_DIAGNOSTICS("GCC diagnostic ignored \"-Wreturn-type\"", __VA_ARGS__)
```
MichaelRawson, line 203: unused

Debug/Assertion.hpp
```diff
@@ -200,11 +200,9 @@ template <typename T>
...
-
 #define ASS(Cond)  {}
-#define ALWAYS(Cond) (void) ( Cond );
-#define NEVER(Cond) (void) ( Cond );
+#define ALWAYS(Cond) if (!(Cond)) __UNREACHABLE;
+#define NEVER(Cond) if ((Cond)) __UNREACHABLE;
```
MichaelRawson, line 205: I assert (without evidence but with reasonable confidence) this has a minimal, if not positive, impact on code generation. It persuades GCC that quite a few warnings aren't problems because they aren't reachable.

FMB/DefinitionIntroduction.hpp
```diff
@@ -138,8 +138,7 @@ namespace FMB {
             // do not add definitions for constants
             if(ts->term()->arity()==0){args.push(*ts);}
             else{
-              Term* argT;
-              ALWAYS(_introduced.find(ts->term(),argT));
+              Term* argT = _introduced.get(ts->term());
```
MichaelRawson, line 141: Equivalent-but-shorter unfailing hashmap lookup.

Indexing/CodeTree.cpp
```diff
@@ -377,10 +377,8 @@ bool CodeTree::CodeOp::equalsForOpMatching(const CodeOp& o) const
     //during insertion into the code tree
     ASS_NEQ(instrSuffix(), SEARCH_STRUCT);
     return instrSuffix()==o.instrSuffix() && arg()==o.arg();
-#if VDEBUG
   default:
     ASSERTION_VIOLATION;
-#endif
```
MichaelRawson, line 383: Not necessary any more, `ASSERTION_VIOLATION` compiles to an unreachable-hint in release mode.

Indexing/SubstitutionTree.cpp
```diff
@@ -682,7 +682,10 @@ SubstitutionTree::UnificationsIterator::UnificationsIterator(SubstitutionTree* p
 svStack(32), literalRetrieval(query->isLiteral()),
   retrieveSubstitution(retrieveSubstitution), inLeaf(false),
 ldIterator(LDIterator::getEmpty()), nodeIterators(8), bdStack(8),
-clientBDRecording(false), tree(parent), useUWAConstraints(useC)
+clientBDRecording(false), useUWAConstraints(useC)
+#if VDEBUG
+  , tree(parent)
+#endif
```
MichaelRawson, line 688: field only used in debug

Indexing/SubstitutionTree_FastGen.cpp
```diff
@@ -477,7 +477,7 @@ bool SubstitutionTree::FastGeneralizationsIterator::findNextLeaf()
   CALL("SubstitutionTree::FastGeneralizationsIterator::findNextLeaf");
 
   Node* curr;
-  bool sibilingsRemain;
+  bool sibilingsRemain = false;
```
MichaelRawson, line 480: I was unable to persuade GCC's control flow analyzer that these are in fact assigned before use, and I'm not completely sure myself because of the nested-loop-goto thing below. Set-to-zero is more or less free on modern hardware, save us some non-deterministic debugging later.

Kernel/FormulaTransformer.cpp
```diff
@@ -438,92 +438,5 @@ void ScanAndApplyFormulaUnitTransformer::updateModifiedProblem(Problem& prb)
   prb.invalidateEverything();
 }
 
-
```
MichaelRawson, line 441: There actually was a bug here related to alias analysis and a pure-virtual call (which incidentally was _really_ hard to find, thank you GCC), but luckily it's dead code, remove it.

Kernel/KBO.cpp
```diff
@@ -803,7 +803,7 @@ KboWeightMap<SigTraits> KboWeightMap<SigTraits>::fromSomeUnsigned(Extractor ex,
   auto nSym = SigTraits::nSymbols();
   DArray<KboWeight> weights(nSym);
 
-  unsigned max = 0;
+  decltype(ex(0)) max = 0;
```
MichaelRawson, line 806: Warning about signed/unsigned comparison, match types.

Kernel/PolynomialNormalizer.cpp
```diff
@@ -25,13 +25,16 @@ struct RenderMonom {
     using Monom        = Monom       <NumTraits>;
     auto& raw = x.raw();
     std::sort(raw.begin(), raw.end());
-    auto fstNum = Option<Numeral>();
+
```
MichaelRawson, line 28: GCC reckons that the value of `fstNum` could be unitialised. I _think_ it's a false-positive, but this version doesn't produce a warning and is clearer in my view.

Shell/FunctionDefinition.cpp
```diff
@@ -282,7 +282,7 @@ bool FunctionDefinition::removeAllDefinitions(UnitList*& units)
         scanIterator.del();
       } else if(_defs.get(d->fun)->twoConstDef){
         Def* d2;
-        _defs.pop(d->fun, d2);
+        ALWAYS(_defs.pop(d->fun, d2));
```
MichaelRawson, line 285: `d2` could be unitialised in the call to `reverse` otherwise.

Shell/TermAlgebra.cpp
```diff
@@ -26,6 +26,7 @@ TermAlgebraConstructor::TermAlgebraConstructor(unsigned functor, Lib::Array<unsi
   : _functor(functor), _hasDiscriminator(false), _destructors(destructors)
 {
   _type = env.signature->getFunction(_functor)->fnType();
+#if VDEBUG
```
MichaelRawson, line 29: Remaining stuff in the function is debug-only checks. Perhaps we should also change the signature of the constructor to not require this stuff in release, but it's a minor optimisation.

Shell/TheoryFinder.cpp
```diff
@@ -205,12 +205,9 @@ bool TheoryFinder::matchCode(const void* obj,
...
 
   // the clause, if any
-  const Clause* clause;
+  const Clause* clause = nullptr;
   // the length of this clause
-  int clength;
-#ifdef VDEBUG
-  bool clength_assigned = false;
-#endif
+  int clength = 0;
```
MichaelRawson, line 210: Again, can't convince the compiler that these are actually assigned, although I think they are. Just assign them, we'll get a quicker crash if `clause` is `nullptr` than if `clause` is unitialised.

Shell/TheoryFinder.cpp
```diff
@@ -205,12 +205,9 @@ bool TheoryFinder::matchCode(const void* obj,
   unsigned cp = 0; // code pointer
 
   // the clause, if any
-  const Clause* clause;
+  const Clause* clause = nullptr;
   // the length of this clause
-  int clength;
-#ifdef VDEBUG
-  bool clength_assigned = false;
-#endif
```
MichaelRawson, line 213: I was able to persuade `clang-tidy` about this in the past using an auxiliary variable. This seems a bit silly in retrospect.

Kernel/PolynomialNormalizer.cpp
MichaelRawson: @joe-hauns - meant to tag you in this, any objections?

joe-hauns: Strange that it gives a false positive there, but yeah looks good!

MichaelRawson:
@quickbeam123 - I think this is ready when you are, @joe-hauns is OK with the change to his stuff.

PR #382, opened 2022-06-08
fix bug temporarily by increasing `SPEC_UPPER_BOUND`

MichaelRawson:
As Martin reports, we get some crashes on long runs of Vampire. The symptom is a segfault in indices where something impossible happens. It looks like it might be a double-delete or suchlike, but the problem is actually that we exceed the constant `TermList::SPEC_UPPER_BOUND` which is used to differentiate between `isSpecialVar` and `isVSpecialVar` (aside: Vampire special var? Very special var?). This has unfortunate consequences.

Fix this temporarily by increasing the constant to the midpoint of its range. Given how much time/memory even these runs use on my machine I expect this to work for at least a few years. However, it would be nice to have a proper fix. I believe @ibnyusuf introduced `VSpecialVar`s for his work with applicative encodings. Any ideas?

### DIFF
--- a/Kernel/Term.hpp
+++ b/Kernel/Term.hpp
@@ -92,7 +92,8 @@ bool operator<(const TermList& lhs, const TermList& rhs);
 class TermList {
 public:
   CLASS_NAME(TermList)
-  static const unsigned SPEC_UPPER_BOUND = 10000000;
+  // divide by 4 because of the tag, by 2 to split the space evenly
+  static const unsigned SPEC_UPPER_BOUND = (UINT_MAX / 4) / 2;
   /** dummy constructor, does nothing */
   TermList() {}
   /** creates a term list and initialises its content with data */


ibnyusuf:
Hmm, interesting. The `VSpecialVar` mechanism is not to do with applicative terms, but linked to a unification with abstraction type of thing that we use in higher-order mode. I have long thought that we should combine it with unification with abstraction instead of having two mechanisms.

I can illustrate it with an example. Imagine that we have terms `p(8)` and `~p(2y)`. Instead of inserting these terms into the indices, we would insert terms `p(#1)` and `~p(#2)` into the indices where `#1` and `#2` are the so-called `VSpecialVar`s. They need to be disjont from both normal vars and special variables. Outside the indices, we mantain a mapping from very special vars to the terms they represent (I can't recall what we do about alpha-equivalence).

When attempting to unify the terms, we succeed and produce the constraint `#1 != #2`. Dereferencing the very special variable then gives us the actual constraint `8 != 2y`. 

Two points:

1. If there is some other, less problematic method of creating a third disjoint set of variables, I am happy to go with that.
2. As noted above, we should really merge the two unification with abstraction mechanism currently functional in Vampire. From before, I recall thinking that the solution I outlined above is cleaner than the other mechanism, but if this is mistaken we can go with the other mechanism.

MichaelRawson:
Thanks for the explanation! I got to applicative since they all seem to be created within `ApplicativeHelper`, but this makes more sense.

I agree that merging two UWA-ish things seems like the best course of action. As I recall UWA manages to "unify" different _functors_ rather than _variables_ (special or otherwise), so maybe they are not so similar? But I don't understand it well so I'm happy to be corrected.

MichaelRawson:
It's also possible this is a non-problem. We only "run out" of special variables because substitution trees do something a bit surprising with `_nextVar` to achieve freshness. When I get around to looking at that it might be possible to do something different and this may go away altogether.

ibnyusuf:
> Thanks for the explanation! I got to applicative since they all seem to be created within `ApplicativeHelper`, but this makes more sense.
> 
> I agree that merging two UWA-ish things seems like the best course of action. As I recall UWA manages to "unify" different _functors_ rather than _variables_ (special or otherwise), so maybe they are not so similar? But I don't understand it well so I'm happy to be corrected.

Correct, and the mechanism I outline can also create constraints between arbitrary terms. The very special variables are used to avoid inserting terms we do not want involved in the unification process, into the substitution tree. 

Unification between very special variables always succeeds and afterwards, the variables are derefernced to create the actual constraint.

ibnyusuf:
And thank you for tracing this, it must have been awful...

MichaelRawson:
> And thank you for tracing this, it must have been awful...

My pleasure, it was indeed "challenging"! But I can now recommend the use of replay debuggers (https://rr-project.org/ in my case) for this kind of thing should you come across another scary one. :-)

Issue #379, opened 2022-06-02
Suboptimal job distribution on multicore


quickbeam123:
`PortfolioMode::performStrategy` + `getExtraSchedules` behaves in a suboptimal way, because as it calls `runSchedule` on a particular schedule it waits for it to fully finish, which, towards the end of its job, will only make part of the available cores busy. 

A solution will require turning the preparation and while-loop logic of `PortfolioMode::performStrategy` into a strategy generating stream to be passed to `runSchedule` once and for all.

MichaelRawson:
I've been meaning to remove some of the scheduling complexity I sinfully added circa 2015 (!). I will do that and try to fix this at the same time.

quickbeam123:
That's a great idea! (In the meantime @filipbartek will add some basic support for loading a schedule from a file.)

filipbartek:
> In the meantime @filipbartek will add some basic support for loading a schedule from a file.

Done in #380.

PR #386, opened 2022-06-15
no break in core usage when getting extra schedules

MichaelRawson:
See #379. There is a kind of "pause" in core usage when we wait for all strategies from the previous schedule to finish, then load more strategies and spool up again. Don't do that.

Also, some general cleanup around `PortfolioMode`. I'd done some stuff a long time ago to make it more general, but we don't need it and it was a bit of a headache. Comments inline.


CASC/PortfolioMode.cpp
```diff
@@ -46,6 +47,14 @@ using namespace Lib;
 using namespace CASC;
 
 PortfolioMode::PortfolioMode() : _slowness(1.0), _syncSemaphore(2) {
+  unsigned cores = System::getNumberOfCores();
```
MichaelRawson, line 50: Inline a function/member from `ScheduleExecutor` to track how many worker processes we should spin up.

CASC/PortfolioMode.cpp
```diff
@@ -163,46 +177,26 @@ bool PortfolioMode::performStrategy(Shell::Property* property)
   // However, in SMTCOMP mode the fallback is universal for all
   // logics e.g. it's not very strong. Therefore, in SMTCOMP
   // mode we do main main_extra fallback fallback_extra
- 
-  Stack<Schedule> schedules;
+
```
MichaelRawson, line 180: I think we "really" want a single schedule made up of the constituent main/fallback parts. Then we can call `getExtraSchedules` once with the single mega-schedule when we need more.

CASC/PortfolioMode.cpp
```diff
@@ -353,72 +347,91 @@ void PortfolioMode::getSchedules(Property& prop, Schedule& quick, Schedule& fall
   }
 }
 
+bool PortfolioMode::runSchedule(Shell::Property *property, Schedule schedule) {
```
MichaelRawson, line 350: Inlined `ScheduleExecutor::run`. This is the main loop of the portfolio mode: keep a pool of processes busy with strategies, and ask for more strategies if we run out.

CASC/PortfolioMode.cpp
```diff
@@ -353,72 +347,91 @@ void PortfolioMode::getSchedules(Property& prop, Schedule& quick, Schedule& fall
...
+  Schedule::BottomFirstIterator it(schedule);
+  Set<pid_t> processes;
+  bool success = false;
+  int remainingTime;
+  while(Timer::syncClock(), remainingTime = env.remainingTime() / 100, remainingTime > 0)
+  {
+    // running under capacity, wake up more tasks
+    while(processes.size() < _numWorkers)
+    {
+      // after exhaustion we replace the schedule
```
MichaelRawson, line 362: The "give me more strategies" logic is here now.

CASC/PortfolioMode.cpp
```diff
@@ -353,72 +347,91 @@ void PortfolioMode::getSchedules(Property& prop, Schedule& quick, Schedule& fall
...
+      {
+        runSlice(code, remainingTime);
+        ASSERTION_VIOLATION; // should not return
+      }
+      ALWAYS(processes.insert(process));
+    }
 
-  vstring chopped;
-  int sliceTime = _mode->getSliceTime(sliceCode, chopped);
+    bool exited, signalled;
```
MichaelRawson, line 383: The logic is simplified a bit from the original since we can no longer have "stopped" processes.

CASC/PortfolioMode.cpp
```diff
@@ -353,72 +347,91 @@ void PortfolioMode::getSchedules(Property& prop, Schedule& quick, Schedule& fall
...
+
+  return success;
 }
 
 /**
  * Run a schedule.
  * Return true if a proof was found, otherwise return false.
  */
-bool PortfolioMode::runSchedule(Schedule& schedule)
+bool PortfolioMode::runScheduleAndRecoverProof(Shell::Property *property, Schedule schedule)
```
MichaelRawson, line 425: Rename to avoid clash, but also a bit more descriptive.

CASC/PortfolioMode.cpp
```diff
@@ -456,76 +469,60 @@ bool PortfolioMode::runSchedule(Schedule& schedule)
 }
 
 /**
- * Return the intended slice time in deciseconds and assign the slice
- * vstring with chopped time limit to @b chopped.
+ * Return the intended slice time in deciseconds
  */
-unsigned PortfolioMode::getSliceTime(vstring sliceCode,vstring& chopped)
```
MichaelRawson, line 462: Nothing used `chopped`, and it's kind of a weird thing to do.

CASC/PortfolioMode.cpp
```diff
@@ -456,76 +469,60 @@ bool PortfolioMode::runSchedule(Schedule& schedule)
...
 
   unsigned pos = sliceCode.find_last_of('_');
   vstring sliceTimeStr = sliceCode.substr(pos+1);
-  chopped.assign(sliceCode.substr(0,pos));
   unsigned sliceTime;
   ALWAYS(Int::stringToUnsignedInt(sliceTimeStr,sliceTime));
   ASS_G(sliceTime,0); //strategies with zero time don't make sense
 
   unsigned time = _slowness * sliceTime + 1;
+  // TODO: huh?
```
MichaelRawson, line 485: I have no idea what's going on here. If it's less than 10 deciseconds, increment by one?

CASC/PortfolioMode.cpp
```diff
@@ -146,6 +155,7 @@ bool PortfolioMode::searchForProof()
   return performStrategy(property);
 }
 
+// TODO this name confuses me
```
MichaelRawson, line 158: Cue bikeshedding about what we should call it. I find the original confusing because it suggests that it does something like `runSlice`, but actually it generates a schedule and runs it.

CASC/PortfolioMode.cpp
```diff
@@ -155,6 +165,10 @@ bool PortfolioMode::performStrategy(Shell::Property* property)
   Schedule main_extra;
   Schedule fallback_extra;
 
+  // TODO this doesn't make a lot of sense to me either
+  // first make *_extra (x3 multiplier with new options) schedules
+  // then later repeat them (x2 multiplier, no new options) until timeout
+  // if this is really what we want, fine, but why?!
```
MichaelRawson, line 171: Like the comment says. Any ideas what the logic is/was?

CASC/PortfolioMode.cpp
```diff
@@ -456,76 +469,60 @@ bool PortfolioMode::runSchedule(Schedule& schedule)
...
 {
   CALL("PortfolioMode::runSlice");
 
-  Options opt = *env.options;
-  opt.readFromEncodedOptions(sliceCode);
-  opt.setTimeLimitInDeciseconds(timeLimitInDeciseconds);
-  int stl = opt.simulatedTimeLimit();
-  if (stl) {
-    opt.setSimulatedTimeLimit(int(stl * _slowness));
+  int sliceTime = getSliceTime(sliceCode);
```
MichaelRawson, line 499: Inlined from `PortfolioSliceExecutor::runSlice`.

CASC/PortfolioMode.cpp
```diff
@@ -635,72 +632,4 @@ void PortfolioMode::runSlice(Options& strategyOpt)
   STOP_CHECKING_FOR_ALLOCATOR_BYPASSES;
 
   exit(resultValue);
-} // runSlice
-
-// BELOW ARE TWO LEFT-OVER FUNCTIONS FROM THE ORIGINAL (SINGLE-CHILD) CASC-MODE
```
MichaelRawson, line 640: Unused, ancient.

CASC/PortfolioMode.hpp
```diff
@@ -17,74 +17,48 @@
...
-  friend void PortfolioSliceExecutor::runSlice(vstring sliceCode, int terminationTime);
 public:
   static bool perform(float slowness);
-  unsigned getSliceTime(vstring sliceCode,vstring& chopped);
 
 private:
-
   // some of these names are kind of arbitrary and should be perhaps changed
-
+  unsigned getSliceTime(const vstring &sliceCode);
```
MichaelRawson, line 47: OK to be private, change signature.

Lib/PriorityQueue.hpp
```diff
@@ -1,72 +0,0 @@
-/*
```
MichaelRawson, line 1: Only used by `ScheduleExecutor`.

Lib/Sys/Multiprocessing.cpp
```diff
@@ -127,101 +127,6 @@ pid_t Multiprocessing::waitForChildTermination(int& resValue)
   return childPid;
 }
 
-/**
```
MichaelRawson, line 130: Unused.

Lib/Sys/Multiprocessing.cpp
```diff
@@ -239,7 +144,7 @@ void Multiprocessing::killNoCheck(pid_t child, int signal)
   ::kill(child, signal);
 }
 
-pid_t Multiprocessing::poll_children(bool &stopped, bool &exited, bool &signalled, int &code)
```
MichaelRawson, line 242: Nothing stops child processes any more.

MichaelRawson:
Also, the `tfork` unit test relied on `Multiprocessing`'s `sleep` and `waitForParticularChildTermination`, both of which I just removed. Oops. However, looking at it, it's very platform-specific and mostly tests behaviour of the system calls, rather than our (very thin) wrappers over them. I removed it altogether for now, if you want it back I'm happy to restore functionality.

CASC/PortfolioMode.cpp
quickbeam123: @selig maybe?

PR #389, opened 2022-06-23
Cleaning up KBO

ibnyusuf:
PR achieves the following:

- Setting different symbol precedence schemes affects type constructors as well (previously they always used default scheme)
- There can now only be a single unary function symbol of weight zero (in line with the theory). Exceptions are made for unary minus symbols until new calculus can be merged.
- For the purpose of the term ordering, a function symbol is considered constant or unary based on the number of _term_ arguments it accepts (in line with the theory).

### DIFF
--- a/Kernel/KBO.cpp
+++ b/Kernel/KBO.cpp
@@ -562,13 +562,16 @@ KBO KBO::testKBO()
 }
 
 void KBO::zeroWeightForMaximalFunc() {
+  CALL("KBO::zeroWeightForMaximalFunc");
   // actually, it's non-constant maximal func, as constants cannot be weight 0
 
   using FunctionSymbol = unsigned;
   auto nFunctions = _funcWeights._weights.size();
+  if (!nFunctions) {
+    return;
+  }
 
   FunctionSymbol maxFn = 0;
-
   for (FunctionSymbol i = 1; i < nFunctions; i++) {
     if (compareFunctionPrecedences(maxFn, i) == LESS) {
       maxFn = i;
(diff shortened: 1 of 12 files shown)


Kernel/KBO.cpp
```diff
@@ -567,72 +554,71 @@ KBO KBO::testKBO()
...
+  for (FunctionSymbol i = 1; i < nFunctions; i++) {
     if (compareFunctionPrecedences(maxFn, i) == LESS) {
-      maximalFunctions.replace(sort, i);
+      maxFn = i;
     }
   }
 
+  // TODO remove unary minus check once new 
+  // theory calculus goes into master
+  auto isUnaryMinus = [](unsigned functor){
```
joe-hauns, line 606: I locally made some changes so that the unary minus doesn't get added to the signature here, if it is not present in the problem. I'm not sure whether i should push it to this branch or make a separate PR. What do you think?
It's a quite small thing.

joe-hauns (approved):
Looks all good to me. 
I added some helpers to get rid of some copy-paste code i wrote at some point.

Kernel/Ordering.cpp
```diff
@@ -891,40 +901,42 @@ DArray<int> PrecedenceOrdering::predLevelsFromOptsAndPrec(Problem& prb, const Op
...
-    out << "% ===== end of function precedences ===== " << std::endl;
-  }
+      out << "%" << std::endl;
+    };
 
-  out << "%" << std::endl;
+  _show("type constructor", 
+      env.signature->typeCons(), 
+      [](unsigned f) { return env.signature->getTypeCon(f); },
+      [&](unsigned l, unsigned r){ return intoComparison(compareFunctionPrecedences(l,r)); });
```
joe-hauns, line 927: ah yeah .. is this the correct way of comparing type constructor precedences?
I copied it from your code, but I reckon that should be compareTypeConPrecedences, right?

Kernel/KBO.cpp
ibnyusuf: Don't have a preference either way. Adding it to this PR is fine by me.

Kernel/Ordering.cpp
ibnyusuf: Thanks, fixed!

Kernel/KBO.cpp
quickbeam123: Yes, @joe-hauns, please add here if you haven't already. Thank you!

PR #391, opened 2022-07-08
Ahmed improved SKIKBO

ibnyusuf:
Improvements to the higher-order SKIKBO ordering to make it more efficient and to facilitate the comparison of a larger number of non-ground terms.

In this PR, I also introduce polymorphic syntax sugar and add some unit tests relating to higher-order and polymorphism. More needs to be done on this.


Kernel/KBO.cpp
```diff
@@ -983,6 +983,7 @@ bool KboSpecialWeights<FuncSigTraits>::tryGetWeight(unsigned functor, unsigned&
 }
 
 template KboWeightMap<FuncSigTraits> KboWeightMap<FuncSigTraits>::dflt();
+template KboWeight KboWeightMap<FuncSigTraits>::symbolWeight(Term*) const;
```
quickbeam123, line 986: What do these actually mean? Are they necessary?

ibnyusuf: @joe-hauns Martin has found that if comment out these template specialisations, everything still compiles fine. Do we really need them?

joe-hauns: So it is there to instantiate the `KboWeightMap<...>::symbolWeight` for `FuncSigTraits`. 
This means that the compiler will monomorphize and precompile it for `FuncSigTraits`, so you can have the implementation in a cpp (not a hpp) file and still use the template instance for `FuncSigTraits` in a different file. 
Probably this function is not used outside from the cpp file (anymore). This means that the explicit instantiation is not used anymore. 
Sooo if it compiles without this, just get rid of it yes :)

PR #394, opened 2022-07-19
remove grounding mode and orphaned code

MichaelRawson:
See #157. `--mode grounding` is not obviously useful (to me at least - shout if you want it!) and doesn't work with e.g. many-sorted problems. It is also the last piece of code using some other parts that can now be removed. Also, clean up `SATClause` a bit.


selig:
Checked with Geoff that grounding mode is not used. He said

> As I didn't even know it exists, so I certainly won't miss it. I have not heard of anyone using it.

PR #399, opened 2022-07-22
Time Profiling Refactoring

joe-hauns:
Major refactoring of the time statistics (option `-tstat on`).

Instead of just having measurements per function call site the new time profiling tracs time in a call-stack-like tree structure.

In order to profile time the `TIME_TRACE(name)` is to be called, which measure the time from the call site, to the end of the current scope. 
If in some sub-block/function there is another call to the macro occurs, this will be rendered as a sub-execution in the final outputting of the time profile.
This gives the possibliity to not only show how much time some function call (e.g. looking up something in an index) takes in the whole execution of vampire, but also how much time it takes up in an individual rule, etc. 

In addition to this tree-like outputting of the time profile a flattened version of the time profile is outputted, that shows the total amount of time some call site takes up over all the branches of the time trace tree.

Example output:

Vampire call:
```
./time-profiling-Debug/bin/vampire_z3_dbg_time-profiling_6312 --decode dis+1010_4_afp=40000:afq=1.1:amm=sco:anc=none:cond=fast:fde=none:gs=on:gsem=on:ile=on:irw=on:nm=4:nwc=1:sp=reverse_arity:uhcvi=on_34 -stat none -sas z3 AUFLIRA/FFT/smtlib.613529.smt2 --input_syntax smtlib2 -tstat on
```

Output:
```
% Time limit reached!
===== start of time trace =====
[root] (total: 3500 ms, avg: 3500 ms, cnt: 1)
  ├──[88%] main loop (total: 3106 ms, avg: 3106 ms, cnt: 1)
  │    ├──[97%] run (total: 3023 ms, avg: 3023 ms, cnt: 1)
  │    │    ├──[44%] activation (total: 1339 ms, avg: 690 μs, cnt: 1940)
  │    │    │    ├──[88%] clause generation (total: 1184 ms, avg: 70 μs, cnt: 16922)
  │    │    │    │    ├──[45%] superposition (total: 540 ms, avg: 29 μs, cnt: 18144)
  │    │    │    │    │    ├──[60%] perform superposition (total: 327 ms, avg: 44 μs, cnt: 7407)
  │    │    │    │    │    │    └──[28%] term sharing (total: 93 ms, avg: 1932 ns, cnt: 48434)
  │    │    │    │    │    └──[ 0%] term sharing (total: 335 μs, avg: 1349 ns, cnt: 249)
  │    │    │    │    ├──[41%] resolution (total: 492 ms, avg: 41 μs, cnt: 11800)
  │    │    │    │    │    └──[ 5%] term sharing (total: 26 ms, avg: 4438 ns, cnt: 6016)
  │    │    │    │    └──[ 0%] term sharing (total: 346 μs, avg: 827 ns, cnt: 419)
  │    │    │    ├──[ 8%] add clause (total: 120 ms, avg: 62 μs, cnt: 1940)
  │    │    │    │    ├──[22%] backward demodulation index maintenance (total: 27 ms, avg: 13 μs, cnt: 1940)
  │    │    │    │    │    └──[ 2%] term sharing (total: 592 μs, avg: 1249 ns, cnt: 474)
  │    │    │    │    ├──[18%] backward superposition index maintenance (total: 21 ms, avg: 11 μs, cnt: 1940)
  │    │    │    │    │    └──[ 1%] term sharing (total: 231 μs, avg: 739 ns, cnt: 313)
  │    │    │    │    ├──[13%] forward demodulation index maintenance (total: 16 ms, avg: 8983 ns, cnt: 1841)
  │    │    │    │    ├──[13%] binary resolution index maintenance (total: 15 ms, avg: 8093 ns, cnt: 1940)
  │    │    │    │    │    └──[ 2%] term sharing (total: 366 μs, avg: 718 ns, cnt: 510)
  │    │    │    │    ├──[12%] unit clause index maintenance (total: 15 ms, avg: 8330 ns, cnt: 1841)
  │    │    │    │    │    └──[ 2%] term sharing (total: 310 μs, avg: 785 ns, cnt: 396)
  │    │    │    │    ├──[ 7%] forward superposition index maintenance (total: 9203 μs, avg: 4744 ns, cnt: 1940)
  │    │    │    │    │    └──[ 0%] term sharing (total: 55 μs, avg: 648 ns, cnt: 86)
  │    │    │    │    └──[ 1%] forward subsumption index maintenance (total: 1996 μs, avg: 20 μs, cnt: 99)
  │    │    │    │         └──[13%] term sharing (total: 267 μs, avg: 1950 ns, cnt: 137)
  │    │    │    ├──[ 0%] clause selection (total: 2253 μs, avg: 1161 ns, cnt: 1940)
  │    │    │    │    └──[52%] literal selection (total: 1189 μs, avg: 613 ns, cnt: 1940)
  │    │    │    ├──[ 0%] splitting (total: 354 μs, avg: 182 ns, cnt: 1940)
  │    │    │    └──[ 0%] redundancy check (total: 325 μs, avg: 167 ns, cnt: 1940)
  │    │    ├──[35%] forward simplification (total: 1073 ms, avg: 111 μs, cnt: 9656)
  │    │    │    ├──[37%] forward demodulation (total: 407 ms, avg: 51 μs, cnt: 7864)
  │    │    │    │    └──[98%] forward demodulation (total: 402 ms, avg: 51 μs, cnt: 7864)
  │    │    │    │         └──[10%] term sharing (total: 41 ms, avg: 1765 ns, cnt: 23242)
  │    │    │    ├──[36%] forward subsumption (total: 393 ms, avg: 40 μs, cnt: 9656)
  │    │    │    │    └──[39%] forward subsumption resolution (total: 154 ms, avg: 19 μs, cnt: 7864)
  │    │    │    ├──[15%] global subsumption (total: 169 ms, avg: 23 μs, cnt: 7343)
  │    │    │    │    └──[17%] term sharing (total: 29 ms, avg: 2177 ns, cnt: 13631)
  │    │    │    ├──[ 0%] term sharing (total: 402 μs, avg: 8387 ns, cnt: 48)
  │    │    │    ├──[ 0%] splitting component index maintenance (total: 276 μs, avg: 2715 ns, cnt: 102)
  │    │    │    │    └──[75%] hvci insert (total: 208 μs, avg: 2044 ns, cnt: 102)
  │    │    │    │         └──[25%] hvci compute hash (total: 52 μs, avg: 514 ns, cnt: 102)
  │    │    │    └──[ 0%] splitting component index usage (total: 157 μs, avg: 3024 ns, cnt: 52)
  │    │    │         └──[70%] hvci retrieve (total: 110 μs, avg: 2130 ns, cnt: 52)
  │    │    │              └──[53%] hvci compute hash (total: 59 μs, avg: 1138 ns, cnt: 52)
  │    │    ├──[ 7%] immediate simplification (total: 221 ms, avg: 22 μs, cnt: 9704)
  │    │    │    ├──[85%] interpreted evaluation (total: 190 ms, avg: 19 μs, cnt: 9574)
  │    │    │    │    └──[21%] term sharing (total: 41 ms, avg: 849 ns, cnt: 48539)
  │    │    │    └──[ 3%] condensation (total: 7030 μs, avg: 1041 ns, cnt: 6750)
  │    │    ├──[ 5%] SAT solver (total: 165 ms, avg: 7216 μs, cnt: 23)
  │    │    ├──[ 4%] backward simplification (total: 123 ms, avg: 63 μs, cnt: 1940)
  │    │    │    ├──[53%] backward demodulation (total: 66 ms, avg: 34 μs, cnt: 1940)
  │    │    │    │    └──[13%] term sharing (total: 8854 μs, avg: 8432 ns, cnt: 1050)
  │    │    │    ├──[ 9%] forward demodulation index maintenance (total: 11 ms, avg: 11 μs, cnt: 991)
  │    │    │    ├──[ 6%] backward demodulation index maintenance (total: 7842 μs, avg: 7913 ns, cnt: 991)
  │    │    │    │    └──[ 0%] term sharing (total: 4177 ns, avg: 1044 ns, cnt: 4)
  │    │    │    ├──[ 6%] backward superposition index maintenance (total: 7737 μs, avg: 7807 ns, cnt: 991)
  │    │    │    │    └──[ 0%] term sharing (total: 3481 ns, avg: 870 ns, cnt: 4)
  │    │    │    ├──[ 4%] binary resolution index maintenance (total: 5705 μs, avg: 5756 ns, cnt: 991)
  │    │    │    ├──[ 4%] unit clause index maintenance (total: 5330 μs, avg: 5379 ns, cnt: 991)
  │    │    │    └──[ 4%] forward superposition index maintenance (total: 5256 μs, avg: 5304 ns, cnt: 991)
  │    │    ├──[ 1%] passive container maintenance (total: 35 ms, avg: 4257 ns, cnt: 8257)
  │    │    ├──[ 0%] minimizing solver time (total: 4358 μs, avg: 189 μs, cnt: 23)
  │    │    └──[ 0%] splitting model update (total: 111 μs, avg: 4837 ns, cnt: 23)
  │    └──[ 2%] init (total: 82 ms, avg: 82 ms, cnt: 1)
  ├──[ 7%] preprocessing (total: 262 ms, avg: 262 ms, cnt: 1)
  │    ├──[52%] property evaluation (total: 136 ms, avg: 34 ms, cnt: 4)
  │    ├──[ 9%] term sharing (total: 24 ms, avg: 2953 ns, cnt: 8346)
  │    └──[ 0%] naming (total: 1844 μs, avg: 2984 ns, cnt: 618)
  │         └──[ 1%] term sharing (total: 30 μs, avg: 3075 ns, cnt: 10)
  ├──[ 2%] parsing (total: 87 ms, avg: 87 ms, cnt: 1)
  │    ├──[20%] term sharing (total: 17 ms, avg: 3107 ns, cnt: 5680)
  │    └──[ 0%] sort sharing (total: 863 μs, avg: 599 ns, cnt: 1440)
  └──[ 1%] property evaluation (total: 37 ms, avg: 37 ms, cnt: 1)
===== end of time trace =====

===== start of flattened time profile =====
[root] (total: 3500 ms, avg: 3500 ms, cnt:      1)
  ├──[88%] main loop                                (total: 3106 ms, avg: 3106 ms, cnt:      1)
  ├──[86%] run                                      (total: 3023 ms, avg: 3023 ms, cnt:      1)
  ├──[38%] activation                               (total: 1339 ms, avg:  690 μs, cnt:   1940)
  ├──[33%] clause generation                        (total: 1184 ms, avg:   70 μs, cnt:  16922)
  ├──[30%] forward simplification                   (total: 1073 ms, avg:  111 μs, cnt:   9656)
  ├──[23%] forward demodulation                     (total:  810 ms, avg:   51 μs, cnt:  15728)
  ├──[15%] superposition                            (total:  540 ms, avg:   29 μs, cnt:  18144)
  ├──[14%] resolution                               (total:  492 ms, avg:   41 μs, cnt:  11800)
  ├──[11%] forward subsumption                      (total:  393 ms, avg:   40 μs, cnt:   9656)
  ├──[ 9%] perform superposition                    (total:  327 ms, avg:   44 μs, cnt:   7407)
  ├──[ 8%] term sharing                             (total:  286 ms, avg: 1816 ns, cnt: 157588)
  ├──[ 7%] preprocessing                            (total:  262 ms, avg:  262 ms, cnt:      1)
  ├──[ 6%] immediate simplification                 (total:  221 ms, avg:   22 μs, cnt:   9704)
  ├──[ 5%] interpreted evaluation                   (total:  190 ms, avg:   19 μs, cnt:   9574)
  ├──[ 4%] property evaluation                      (total:  174 ms, avg:   34 ms, cnt:      5)
  ├──[ 4%] global subsumption                       (total:  169 ms, avg:   23 μs, cnt:   7343)
  ├──[ 4%] SAT solver                               (total:  165 ms, avg: 7216 μs, cnt:     23)
  ├──[ 4%] forward subsumption resolution           (total:  154 ms, avg:   19 μs, cnt:   7864)
  ├──[ 3%] backward simplification                  (total:  123 ms, avg:   63 μs, cnt:   1940)
  ├──[ 3%] add clause                               (total:  120 ms, avg:   62 μs, cnt:   1940)
  ├──[ 2%] parsing                                  (total:   87 ms, avg:   87 ms, cnt:      1)
  ├──[ 2%] init                                     (total:   82 ms, avg:   82 ms, cnt:      1)
  ├──[ 1%] backward demodulation                    (total:   66 ms, avg:   34 μs, cnt:   1940)
  ├──[ 1%] passive container maintenance            (total:   35 ms, avg: 4257 ns, cnt:   8257)
  ├──[ 0%] backward demodulation index maintenance  (total:   34 ms, avg:   11 μs, cnt:   2931)
  ├──[ 0%] backward superposition index maintenance (total:   29 ms, avg:   10 μs, cnt:   2931)
  ├──[ 0%] forward demodulation index maintenance   (total:   27 ms, avg: 9791 ns, cnt:   2832)
  ├──[ 0%] binary resolution index maintenance      (total:   21 ms, avg: 7303 ns, cnt:   2931)
  ├──[ 0%] unit clause index maintenance            (total:   20 ms, avg: 7297 ns, cnt:   2832)
  ├──[ 0%] forward superposition index maintenance  (total:   14 ms, avg: 4933 ns, cnt:   2931)
  ├──[ 0%] condensation                             (total: 7030 μs, avg: 1041 ns, cnt:   6750)
  ├──[ 0%] minimizing solver time                   (total: 4358 μs, avg:  189 μs, cnt:     23)
  ├──[ 0%] clause selection                         (total: 2253 μs, avg: 1161 ns, cnt:   1940)
  ├──[ 0%] forward subsumption index maintenance    (total: 1996 μs, avg:   20 μs, cnt:     99)
  ├──[ 0%] naming                                   (total: 1844 μs, avg: 2984 ns, cnt:    618)
  ├──[ 0%] literal selection                        (total: 1189 μs, avg:  613 ns, cnt:   1940)
  ├──[ 0%] sort sharing                             (total:  863 μs, avg:  599 ns, cnt:   1440)
  ├──[ 0%] splitting                                (total:  354 μs, avg:  182 ns, cnt:   1940)
  ├──[ 0%] redundancy check                         (total:  325 μs, avg:  167 ns, cnt:   1940)
  ├──[ 0%] splitting component index maintenance    (total:  276 μs, avg: 2715 ns, cnt:    102)
  ├──[ 0%] hvci insert                              (total:  208 μs, avg: 2044 ns, cnt:    102)
  ├──[ 0%] splitting component index usage          (total:  157 μs, avg: 3024 ns, cnt:     52)
  ├──[ 0%] hvci compute hash                        (total:  111 μs, avg:  724 ns, cnt:    154)
  ├──[ 0%] splitting model update                   (total:  111 μs, avg: 4837 ns, cnt:     23)
  └──[ 0%] hvci retrieve                            (total:  110 μs, avg: 2130 ns, cnt:     52)
===== end of flattened time profile =====
```


joe-hauns:
There is one blocking question i wanted to discuss. 
In contrast to the original time proifiling, all time-profiling related behaviour is hidden behind marcos. This allos us to compile away all time tracing for release builds.
Right now there is a flat `ENABLE_TIME_TRACING` that is meant to be used for that. It's untested atm as I don't really know how we want to implement this. 
The obvious first choice would be that `ENABLE_TIME_PROFILING` could always be on in debug mode and always off in release mode, as we don't want to have any time profiling going on when we run say in SMT-COMP.
The issue with that is though that we might wanna see the time profile in release mode still as it might be quite different from the one in debug mode and we might wanna analyse it as well. Therefore I'd propose introducing some new a compile option a lá `VZ3` in order to enable/disable time profiling. 
The question is what we want for default, and what the name for this flag should be. Do you have any suggestions?

quickbeam123:
First impression - the output looks very cool! :) Looking forward to try it out myself (and checking out the code - out of time for that now). 

But to your blocking question:
1) it IMO almost does not make sense to do time profiling in debug-mode. (It's not that debug is just uniformly slower because of some assertion checking. The assertion checking varies in how expensive it is, so it's hard to generalized anything from a performance profile of debug to release, where performance matters).
2) the experience with the original time statistics was that it was so cheap that it didn't make much sense to have release without it. If this new one is a bit similar, we could have profiling by default in release and only turn if off (if necessary) for the competitions... (this needs to be measured).

MichaelRawson:
This is cool! I definitely don't object and I will get around to review soon.


My big question is what this gets us over, say, a profiling tool like `perf`. This isn't necessarily a blocker: we have the `CALL` macros even though we could probably achieve the same thing via system APIs or a core dump. But I admit my imagination for the `TimeCounter` stuff was that I would eventually kill it, squashing any resistance with "use a profiler".

joe-hauns:
Right I used a perf-ish thing before, and there were some issues for me there.
- you always get the whole trace, which contains looooooaaaads of rubbish helper methods which make it reaaally hard to understand what's going on. With this we just have explicitly enabled nodes in the call stack that provide more of a semantic overview of what's going on
- you have to get perf or a similar tool running on whatever server you want to run vampire on, and somehow manage to include that into your workflow of benchmarking and so on. With this we get the traces straight in vampires stdout which is easily to capture by whatever script. I was thinking of potentially adding an option for serializing the tree machine instead of human-readable (e.g. as json or so) if we want to automate merging the traces of multiple runs into a big one. This would (probably also not be so easy with a perf-like tool). 
- We can make our trace consider two different call-sites the same. e.g. in `timeTraceIter` both `hasNext` and `next` go into the same node, which is really handy as we don't know when the actual work will be performed.

joe-hauns:
> But to your blocking question:
> 
>     1. it IMO almost does not make sense to do time profiling in debug-mode. (It's not that debug is just uniformly slower because of some assertion checking. The assertion checking varies in how expensive it is, so it's hard to generalized anything from a performance profile of debug to release, where performance matters).
>
>     2. the experience with the original time statistics was that it was so cheap that it didn't make much sense to have release without it. If this new one is a bit similar, we could have profiling by default in release and only turn it off (if necessary) for the competitions... (this needs to be measured).

Yeah true say. We surely wanna a measure time without assertions and that.
But enabling it by default would mean that users of vampire who just compile in release mode as the readme suggests will run vampire with time profiling on. 
So maybe we should have three levels of debug/release. Like `mode = {Default,Profiling,Release}` or something?

MichaelRawson:
>Right I used a perf-ish thing before, and there were some issues for me there.

Cool - as long as you've thought about it, sounds great! On the review pile for tomorrow. :-)

Debug/TimeProfiling.hpp
```diff
@@ -0,0 +1,208 @@
...
+
+
+/**
+ * A class to trace time for particular blocks. this class shall never be used directly,
+ * as it is preprocessed away when the flag ENABLE_TIME_PROFILING is set to false.
+ * Use the macros TIME_TRACE, TIME_TRACE_EXPR, and TIME_TRACE_NEW_ROOT instead.
+ */
+class TimeTrace
+{
+  using Clock = std::chrono::high_resolution_clock;
```
MichaelRawson, line 83: [cppreference](https://en.cppreference.com/w/cpp/chrono/high_resolution_clock) maintains that `std::chrono::steady_clock` might be preferred here. Any thoughts?

Debug/TimeProfiling.hpp
```diff
@@ -0,0 +1,208 @@
...
+ * The parameter \param name is a human-readable name to be presented when the time trace is being
+ * outputed.
+ * If you want two blocks to contribute to the same node in the tree, they have to occur in the same
+ * subtree, and have to have the same const char* name. Note that just specifying the same literal
+ * twice does not guarantee that const char* are the same, even though that is the case for some
+ * compilers. Therefore one should use one static const char* for these kind of measurements.
+ *
+ * In addition to the time profiling tree a flattened version of the tree is being outputted at the
+ * end of a program execution, that presents the sum of all the runtimes a node was encountered on
+ * any branch of the tree.
```
MichaelRawson, line 41: It may be worth noting here that construction of these things has non-trivial overhead, so probably they should not be used in tight loops...right?

Debug/TimeProfiling.hpp
```diff
@@ -0,0 +1,208 @@
...
+      _cnt += other._cnt;
+    }
+  };
+
+
+  struct Node {
+    CLASS_NAME(Node)
+    USE_ALLOCATOR(Node)
+    const char* name;
+    Lib::Stack<unique_ptr<Node>> children;
```
MichaelRawson, line 117: How does the compiler resolve `std::unique_ptr` from here?! I guess something in the includes.

Indexing/TermSharing.cpp
```diff
@@ -216,11 +218,11 @@ AtomicSort* TermSharing::insert(AtomicSort* sort)
   ASS(!sort->isSpecial());
   ASS(sort->isSort());
 
-  // cannot use TC_TERM_SHARING
+  // cannot use TERM_SHARING
```
MichaelRawson, line 221: Is this still true? Could we fix it now?

Inferences/FastCondensation.cpp
```diff
@@ -71,7 +71,7 @@ Clause* FastCondensation::simplify(Clause* cl)
 {
   CALL("FastCondensation::perform");
 
-  TimeCounter tc(TC_CONDENSATION);
+  TIME_TRACE("condensation");
```
MichaelRawson, line 74: Does this want to be the same `"condensation"` as slow-condensation? If so should be `static const char *`, if not maybe could be `"fast condensation"`?

Inferences/ForwardDemodulation.cpp
```diff
@@ -73,8 +73,9 @@ template <bool combinatorySupSupport>
 bool ForwardDemodulationImpl<combinatorySupSupport>::perform(Clause* cl, Clause*& replacement, ClauseIterator& premises)
 {
   CALL("ForwardDemodulation::perform");
+  TIME_TRACE("forward demodulation")
 
-  TimeCounter tc(TC_FORWARD_DEMODULATION);
+  TIME_TRACE("forward demodulation");
```
MichaelRawson, line 78: Duplicate?

Lib/MacroUtils.hpp
```diff
@@ -0,0 +1,23 @@
...
+ * copyright laws.
+ *
+ * This source code is distributed under the licence found here
+ * https://vprover.github.io/license.html
+ * and in the source directory
+ */
+/**
+ * @file Reflection.hpp
+ * Defines class Reflection.
+ */
```
MichaelRawson, line 13: Love the file! Doesn't implement `Reflection`. ;-)

Lib/Stack.hpp
```diff
@@ -613,6 +613,21 @@ class Stack
...
+    C next() { return ConstRefIterator::next(); }
+  };
+
+  RefIterator iter() &
+  { return RefIterator(*this); }
+
+  ConstRefIterator iter() const&
+  { return ConstRefIterator(*this); }
+
+
```
MichaelRawson, line 629: Could you explain what's going on here for me? Not sure how it relates to current change.

MichaelRawson (changes requested):
Looks fine to me, just a few comments. 84 files looks scary but most are straightforward. :-) Let me get to the preprocessor/CMake stuff now...

Debug/TimeProfiling.hpp
joe-hauns: Agreed this should be changed!

Debug/TimeProfiling.hpp
joe-hauns: You mean the construction of the flattened profile or the normal one? 
The flattened one is constructed in post-processing, and the other one is not too expensive (no allocations except for the first call, lookup in a `Stack` with compile-time bound (small) size, and adding up two durations). But you are right, we should deffo mention that it's not for free and this can surely make quite some impact for very short loops.

joe-hauns: Speaking of that. I was actually also thinking of whether it's the right thing to have this in term sharing, as I think this is a part that is called veeeery often. I only included it as there was a `TimeCounter` in the old implemenation on the same place.

Debug/TimeProfiling.hpp
joe-hauns: You mean why i could avoid the `std::`? There's a `using namespace std;` in quite a bunch of vampire headers...

Indexing/TermSharing.cpp
joe-hauns: I actually didn't look into this, because i just did a semi-automated refactoring for replacing `TimeCounter` vs `TIME_TRACE`. 
We can definitely use `TERM_SHARING` here with the new impl. The question is whether we want to. I think as probably noone is using the time statistics right now we can just remove the comment and if someone needs to know some specific details they can just change the macro call.

Inferences/FastCondensation.cpp
joe-hauns: yeah true .. i wasn't sure about that so i did the only wrong thing apparently hehe
I'll change it to `"fast condensation"`

Lib/Stack.hpp
joe-hauns: Soooo the issue was that the `Stack` iterators were not consistent with vampire's (informal) iterator interface. 
That is `Stack::Iterator` used to have `DECL_ELEMENT_TYPE(C)` but the function `next` did return `C&`. 
This means in cases where `C` was not copy constructible, one couldn't use our iterator adaptors (`MappingIterator` etc). Still some parts of the codebase (i.e. the implemenation of constraints in `Options.hpp`) relied on the iterators returning a reference on next (because they dealt with `unique_ptr` which are not copyable). In order to resolve this I created two versions of iterators: `Stack::Iterator` and `Stack::RefIterator` (where one is just a copying wrapper around the other one), that both conform to the iterator interface.
(And of course the same for `ConstRefIterator` vs `ConstIterator`).

(Sorry that should have maybe been a separate PR? Just changed it bc I needed it.)

joe-hauns:
> I'm moderately against having a new "optimisation" level like Debug, Release, ReleaseProfile - not for any good reason, but because it goes against the CMake convention, and in the past my attempts to deviate from what CMake expects have been punished.

Right I feel you there. So I'll go with just having a flag `VTIME_PROFILING` for now. I'll let you know when it's tested.

Debug/TimeProfiling.hpp
MichaelRawson: Sure, I just meant that invoking `TIME_TRACE` inside a _really_ hot loop will distort results (and performance) - I think perf is actually pretty good like you say, good job! Probably a good shout with term sharing, it's pretty hot I think.

Debug/TimeProfiling.hpp
MichaelRawson: Ah, right. I wondered if this was some daft ADLish thing, I still have no idea how C++ name lookup works. OK, I'll try and purge those at some point then...

Lib/Stack.hpp
MichaelRawson: Oh, right. Yeah, that's gross - good catch, no objections here.

MichaelRawson (approved):

MichaelRawson:
All looks fine to me. @quickbeam123 - I think Joe has a few little bits to do, then it should be ready for your scrutiny. I might at some point see if there's any easy perf wins to reduce the distortion we get from the instrumentation but that should be a separate PR.

ibnyusuf:
> > I'm moderately against having a new "optimisation" level like Debug, Release, ReleaseProfile - not for any good reason, but because it goes against the CMake convention, and in the past my attempts to deviate from what CMake expects have been punished.
> 
> Right I feel you there. So I'll go with just having a flag `VTIME_PROFILING` for now. I'll let you know when it's tested.

If time profiling is cheap, could we not just hide it behind a standard user option instead of a compiler flag?

MichaelRawson:
>If time profiling is cheap, could we not just hide it behind a standard user option instead of a compiler flag?

Good idea. I think it might be expensive enough that we want it off for competitions, but paging @joe-hauns - any insight?

joe-hauns:
Yeah so there is already the option `-tstat`. If it that off nothing's being profiled.

I don't think the profiling itself is expensive, but that it might block optimizations, as the compiler needs to keep the time profiling scopes and such.

joe-hauns:
Okay I just introduced the compile flag `VTIME_PROFILING`, which is for now unconditionally set to `1` in the CMakeFile. 
It's also tested to compile with setting it to `0`.
From my side the PR is now good to go :)

quickbeam123:
I was just wondering, since we in many cases declare `const char*` statics for the textual descriptions of the measured scopes (as is advised as soon as we refer to it more than once), wouldn't it be nice to have an in-house rule that these should all appear in one place?

`TimeTrace::Groups` was a bit confusing for me at first, as I was expecting "Groups" to have special meaning, different from plain `const char*` scope.

How about we always declare a new static constant directly under `TimeTrace`? And if were to go
`static constexpr const char* const ID`
we could even initialize them in place and from C++17 we even get away with no module definition.

Opinions?

joe-hauns:
> I was just wondering, since we in many cases declare const char* statics for the textual descriptions of the measured scopes (as is advised as soon as we refer to it more than once), wouldn't it be nice to have an in-house rule that these should all appear in one place?

Yes I thought about that. So what I thought of as rules was 
1. if they are used multiple times across files use a static in `TimeProfiling.hpp`
2. it they are used multiple times with one file use a static in that file (so in that case u can be sure that noone else uses the same scope name)
3. if u use it only once use a string literal

Of course we can drop 2. If not we should probably make these guidelines explicit in the comments, and have a naming/location convention for the file-local statics.

> How about we always declare a new static constant directly under TimeTrace? And if were to go
> static constexpr const char* const ID
> we could even initialize them in place and from C++17 we even get away with no module definition.

Yeah that sounds like a shout. The thing with `static constexpr ...` was that I wasn't sure of the semantics of pointeres in in a constexpr context. i.e. I wasn't sure if u r guaranteed to get the same pointer if u use the same value or so... but actually thinking about it now seems like there shouldn't be any issues.

quickbeam123:
I implemented and pushed:

"A suggestion: let's drop the Group struct and eventually declare all the TIME_TRACE strings as 'static constexpr const char* const' in TimeProfiling.hpp"

a commit which does this with those already present there (PREPROCESSING/PARSING/LITERAL_ORDER_AFTERCHECK).

Unless nobody hates the idea to much (feel free to shout out) I am happy to do the similarly with all the other strings.

joe-hauns:
I'm happpy about not having the group struct. 
But I don't think we should have a static for *every* string. For the ones that are being used only once I think we should definitely use a string literal, so someone who reads the code knows that really only this one block is being time measured and nothing else.

quickbeam123:
Hi Joe, welcome back from vacation... Glad you don't mind the group removed. How about I finish it off by fixing the "multi-use" ones? (I find it especially ugly to create a new module cpp file just to store the static there - as is done in one place, if I remember correctly.)

Inferences/ArithmeticSubtermGeneralization/VariableMultiplicationGeneralizationImpl.cpp
```diff
@@ -265,7 +265,8 @@ SimplifyingGeneratingInference1::Result applyRule(Clause* cl, bool doOrderingChe
 
       /* one variable with power one needs to be kept, per varible region */
       auto var = iterTraits(region.iter())
-        .filter([](auto p) { return p.apply([](auto& t){ return t.tryVar(); }).isSome(); })
+        .map([](auto& x) { return x; })
```
quickbeam123, line 268: Is this some overlooked leftover?

Inferences/ForwardSubsumptionAndResolution.cpp
```diff
@@ -278,14 +278,12 @@ bool ForwardSubsumptionAndResolution::perform(Clause *cl, Clause *&replacement,
       }
     }
 
-    tc_fs.stop();
```
quickbeam123, line 281: What's the reason for deleting this?

JakobR:
I really like the new output, but I have a question about this change. Forward subsumption resolution (FSR) is now counted as part of forward subsumption (FS):
```
  │    │    │    ├──[36%] forward subsumption (total: 393 ms, avg: 40 μs, cnt: 9656)
  │    │    │    │    └──[39%] forward subsumption resolution (total: 154 ms, avg: 19 μs, cnt: 7864)
```
Was that changed intentionally? They were separate before.

(I don't have a strong opinion either way. On the one hand they're different rules; but on the other hand FSR re-uses work from FS. And FSR requires FS to be enabled.)

quickbeam123:
Right, we had 

```
TimeCounter tc_fs(TC_FORWARD_SUBSUMPTION);
...
tc_fs.stop();

```

there before.

@joe-hauns , do we have a way to do something similar with your solution?

(I also don't mind for this particular case, but am curious about the possibility.)

PR #400, opened 2022-07-25
remove `InverseLookup`, not useful

MichaelRawson:
`InverseLookup` is only used to find the location of a particular literal within a clause. I could not find a case where this is faster than linear search, particularly as clauses are typically relatively short and linear scan for words is fast on most platforms. Suggestions for test cases where this does have a performance impact welcome.


quickbeam123:
Hi Michael, what kind of tests did you run check there is no impact?

MichaelRawson:
Hi Martin, I'll admit to going a little "on feel" with this one. Performing a hashtable lookup has a very high constant factor compared to a very tight loop. I tried a couple hundred problems with a shell script and couldn't find a reproducible difference, but that's not to say that there isn't one - for there to be a difference we'd need very large clauses, I'd expect hundreds or thousands of literals at least.

selig:
Clauses can get very very long (definitely hundreds of literals, maybe thousands). Perahps not a common case but it is a case that we care about as those are hard problems.

Do we have a good way of searching statistics of solved problems? Ideally here we'd do a run to record max clause length and use this to select a subset of problems to do some performance checks on.

Of course, it's likely that the best solution is to use linear search for short clauses and inverse lookup for long clauses!

MichaelRawson:
Performance testing inconclusive, to my surprise. Let's shelve this for now.

PR #401, opened 2022-07-25
remove hypersuperposition

MichaelRawson:
See #157. `Inferences::HyperSuperposition` does not appear in schedules and nobody seems to think it is used. Please object if you do want/need this.


quickbeam123:
Hmm, I just tested it over the 870 FOF 2020+2021 problems on top of casc mode getting 6 unique solutions problems (casc mode 776, casc+hs-on 746). Maybe we could just test it and incorporate in schedules next years?

MichaelRawson:
OK - it looks a little unloved so maybe it doesn't work for e.g. polymorphic problems, but happy to leave it for now. Thanks for testing!

MichaelRawson:
Reopened at the request of @quickbeam123.

quickbeam123:
I think vampire can live without hypersuperposition. I don't think I ever tested it, so it's potentially dangerous. So unless @easychair remembers what it was and feels some regret, I think it should go.

MichaelRawson:
@easychair - this one's hyper-**superposition** - I think hyperresolution is already gone or never existed. Thanks for confirmation anyway, axing hypersuperposition.

PR #404, opened 2022-08-03
Changing UWA Implementation

ibnyusuf:
This PR changes UWA implementation to use very special variable mechanism instead of old mechanism. Benefits are:

1. UWA is no longer dependant on the shape of the tree
2. Substitution tree code is simplified as all constraint creation now occurs within RobSubstitution and the various handlers
3. Constraints are now part of the substitution object which is their logical location
4. We avoid creating constraints that will immediately be simplified away such as `a != a`

Currently, there are only two options for uwa. It can either be set to `interpreted_only` in which case both terms must be interpreted for a constraint to be created or it can be set to `one_side_interpreted` in which case at least one of the terms must be interpreted.

Other options such as `ground` or `constant` can be added without great difficulty once we have decide on precisely what behaviour we want from them.

The code has been lightly tested, but requires much more testing to be merge-ready.


ibnyusuf:
We fail sanity check at the moment on problem `int_power_0_all_0`. My investigations haven't uncovered a bug, and if we run with the same strategy as currently, but slowness 1 instead of 0.5 we pass the test.

Kernel/TermTransformer.hpp
```diff
@@ -38,14 +38,25 @@ namespace Kernel {
...
   Literal* transform(Literal* lit);
+  TermList transform(TermList ts);
 protected:
   virtual TermList transformSubterm(TermList trm) = 0;
   Term* transformSpecial(Term* specialTerm);
-  TermList transform(TermList ts);
+  // currently this can be used to transform the
+  // first-order / "green" subterms of an applicative term.
+  // TODO update to allow transformation of prefix subterms
+  // TODO try and code share with transform(Term*)
```
joe-hauns, line 53: I think this should definitely be done before merging. It looks like a big bunch of complicatied code that needs maintaining and can diverge otherwise.

Lib/MaybeBool.hpp
```diff
@@ -37,6 +37,7 @@ class MaybeBool
   MaybeBool(Value val) : _value(val) {}
 
   bool known() const { return _value!=UNKNOWN; }
+  bool maybe() const { return _value==UNKNOWN; }
```
joe-hauns, line 40: Should we maybe rename that to `unknown`? In order to be consistent with the names of the enum variants.

UnitTests/tUnificationWithAbstraction.cpp
```diff
@@ -25,223 +25,290 @@
...
-  return TermList(Term::create(env.signature->addIntegerConstant(n,false),0,0));
-}
-TermList var(unsigned i)
+Clause* unit(Literal* lit)
 {
-  return TermList(i,false);
+  static Inference testInf = Kernel::NonspecificInference0(UnitInputType::ASSUMPTION, InferenceRule::INPUT); 
+  Clause * cl = new(1) Clause(1,testInf);
+  (* cl)[0] = lit;
+  return cl;
```
joe-hauns, line 44: This can be replaced by `return clause({ lit });`, from `SyntaxSugar.hpp`

UnitTests/tUnificationWithAbstraction.cpp
```diff
@@ -25,223 +25,290 @@
...
+    cout << "and constraints: "<< endl;
+    qr.substitution->numberOfConstraints();
+    auto constraints = qr.substitution->getConstraints();
+    while(constraints.hasNext()){
+      Literal* constraint = constraints.next();
+      cout << "> " << constraint->toString() << endl;
+    }
+  }
+  cout << endl;
 }
```
joe-hauns, line 83: I'm not 100% happy with these unit tests, as they require one to check the stdout of the test case, hence their correctness is not checked by our CI. 
It would be nice to replace the function `reportTermMatches` by `checkTermMatches`, and add an argument `Stack<TermQueryResult> expected`, that is being checked against the iterator.

Anyways this unit test is better than what we have so I'm very happy about having it as is as well!

joe-hauns (commented):
Okay.. first of all! This does a great deal of simplifying how UWA is being handled, which i reeeeaaally like!!!! well done. 

I added a couple of rather minor comments as I went through the code. Lemme know what you think.

Inferences/EqualityResolution.cpp
```diff
@@ -58,76 +59,45 @@ struct EqualityResolution::IsNegativeEqualityFn
...
-
-    if(use_uwa_handler){
-      UWAMismatchHandler hndlr(constraints);
-      if(!subst.unify(arg0,0,arg1,0,&hndlr)){ 
-        return 0; 
-      }
+       
+    if(_handler){
+      arg0 = _handler->transform(arg0);
+      arg1 = _handler->transform(arg1);
```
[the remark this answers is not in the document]
ibnyusuf: I don't see any comment on this code later?

Lib/MaybeBool.hpp
ibnyusuf: `maybe` makes more sense semantically than `unknown`.  If  `isConstraintTerm(t).maybe()` suggests that `t` could possibly be a constraint term, whilst `isConstraintTerm(t).unknown()` just says that I don't know whether `t` is or is not.

I would suggest changing the names of the enum variants.

Indexing/IndexManager.hpp
```diff
@@ -94,6 +96,7 @@ class IndexManager
   DHMap<IndexType,Entry> _store;
 
   LiteralIndexingStructure* _genLitIndex;
+  CompositeMismatchHandler* _handler;
```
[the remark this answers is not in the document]
ibnyusuf: Agreed, resolved

Kernel/MismatchHandler.hpp
```diff
@@ -18,48 +18,107 @@
...
+
+  // With polymorphism, a term may end up being a constraint term
+  // depending on type substitutions.
+  // Also a term such as f(a,b) : $int may be a constraint term
+  //
+  // Function used as follows:
+  // if returns true, term is ONLY involved in constraints and never unified
+  // if returns false, term is ONLY unified and never involved in constraints
+  // if returns maybe, term is involved in constraints and unified
+  virtual MaybeBool isConstraintTerm(TermList t) = 0;
```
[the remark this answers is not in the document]
ibnyusuf: Have attempted to clarify and expand the comment. Let me know if it suficient.

UnitTests/tUnificationWithAbstraction.cpp
ibnyusuf: That would be good, but a bit of work. We don't want to make the test dependant on the _order_ that the results are returned.

Indexing/IndexManager.cpp
```diff
@@ -33,10 +33,26 @@
...
+
+  // urther handlers can be added here
+  if(uwa || eba){
+    _handler = new CompositeMismatchHandler();
+    
+    if(uwa){
+      _handler->addHandler(new UWAMismatchHandler());
+    }
+    if(eba){
+      _handler->addHandler(new HOMismatchHandler());
```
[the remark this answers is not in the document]
ibnyusuf: Is there any downside to calling the destructors of the handlers from the destructor of `CompositeHandler`? New implementors are not required to do anything.

Kernel/TermTransformer.hpp
ibnyusuf: It is actually a big simplification in that it allowed me to get rid of three other big blobs of code.

Any suggestions on how to code share with `transform(Term*)`. The two are almost identical, but I don't want to merge them into one, in order to avoid slowing down `transform(Term*)`

Kernel/MismatchHandler.hpp
```diff
@@ -18,48 +18,107 @@
...
+class MismatchHandler : public TermTransformer
 {
 public:
+
+  MismatchHandler() : TermTransformer(false) {}
+
   // returns true if the mismatch was handled.
-  virtual bool handle(RobSubstitution* sub, TermList t1, unsigned index1, TermList t2, unsigned index2) = 0;
+  virtual bool handle(TermList t1, unsigned index1, TermList t2, unsigned index2, 
+    UnificationConstraintStack& ucs, BacktrackData& bd, bool recording) = 0;
```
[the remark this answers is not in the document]
ibnyusuf: Precisely. Added a comment to code to explain.

Inferences/EqualityFactoring.cpp
```diff
@@ -115,28 +123,17 @@ struct EqualityFactoring::ResultFn
...
-      }
-    } else {
-      if(!subst.unify(sLHS,0,fLHS,0)) {
-        return 0;
-      }
+    if(_handler){
+      // replacing subterms that could be part of constraints with very special variables
+      // for example f($sum(1, Y)) -> f(#)
+      sLHS = _handler->transform(sLHS);
+      fLHS = _handler->transform(fLHS);
```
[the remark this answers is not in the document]
ibnyusuf: Done

Kernel/MismatchHandler.hpp
```diff
@@ -18,48 +18,107 @@
...
+  // Function used as follows:
+  // if returns true, term is ONLY involved in constraints and never unified
+  // if returns false, term is ONLY unified and never involved in constraints
+  // if returns maybe, term is involved in constraints and unified
+  virtual MaybeBool isConstraintTerm(TermList t) = 0;
+
+  bool areIdentical(Term* t1, Term* t2, unsigned idx1, unsigned idx2);
+
+  virtual Term* get(unsigned var){ NOT_IMPLEMENTED; }
+
```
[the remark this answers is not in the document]
ibnyusuf: We do use `get` in release code, but only via the `compositeHandler` since no class should have direct access to any handler object other than `compositeHandler`.

For unit testing, we take shortcuts and don;t bother creating a `compositeHandler` which is why other handlers need to implement `get`

Kernel/MismatchHandler.cpp
```diff
@@ -45,52 +57,227 @@ bool UWAMismatchHandler::checkUWA(TermList t1, TermList t2)
...
+  if(!tm2) tm2 = t2.term();
+
+  if(isConstraintTerm(TermList(tm1)).isFalse() || isConstraintTerm(TermList(tm2)).isFalse())
+    return false;
+  
+  if(areIdentical(tm1,tm2,index1,index2))
+    return true;
+
+  return introduceConstraint(TermList(tm1),index1,TermList(tm2),index2,ucs,bd,recording);
+}
```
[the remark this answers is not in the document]
ibnyusuf: Modified implementation. Now most of the logic inside the composite handler (at the possible cost of some flexibility).

ibnyusuf:
@joe one concern I have with current PR is that we never remove terms from these maps. I can imagine that on large problems this could be issue. However, iwth current implementation removing terms from the maps is very tricky since we have one map per handler rather than one map per tree. What do you think?

quickbeam123:
Is it correct I consider this PR, for the time being, on-hold-don't-merge?

Indexing/IndexManager.cpp
```diff
@@ -49,6 +65,10 @@ IndexManager::~IndexManager()
   if(_alg) {
     release(GENERATING_SUBST_TREE);
   }
+
+  if(_handler){
+    delete _handler;
+  }
 }
```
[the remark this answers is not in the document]
ibnyusuf: Using a pointer for handler is useful since it allows us to run checks such as `if(_handler && _handler->...)`. Is there a particular downside to explicitly deleting the handler in the destructor?

ibnyusuf:
> Is it correct I consider this PR, for the time being, on-hold-don't-merge?

This PR is ready for some serious performance testing.

quickbeam123:
Hmm, did I just break something (performance-wise) by making the merge with the most recent master? (Seeing a failed induction integration test.) Or was that there already before?

ibnyusuf:
Seems to randomly come and go, see comment above https://github.com/vprover/vampire/pull/404#issuecomment-1204938466 

Bit odd, not sure why this happens. Also always succeeds on my Apple machine.

Shell/Skolem.cpp
```diff
@@ -483,6 +493,7 @@ Formula* Skolem::skolemise (Formula* f)
         reuse_key = name_reuse->key(before);
         successfully_reused = name_reuse->get(reuse_key, sym);
         if (successfully_reused) { // only counts one per the whole quantifier block
+          cout << "SYM " << sym << endl;
```
quickbeam123, line 496: Possibly a leftover debug printout?

Shell/NewCNF.cpp
```diff
@@ -1244,43 +1245,46 @@ void NewCNF::process(QuantifiedFormula* g, Occurrences &occurrences)
...
+  CALL("NewCNF::processBoolterm");
 
   if (ts.isVar()) {
     processBoolVar(POSITIVE, ts.var(), occurrences);
     return;
   }
 
   Term* term = ts.term();
-  ASS_REP(term->isSpecial(), term->toString());
+  //ASS_REP(term->isSpecial(), term->toString());
```
quickbeam123, line 1256: So, processBoolterm may now be called on non-bool  terms (which don't need any processing)?

ibnyusuf: Not quite, `processBoolterm` can be called on a term of type `$o` that is not a special.

Shell/Options.hpp
```diff
@@ -2403,6 +2404,7 @@ bool _hard;
   bool prioritiseClausesProducedByLongReduction() const { return _priortyToLongReducts.actualValue; }
   int maxXXNarrows() const { return _maximumXXNarrows.actualValue; }
   FunctionExtensionality functionExtensionality() const { return _functionExtensionality.actualValue; }
+  void setFE(FunctionExtensionality value){ _functionExtensionality.actualValue = value; }
```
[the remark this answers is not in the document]
ibnyusuf: Can't see any use. Removed

Shell/NewCNF.cpp
ibnyusuf: Only called on Bool terms, but in some cases, the term may not be a special. If this is case, we are guaranteed that it will be a constant or variable. 

As an alternative, we could mark a problem as higher-order as soon as we read the `thf` tag in parsing which would mean that NewCNF would never be called on a quasi-THF problem

ibnyusuf:
@joe-hauns would liek to investigate some alternative implementations that do not involve very special variables. Whilst he does that, this PR is on hold.

MichaelRawson:
@ibnyusuf @joe-hauns - I'm doing some post-CADE janitorial work on the open PRs. Can I get a quick status update here? Looks to me like this one might be superseded by new work in the HOL/UWA/theory space.

joe-hauns:
Yessss I think we abandoned this one. There's a different one called `joe-uwa-refactor` in my PR pipeline. Martin's currently running performance checks with it.

MichaelRawson:
Cool, thanks! I'll close for now but leave the branch alive, we can reopen if @ibnyusuf wants it.

PR #405, opened 2022-08-04
fix evaluation option command line parsing

joe-hauns:
Fixes the evaluation  command line option. 
Command line parsing for this option was broken up to now. That is, whenever options were mapped to evaluation modes as follows:
- `"simple" -> EvaluationMode::OFF`
- `"force" -> EvaluationMode::SIMPLE`
- `"cautious" -> EvaluationMode::POLYNOMIAL_FORCE`
- (`off` would lead to a crash I think)

This PR fixes this, and adjusts the portfolios accordingly. 
The default value for this option was not affected so all our results should be stable (*blessssss*).

Is there any other things than the porfolio and default options that I might have forgotten, that might have been affected by this?

### DIFF
--- a/CASC/PortfolioMode.cpp
+++ b/CASC/PortfolioMode.cpp
@@ -243,7 +243,7 @@ void PortfolioMode::getExtraSchedules(Property& prop, Schedule& old, Schedule& e
    if(prop.props() & (Property::PR_HAS_INTEGERS | Property::PR_HAS_RATS | Property::PR_HAS_REALS)){
 
     extra_opts.push("hsm=on");             // Sets a sensible set of Joe's arithmetic rules (TACAS-21) 
-    extra_opts.push("gve=force:asg=force:canc=force:ev=force:pum=on"); // More drastic set of rules
+    extra_opts.push("gve=force:asg=force:canc=force:ev=simple:pum=on"); // More drastic set of rules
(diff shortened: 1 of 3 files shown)


MichaelRawson (approved):
Great catch! I assume that you've found all the instances in schedule strings. Double-check that for me then good to go. :-)

quickbeam123:
Guys, this doesn't look like a fix to me (yet). There is no "OFF" in the EvaluationMode enum!

joe-hauns:
ohhh .. sorry .. this variant only exists in the branch i found the bug on (`inequality-resolution`), and i didn't check whether it actually exists in master. 
Which means everythign was correct in master, but only broken on my branch. So all the portfolios should stay as they are, and i only need to add the string `"off"` on the branch. 
Or we add the string and the enum variant on master, as well. in any case this means that the portfolios should stay as they are. 
Sorry for that.

quickbeam123:
No worries! 

It still does look suspicious not to have the OFF value. But what's the conclusion again? We close this PR and your branch, when ready, will come with OFF and all other niceties later?

joe-hauns:
Yeah I think that's best to do. 
It's indeed a bit odd that we don't have an off for this, but we always had, because simplificaiton was unconditionally always applied back before I added new opioins for it.
I think we probably want to have an OFF option at some point, but if there's no need for fixing anything we can safe work with not being forced to test the new option now ;)

PR #406, opened 2022-08-11
Fix sine utils performance

joe-hauns:
Fixes a subtle but massive performance bug in `Shell/SineUtils` for formulas.
The issue is that for "special", i.e. formula terms `void SineSymbolExtractor::addSymIds(Term* term, Stack<SymId>& ids)`, calls itself recursively for every subterm of a term, which results in factorial runtime. 
This PR fixes this behaviour, by only calling it recursively for every immediate subterm, yielding the same result with linear runtime. 

Additionally the symbols are being collected into a `DHSet`, instead of a `Stack` now, in order to save memory. This was changed as the bug first showed up as an out-of-memory bug, which was resolved by this change.

A toy example that shows the performance boost is running `$vampire --input_syntax smtlib2 sine-utils-performance.smt2 -s2a on -t 60 -newcnf on` on this benchmark:
```

(set-info :smt-lib-version 2.6)
(set-logic UFLIA)
(assert 
(let (

  (?v_0 true) 
  (?v_1 true) 
  (?v_2 true) 
  (?v_3 true) 
  (?v_4 true) 
  (?v_5 true) 
  (?v_6 true) 
  (?v_7 true) 
  (?v_8 true) 
  (?v_9 true) 
  (?v_10 true) 
  (?v_11 true) 
  (?v_12 true) 
  (?v_13 true) 
  (?v_14 true) 
  (?v_15 true) 
  (?v_16 true) 
  (?v_17 true) 
  (?v_18 true) 
  (?v_19 true) 
  (?v_20 true) 
  (?v_21 true) 
  (?v_22 true) 
  (?v_23 true) 

  ) true))
(check-sat)
(exit)
```

On master (in debug mode) we get:
```
...
% Time elapsed: 31.170 s
...
```
(Note that adding another let variable in the benchmark ~doubles runtime here.)
On the branch we now get:
```
...
% Time elapsed: 0.010 s
...
```

Before merging we should consider how this interferes with current schedules. I came across this bug, because running vvampire in portfolio gave an an out of memory for one of the strategies in the schedule. This means there were some strategies in there that were never correctly executed. That is they might have exited early due to out of memory, instead of taking their whole runtime slice, which could quite change the performance of the overall schedules i think.

### DIFF
--- a/Shell/SineUtils.cpp
+++ b/Shell/SineUtils.cpp
@@ -59,7 +59,7 @@ SineSymbolExtractor::SymId SineSymbolExtractor::getSymIdBound()
          max(env.signature->functions()*3, env.signature->typeCons()*3));
 }
 
-void SineSymbolExtractor::addSymIds(Term* term, Stack<SymId>& ids)
+void SineSymbolExtractor::addSymIds(Term* term, DHSet<SymId>& ids)
 {
   CALL("SineSymbolExtractor::addSymIds");
 
@@ -94,26 +94,27 @@ void SineSymbolExtractor::addSymIds(Term* term, Stack<SymId>& ids)
       }
     } else {
       //all sorts should be shared
-      ids.push(term->functor() * 3 + 1);
+      ids.insert(term->functor() * 3 + 1);
     }
-    NonVariableIterator nvi(term);
-    while (nvi.hasNext()) {
-      addSymIds(nvi.next().term(), ids);
+
+    for (auto t : concatIters(typeArgIter(term), termArgIter(term))) {
+      if (t.isTerm())
+        addSymIds(t.term(), ids);
     }
   } else {
     if(term->isSort()){
-      ids.push(term->functor() * 3 + 2);
+      ids.insert(term->functor() * 3 + 2);
     } else {
-      ids.push(term->functor() * 3 + 1);
+      ids.insert(term->functor() * 3 + 1);
     }
 
     NonVariableIterator nvi(term);
     while (nvi.hasNext()) {
       Term* t = nvi.next().term();
       if(t->isSort()){
-        ids.push(t->functor() * 3 + 2);
+        ids.insert(t->functor() * 3 + 2);
       } else {
-        ids.push(t->functor() * 3 + 1);
+        ids.insert(t->functor() * 3 + 1);
       }
     }
   }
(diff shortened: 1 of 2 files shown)


Shell/SineUtils.cpp
```diff
@@ -252,7 +253,11 @@ SineSymbolExtractor::SymIdIterator SineSymbolExtractor::extractSymIds(Unit* u)
     FormulaUnit* fu=static_cast<FormulaUnit*>(u);
     extractFormulaSymbols(fu->formula(),itms);
   }
-  return pvi( getUniquePersistentIterator(Stack<SymId>::Iterator(itms)) );
+  Stack<SymId> ids(itms.size());
+  DHSet<SymId>::Iterator iter(itms);
+  ids.loadFromIterator(iter);
+  std::sort(ids.begin(), ids.end()); // <- make order deterministic
```
MichaelRawson, line 259: I'm not super-familiar with the SInE code - does changing the order we return symbols in break anything? Maybe @quickbeam123 knows.

MichaelRawson (approved):
Looks great to me, that's one hell of a bug. I added a note about determinism but I don't expect it's a problem.

Shell/SineUtils.cpp
selig: Theoretically, it shouldn't. The definition of relevant symbols in SInE is independent on 'order' ... I'm guessing non-deterministic traversal here would only lead to a non-deterministic performance profile. But @quickbeam123 has looked at SInE more recently than me... :)

selig:
Would be cool to quantify the performance impact. For large problems I expect this to be very large. For most problems, it probably won't register. I think you'll find the danger problems in the TPTP HWV category (there are problems there that take a long time to process before proving even starts).

quickbeam123:
I don't think special terms are really an issue in TPTP. Guessing this would mostly be hurting us on SMTLIB...

Shell/SineUtils.cpp
quickbeam123: `SymId` is simply an unsigned. It's nice to care about reproducibility, but unless we hash on pointers, we are deterministic (if a bit chaotic). Unless I am overlooking something, I wouldn't even bother sorting here.

quickbeam123:
FOF_master6344_casc 
775 / 869
FOF_fsup6346_casc 
775 / 869
where FOF means the union of CASC FOF problems from 2020 and 2021.

easychair:
We rarely use SINE on CASC problems. Great that this was spotted but
it is unlikely to give any improvement on CASC.

A

PR #407, opened 2022-08-13
Finally, let's have ActiveClauseContainer store its clauses explicitly

quickbeam123:
Vampire has been using a trick in which the ActiveClauseContainer does not explicitly store its clauses and relies on the implicitly present indexes when active clause enumeration is needed (mainly in `SaturationAlgorithm::collectSaturatedSet`). However, the index we have been using for this was `GENERATING_SUBST_TREE` a.k.a. BINARY_RESOLUTION_INDEX).

Ahmed recently noticed it's suboptimal to index equalities into this index and but with the corresponding fix I didn't realize we would lose pure equational clauses from collectSaturatedSet! Now, reverting the commit was the first reaction, however, the original solution was ugly anyway, because with multi-literal-selections collectSaturatedSet would list a clause more than once. Long story short, with this PR:
-  ActiveClauseContainer maintains its content explicitly in a DHSet
- this set is used by collectSaturatedSet and few other places (some of which could be streamlined) and the above bug / deficiency are gone
- we no longer reserve GENERATING_SUBST_TREE to be always present on behalf of SaturationAlgorithm, so, with `-br off` this index will not even be requested


MichaelRawson (approved):
This is nicer, shorter and matches my intuition about `ActiveClauseContainer` much better. Sometimes I want "gimme active clauses" and now I can do it without remembering the index hack. Good job!

MichaelRawson:
Although what's with the branch name?

quickbeam123:
Encompassment demodulation (from Andre and Kostya) is coming next... Well, provided you can have two PRs on a single branch! ;)

ibnyusuf:
Have a sense of deja vu. Some years ago, I noticed that we insert terms into demodulation subterm index once per a literal, rather than once per a clause as required. We carried out the modification to master only to realise that because order of literals doesn't remain stable this wasn't a safe thign to do...

Anyway, this PR looks good. Haven't viewed it closely yet.

Indexing/IndexManager.cpp
```diff
@@ -91,9 +56,6 @@ void IndexManager::release(IndexType t)
 
   e.refCnt--;
   if(e.refCnt==0) {
-    if(t==GENERATING_SUBST_TREE) {
```
ibnyusuf, line 94: Can we take this opportunity to chaneg the name of the index? Perhaps to `BinaryResolutionTree`?

selig:
This makes much more sense in modern Vampire. The previous approach was very resolution-centric.

Much needed improvement to code comprehension as well.

Saturation/SaturationAlgorithm.cpp
```diff
@@ -1308,14 +1307,12 @@ UnitList* SaturationAlgorithm::collectSaturatedSet()
...
-  LiteralIndexingStructure* gis=getIndexManager()->getGeneratingLiteralIndexingStructure();
-
   UnitList* res = 0;
-  SLQueryResultIterator qrit = gis->getAll();
-  while (qrit.hasNext()) {
-    SLQueryResult qres = qrit.next();
-    UnitList::push(qres.clause, res);
-    qres.clause->incRefCnt();
+  ClauseIterator it = _active->clauses();
+  while (it.hasNext()) {
```
selig, line 1312: We're iterating over a set. This makes me worried about non-determinism. If we have two proof searches that end in the same saturated set do these get printed in exactly the same way? If not that's a potential sub-optimal user experience.

MichaelRawson: Yes, this won't be guaranteed. We need to hash the _content_ of a `Clause` somehow, probably via `DerefPtrHash`.

Indexing/IndexManager.cpp
quickbeam123: A good idea! I will add some name changes to the PR...

Saturation/SaturationAlgorithm.cpp
quickbeam123: Well spotted! I think we could easily hash on `clause->number()` as these are as unique as the pointers themselves. @MichaelRawson , you are the primary/secondary hash master now, what should I override/overload and where, to get the clauses' hash here be derived from `number()`?

PR #409, opened 2022-08-16
Encompassment demodulation check (as an option)

quickbeam123:
Implements the encompassment demodulation checks according to 

André Duarte, Konstantin Korovin:
[Ground Joinability and Connectedness in the Superposition Calculus.](https://link.springer.com/chapter/10.1007/978-3-031-10769-6_11) IJCAR 2022: 169-187

After some experiments with casc mode (2minutes 8 cores) on the 869 problems from the union of FOF 2020+2021, I got a greedy cover with
FOF_encompassment6348_casc_fwd-de-on.pkl contributes 781 total 781
FOF_encompassment6351_casc_bwd-de-on.pkl contributes 7 total 777
FOF_encompassment6349_casc_both-de-on.pkl contributes 3 total 779
FOF_encompassment6346_casc.pkl contributes 2 total 776
Total 793
So an overall improvement (of a portfolio!) and quite a few new problems solved as well. (Expected to have impact on UEQ as well.)

### DIFF
--- a/Inferences/BackwardDemodulation.cpp
+++ b/Inferences/BackwardDemodulation.cpp
@@ -242,7 +242,7 @@ private:
   bool _encompassing;
 
   BackwardDemodulation& _parent;
-  Ordering& _ordering;  
+  Ordering& _ordering;
 };
 
 
--- a/Inferences/BackwardDemodulation.cpp
+++ b/Inferences/BackwardDemodulation.cpp
@@ -96,6 +96,7 @@ struct BackwardDemodulation::ResultFn
     _eqLit=(*_cl)[0];
     _eqSort = SortHelper::getEqualityArgumentSort(_eqLit);
     _removed=SmartPtr<ClauseSet>(new ClauseSet());
+    _encompassing = parent.getOptions().demodulationEncompassment();
   }
   /**
    * Return pair of clauses. First clause is being replaced,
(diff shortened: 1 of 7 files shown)


Inferences/BackwardDemodulation.cpp
```diff
@@ -168,32 +169,41 @@ struct BackwardDemodulation::ResultFn
...
-        Literal* eqLitS=Literal::createEquality(true, lhsS, rhsS, eqSort);
-        bool isMax=true;
-        Clause::Iterator cit(*qr.clause);
-        while(cit.hasNext()) {
-          Literal* lit2=cit.next();
-          if(qr.literal==lit2) {
-            continue;
+        if (_encompassing) {
+          if (qr.substitution->isRenamingOn(lhs,false /* we talk of a non-result, i.e., a query term */)) {
+            // under _encompassing, we know there are no other literals in qr.clause
```
ibnyusuf, line 180: This doesn't capture the full generality of the rule as given in the paper. The substitution can be a renaming so long as we are rewriting a proper subterm (or the smaller side of the equality). Perhaps we don't want / need the full generality?

Inferences/ForwardDemodulation.cpp
```diff
@@ -169,29 +175,48 @@ bool ForwardDemodulationImpl<combinatorySupSupport>::perform(Clause* cl, Clause*
...
           if(tord!=Ordering::LESS && tord!=Ordering::LESS_EQ) {
-            Literal* eqLitS=qr.substitution->applyToBoundResult(qr.literal);
-            bool isMax=true;
-            for(unsigned li2=0;li2<cLen;li2++) {
-              if(li==li2) {
-          continue;
+            if (_encompassing) {
+              // last chance, if the matcher is not a renaming
+              if (qr.substitution->isRenamingOn(qr.term,true /* we talk of result term */)) {
+                continue; // under _encompassing, we know there are no other literals in cl
```
ibnyusuf, line 197: Implementation doesn't seem to capture full generality of rule. Substitution can be a renaming on the lareger side, so long as `r\theta < t` (condition (v) in the paper)

ibnyusuf:
Encompassment demodulation seems to be a strict generalisation of demodulation. Is there any point in keeping both? Could we not just keep encompassment demodulation as it maintains completeness whilst somplifying more clauses?

ibnyusuf (approved):

quickbeam123:
Thank you, Ahmed, for the careful check!

You are right, encompassment demodulation is a strict generalization and we might want to use it exclusively. This is not done in this PR (yet), however, for the usual reason of "let's not disturb the old and tested by quickly switching to something new". (Old schedules could suffer, etc.)

Inferences/BackwardDemodulation.cpp
quickbeam123: The same reason is in Forward?

MichaelRawson:
> (At some point, I made simultaneous superposition -- which was not in vampire in the olden days -- the default and we survived the switch ;))

Not to get off-topic, but in a similar vein to hypersuperposition: is there ever a case when we don't want simultaneous superposition? It's not in schedules so my usual heuristic is "no"...

ibnyusuf:
Always adding new stuff as an option in order to avoid disturbing the current status quo sounds lika recipe for code bloat.

In the long term, we need to a method to evaluate whether a new feature is sufficiently powerful to justify dropping some previous version of it. This will allow us to keep the codebase streamlined and the number of options managebale.

Inferences/BackwardDemodulation.cpp
```diff
@@ -229,8 +239,10 @@ struct BackwardDemodulation::ResultFn
   Clause* _cl;
   SmartPtr<ClauseSet> _removed;
 
+  bool _encompassing;
+
   BackwardDemodulation& _parent;
-  Ordering& _ordering;
```
MichaelRawson, line 233: Extraneous whitespace/diff.

MichaelRawson (approved):
I won't pretend to know much about this - Ahmed is better qualified and very thorough! Looks plausible to me.

MichaelRawson:
> Always adding new stuff as an option in order to avoid disturbing the current status quo sounds lika recipe for code bloat.

Broadly agree, although I'm a little less stern as I tend to remove such stuff eventually - @quickbeam123, are there any problems for which this hurts?

ibnyusuf:
Didn't intend to sound harsh. I can see the benefits of not changing whats working. However, at some point we do have to evaluate between similar options (not necessarily immediately).

Inferences/ForwardDemodulation.cpp
ibnyusuf: Agreed

Inferences/BackwardDemodulation.cpp
ibnyusuf: Your comment there clarified. Think it is fine now.

easychair:
I'd suggest that everybody keep experimenting with new options.

We will use this creature to weed out those we don't need.

Our new slogan:

  WE DON'T NEED ALL OPTIONS - ONLY THOSE OPTIONS A TRUE VAMPIRE NEEDS!

Cheers,
Andrei

Inferences/BackwardDemodulation.cpp
quickbeam123: Should be OK now.

quickbeam123:
> Not to get off-topic, but in a similar vein to hypersuperposition...
My guess is that hypersuperposition is not in the schedules, because nobody tried to put it there... So I wanted to first give it a chance and only then possibly discard it.

And yes, I agree about options in general. Maybe let's continue our discussion on slack about how we identify options that could/should get eliminated.

PR #410, opened 2022-08-17
Adding a coefficient/option for tweaking LRS' aggressiveness

quickbeam123:
Sometimes LRS throws away clauses and finishes too early. Some other times, maybe interesting problems could be solved by more aggressive discarding? This is a trivial modification of LRS' estimate (multiply it by a float, by default of value 1.0) which allows us to tweak this. With coef -> \infty, LRS -> otter. 

Some greedy cover results (on my now usual 870 FOF 2020+2021 casc problems) with a casc mode (2minutes 8cores):
FOF_master6348_casc.pkl contributes 776 total 776
FOF_6348lecc_casc_lecc1.2.pkl contributes 5 total 776
FOF_6348lecc_casc_lecc0.8.pkl contributes 2 total 776
FOF_6348lecc_casc_lecc0.9.pkl contributes 2 total 774
FOF_6348lecc_casc_lecc1.1.pkl contributes 1 total 776
Total 786
So some potential for strategy building is here...

### DIFF
--- a/Shell/Options.cpp
+++ b/Shell/Options.cpp
@@ -1093,7 +1093,7 @@ void Options::init()
     _simulatedTimeLimit.tag(OptionTag::LRS);
 
     _lrsEstimateCorrectionCoef = FloatOptionValue("lrs_estimate_correction_coef","lecc",1.0);
-    _lrsEstimateCorrectionCoef.description = "Make lrs more (<1.0) or less (>1.0) agressive by multiplying this coef its estimate of how many clauses are still reachable.";
+    _lrsEstimateCorrectionCoef.description = "Make lrs more (<1.0) or less (>1.0) agressive by multiplying by this coef its estimate of how many clauses are still reachable.";
     _lookup.insert(&_lrsEstimateCorrectionCoef);
     _lrsEstimateCorrectionCoef.tag(OptionTag::SATURATION);
     _lrsEstimateCorrectionCoef.addConstraint(greaterThan(0.0f));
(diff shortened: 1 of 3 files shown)


Shell/Options.cpp
```diff
@@ -1092,6 +1092,13 @@ void Options::init()
     _lookup.insert(&_simulatedTimeLimit);
     _simulatedTimeLimit.tag(OptionTag::LRS);
 
+    _lrsEstimateCorrectionCoef = FloatOptionValue("lrs_estimate_correction_coef","lecc",1.0);
+    _lrsEstimateCorrectionCoef.description = "Make lrs more (<1.0) or less (>1.0) agressive by multiplying this coef its estimate of how many clauses are still reachable.";
```
MichaelRawson, line 1096: "by multiplying this coef _by_", right?

MichaelRawson (approved):
Great! Nice feature, minimal code change. :-) Some typo in the option description but good to go from me.

easychair:
We already have an option for this (simulated_time_limit) - what is the
reason for an extra option doing the same?

A

Shell/Options.cpp
quickbeam123: Yes, one more "by" in needed. Thanks!

quickbeam123:
I don't think they are quite the same, although similar for sure. 

Let's say we have `--time_limit 10` and `--simulated_time_limit 12`. If the first estimate is done close to the beginning of the run, the result with be comparable to having `-lecc 1.2`, but when there is another estimate at second 8, the simulated_time_limit run will think it has 4 seconds left while only 2 are actually remainig, which corresponds to `-lecc 2.0` now.

In some sense, `lecc` is uniform along the time axis while `stl` gradually interpolates lrs to otter within a single run.

PR #411, opened 2022-08-18
Shuffling

quickbeam123:
The randomization options described in the Vampire Getting Noisy paper. Possible application in `getExtraSchedules` as already using in 2021. Also, the snake schedules (to come in a separate PR rely on this).

This PR also replaces, in `Random.hpp`, the randomizing routines based on C standard library's `random()` by the C++11 solution.


Makefile
```diff
@@ -203,7 +203,8 @@ VK_OBJ= Kernel/Clause.o\
         Kernel/Matcher.o\
         Kernel/MaximalLiteralSelector.o\
         Kernel/SpassLiteralSelector.o\
-        Kernel/ELiteralSelector.o\
+  Kernel/ELiteralSelector.o\
```
MichaelRawson, line 206: Indentation is off. GNU make might be OK with this, others would be in their rights to whinge. If I recall POSIX correctly...ignore if I misremembered, you are the main make-user!

SAT/SATSolver.hpp
```diff
@@ -134,6 +134,8 @@ class SATSolver {
   
   virtual void suggestPolarity(unsigned var, unsigned pol) = 0;
 
+  virtual void requestRandomization(unsigned seed) {}
```
MichaelRawson, line 137: Maybe a pure-virtual here isn't so bad? `suggestPolarity` is pure virtual, for example.

SAT/Z3Interfacing.cpp
```diff
@@ -852,10 +852,15 @@ struct ToZ3Expr
       symb = env.signature->getPredicate(trm->functor());
       range_sort = AtomicSort::boolSort();
       // check for equality
-      if( trm->functor()==0 || symb->equalityProxy()){
+      if( trm->functor()==0 || (symb->equalityProxy())){
```
MichaelRawson, line 855: Unnecessary parenthesis.

SAT/Z3Interfacing.cpp
```diff
@@ -852,10 +852,15 @@ struct ToZ3Expr
...
         ASS(trm->numTermArguments()==2);
         // both equality and equality proxy translated as z3 equality
-        return args[0] == args[1];
+        if (symb->wasFlipped()) {
```
MichaelRawson, line 858: I support this feature and I don't see a better way of doing it. But I will add this to the long list of ways to shoot oneself in the foot, Vampirically.

Saturation/AWPassiveClauseContainer.cpp
```diff
@@ -67,6 +68,9 @@ AWPassiveClauseContainer::AWPassiveClauseContainer(bool isOutermost, const Shell
     _ageRatio = 1;
     _weightRatio = 1;
   }
+
+  _randomize = _opt.randomAWR();
```
MichaelRawson, line 72: I would remove the boolean field, `_opt.randomAWR()` will compile to a single load quite happily.

Saturation/AWPassiveClauseContainer.cpp
```diff
@@ -295,7 +299,9 @@ Clause* AWPassiveClauseContainer::popSelected()
   _size--;
 
   Clause* cl;
-  if (byWeight(_balance)) {
+  bool selByWeight = _randomize ? (Random::getInteger(_ageRatio+_weightRatio) < _weightRatio) : byWeight(_balance);
```
MichaelRawson, line 302: A comment would help me here - I expected "fully" (i.e. flip a coin for age or weight) random AWR, but it's actually random-but-respecting-AWR-ratio, I think.

Saturation/PredicateSplitPassiveClauseContainer.cpp
```diff
@@ -44,17 +48,27 @@ PredicateSplitPassiveClauseContainer::PredicateSplitPassiveClauseContainer(bool
     USER_ERROR("Queue " + name + ": The number of cutoffs needs to match the number of queues, but " + Int::toString(_cutoffs.size()) + " != " + Int::toString(_queues.size()));
   }
 
+  if (_randomize) {
```
MichaelRawson, line 51: I trust you know what you're doing here, I've no idea what `PredicateSplitPassiveClauseContainer` does!

Saturation/SaturationAlgorithm.cpp
```diff
@@ -1189,44 +1208,52 @@ void SaturationAlgorithm::activate(Clause* cl)
   if (!cl->numSelected()) {
     TimeCounter tc(TC_LITERAL_SELECTION);
 
+    if (env.options->randomTraversals()) {
+      TimeCounter tc(TC_SHUFFLING);
+
+      Shuffling::shuffle(cl);
```
MichaelRawson, line 1214: Doesn't this mean that a clause could be shuffled a few times? At least when new, then when activated? Possibly many times with AVATAR. This might be desirable, just checking...

Saturation/SaturationAlgorithm.cpp
```diff
@@ -1387,6 +1414,8 @@ MainLoopResult SaturationAlgorithm::runImpl()
       if (env.timeLimitReached()) {
         throw TimeLimitExceededException();
       }
+
+      env.statistics->activations = l;
```
MichaelRawson, line 1418: Huh? `+= 1`?

Shell/Options.cpp
```diff
@@ -92,6 +92,11 @@ void Options::init()
   _instructionLimit = UnsignedOptionValue("instruction_limit","i",0);
   _instructionLimit.description="Limit the number (in millions) of executed instructions (excluding the kernel ones).";
   _lookup.insert(&_instructionLimit);
+
+  _parsingDoesNotCount = BoolOptionValue("parsing_does_not_count","",false);
```
MichaelRawson, line 96: Cute, I could use this!

Shell/Options.cpp
```diff
@@ -981,13 +989,13 @@ void Options::init()
...
 
-    _sineLevelSplitQueueCutoffs = StringOptionValue("sine_level_split_queue_cutoffs", "slsqc", "0,1");
+    _sineLevelSplitQueueCutoffs = StringOptionValue("sine_level_split_queue_cutoffs", "slsqc", "0");
     _sineLevelSplitQueueCutoffs.description = "The cutoff-values for the sine-level-split-queues (the cutoff value for the last queue is omitted, since it has to be infinity).";
     _lookup.insert(&_sineLevelSplitQueueCutoffs);
     _sineLevelSplitQueueCutoffs.onlyUsefulWith(_useSineLevelSplitQueues.is(equal(true)));
     _sineLevelSplitQueueCutoffs.tag(OptionTag::SATURATION);
 
-    _sineLevelSplitQueueRatios = StringOptionValue("sine_level_split_queue_ratios", "slsqr", "1,2,3");
+    _sineLevelSplitQueueRatios = StringOptionValue("sine_level_split_queue_ratios", "slsqr", "1,1");
```
MichaelRawson, line 998: I...guess this kind of thing is motivated by something, so I'll leave you to it!

Shell/Preprocess.cpp
```diff
@@ -107,6 +108,17 @@ void Preprocess::preprocess(Problem& prb)
...
   }
 
+  if (_options.shuffleInput()) {
+    TimeCounter tc(TC_SHUFFLING);
+    env.statistics->phase=Statistics::SHUFFLING;    
+
+    if (env.options->showPreprocessing())
+      env.out() << "shuffling1" << std::endl;
+
+    // shuffling at the very beginning
```
MichaelRawson, line 118: Ah, so we do have a shuffle here! So...why also in the various algorithms?

Shell/Shuffling.hpp
```diff
@@ -0,0 +1,148 @@
...
+using namespace Kernel;
+
+/**
+ * Class implementing shuffling-related procedures.
+ * @since 09/06/2021 Prague
+ */
+class Shuffling
+{
+private:
+  typedef Coproduct<Formula*, Literal*, TermList> Shufflable;
```
MichaelRawson, line 39: Joe-code is gaining ground! I'm fine with it, but does this get us something versus e.g. overloads, template specialisation?

Shell/Shuffling.hpp
```diff
@@ -0,0 +1,148 @@
...
+
+  static void shuffle(Unit*);     // shuffle a unit; just dispatches to either shuffle(Clause*) or shuffle(Formula*)
+  static void shuffle(Clause*);   // shuffling a clause; it assumes literals are shared (i.e. nothing "special" in them anyway) so it does not touch those
+
+  static void shuffle(Formula* f) { shuffleIter(Shufflable(f)); }  // shuffling a formula; will try to descend also to the term level, to shuffle around commutative operators and formulas inside terms (if applicable)
+  static void shuffle(Literal* l) { shuffleIter(Shufflable(l)); }
+  static void shuffle(TermList tl) { shuffleIter(Shufflable(tl)); }
+
+  template<typename Arrayish>
+  static void shuffleArray(Arrayish& a, unsigned len) {
```
MichaelRawson, line 59: This is Fisher-Yates, or something like it, right? Might be worth writing down exactly what it is for later Googling.

MichaelRawson (approved):
Great work! I enjoyed the paper so it's nice to see the nitty-gritty. Some comments, but mostly questions rather than changes.

quickbeam123:
Thank you, @MichaelRawson , for the comments! Some decisions are definitely up for discussion. Let me reply one by one...

CASC/PortfolioMode.cpp
```diff
@@ -132,8 +133,13 @@ bool PortfolioMode::searchForProof()
 
     //we normalize now so that we don't have to do it in every child Vampire
     ScopedLet<Statistics::ExecutionPhase> phaseLet(env.statistics->phase,Statistics::NORMALIZATION);
-    Normalisation().normalise(*_prb);
     
+    if (env.options->shuffleInput()) { // instead to "combing things into shape" we shuffle and ruffle them
+      Shuffling().shuffle(*_prb);
```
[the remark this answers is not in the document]
quickbeam123: Well spotted! This was from one of the very first commits and I didn't want to go through the pains of editing it out. However, I plan a separate PR with my other changes to `PortfolioMode` and strategy related things which will override this... So, since this change is harmless for non-shuffling workflows I guess we can simply ignore its effect for now.

FMB/FiniteModelBuilder.cpp
```diff
@@ -1667,6 +1669,9 @@ MainLoopResult FiniteModelBuilder::runImpl()
     // pass clauses and assumption to SAT Solver
     {
       TimeCounter tc(TC_FMB_SAT_SOLVING);
+      if (_opt.randomTraversals()) {
+        Shuffling::shuffleArray(_clausesToBeAdded,_clausesToBeAdded.size());
```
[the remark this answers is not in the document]
quickbeam123: If I remember correctly, I did. However, I was happy to also notice the possibility of using `randomizeForNextAssignment`, which does not hurt performance-wise for one off solving (as it does in the incremental way of using a SAT solver in AVATAR) and really makes things nicely stochastic here in FMB.

InstGen/IGAlgorithm.cpp
```diff
@@ -184,6 +184,12 @@ void IGAlgorithm::init()
     _equalityProxy->apply(_prb);
   }
 
+  if (env.options->randomTraversals()) {
+    TimeCounter tc(TC_SHUFFLING);
+
+    Shuffling::shuffle(_prb);
```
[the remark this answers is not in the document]
quickbeam123: Well, it is already done under `shuffleInput` during preprocessing. The question is whether we want to shuffle the problem just before saturation under (`shuffleInput` off && `randomTraversals ` on)

Kernel/Formula.hpp
```diff
@@ -276,6 +281,11 @@ class BinaryFormula
   /** Return the rhs subformula of this formula */
   Formula* rhs() { return _right; }
 
+  // careful, this really swaps the left and right subformulas
```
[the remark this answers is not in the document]
quickbeam123: Agree, it's not very deep. I think I indeed meant to warn it's destructive. (Not very good for proof documenting where you would parse A&B and print it as B&A at the end. If this becomes a real concern we need a lot of copying and inference objects everywhere we shuffle.)

Kernel/RCClauseStack.hpp
```diff
@@ -137,6 +137,11 @@ class RCClauseStack {
   
 private:
   ClauseStack _s;
+
+public:
+  ClauseStack& naked() {
```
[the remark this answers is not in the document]
quickbeam123: Is it worth the work? 

`RCClauseStack` is just a `ClauseStack` with reference counting on `push`/`pop`. It even brazenly allows `pushWithoutInc`/`popWithoutDec`. We are not using clause reference counting very well at the moment anywhere anyway. (I think a normal Stack with smart pointers to clauses would be right design to replace `RCClauseStack`). 

Given all this, I think I was even a bit enjoying breaking the interface of this already quite broken data structure... (Although, let's by all means discuss more if you think otherwise.)

Kernel/RndLiteralSelector.hpp
```diff
@@ -0,0 +1,53 @@
...
+#include "Ordering.hpp"
+
+#include "LiteralSelector.hpp"
+
+namespace Kernel {
+
+/**
+ * Class SpassLiteralSelector implements literal
+ * selectors as understood from the code of SPASS 3.7.
+ */
```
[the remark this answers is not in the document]
quickbeam123: Oh, wrong comment (copy-paste error). Adding the one that should have been there.

And yes, we have `SpassLiteralSelector` just next to this class and it would make sense to add the proper explanation there, but I have honestly forgotten since 2016...

Kernel/Term.cpp
```diff
@@ -33,6 +33,15 @@ const unsigned Term::SF_FORMULA;
 const unsigned Term::SF_LAMBDA;
 const unsigned Term::SPECIAL_FUNCTOR_LOWER_BOUND;
 
+void Term::setId(unsigned id)
+{
+  CALL("Term::setId");
+  if (env.options->randomTraversals()) {
+    id += Random::getInteger(1 << 12) << 20; // the twelve most significant bits are randomized
```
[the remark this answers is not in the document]
quickbeam123: That's the idea! 

(Although, on a different branch, I was forced and able to get rid of `Term::id` completely so not sure how long this will last.)

Kernel/Term.hpp
```diff
@@ -92,7 +92,8 @@ bool operator<(const TermList& lhs, const TermList& rhs);
 class TermList {
 public:
   CLASS_NAME(TermList)
-  static const unsigned SPEC_UPPER_BOUND = 10000000;
+  // divide by 4 because of the tag, by 2 to split the space evenly
+  static const unsigned SPEC_UPPER_BOUND = (UINT_MAX / 4) / 2;
```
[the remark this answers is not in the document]
quickbeam123: It's fixed in master as well. The diff shows what the edit was back then not how it compares to master now (I was also confused). 

In any case, Github boldly claims it can do the merge. And it's hard to imagine that it could mess up given that both master and this branch insist on the new version...

SAT/Z3Interfacing.cpp
quickbeam123: Yes, this can go unnoticed for ages... Unless you sample strategies randomly (sas=z3,ep!=off,rp=on is sufficient) and then prove something really hard! :)

Saturation/AWPassiveClauseContainer.cpp
quickbeam123: Indeed! Comment added..

Saturation/PredicateSplitPassiveClauseContainer.cpp
quickbeam123: It's a meta-container with its own ratio. The change is in that it may hold more than two containers, so that ratio is a tuple of things numbers not just a pair

Saturation/SaturationAlgorithm.cpp
quickbeam123: Yes, the current code is definitely slightly over-shuffled. For instance, also during preprocessing we shuffle more than once (but that could have subtle implications for skolem names, etc...)

Saturation/SaturationAlgorithm.cpp
quickbeam123: You mean MOV vs INC? Is it a big difference?

Shell/Options.cpp
quickbeam123: Just note it only works with instruction limit at the moment.

Shell/Options.cpp
quickbeam123: Indeed. Basically, we didn't want to write code that samples multi-ratios.

Shell/Preprocess.cpp
quickbeam123: Shuffling the input is what Geoff (allegedly) does to our inputs in the competition. 

But it's also interesting to study a fixed input and noise in the algorithms. 

Didn't have a careful look at the difference between the two yet, but the two options create some conceptual separation.

Shell/Shuffling.hpp
quickbeam123: If I remember correctly, I interpreted this is a modern (type safe-ish and better documenting) way of doing tagged union. For that use case, it was a very pleasant experience!

Shell/Shuffling.hpp
quickbeam123: Didn't know it was called that way. (Although I've heard of Knuth shuffle and of various ways of making mistakes in trying to draw a permutation uniformly at random.)

Kernel/RCClauseStack.hpp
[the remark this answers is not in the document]
MichaelRawson: Yeah, looking more broadly at `RCClauseStack` it's already terrifying. Let's put this on the to-kill pile and move on, you're right.

Saturation/SaturationAlgorithm.cpp
MichaelRawson: Oh, now I see. I read that as `= 1`, not `= l`, but there is an `l` in scope. Still - if we throw an exception for e.g. timeout, do we not want to report the number of activations? But no big deal, I've got it now.

Shell/Shuffling.hpp
MichaelRawson: Ahah, you want a stack of heterogenous things in `shuffleIter`, right. OK, sounds good!

MichaelRawson:
Thanks for changes! All happy here. :-)

PR #412, opened 2022-08-23
stacktrace without `CALL`

MichaelRawson:
We use manually-inserted `CALL` macros to keep a call stack available in debug builds. This is primarily used to generate backtraces when crashing, but also for some old-school watchpoint stuff that I don't fully understand and suspect isn't used. Backtraces are useful (rather than debugging) when e.g. Vampire crashes on a compute cluster or in a competition, where debugging or even getting a core dump is not so easy.

However, in most cases and on most platforms it's possible to get a callstack without these macros. This PR implements `CALL`-free backtraces for GNU-flavoured things (and MacOS - thanks @quickbeam123!) by (i) inspecting the call stack, (ii) pulling out symbol names and (iii) demangling them according to the C++ ABI. It should be possible to add support for other platforms incrementally.

Pros:
- no need to write or update `CALL` macros
- decrease in binary size, about 2%
- significant increase in debug performance: 20-40% on my machine
- explicit types in backtrace (although this might be considered a disadvantage in some cases)
- get a backtrace in release builds too

Cons:
- more platform-specific code, a small amount of additional complexity
- need to link to `libdl` on GNU (extra library shipped with glibc)
- now displays the real callstack, not an idealised version: optimising transformations may cause some surprises
- no backtraces on unsupported platforms until somebody implements it

Let me know what you think, I've no desire to mess with people's workflows, but the lack of `CALL` and the increase in debug performance are compelling in my view. If people like it I will also remove the `CALL` macros and clean up `Debug::Tracer` a bit.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,11 +9,8 @@ set(CMAKE_CXX_STANDARD 14)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
 
-set(CMAKE_ENABLE_EXPORTS ON)
-
 # configure compilers
 if(CMAKE_CXX_COMPILER_ID STREQUAL GNU OR CMAKE_CXX_COMPILER_ID MATCHES Clang$)
-  add_compile_options(-ggdb)
   add_compile_options(-Wall)
   # we don't use multithreading in (mainline!) Vampire
   add_compile_options(-fno-threadsafe-statics)
@@ -44,6 +41,12 @@ if (NOT BUILD_SHARED_LIBS)
   set(CMAKE_EXE_LINKER_FLAGS -static)
   set(VAMPIRE_BINARY_STATIC _static)
 endif()
+
+# get symbol names in binary for backtraces
+set(CMAKE_ENABLE_EXPORTS ON)
+
+# link to libdl for dladdr() on POSIXy things, used for backtraces
+link_libraries(${CMAKE_DL_LIBS})
 
 # We compile tests only in debug mode, since in release mode assertions are NOPs anyways.
 if(CMAKE_BUILD_TYPE STREQUAL Debug)
@@ -913,10 +916,6 @@ if (COMPILE_TESTS)
   endforeach()
 
 endif() # COMPILE_TESTS
-
-if(CMAKE_BUILD_TYPE STREQUAL "Debug")
-  link_libraries(${CMAKE_DL_LIBS})
-endif()
 
 #################################################################
 # automated generation of Vampire revision information from git #
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,8 +9,11 @@ set(CMAKE_CXX_STANDARD 14)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
 
+set(CMAKE_ENABLE_EXPORTS ON)
+
 # configure compilers
 if(CMAKE_CXX_COMPILER_ID STREQUAL GNU OR CMAKE_CXX_COMPILER_ID MATCHES Clang$)
+  add_compile_options(-ggdb)
   add_compile_options(-Wall)
   # we don't use multithreading in (mainline!) Vampire
   add_compile_options(-fno-threadsafe-statics)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -911,6 +911,10 @@ if (COMPILE_TESTS)
 
 endif() # COMPILE_TESTS
 
+if(CMAKE_BUILD_TYPE STREQUAL "Debug")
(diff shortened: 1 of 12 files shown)


quickbeam123:
Looking cool :)

I guess we should all play with it a bit and then discuss!

selig:
There are a lot of benefits there.

One point in the discussion is that I have relied on control points for debugging before. The problem with debug output is often that the printing is far too slow and far too much when the point you're interested in is quite far into proof search.

The way the feature works is that when you get a violation in debug mode you get the number of 'control points' passed. You can then define CP_FIRST_POINT and CP_LAST_POINT in Debug/Tracer to only print debug stuff within a window. A control point is created for every CALL macro. If we don't have the CALL macros then we can't use this feature. Although looking at the code I can't really work out how this is all achieved... maybe it doesn't work anymore! I think it's been a few years since I've used the feature.

The question is whether the benefits outweigh losing the feature (which perhaps only I use?)... perhaps there's another way of achieving this

MichaelRawson:
>One point in the discussion is that I have relied on control points for debugging before.

>The question is whether the benefits outweigh losing the feature (which perhaps only I use?)... perhaps there's another way of achieving this

Fair enough! From your description I would reach for a debugger with a watchpoint set on some variable, but this doesn't map exactly.

selig:
Right... I wonder if there's a way of saying "only break on this watchpoint after reaching it 5000th time"... that's the core of the issue - when runs are long and you still want to do detailed debugging.

MichaelRawson:
Broadly yes, see below. Or you could use a conditional breakpoint or something. I don't think the question is really whether a debugger _can_ do it, more whether the pain of changing the way people debug stuff is worth the other positives.

```console
michael@desktop:/tmp$ cat test.c
int main()
{
	int x;
	for(int i = 0; i < 5003; i++)
		x = i;
	return 0;
}
michael@desktop:/tmp$ cc -o test test.c -ggdb
michael@desktop:/tmp$ gdb ./test
(gdb) start
Temporary breakpoint 1 at 0x1129: file test.c, line 4.
Starting program: /tmp/test 

Temporary breakpoint 1, main () at test.c:4
4		for(int i = 0; i < 5003; i++)
(gdb) watch x
Hardware watchpoint 2: x
(gdb) ignore 2 5000
Will ignore next 5000 crossings of breakpoint 2.
(gdb) continue
Continuing.

Hardware watchpoint 2: x

Old value = 5000
New value = 5001
main () at test.c:4
4		for(int i = 0; i < 5003; i++)
(gdb) continue
Continuing.

Hardware watchpoint 2: x

Old value = 5001
New value = 5002
main () at test.c:4
4		for(int i = 0; i < 5003; i++)
(gdb) continue
Continuing.

Watchpoint 2 deleted because the program has left the block in
which its expression is valid.
0x00007ffff7e1dd0a in __libc_start_main (main=0x555555555125 <main>, argc=1, 
    argv=0x7fffffffe288, init=<optimized out>, fini=<optimized out>, 
    rtld_fini=<optimized out>, stack_end=0x7fffffffe278)
    at ../csu/libc-start.c:308
308	../csu/libc-start.c: No such file or directory.
(gdb) continue
Continuing.
[Inferior 1 (process 1897) exited normally]
(gdb) 
```

selig:
I can update my workflow to adopt the above gdb magic so if it's just me who uses it please consider it unused.

ibnyusuf:
The CALL macros are a pain to maintain, so I think this is a good step.

One functionality that I  have used is the ability to print the call stack without breaking program flow. For example, if I want to analyse all calls to function `foo`, I add a statement in `foo`'s body `Debug::tracer::printStack(cout)`. Is this something that we can do with new method?

MichaelRawson:
Seems like 100% (3 out of 3 people) are keen, so let's try and get this in. I've neatened it up a bit, rebased onto master and implemented backtrace-in-release-builds. I believe this should work on GNUish things and Mac OS, which covers most of our developers and StarExec. In the event of an unsupported platform, we "gracefully degrade" to printing whatever information we do actually know how to get. If we ever have a Windows developer that wants this I would be delighted to help them patch it, other platforms too (BSD?!).

@joe-hauns - one thing I did break a bit was your fancy `DBG` macros: they still work, but because they don't know how deep they are it doesn't indent nicely like it used to. The current function name I get from `__func__`, which probably isn't quite as nice either. If you want these fixed and don't mind a slight performance hit we can query this from the backtrace info. I hope you can forgive me!

I'll advertise this a bit internally and leave some more time for discussion, then I'll consider merging it.

joe-hauns:
*very happy to see these changes come btw

quickbeam123:
I am also fine with the transition...

MichaelRawson:
In the best traditions of "deploy just before the weekend", any final objections? Otherwise I'll merge this afternoon.

MichaelRawson:
Andrei-approved. This goes in, thanks all!

PR #413, opened 2022-08-23
Some updates in PortfolioMode and snake schedules

quickbeam123:
Portfolio mode made compatible with
- instruction limiting
- shuffled strategies

Additionally, some slight improvements in `getExtraSchedules` and the latest snake schedules: a bit better still than the CASC ones (yes already "rotting" as well, due to changes in master).


CASC/PortfolioMode.cpp
```diff
@@ -224,80 +222,116 @@ bool PortfolioMode::performStrategy(Shell::Property* property)
...
-void PortfolioMode::getExtraSchedules(Property& prop, Schedule& old, Schedule& extra, bool add_extra, int time_multiplier)
+void PortfolioMode::getExtraSchedules(Property& prop, Schedule& old, Schedule& extra, bool add_extra, float time_multiplier)
 {
   CALL("PortfolioMode::getExtraSchedules");
 
   // Add new extra_opts here
   Stack<vstring> extra_opts;
 
   if(add_extra){
+    /*
```
MichaelRawson, line 233: More generally this fuzzing seem quite ad-hoc and some are only here because of the old schedules. Can some of these go? Can this whole function disappear?

CASC/PortfolioMode.cpp
```diff
@@ -513,6 +577,14 @@ void PortfolioMode::runSlice(vstring sliceCode, int timeLimitInDeciseconds)
   try
   {
     Options opt = *env.options;
+
+    // opt.randomSeed() would normally be inherited from the parent
+    // addCommentSignForSZS(cout) << "runSlice - seed before setting: " << opt.randomSeed() << endl;    
+    if (env.options->randomizedPortfolioWorkers()) {
+      // but here we want each worker to have their own seed
+      opt.setRandomSeed(getpid());
```
MichaelRawson, line 585: This seems dubious as PIDs might be recycled. Maybe `std::random_device` again, or a combination of PID and UNIX time?

Lib/Environment.cpp
```diff
@@ -127,9 +127,9 @@ bool Environment::timeLimitReached() const
  */
 int Environment::remainingTime() const
 {
-  // If time limit is set to 0 then assume we always have a minute left
+  // If time limit is set to 0 then assume we always have an hour left
```
MichaelRawson, line 130: No problem with it, but why?

Shell/Options.cpp
```diff
@@ -3215,7 +3222,6 @@ void Options::readOptionsString(vstring optionsString,bool assign)
                 USER_ERROR("value "+value+" for option "+ param +" not known");
                 break;
               case IgnoreMissing::WARN:
-                env.beginOutput();
```
MichaelRawson, line 3218: I assume deliberate.

Shell/Options.cpp
```diff
@@ -179,6 +181,11 @@ void Options::init()
     _lookup.insert(&_slowness);
     _slowness.onlyUsefulWith(UsingPortfolioTechnology());
 
+    _randomizedPortfolioWorkers = BoolOptionValue("randomized_portfolio_workers","",true);
+    _randomizedPortfolioWorkers.description = "In portfolio mode, let each worker process start from its own independent random seed.";
+    _lookup.insert(&_randomizedPortfolioWorkers);
+    _randomizedPortfolioWorkers.onlyUsefulWith(UsingPortfolioTechnology());
```
MichaelRawson, line 187: No issue here, I'm just really pleased with how the option constraint stuff is going, this line is really beautiful!

MichaelRawson (approved):
Nice changes, fixes some of my gripes. :-)

CASC/PortfolioMode.cpp
quickbeam123: For instance, most of it could go from the point of view of the new snake schedules, but even these are already slowly getting outdated, so I think the function "as a concept" is nice to have! And that's for add_extra part.

The grow_time part is, in my opinion, generally useful. I can imagine a user with a lot of time and not much inspiration asking a schedule to run for days ... then it seems the best we can possibly offer is to repeat the strategies with larger and larger time/instruction limits.

Lib/Environment.cpp
quickbeam123: Thinking of the use case with time limit set to 0 (in the master process) - which means "run forever". I want to run strategies scheduled with a time limit greater than 1 minute each...

CASC/PortfolioMode.cpp
quickbeam123: A good point.

Shell/Options.cpp
MichaelRawson: Yes, absolutely! Pet peeve.

CASC/PortfolioMode.cpp
```diff
@@ -143,12 +144,16 @@ bool PortfolioMode::searchForProof()
...
-    if (env.options->shuffleInput()) { // instead to "combing things into shape" we shuffle and ruffle them
-      Shuffling().shuffle(*_prb);
-    } else {
+    if (env.options->normalize()) { // set explicitly by CASC(SAT) and SMTCOMP modes
       Normalisation().normalise(*_prb);
     }
 
+    // note that this is shuffleInput for the master process (for exceptional/experimental use)
+    // the usual way is to have strategies request shuffling explicitly in the schedule strings
+    if (env.options->shuffleInput()) {
```
selig, line 153: Out of interest, what's the interaction between normalization and shuffling... I guess that if we normalize before shuffling we'll get something different than if we don't normalize? But, in general, it seems a waste of work to normalize then shuffle (?)

CASC/PortfolioMode.cpp
```diff
@@ -224,80 +223,116 @@ bool PortfolioMode::performStrategy(Shell::Property* property)
...
+    {
+      // Don't try this in SMT-COMP mode as it requires a goal
+      extra_opts.push("slsq=on");
+      extra_opts.push("tgt=full");
+    }
 
+    // (almost) all of the above are integrated into SNAKE schedules
+    if (env.options->schedule() == Options::Schedule::SNAKE_TPTP_UNS) 
+    {
+      extra_opts.reset();
```
selig, line 283: Instead of adding them and resetting it, perhaps this should be a branch at the top level, perhaps with different functions? I'm thinking about readability... the logical flow of branching on options->schedule isn't clear here.

CASC/PortfolioMode.cpp
```diff
@@ -224,80 +223,116 @@ bool PortfolioMode::performStrategy(Shell::Property* property)
...
+      // Add a copy with increased time limit ...
+      vstring new_time_suffix = Lib::Int::toString((int)(t*time_multiplier));
+
+      if (add_extra) {
+        // ... for every possible extra_opt
         Stack<vstring>::Iterator addit(extra_opts);
-        while(addit.hasNext()){
-          vstring new_s = prefix + ":" + addit.next() + "_" + Lib::Int::toString(t);
+        while(addit.hasNext()) {
+          vstring new_s = prefix + ((prefix.back() != '_') ? ":" : "") + addit.next() + "_" + new_time_suffix;
```
selig, line 326: Can I check what the prefix.back() != '_' case is?

Lib/Environment.cpp
selig: Makes sense... btw most users don't know about the time limit = 0 means forever behavior

Shell/Options.cpp
selig: If I recall, one of the things beginOutput does is take a lock on output stream so that multiple processes can't write at the same time... which is useful/important.

Shell/Options.cpp
```diff
@@ -179,6 +181,11 @@ void Options::init()
     _lookup.insert(&_slowness);
     _slowness.onlyUsefulWith(UsingPortfolioTechnology());
 
+    _randomizedPortfolioWorkers = BoolOptionValue("randomized_portfolio_workers","",true);
```
selig, line 184: Is this option name clear? From the option name, I assumed some extra randomization going on but that's not quite what's going on.

Out of interest... portfolio mode isn't really deterministic but I'm guessing setting this to true makes it 'less' deterministic (?)

CASC/PortfolioMode.cpp
quickbeam123: A good shuffle should make any previous normalization a waste of effort. A good normalization should undo any shuffle. 

However, unfortunately, our Normalization wasn't perfect last time I looked and I didn't find a way to fix it (cheaply). It boils down to graph isomorphism problem again, so somewhere down the parse tree it must decided, e.g., that "a" > "b" or vice versa. But that might be a detail compared to the fact we cannot "normalize" the signature order, so ---most glaringly---`sp=occurrence` will always refer to the occurrence in the original (perhaps Geoff-shuffled) input, not in the normalized one.

CASC/PortfolioMode.cpp
quickbeam123: A good point! Although I was thinking of the part of the code as a very temporary mess (with lower coding standards), now I maybe have an idea how to streamline things a bit... Will push a fix in a bit!

CASC/PortfolioMode.cpp
```diff
@@ -224,80 +223,116 @@ bool PortfolioMode::performStrategy(Shell::Property* property)
...
-        // Add copies with new extra options (keeping the old time limit) 
+    vstring s = it.next();
+
+    if (time_multiplier > 1.0) { // also blow up any present instruction limits      
+      size_t bidx = s.rfind(":i=");
+      if (bidx == vstring::npos) {
+        bidx = s.rfind("_i=");
+      }
+      if (bidx != vstring::npos) {
+        bidx += 3; // advance past the "*i=" bit
```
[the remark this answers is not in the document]
quickbeam123: It's meant as a wild-card. The string could be, e.g., "lrs+10_1:1_i=50_0", but also the more typical "lrs+10_1:1_av=off:i=50_0".

CASC/PortfolioMode.cpp
quickbeam123: That for inserting into the empty main part of the strat string: "fmb+10_1:1__100"

vampire.cpp
```diff
@@ -706,6 +707,8 @@ int main(int argc, char* argv[])
       env.options->setOutputMode(Options::Output::SZS);
       env.options->setProof(Options::Proof::TPTP);
       env.options->setOutputAxiomNames(true);
+      env.options->setNormalize(true);
```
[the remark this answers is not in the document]
quickbeam123: Yes! And no... 

I would argue that the use case of running Vampire on problems we have seen in training (of the respective competition mode), we know are solvable beforehand, and moreover, can only be solved by Vampire when normalized, is not an interesting one. We want Vampire to work on previously unseen problems!

And on unseen problems, normalization is not a useful heuristic, it's effectively a shuffle with a particular fixed seed. So we can as well not normalize and save the time...

For competitions, I would like to help and provide schedules which don't rely on normalization. 

But yes, with the current spider schedules at least, we need to remember things if we depart from comp_modes and use portfolio_mode + options.

Shell/Options.cpp
quickbeam123: Happy to discuss a better name here! "randomize_seed_for_portfolio_workers"?

The main use case is this: we have strategy A, later, thanks to getExtraSchedules, we have strategy A+shuffling and later still we have it again (perhaps with higher limit). If the workers inherit a fixed seed from the master process running A+shuffling twice in a row will not be as effective as with different seeds. 

But let's also discuss, if it should be on by default. It though it's fine, as long as proper competition modes have it turned off. Also, without explicit shuffling there a little use of the random bits by vampire. If I remember correctly, it's in the skiplists of the passive clause containers (only influences timing, not decisions) and in global_subsumption_explicit_minim which is actually "randomized" by default.

Shell/Options.cpp
quickbeam123: This one is actually (mildly) interesting.

`beginOutput() / endOutput()` do some locking but only if a `pipe` has been set as the output sink. Which is done by `setPipeOutput` only ever called by `tTwoVampires.ccp` close to a very historic line 

`runChild(units, "dis+10_32_nwc=2.0:sac=on:spl=backtracking_20"); //start proving`

we also have `Environment::setPriorityOutput` never called at all. So for all practical purposes `Environment::out()` is the same as `cout` and `beginOutput() / endOutput()` never does anything useful. (Unless we want to go back to pipes and priorities, of course.)

quickbeam123:
I just changed the way we do "getExtraSchedules" in PortfolioMode, in case you wanted to have an additional look. (The said function is actually gone.)

PR #414, opened 2022-09-01
Joe uwa stack of handlers

joe-hauns:
The biggest change is the change the change in the `MisatchHandler` class hierarchy.
What used to be `CompositieMismatchHandler` is now `MismatchHandler`. No inheritance or subtyping there, because this is the only kind of mismatch handler that is actually being used in client code (i.e. Indexing and Robsubstitution). 

What used to be the abstract class `MisnmatchHandler` is now the abstract class `AtomicMismatchHandler` (maybe we have a better name for this (?), like AbstractionStrategy, ...).
This abstract class is implemeneted by `HOMismatchHandler` and `UWAMismatchHandler`. it is not being interacted with anywhere in the codebase, except for in `MismatchHandler.cpp`.

These changes fix the issue that some function in the old `MismatchHandler would have to throw an exception  before, therefore we get compile-time safety that different handlers cannot be "misused".

### DIFF
--- a/Indexing/IndexManager.cpp
+++ b/Indexing/IndexManager.cpp
@@ -37,21 +37,13 @@ IndexManager::IndexManager(SaturationAlgorithm* alg) : _alg(alg), _handler()
 {
   CALL("IndexManager::IndexManager");
 
-  static bool const uwa = env.options->unificationWithAbstraction()!=Options::UnificationWithAbstraction::OFF;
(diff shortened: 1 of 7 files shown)


joe-hauns:
Further I made the uwa tests being automatically checked in unit testing (results aren't just printed to stdout anymore). 
There are some little TODO's in `tUnificaitonWithAbstraction` for the `HOMismatchHandler`, as I'm not 100% sure what the implementation should do there.

PR #420, opened 2022-10-06
Parametric datatypes

mezpusz:
The SMTLIB2 standard specifies parametric datatypes and sort definitions, the main improvement of this PR is adding support for those in the following form:
```
(declare-datatype list (par (a) ((Nil) (Cons (Head a) (Tail (list a))))))
(define-sort lists (a) (list (list a)))
```
Note that the `declare-datatypes` keyword has not been extended, it remains as a future work. The datatypes are truely polymorphic in a sense that the datatype axioms are added with type variables, although according to the standard it makes only sense to use them in this way is via theory declarations, which Vampire does not support. Trying to reach a halfway point between full usability and effort invested results in a non-standard extension of SMTLIB2 parsing by being able to parameterize function declarations/definitions and formulas. A couple of examples for this:
```
(declare-fun append (par (a) ((list a) (list a)) (list a)))
(assert (par (a) (forall ((ys (list a))) (= (append a (Nil a) ys) ys))))
(assert (par (a) (forall ((x a) (xs (list a)) (ys (list a))) (= (append a (Cons a x xs) ys) (Cons a x (append a xs ys))))))

(define-fun-rec len (par (a) ((xs (list a))) nat
  (match xs (((Nil a) zero)
             ((Cons a y ys) (s (len a xs)))))))
```
Note that the position of `par` blocks in `declare-fun` and `define-fun[-rec]` blocks is fixed, while in `assert` blocks it can be anywhere similar to a `forall` quantifier and its meaning is the same as universally quantifying over the type variables in its block. Type parameters are optional, the original SMTLIB2 parsing also works as before. One can notice that the type parameters need to be explicitly given like in `(Nil a)`, `(len a xs)` instead of only `Nil` and `(len xs)`. Implementing some kind of type deduction to avoid this is also future work, probably the current parsing needs to be restructured to achieve this.

A note on `let` expressions. SMTLIB2 defines `let` binders as variables rather than functions as in TPTP. We define the sort of a binder as the result sort of the bound expression, and therefore the type arguments are the free variables of the bound expression sort. Due to the current internal working of `let`s in Vampire, the type arguments need to be explicitly stated for the binders but since they also have to behave like variables (e.g. shadowing), the result is a peculiar combination of both. An example, where `ys`, `zs` and `ts` are used as `(ys a)`, `(zs a)` and `(ts a)`, respectively:
```
(let ((ys (take a (pow2 k) xs)) (zs (drop a (pow2 k) xs)))
    (let ((ts (brauns a (s k) (zs a)))) (nodes a (ts a) (ys a) (drop (tree a) (pow2 k) (ts a)))))
```

PS: a couple of related bugs were also caught, thanks @ibnyusuf for helping with the code!


MichaelRawson:
Looks great from a first glance! I will take a little while to review this, I remember that I wrote some code assuming that datatypes would be monomorphic (since the SMTLIB standard says that they are!) so probably I need to go and fix that.

MichaelRawson (changes requested):
Bravo! Looks mostly good to me, a few comments here and there. A particularly heroic effort extending the SMTLIB parser.

A slight hitch: `Z3Interfacing`. I think it treats datatype constructors with zero _term_ arity as a constant for Z3, see lines 869-888 of `Z3Interfacing.cpp`. I'm not sure why it checks for term arity rather than general arity: in principle we could have polymorphic constants. If you are able to look and fix it, great - otherwise I will file a bug and get to it at some point.

Kernel/FormulaVarIterator.cpp
```diff
@@ -165,6 +165,8 @@ bool FormulaVarIterator::hasNext()
             case Term::SF_ITE:
               _instructions.push(FVI_FORMULA);
               _formulas.push(sd->getCondition());
+              _instructions.push(FVI_TERM_LIST);
```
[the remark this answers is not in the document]
mezpusz: Yes, in particular special terms do not necessarily contain the type variables anywhere in the term itself, so we check the sort as well.

Kernel/LPO.cpp
```diff
@@ -64,6 +64,22 @@ Ordering::Result LPO::comparePredicates(Literal* l1, Literal *l2) const
   return (predicatePrecedence(p1) > predicatePrecedence(p2)) ? GREATER : LESS;
 } // LPO::comparePredicates()
 
+Ordering::Result LPO::comparePrecedences(Term* t1, Term* t2) const
```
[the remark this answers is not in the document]
mezpusz: As far as I know LPO hasn't been used much for TPTP benchmarks, so it remained undetected so far.

Parse/TPTP.cpp
```diff
@@ -3598,6 +3598,7 @@ void TPTP::endType()
     break;
   case TT_QUANTIFIED:
     VList* vl = _varLists.pop();
+    _sortLists.pop();
```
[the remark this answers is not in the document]
mezpusz: It was something @ibnyusuf fixed, so he might know what happened here. And this reminds me that maybe a TPTP benchmark with type variables should be added as a parser check.

Shell/InterpretedNormalizer.cpp
```diff
@@ -278,6 +278,8 @@ class InterpretedNormalizer::NLiteralTransformer : public TermTransformer
   {
     CALL("InterpretedNormalizer::NLiteralTransformer::transformSubterm");
 
+    if(trm.isTerm() && trm.term()->isSort()) return trm;
```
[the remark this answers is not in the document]
mezpusz: Hm, I don't remember this part, @ibnyusuf ?

mezpusz:
> A slight hitch: `Z3Interfacing`. I think it treats datatype constructors with zero _term_ arity as a constant for Z3, see lines 869-888 of `Z3Interfacing.cpp`. I'm not sure why it checks for term arity rather than general arity: in principle we could have polymorphic constants. If you are able to look and fix it, great - otherwise I will file a bug and get to it at some point.

Thanks for spotting this @MichaelRawson ! Honestly I haven't tried the code with Z3, so I'll have to fix those parts. I have a look shortly and see if I can fix it myself.

Parse/TPTP.cpp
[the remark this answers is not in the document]
ibnyusuf: This was an existing error, that somehow never manifested until now. The size of `_varLists` and `_sortLists` should always remain the same. This fix ensures that that invariant is maintained here.

@mezpusz , I will try and find a simple polymorphic problem to add to our sanity checks.

Parse/SMTLIB2.cpp
```diff
@@ -888,87 +719,90 @@ SMTLIB2::DeclaredFunction SMTLIB2::declareFunctionOrPredicate(const vstring& nam
...
+    type = OperatorType::getPredicateType(argSorts.size(),argSorts.begin(),taArity);
 
     LOG1("declareFunctionOrPredicate-Predicate");
   } else { // proper function
-    if (argSorts.size() > 0) {
-      symNum = env.signature->addFunction(name, argSorts.size(), added);
+    if (argSorts.size() > 0 || taArity > 0) {
+      symNum = env.signature->addFunction(name, argSorts.size()+taArity, added);
     } else {
+      // TODO what about taArity here?
```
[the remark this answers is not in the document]
mezpusz: Right, but I was thinking that maybe constants are handled differently which should be preserved even for `taArity > 0`, so for "constants" modulo type arguments. But as far as I can tell, there's no such special handling anywhere.

Parse/SMTLIB2.cpp
```diff
@@ -1002,64 +837,101 @@ void SMTLIB2::readDefineFun(const vstring& name, LExprList* iArgs, LExpr* oSort,
   UnitList::push(fu, _formulas);
 }
 
+void SMTLIB2::tryReadTypeParameters(LispListReader& rdr, TermLookup* lookup, TermStack* ts)
```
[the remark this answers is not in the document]
mezpusz: Good observation! It started as a "try" function but now it should be only called when there *is* a parametric block ahead. Maybe this should be stated in a comment somewhere.

Parse/SMTLIB2.cpp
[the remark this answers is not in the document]
MichaelRawson: Yeah, as far as I know (with the possible exception of `Z3Interfacing`!) these "constants" are not treated specially. The whole point of doing the whole TFF1 hack is so that sorts are kinda-terms, so in principle there should be very little special handling.

Kernel/FormulaVarIterator.cpp
```diff
@@ -214,6 +220,8 @@ bool FormulaVarIterator::hasNext()
               _termLists.push(sd->getLambdaExp());
               _instructions.push(FVI_BIND);
               _vars.push(sd->getLambdaVars());
+              _instructions.push(FVI_TERM_LIST);
```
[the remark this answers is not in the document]
ibnyusuf: These instructions are not required for SF_LAMBDA since we are already checking the sort with some the instructions above. I will fix this shortly.

MichaelRawson (approved):

Inferences/TermAlgebraReasoning.cpp
```diff
@@ -248,16 +246,22 @@ namespace Inferences {
...
-  bool NegativeInjectivityISE::litCondition(Clause *c, unsigned i) {
+  bool NegativeInjectivityISE::litCondition(Clause *c, unsigned i)
+  {
+    CALL("NegativeInjectivityISE::litCondition");
     Literal *lit = (*c)[i];
     if (sameConstructorsEquality(lit) && !lit->polarity()) {
       unsigned arity = lit->nthArgument(0)->term()->arity();
-      OperatorType *type = env.signature->getFunction(lit->nthArgument(0)->term()->functor())->fnType();
       for (unsigned j = 0; j < arity; j++) {
+        if (j < lit->nthArgument(0)->term()->numTypeArguments()) {
```
mezpusz, line 256: I think this is not necessary due to the well-typedness which is performed on each shared term. @ibnyusuf what do you think?

ibnyusuf: Agreed. This check is unnecessary, since if the type arguments differed, the equality literal wouldn't type check.

Inferences/TheoryInstAndSimp.cpp
```diff
@@ -171,8 +170,8 @@ bool TheoryInstAndSimp::isSupportedFunction(Term* trm) {
   auto sym = env.signature->getFunction(trm->functor());
```
ibnyusuf, line 170: Can't add a comment to line 210 of this file, so commenting here.

I don't think we should be using `SubtermIterator` on line 210, but rather `NonVariableNonTypeIterator`

Kernel/FormulaVarIterator.cpp
```diff
@@ -180,13 +182,17 @@ bool FormulaVarIterator::hasNext()
 
               _instructions.push(FVI_BIND);
               _vars.push(sd->getVariables());
+              _instructions.push(FVI_TERM_LIST);
```
ibnyusuf, line 185: You can remove this line and the next one, since the sort of an `SF_FORMULA` term must be` $o `and therefore cannot contain any variables.

Kernel/Signature.hpp
```diff
@@ -884,9 +884,9 @@ class Signature
   unsigned formulaCount(Term* t);
 
 
-  bool isTermAlgebraSort(TermList sort) { return _termAlgebras.find(sort); }
-  Shell::TermAlgebra *getTermAlgebraOfSort(TermList sort) { return _termAlgebras.get(sort); }
-  void addTermAlgebra(Shell::TermAlgebra *ta) { _termAlgebras.insert(ta->sort(), ta); }
+  bool isTermAlgebraSort(TermList sort) { return sort.isTerm() && _termAlgebras.find(sort.term()->functor()); }
+  Shell::TermAlgebra *getTermAlgebraOfSort(TermList sort) { return _termAlgebras.get(sort.term()->functor()); }
+  void addTermAlgebra(Shell::TermAlgebra *ta) { _termAlgebras.insert(ta->sort().term()->functor(), ta); }
```
ibnyusuf, line 889: What is the reason for these changes?

ibnyusuf (commented):
I haven't looked at the parser code, assuming that to be correct.

Kernel/FormulaVarIterator.cpp
mezpusz: This is the `SF_LET` case, I think the GitHub code viewer tricked you!

Inferences/Induction.cpp
```diff
@@ -1043,12 +1046,12 @@ void InductionClauseIterator::performStructInductionOne(const InductionContext&
   for(unsigned i=0;i<ta->nConstructors();i++){
     TermAlgebraConstructor* con = ta->constructor(i);
     unsigned arity = con->arity();
-      Stack<TermList> argTerms;
+      Stack<TermList> argTerms = typeArgs;
```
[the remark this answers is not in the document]
mezpusz: Good point, calling (the new, efficient) `getTypeArguments` instead.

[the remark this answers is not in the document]
mezpusz: Noted.

Kernel/SortHelper.cpp
```diff
@@ -73,6 +73,14 @@ void SortHelper::getTypeSub(const Term* t, Substitution& subst)
   }  
 } // getTypeSub
 
+TermStack SortHelper::getTypeArguments(TermList sort)
+{
+  CALL("SortHelper::getTypeArguments");
+  ASS(sort.isTerm() && sort.term()->isSort());
+
+  return TermStack::fromIterator(Term::Iterator(sort.term()));
```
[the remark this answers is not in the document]
mezpusz: Changed it.

Inferences/TheoryInstAndSimp.cpp
mezpusz: Good point, changed it.

Kernel/Signature.hpp
mezpusz: Since datatypes can be polymorphic now, it's easier to track them via their sort functor (otherwise we have to match sorts when we want to get the term algebra corresponding to them).

Kernel/Signature.hpp
```diff
@@ -884,9 +884,9 @@ class Signature
   unsigned formulaCount(Term* t);
 
 
-  bool isTermAlgebraSort(TermList sort) { return _termAlgebras.find(sort); }
-  Shell::TermAlgebra *getTermAlgebraOfSort(TermList sort) { return _termAlgebras.get(sort); }
-  void addTermAlgebra(Shell::TermAlgebra *ta) { _termAlgebras.insert(ta->sort(), ta); }
+  bool isTermAlgebraSort(TermList sort) { return _termAlgebras.find(sort.term()->functor()); }
```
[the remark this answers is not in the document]
mezpusz: How about short circuiting the return value with `sort.isTerm()` for `isTermAlgebraSort`? Some code calls this with arbitrary `TermList` sorts, which can be variables. For `getTermAlgebraSort` I agree with adding an assertion.

Kernel/FormulaVarIterator.cpp
ibnyusuf: Quite right!

Kernel/Signature.hpp
ibnyusuf: Right, makes sense.

Kernel/SortHelper.cpp
```diff
@@ -73,12 +73,12 @@ void SortHelper::getTypeSub(const Term* t, Substitution& subst)
...
 } // getTypeSub
 
-TermStack SortHelper::getTypeArguments(TermList sort)
+void SortHelper::getTypeArguments(TermList sort, TermStack& res)
 {
   CALL("SortHelper::getTypeArguments");
   ASS(sort.isTerm() && sort.term()->isSort());
 
-  return TermStack::fromIterator(Term::Iterator(sort.term()));
+  return res.loadFromIterator(Term::Iterator(sort.term()));
```
ibnyusuf, line 81: Does this `return` from a `void` function serve a purpose?

mezpusz: Nope, thank for catching it!

Kernel/SortHelper.cpp
```diff
@@ -73,6 +73,14 @@ void SortHelper::getTypeSub(const Term* t, Substitution& subst)
   }  
 } // getTypeSub
 
+void SortHelper::getTypeArguments(TermList sort, TermStack& res)
```
quickbeam123, line 76: I am not understanding the name of this function. 

It seems it assumes sort is a sort and returns all its arguments. Is this really what "getTypeArguments" means? Could that be explained in a comment?

mezpusz: Putting it like this, neither the name nor adding a function for this purpose make much sense. How about I inline this everywhere? It is only used on sorts in `Induction`, so the assertions are also kind of unnecessary.

Kernel/Signature.hpp
```diff
@@ -970,7 +973,7 @@ class Signature
   /**
    * Map from sorts to the associated term algebra, if applicable for the sort
```
quickbeam123, line 974: Could you please also update the comment?

Kernel/SortHelper.cpp
```diff
@@ -453,7 +459,7 @@ void SortHelper::collectVariableSortsIter(CollectTask task, DHMap<unsigned,TermL
 
             newTask.ts = binding;
             if (!isPredicate) {
-              newTask.contextSort = symbol->fnType()->result();
+              newTask.contextSort = SubstHelper::apply(type->result(),subst);
```
mezpusz, line 462: During looking for an error, I fixed this part which wasn't correct I think. The question is whether this might have caused any issues, and if it should be also fixed for let tuples or should I just revert it.

mezpusz:
Update: a couple of things have been modified after discussing $lets with @quickbeam123 and discovering a couple of errors.

First, the $let syntax is now the same as in the SMTLIB2 standard, no type arguments are needed when using the binder in the $let body. Also, some of the $let expressions were ill-formed, this has been fixed at the cost of adding all variables present in the bound expression to the $let binder to ensure that no variable having a certain type from an earlier quantification has a different type because of the fresh type variables in the $let binder.

Second, it turned out that SMTLIB allows an overlap between function/predicate and type constructor identifiers which was not taken into account in this PR. So there can be e.g. a datatype `dt` with a constructor `dt` and this fact is massively used in SMTLIB benchmarks. This has been fixed by adding the `()` postfix to type constructors.

As a result, now the branch gives more or less the same result for SMTLIB benchmarks as master, with only minor differences (ran with `-sa discount -t 10`:

- master: 55287 (302 unique) / 93347
- branch: 55468(483 unique) / 93347

A regression run for TPTP showed basically no difference.

SAT/Z3Interfacing.cpp
```diff
@@ -567,54 +567,65 @@ vstring to_vstring(A const& a)
...
+  for (auto taSort : taSorts) {
+    _createdTermAlgebras.insert(taSort);
+    ASS(taSort.isTerm());
+    auto taSortT = taSort.term();
+    auto ta = env.signature->getTermAlgebraOfSort(taSort);
+    Substitution typeSubst;
+    vstring typePostfix = "$";
+    ta->getTypeSub(taSortT, typeSubst);
+    for(unsigned i = 0; i < taSortT->arity(); i++) {
+      typePostfix += taSortT->nthArgument(i)->toString();
```
quickbeam123, line 613: Is it OK not to have a separator between the arguments' toStrings?

PR #422, opened 2022-10-18
cleanup `Forwards.hpp`

MichaelRawson:
See #157. `Forwards.hpp` is included almost everywhere to get some common declarations. There's some old stuff forward-declared in `Forwards.hpp`, which has been removed. Also, anything that perhaps doesn't belong there and is used only in a few places is moved into a more appropriate location. While doing this I also noticed some dead code, removed.


Lib/DArray.hpp
```diff
@@ -209,7 +209,9 @@ class DArray
     C* afterLast=_array+_capacity;
 
     while(nptr!=firstEmpty) {
-      Relocator<C>::relocate(optr++, nptr++);
+      C *oldAddr = optr++, *newAddr = nptr++;
```
MichaelRawson, line 212: This is probably the only potentially-hazardous bit, but it seems to be fine. Inline the definition of `Relocator`, copying the specialisation for `Stack` rather than the basic version, since with `std::move` it ought to be the same anyway. This is the last instance of `Relocator` so it can go now.

Saturation/ConsequenceFinder.hpp
```diff
@@ -27,6 +27,10 @@
 
 #include "Inferences/TautologyDeletionISE.hpp"
 
+namespace Lib {
```
quickbeam123, line 30: Generally in favour of local forward declarations, where necessary, but when one gets longer than a oneliner ... wouldn't 

`#include "Lib/SkipList.hpp"`

be easy to understand?

(In this particular case, by the way, the SkipList seems to be coming from some nested include anyway...)

MichaelRawson: Sure, depends on the tradeoff between "touched one header, recompile everything" and ease of maintenance. Let me know what you want to do in general, this one in particular can just go, as you say.

Lib/DArray.hpp
quickbeam123: The stack version used `::new` here you use `new`, could you remind me what the difference is and does it make a difference here? :)

MichaelRawson: This took a bit longer to figure out than I'd like! As I'm sure you recall (I didn't), this is a placement-new expression, which constructs `C` in the memory returned by `operator new(sizeof(C), newAddr)`. By default this function just returns `newAddr`, so this constructs `C` in-place at `newAddr`.

There are two points of subtlety:
1. This function can be overloaded. Doing so is not advisable.
2. If there is a class overload in `C` for `operator new(size_t count, void *ptr)`, then this is called, otherwise the global version is called. If instead you say `::new`, you skip the check for the class overload.

In Vampire we do overload the placement-new operators, but only via the macro `DECLARE_PLACEMENT_NEW` defined in `Allocator.hpp`. However, this overload simply implements the behaviour that would be the default anyway. I'm not sure why we have this: I think it can go altogether, which I will do in a separate PR.

So, to summarise: calling `::new` would skip the check for the `C`-overload placement-new. There may be such an overload, but at present it always does the same as the global version. If in future somebody does implement a non-trivial placement-new operator for `C`, they probably want `Stack` to call that rather than the global version, although I fervently hope that their implementation only e.g. increments some statistic and returns `newAddr`.

quickbeam123: Wow, thanks a lot for the detailed research and explanation! 

I agree overloading placement new for anything than maybe stats update is evil. 

I also agree getting rid of `DECLARE_PLACEMENT_NEW` would be nice. But separate PR might be a good idea, since some complications are to be expected given, e.g.:

```
class SDClauseMatches
{
  CLASS_NAME(SDClauseMatches);
  USE_ALLOCATOR(SDClauseMatches);

  public:
    // required to use std::vector::emplace_back
    DECLARE_PLACEMENT_NEW;
```

PR #423, opened 2022-10-27
cleanup `Lib::List`

MichaelRawson:
See #157. It was brought to my attention that there is a default `List<T>` constructor, which is almost never what you want. Remove it. While I was there, I also:
- removed liberal use of the `inline` keyword, the compiler probably decides anyway so this is just noise
- removed a few functions that weren't used and didn't seem immediately useful
- removed the last few instances of `FIFO::push` in favour of `FIFO::pushBack`
- removed the default `List<T>::Iterator` constructor, almost never what you want
- removed some const-correctness bugs that were there for some reason

### DIFF
--- a/Kernel/Formula.cpp
+++ b/Kernel/Formula.cpp
@@ -294,7 +294,7 @@ VList* Formula::freeVariables () const
   VList* result = VList::empty();
   VList::FIFO stack(result);
   while (fvi.hasNext()) {
-    stack.push(fvi.next());
+    stack.pushBack(fvi.next());
   }
   return result;
 } // Formula::freeVariables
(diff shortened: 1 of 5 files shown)


quickbeam123:
All looking good! 

I am just a bit hesitant to delete the two functions, which, I assume, were once debugged and someone might want to use in the future (separated in a library header file, do they hurt much?)

MichaelRawson:
> I am just a bit hesitant to delete the two functions, which, I assume, were once debugged and someone might want to use in the future (separated in a library header file, do they hurt much?)

I couldn't decide either! There is a slight compile-time penalty but if the compiler is any good they shouldn't end up in the final binary. I'll restore them.

PR #425, opened 2022-12-07
function definition introduction

MichaelRawson:
Well, we have function definition _elimination_, why not the opposite? More seriously, we found this was useful in a very specialised application of Vampire. Opening this to see if anyone's interested, not necessarily to be merged.

The general idea is to detect when a term is used frequently and introduce a definition for it.

The code works by looking at terms in clauses as they are activated: if a term has been seen more than a threshold, introduce a definition for it. This includes generalisations: if we see e.g. `f(a, g(a))` and `f(b, g(b))` frequently, we could define `d(X) = f(X, g(X))`. I use a cheap approximate technique to detect when we need a definition, which seems to work well enough. Definitions are added to proof search as a unit equation, as if it were an axiom. Demodulation typically (always?) rewrites the rest of the search space as a result, since the definition is smaller than what it denotes. Since these are just definitions, the technique is sound and complete.

This is not a good idea in general. There is non-zero overhead (although relatively small, at least in the beginning of proof search), but the bigger problem is that the disturbance to proof search is significant, particularly the SAT-based methods which don't tend to like new function symbols appearing. There are however some problems for which this is really beneficial.

I ran 1-second DISCOUNT runs over ~5000 problems in TPTP, with the threshold set to 0 (i.e. off), 1, 2, 4, and 8 occurrences. In hindsight maybe a few more powers of 2 would have been useful. This had no effect on the majority of problems. Of the rest, most were hindered by introducing definitions. However, 60 problems were not solved without definition introduction, but were solved with some threshold. They are:

```
ALG370-1
ARI105=1
ARI535=1
ARI674=1
ARI721=1
ARI726=1
CAT004-1
DAT080=1
GEO112-1
GEO568+1
GEO619+1
GEO632+1
GEO642+1
GRP402-1
GRP498-1
GRP519-1
HWV031-1
ITP012+2
ITP305_1
ITP325_1
KLE010+1
KLE015+1
KLE144+1
KLE150+1
LCL641+1.020
LCL896+1
MGT035-1
NUM569+3
NUM926_1
NUN086+1
PRO009+2
SCT085-1
SET021-6
SET108-7
SET235-6
SET473-6
SET669+3
SET844-1
SEU141+1
SEU147+2
SEU190+1
SWC031+1
SWC058+1
SWC065+1
SWC110+1
SWC355-1
SWC370+1
SWC372-1
SWV053+1
SWV153+1
SWV221+1
SWV394+1
SWV449+1
SWV454+1
SWV473+1
SWV592-1
SWW478_3
SWW619=2
TOP018-1
TOP019-1
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -423,6 +423,7 @@ set(VAMPIRE_INFERENCE_SOURCES
     Inferences/TautologyDeletionISE.cpp
     Inferences/TermAlgebraReasoning.cpp
     Inferences/URResolution.cpp
+    Inferences/DefinitionIntroduction.cpp
     Inferences/BackwardDemodulation.hpp
     Inferences/BackwardSubsumptionResolution.hpp
     Inferences/BinaryResolution.hpp
@@ -458,6 +459,7 @@ set(VAMPIRE_INFERENCE_SOURCES
     Inferences/TautologyDeletionISE.hpp
     Inferences/TermAlgebraReasoning.hpp
     Inferences/URResolution.hpp
+    Inferences/DefinitionIntroduction.hpp
     Inferences/TheoryInstAndSimp.hpp
     Inferences/TheoryInstAndSimp.cpp  # this is theory instantiation
     Inferences/ArithmeticSubtermGeneralization.hpp
--- a/Inferences/DefinitionIntroduction.cpp
+++ b/Inferences/DefinitionIntroduction.cpp
@@ -130,7 +130,8 @@ void DefinitionIntroduction::process(Term *t) {
       continue;
 
     entry.term = gen;
-    if(++entry.count > env.options->functionDefinitionIntroduction()) {
+    entry.weight += t->weight();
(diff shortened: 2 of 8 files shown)


ibnyusuf:
Interesting. Couple of questions:

1. Do I understand correctlu, that this is similar to naming, but for terms instead of formulas? Do you estimate how many times a term is going to occur in clausified set and decide to introduce a definition based on that?
2. How do you prevent elimination undoing introduction, or introduction undoing elimination?

ibnyusuf:
Ignore my earlier comment. Just looked at the code and realised that this is not a preprocessing technique, but an inference. Looks interesting.

MichaelRawson:
Yep, exactly - although you could definitely argue that perhaps it should also be a preprocessing step like naming, the opposite of `fde`.

MichaelRawson:
@quickbeam123 - reasonable range of settings for `fdi` is probably powers of two in the range 4-1024 (possibly even larger, hard to say - depends how much patience or compute you have), and 0 to switch it off.

PR #429, opened 2022-12-16
check for body of quantified SMTLIB2 formula

MichaelRawson:
There is currently no check that an SMTLIB quantified formula actually contains a sub-formula. This can be quite difficult to debug. Add this check as I doubt there is a performance concern here.

### DIFF
--- a/Parse/SMTLIB2.cpp
+++ b/Parse/SMTLIB2.cpp
@@ -1829,6 +1829,9 @@ void SMTLIB2::parseQuantBegin(LExpr* exp)
     }
   }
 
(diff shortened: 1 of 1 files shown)


quickbeam123:
You mean the master version correctly crashes but with a useless / confusing error message? Or something even worse?

MichaelRawson:
We get a semi-sensible stacktrace for debug master, and hard-crash (segmentation fault, usually) for release.

quickbeam123:
Oh, I see! Thank you for noticing!

PR #431, opened 2023-01-10
Substitution Tree Refactor

joe-hauns:
This PR is a big cleanup of SubstitutionTree and friends.

The big higlights are 
- getting rid of 1200 lines of code while
- making the implementation cleaner
- solving more problems on average
- adding nice readable outputting of substitution tree (great for debugging)
- adding `Recycler`, a smart pointer that keeps allocated memeory aroung (for todo stacks & similar)
- a optimized version of finding variable variants, that does not only work for literals, but also for terms now

The main changes are the following. 

`LiteralSubstitutionTree`, and `TermSubstitutionTree` contain hardly any logic anymore. They are just wrappers around `SubstitutionTree` in order to make it conform with the interfaces `TermIndexingStructure` and `LiteraIndexingStructure`. 
(As all logic was moved down to `SubstitutionTree`, we can now do the same things for terms we can do for literals, like for example get variable varints using `getVariants`.)

`TypeSubstitutionTree` does not exist anymore. It used to be used inside of `TermSubstitutionTree` in order to make it possible to associate variables with a sort inside of the tree. 
We can do the same still, but instead of using an additional tree, we do the following:
For a term `t` of sort `s`, we do not only insert `{ S0 -> t }` into the tree, but instead insert `{ S0 -> t, S1 -> s }`. 
Therefore the substitution tree itself takes over matching sorts.
Further it does the same not only for unification but automatically for Instance, Generalization and Variant retrievals.
Similar issues arise when inserting two-var equalities. Therefore for equalities `t1 = t2` of sort `s` we insert `{ S0 -> t1; S1 -> t2; S2 -> s}` into the tree.

Further I added a output operator for substitution trees, which is neat for debugging.
And example: 
```
SubstitutionTree& tree = ...;
out << tree << std::endl;
out << multiline(tree) <<  std::endl;
```
gives outputs like 
```
S1 -> $i ; S0 -> [ inverse(X0) | multiply(S2,S3) ; S2 -> [ a3 ; S3 -> b3 | multiply(a3,b3) ; S3 -> c3 ] | a3 | b3 | c3 ]
S1 -> $i ; S0 -> [
    | inverse(X0)
    | multiply(S2,S3) ; S2 -> [
        | a3 ; S3 -> b3
        | multiply(a3,b3) ; S3 -> c3
    ]
    | a3
    | b3
    | c3
]
```
(which is the repr of a tree that contains `{ inverse(X0); multiply(a3,b3); mulitply(multiply(a3,b3),c3); a3; b3; c3 }`)

Besides these changes in the implementation I made some optimizations:
We used to very often re-allocate memory in e.g. `FastInstancesIterator`: Whenever an object of this class is created, a couple of stacks and arrays of fixed size would be allocated, and destroyed again when  the iterator is destroyed. As we iterate very often this is quite undesirable. 
We can get around this issue using `Recycler`. In order to implement things more neatly, I introduced a smart pointer `Recycled` that uses `Recycler` to allocate an object and automatically returns it to the `Recycler` when the pointer is destroyed. 
Further a big performance issue was that our implementation of `DHMap` would allocate as soon as an object is created, even when the object is created on the heap, and nothing is ever inserted into it. This makes default constructor of other objects that contain `DHMap`, like RobSubstitution quite expensive. Therefore I changed this behaviour such that the map will only allocate as soon as the first object is inserted.

Some experiments martin ran for me with his snake experiment engine, and some experiments I ran in snake portfolio mode, show that we can solve more problems on average after this refactoring.


Indexing/TermSubstitutionTree.hpp
```diff
@@ -32,9 +31,12 @@ namespace Indexing {
  * Note that unlike LiteralSubstitutionTree, TermSubstitutionTree does
  * not (yet) carry out sort checking when attempting to find unifiers, generalisations
  * or instances. In particular, if the query or result is a variable, it is the callers'
- * responsibility to ensure that the sorts are unifiable/matchable
+ * responsibility to ensure that the sorts are unifiable/matchable 
+ * (edit: if the caller inserts a TypedTermList instead of a TermList, this will be handled automatically.)
```
MichaelRawson, line 35: Should we ever insert an untyped `TermList`, then?

Indexing/TermSubstitutionTree.hpp
```diff
@@ -54,77 +56,54 @@ class TermSubstitutionTree
    */
   TermSubstitutionTree(bool useC=false, bool replaceFunctionalSubterms = false, bool extra = false);
 
-  void insert(TermList t, Literal* lit, Clause* cls);
-  void remove(TermList t, Literal* lit, Clause* cls);
-  void insert(TermList t, TermList trm);
-  void insert(TermList t, TermList trm, Literal* lit, Clause* cls);
+  void handle(TypedTermList tt, Literal* lit, Clause* cls, bool insert);
```
MichaelRawson, line 59: A lot of these functions are now trivial and the implementation could be inlined here.

Indexing/TermSubstitutionTree.cpp
```diff
@@ -35,428 +24,48 @@ using namespace Lib;
...
-    return pvi(getConcatenatedIterator(getConcatenatedIterator(it1, it2), it3));
-  }
-}
-
-//TODO code sharing with getUnifications
-TermQueryResultIterator TermSubstitutionTree::getUnificationsWithConstraints(TermList t,
-          bool retrieveSubstitutions)
-{
-  CALL("TermSubstitutionTree::getUnificationsWithConstraints");
+template<class TypedOrUntypedTermList>
```
MichaelRawson, line 49: I can't figure out how this links, since we don't provide manual template specializations. But any hypothetical linkage problems would be fixed by lifting it to the header file.

Indexing/LiteralSubstitutionTree.hpp
```diff
@@ -36,134 +47,27 @@ class LiteralSubstitutionTree
...
-  template<class Iterator, class Filter>
-  SLQueryResultIterator getResultIterator(Literal* lit,
-	  bool complementary, bool retrieveSubstitutions, bool useConstraints);
-
-  unsigned getRootNodeIndex(Literal* t, bool complementary=false);
-  bool _polymorphic;
+  template<class Iterator>
+  SLQueryResultIterator getResultIterator(Literal* lit, bool complementary, bool retrieveSubstitutions, bool useConstraints);
+
+  Stack<unique_ptr<SubstitutionTree>> _trees;
```
MichaelRawson, line 69: Can this just be a `Stack<SubstitutionTree>`?

MichaelRawson: I know `getTree` returns a reference that would be invalidated on resize, but it seems like we shouldn't resize while the reference is live.

Indexing/SubstitutionTree.hpp
```diff
@@ -678,71 +733,362 @@ class SubstitutionTree
...
+  Leaf* findLeaf(BindingMap& svBindings)
+  { ASS(!_root || !_root->isLeaf() )
+    return _root ? findLeaf(_root, svBindings) : nullptr; }
 
   Leaf* findLeaf(Node* root, BindingMap& svBindings);
 
-  void insert(Node** node,BindingMap& binding,LeafData ld);
-  void remove(Node** node,BindingMap& binding,LeafData ld);
+  // TODO document
+  void setKey(TypedTermList const& term, LeafData& ld)
```
MichaelRawson, line 748: Yes please! I've no idea what the "key" is here.

Indexing/SubstitutionTree.hpp
```diff
@@ -678,71 +733,362 @@ class SubstitutionTree
...
+
+  protected:
+    static const unsigned BACKTRACK_SEPARATOR=0xFFFFFFFF;
+
+    struct Binder;
+    struct Applicator;
+    class Substitution;
+
+    Recycled<VarStack> _boundVars;
+    Recycled<DArray<TermList>, NoReset> _specVars;
```
MichaelRawson, line 982: Why `NoReset`?

Indexing/SubstitutionTree.hpp
```diff
@@ -678,71 +733,362 @@ class SubstitutionTree
...
+
+    /**
+     * @b nextSpecVar Number higher than any special variable present in the tree.
+     * 	It's used to determine size of the array that stores bindings of
+     * 	special variables.
+     */
+    template<class TermOrLit>
+    GenMatcher(TermOrLit query, unsigned nextSpecVar)
+      : _boundVars()
+      , _specVars()
```
MichaelRawson, line 923: These should happen by default for these as they are non-POD, right?

Indexing/SubstitutionTree.hpp
```diff
@@ -678,71 +733,362 @@ class SubstitutionTree
...
   {
   public:
-    FastGeneralizationsIterator(SubstitutionTree* parent, Node* root, Term* query,
-            bool retrieveSubstitution, bool reversed,bool withoutTop,bool useC, 
-            FuncSubtermMap* fstm = 0);
+    FastGeneralizationsIterator(FastGeneralizationsIterator&&) = default;
+    FastGeneralizationsIterator& operator=(FastGeneralizationsIterator&&) = default;
+    DECL_ELEMENT_TYPE(QueryResult);
+    /**
+     * @b nextSpecVar is the first unassigned special variable. Is being used
```
MichaelRawson, line 1058: `nextSpecVar` no longer exists here, right?

Indexing/SubstitutionTree.hpp
```diff
@@ -756,60 +1102,271 @@ class SubstitutionTree
...
-    FastInstancesIterator(SubstitutionTree* parent, Node* root, Term* query,
-	    bool retrieveSubstitution, bool reversed, bool withoutTop, bool useC, 
-      FuncSubtermMap* fstm = 0);
-    ~FastInstancesIterator();
+    FastInstancesIterator(FastInstancesIterator&&) = default;
+    FastInstancesIterator& operator=(FastInstancesIterator&&) = default;
+    DECL_ELEMENT_TYPE(QueryResult);
+
+    /**
+     * @b nextSpecVar is the first unassigned special variable. Is being used
```
MichaelRawson, line 1318: Same here.

Indexing/SubstitutionTree.hpp
```diff
@@ -836,83 +1393,94 @@ class SubstitutionTree
...
-     * where the order of special variables is reversed.
-     */
-    void createReversedInitialBindings(Term* t);
     bool findNextLeaf();
     bool enter(Node* n, BacktrackData& bd);
 
 
     static const int QUERY_BANK=0;
     static const int RESULT_BANK=1;
-    static const int NORM_QUERY_BANK=2;
```
MichaelRawson, line 866: Is `NORM_QUERY_BANK` unused? If so perhaps `NORM_RESULT_BANK` could equal 2?

Indexing/SubstitutionTree.cpp
```diff
@@ -52,14 +52,18 @@ using namespace Indexing;
  * Initialise the substitution tree.
  * @since 16/08/2008 flight Sydney-San Francisco
  */
-SubstitutionTree::SubstitutionTree(int nodes,bool useC, bool rfSubs)
-  : tag(false), _nextVar(0), _nodes(nodes), _useC(useC), _rfSubs(rfSubs)
+SubstitutionTree::SubstitutionTree(bool useC, bool rfSubs)
+  : _nextVar(0)
+  , _useC(useC)
+  , _functionalSubtermMap(rfSubs ? Option<FuncSubtermMap>(FuncSubtermMap()) : Option<FuncSubtermMap>())
```
MichaelRawson, line 58: Maybe this could be a new method for `Option<T>`?

Indexing/SubstitutionTree.cpp
```diff
@@ -72,10 +76,8 @@ SubstitutionTree::~SubstitutionTree()
   CALL("SubstitutionTree::~SubstitutionTree");
   ASS_EQ(_iteratorCnt,0);
 
-  for (unsigned i = 0; i<_nodes.size(); i++) {
-    if(_nodes[i]!=0) {
-      delete _nodes[i];
-    }
+  if (_root) {
```
MichaelRawson, line 79: My understanding is that `delete nullptr;` has no effect, so this should just be `delete _root`.

Indexing/SubstitutionTree.cpp
```diff
@@ -992,5 +862,91 @@ SubstitutionTree::NodeIterator
   }
 }
 
+template<class T>
```
MichaelRawson, line 865: Similar comments about a common header for this kind of output niceties.

MichaelRawson (changes requested):
Hey Joe, lots of comments inline but for the size of change they're relatively minor. Maybe we can meet in person sometime this week and go over the major changes in more detail, but I'm optimistic. :-)

Indexing/TermIndex.cpp
```diff
@@ -141,7 +118,7 @@ void DemodulationSubtermIndexImpl<combinatorySupSupport>::handleClause(Clause* c
       NonVariableNonTypeIterator,
       FirstOrderSubtermIt>::type it(lit);
     while (it.hasNext()) {
-      TermList t=it.next();
+      TermList t=TermList(it.next());
```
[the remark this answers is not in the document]
MichaelRawson: Ah, got it. We now have `Term *`s coming from iterators where they should be compound terms.

Indexing/Index.hpp
```diff
@@ -70,20 +80,30 @@ struct TermQueryResult
...
   TermQueryResult(TermList t, Literal* l, Clause* c, ResultSubstitutionSP s,UnificationConstraintStackSP con, bool b)
-  : term(t), literal(l), clause(c), substitution(s), constraints(con), isTypeSub(b) {}
+  : term(t), literal(l), clause(c), substitution(s), constraints(con) {}
 
   TermList term;
   Literal* literal;
   Clause* clause;
   ResultSubstitutionSP substitution;
   UnificationConstraintStackSP constraints;
-  bool isTypeSub = false; //true if the substitution only unifies the types of the terms
```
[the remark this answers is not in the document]
joe-hauns: So the thing is as sorts are terms the distinction does not really make any sense right? 
It was only needed  originallly as a work-around for using `TermSubstitutionTree` if I recall correctly.

ibnyusuf:
Looks like good work! I don't understand how the replacement to `typeSubstitutionTree` works. Say the query term is `union(a(Y), b(Y)) : set(Y) ` and I want to find all terms in the tree whose sort unifies with `set(Y)`, and create constraints between the result and the query, how does this work?

I have looked through the code, but can't see where this happens. Thanks

joe-hauns:
> Looks like good work! I don't understand how the replacement to typeSubstitutionTree works. Say the query term is union(a(Y), b(Y)) : set(Y) and I want to find all terms in the tree whose sort unifies with set(Y), and create constraints between the results and the query, how does this work?
> 
> I have looked through the code, but can't see where this happens. Thanks

So this happens as follows: we'd insert `{ S0 -> union(a(Y), b(Y)); S1 -> set(Y) }` into the tree, and to retrieve it by calling `tree.getUnificationsUsingSorts(TypedTermList(X, set(Y)), ...)` which will querying `{ S0 -> X, S1 -> set(Y) }` from the underlying SubstitutionTree.
I hope that answers your question!

The relevant pointers to the code base should be:
- https://github.com/vprover/vampire/blob/18cd7516e5a8c37516c509da9b10ab2c633d1f1d/Indexing/SubstitutionTree.hpp#L999
- https://github.com/vprover/vampire/blob/18cd7516e5a8c37516c509da9b10ab2c633d1f1d/Indexing/TermSubstitutionTree.hpp#L74

Forwards.hpp
```diff
@@ -70,6 +70,9 @@ typedef List<TermList> SList; // a list of sorts (which are now, with polymorphi
 typedef const SharedSet<unsigned> VarSet;
 
 typedef std::pair<std::pair<TermList,unsigned>,std::pair<TermList,unsigned>> UnificationConstraint;
+template<class A, class B>
+std::ostream& operator<<(std::ostream& out, std::pair<A,B> const& self)
+{ return out << "(" << self.first << ", " << self.second << ")"; }
```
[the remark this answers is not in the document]
joe-hauns: Yess good shout! I'll create a file `Debug/Output.hpp` for this sort of stuff.

Indexing/Index.hpp
```diff
@@ -61,6 +61,16 @@ struct SLQueryResult
...
 
+// TODO move somewhere else 
+template<class T>
+struct OutputPtr {
+  T* self;
+  friend std::ostream& operator<<(std::ostream& out, OutputPtr const& self)
+  { return self.self ? out << *self.self : out << "NULL"; }
+};
+template<class T>
+OutputPtr<T> outputPtr(T* self) { return { .self = self, }; }
```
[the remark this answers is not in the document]
joe-hauns: yes exactly the same as `std::make_pair` in order to be able to skip writing the type parameter.

Indexing/ResultSubstitution.hpp
```diff
@@ -139,61 +139,12 @@ class ResultSubstitution
...
   }
 
-  static ResultSubstitutionSP fromSubstitution(RobSubstitution* s,
-	  int queryBank, int resultBank);
-#if VDEBUG
-  virtual vstring toString(){ NOT_IMPLEMENTED; }
-#endif
+  static ResultSubstitutionSP fromSubstitution(RobSubstitution* s, int queryBank, int resultBank);
+  virtual void output(std::ostream& ) const = 0;
+  friend std::ostream& operator<<(std::ostream& out, ResultSubstitution const& self)
```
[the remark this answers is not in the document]
joe-hauns: Yess exactly!

Indexing/TermIndex.cpp
```diff
@@ -109,13 +91,8 @@ void SuperpositionLHSIndex::handleClause(Clause* c, bool adding)
...
     TermIterator lhsi=EqHelper::getSuperpositionLHSIterator(lit, _ord, _opt);
     while (lhsi.hasNext()) {
-      TermList lhs=lhsi.next();
-      if (adding) {
-	_is->insert(lhs, lit, c);
-      }
-      else {
-	_is->remove(lhs, lit, c);
-      }
+      // TODO get rid of cast here
```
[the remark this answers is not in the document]
joe-hauns: Yes. The thing is that we already had handle vs add/remove on a couple of layers. So I just made all the layers that were lower than the ones i touched use `handle`. 
I want to refactor the more high-level indexing interface some time soon as well. I'll sort this out then.

Indexing/TermIndex.cpp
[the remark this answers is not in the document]
joe-hauns: Yess exactly!

Indexing/TermIndexingStructure.hpp
```diff
@@ -29,16 +30,10 @@ class TermIndexingStructure {
...
-	  bool retrieveSubstitutions = true) { NOT_IMPLEMENTED; }
-  virtual TermQueryResultIterator getUnificationsUsingSorts(TermList t, TermList sort,
-    bool retrieveSubstitutions = true) { NOT_IMPLEMENTED; }  
-  virtual TermQueryResultIterator getUnificationsWithConstraints(TermList t,
-    bool retrieveSubstitutions = true) { NOT_IMPLEMENTED; }
-  virtual TermQueryResultIterator getGeneralizations(TermList t,
-	  bool retrieveSubstitutions = true) { NOT_IMPLEMENTED; }
-  virtual TermQueryResultIterator getInstances(TermList t,
-	  bool retrieveSubstitutions = true) { NOT_IMPLEMENTED; }
+  virtual TermQueryResultIterator getUnifications(TermList t, bool retrieveSubstitutions = true, bool withConstraints = false) { NOT_IMPLEMENTED; }
```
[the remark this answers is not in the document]
joe-hauns: True. I don't like the `NOT_IMPLEMENTED` pattern at all, but I wasn't bothered to try removing them. There's probably a comple of (hopefully minor) reshuffeling we have to do if we replace it by `= 0`.

Inferences/BackwardDemodulation.cpp
```diff
@@ -131,17 +131,24 @@ struct BackwardDemodulation::ResultFn
...
-      ASS(sub)
-      //rather than 0 and 1, we should use the constants delared in
-      //substitution tree
-      if(!sub->match(_eqSort, 0, qrSort, 1)){
+      if(!qr.substitution->matchSorts(_eqSort, qrSort)) {
         return BwSimplificationRecord(0);        
       }
     }
 
+    // TODO remove this old impl
```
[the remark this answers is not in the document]
joe-hauns: very true! `tryGetRobSubstitution` is eliminated now :)

Kernel/SortHelper.cpp
```diff
@@ -916,16 +918,12 @@ bool SortHelper::areImmediateSortsValidPoly(Term* t)
...
-#if VDEBUG
-      cout << "the term is " + t->toString() << endl;
-      cout << "the type of function " + env.signature->getFunction(t->functor())->name() + " is: " + type->toString() << endl;
-      //cout << "function name : "+ env.signature->getFunction(t->functor())->name() << endl;
-      //cout << "function name 2 :" + t->functionName() << endl;
-      cout << "error with expected " << instantiatedTypeSort.toString() << " and actual " << argSort.toString() << " when functor is " << t->functor() << " and arg is " << arg << endl;
-      ASSERTION_VIOLATION;
-#endif
-*/
+      DBGE(type->toString())
```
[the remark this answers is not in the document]
joe-hauns: Oooops i meant to delete this actually.

Kernel/Renaming.cpp
```diff
@@ -121,24 +121,49 @@ void Renaming::makeInverse(const Renaming& orig)
...
 }
 
+TypedTermList Renaming::normalize(TypedTermList l)
+{
+  CALL("Renaming::normalize(Literal*)");
+  if (l.isTerm()) {
+    return TypedTermList(normalize(l.term()));
+  } else {
+    Recycled<Renaming> n;
+    n->reset();
```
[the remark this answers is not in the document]
joe-hauns: Absolutely. Fixed it.

Kernel/EqHelper.hpp
```diff
@@ -24,6 +24,7 @@
 #include "Lib/DHSet.hpp"
 
 #include "Term.hpp"
+#include "Kernel/BottomUpEvaluation/TypedTermList.hpp"
```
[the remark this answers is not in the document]
joe-hauns: indeed it is not :)

Kernel/BestLiteralSelector.hpp
```diff
@@ -261,14 +261,19 @@ class CompleteBestLiteralSelector
...
+    // it is definitely dead/unnecessary code at the time of refactoring
+    //
+    // FirstOrderSubtermIt fsi(lit);
+    //
+    // while(fsi.hasNext()){
+    //   TermList t = TermList(fsi.next());
+    //   if(t.isVar() && maxTermHeads->contains(t.var())){
+    //     return false;
+    //   }
+    // }
```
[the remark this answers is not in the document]
joe-hauns: @ibnyusuf Any thoughts on this one?

Indexing/SubstitutionTree.hpp
```diff
@@ -54,47 +58,103 @@ using namespace Lib;
 using namespace Kernel;
 
 #define UARR_INTERMEDIATE_NODE_MAX_SIZE 4
+template<class T>
+struct OutputMultiline
+{ T const& self; };
+
+template<class T>
+OutputMultiline<T> multiline(T const& self)
+{ return { self }; }
```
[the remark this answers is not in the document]
joe-hauns: I moved it into `Debug/Output.hpp`

Indexing/SubstitutionTree.hpp
```diff
@@ -54,47 +58,103 @@ using namespace Lib;
...
+
+template<class Key> struct SubtitutionTreeConfig;
 
 /**
- * Class of substitution trees. In fact, contains an array of substitution
- * trees.
- * @since 16/08/2008 flight Sydney-San Francisco
+ * Class of substitution trees. 
+ *
+ * We can either typed terms, or literals into a subtitution tree.
```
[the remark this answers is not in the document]
joe-hauns: fixed

Indexing/SubstitutionTree.hpp
```diff
@@ -54,47 +58,103 @@ using namespace Lib;
...
+  friend class IterCounter;
 public:
+  static constexpr int QRS_QUERY_BANK = 0;
+  static constexpr int QRS_RESULT_BANK = 1;
   CLASS_NAME(SubstitutionTree);
   USE_ALLOCATOR(SubstitutionTree);
 
-  SubstitutionTree(int nodes,bool useC=false, bool rfSubs=false);
-  ~SubstitutionTree();
+  SubstitutionTree(bool useC, bool rfSubs);
```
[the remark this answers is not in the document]
joe-hauns: In the branch I'm currently working on (`joe-uwa-refactor`) both these flags will be completely eliminated. So I think we can just leave them as is for now.

Indexing/SubstitutionTree.hpp
joe-hauns: The reason for this is that we do not need to, as we override all values that will be read anyways. Therefore we will safe time by not resetting. This behaviour was there before, I just implemented it using `Recycled`.

joe-hauns: *I added a comment

Indexing/SubstitutionTree.hpp
joe-hauns: Yess true, because we don't normalize variable names of queries anymore (as this is unnecessary work that does not help us). 
I think I didn't change the constants because I'm a bit concerned that someone might check `if (bank == 3)` or similar somewhere and I didn'
t wanna look for bugs like this haha

ibnyusuf:
Now that `TypedTermList` plays a larger role in the code than it originally did, would it be an idea to move the class to a different location?

joe-hauns:
> Now that `TypedTermList` plays a larger role in the code than it originally did, would it be an idea to move the class to a different location?

yes absolutely! I actually planned on doing so and forgot and michael also pointed it out to me yesterday!

Kernel/ApplicativeHelper.hpp
```diff
@@ -73,6 +74,18 @@ class ApplicativeHelper {
   static bool isExactApplied(TermList head, unsigned argNum);
   static bool isOverApplied(TermList head, unsigned argNum);
   static bool isSafe(TermStack& args);
+  // TODO is this right?
```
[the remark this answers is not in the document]
joe-hauns: @ibnyusuf What do you think of this?

Indexing/SubstitutionTree.hpp
```diff
@@ -109,6 +169,26 @@ class SubstitutionTree
...
+    UnificationConstraintStackSP constr;
+
+    QueryResult(LeafData const& ld) : data(&ld), subst(), constr() {};
+    QueryResult(LeafData const& ld, ResultSubstitutionSP subst, UnificationConstraintStackSP constr) : data(&ld), subst(subst), constr(constr) {}
+  };
+
+  using QueryResultIterator = VirtualIterator<QueryResult>;
+  // TODO make const function
+  template<class Iterator, class TermOrLit> 
+  VirtualIterator<QueryResult> iterator(TermOrLit query, bool retrieveSubstitutions, bool withConstraints, bool reversed = false)
```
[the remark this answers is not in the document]
joe-hauns: Changed the QueryResultIterator thing. Making as const verison requires touching a looot of code .. so I'll leave the TODO for now ;)

Indexing/TermSubstitutionTree.hpp
MichaelRawson: Joe says that this is legal, and that the "expected thing" happens. He may consider changing the interface at some later point to ensure that a sort is always inserted explicitly.

Indexing/SubstitutionTree.hpp
```diff
@@ -54,47 +58,103 @@ using namespace Lib;
...
 public:
+  static constexpr int QRS_QUERY_BANK = 0;
+  static constexpr int QRS_RESULT_BANK = 1;
   CLASS_NAME(SubstitutionTree);
   USE_ALLOCATOR(SubstitutionTree);
 
-  SubstitutionTree(int nodes,bool useC=false, bool rfSubs=false);
-  ~SubstitutionTree();
+  SubstitutionTree(bool useC, bool rfSubs);
+  SubstitutionTree(SubstitutionTree const&) = delete;
```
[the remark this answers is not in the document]
MichaelRawson: This should not be allowed because we don't want to copy substitution trees. But we might want to _move_ them, so Joe will do that.

Kernel/ApplicativeHelper.hpp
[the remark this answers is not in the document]
ibnyusuf: Short answer: no it can't, this is fine.

Long answer: we are incomplete either way, and I doubt that it would help. I am also unsure whether it would bring us any 'closer' to completeness.

Kernel/BestLiteralSelector.hpp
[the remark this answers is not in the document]
ibnyusuf: Yes, this is a higher-order concern. If you look at the calls to the function, they are only made when `combSup` is `true`, so first-order proof search should not be affected. 

In the higher-order calculus, due to some wierd requirement of the completeness proof, we should not be selecting a literals ` s = t` if `s` has  a variable head ` x` and `x` also occurs unapplied in the literal (or clause, I dont remember exactly!)

In any case, if you feel confident that it shouldn't interact with first-order stuff, it would be good if the code could be reset to old version.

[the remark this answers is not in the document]
MichaelRawson: Ahmed - maybe I misunderstood you, but I think Joe's saying this should be equivalent to `return true`, because the `FirstOrderSubtermIt` returns only non-variable terms. I think this is true even in the HOL case - should this use a different iterator or something instead?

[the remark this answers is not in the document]
ibnyusuf: Which is of course 100% true. This means that the code above has never carried out the role taht I thought it did. Let's leave it commented out for now until I find the time to decide what to do here. Thanks for bringing this to light!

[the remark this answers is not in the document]
joe-hauns: okidoke then let's leave it as is for now :)

MichaelRawson (approved):

MichaelRawson:
Hey @joe-hauns - any luck getting the CI to pass, or should I take a look into it?

joe-hauns:
well so the last fix was to just push another unrelated commit so the CI was run again and the test passed.
I've just been doing some checks on my local machine, and the test passes in 100 out of 100 repetitions..
Could you maybe check whether it works out on yours?

MichaelRawson:
Ah, right - yeah it looks fine to me, I've re-triggered the CI.

ibnyusuf:
Just pushed a change that enforces typedness of terms inserted into the tree and query terms. This simplifies the interface as we no longer have `getUnifications` and `getUnificationsUsingSort`. It also protects against subtle bugs. For example, if someone was to insert terms into a tree from two places, some with types, some without, then querying would always return a subset of the correct set.

The one place where I had difficulty in updating was in` /Indexing/AcylicityIndex.cpp`. I made the assumption, that we only insert terms (and never variables) into the index `_tis`. However, I am not confident that this assumption is valid. If someone has more knowledge of the code could confirm / reject that would be good. Alternatively, I could program defensively on the assumption that there could be variables, but the code may be a little slower in that case.

joe-hauns:
> Just pushed a change that enforces typedness of terms inserted into the tree and query terms. This simplifies the interface as we no longer have `getUnifications` and `getUnificationsUsingSort`. It also protects against subtle bugs. For example, if someone was to insert terms into a tree from two places, some with types, some without, then querying would always return a subset of the correct set.
> 
> The one place where I had difficulty in updating was in` /Indexing/AcylicityIndex.cpp`. I made the assumption, that we only insert terms (and never variables) into the index `_tis`. However, I am not confident that this assumption is valid. If someone has more knowledge of the code could confirm / reject that would be good. Alternatively, I could program defensively on the assumption that there could be variables, but the code may be a little slower in that case.

Great thanks a lot! I was planning on doing that but never got around to it! Thankss :)

Indexing/TermIndex.cpp
```diff
@@ -177,11 +143,13 @@ void DemodulationLHSIndex::handleClause(Clause* c, bool adding)
   Literal* lit=(*c)[0];
   TermIterator lhsi=EqHelper::getDemodulationLHSIterator(lit, true, _ord, _opt);
   while (lhsi.hasNext()) {
+    auto lhs = lhsi.next();
+    TypedTermList tt(lhs, SortHelper::getTermSort(lhs, lit));
```
joe-hauns, line 147: This is very expensive. We should change getDemodulationLHSIterator to retur a TypedTermList straight away. 
The same pattern is used in many places.

Inferences/BackwardDemodulation.cpp
```diff
@@ -76,13 +76,15 @@ struct BackwardDemodulation::RemovedIsNonzeroFn
 
 struct BackwardDemodulation::RewritableClausesFn
 {
-  RewritableClausesFn(DemodulationSubtermIndex* index) : _index(index) {}
+  RewritableClausesFn(DemodulationSubtermIndex* index, Literal* lit) : _index(index), _lit(lit) {}
   VirtualIterator<pair<TermList,TermQueryResult> > operator() (TermList lhs)
   {
-    return pvi( pushPairIntoRightIterator(lhs, _index->getInstances(lhs, true)) );
+    TermList sort = SortHelper::getTermSort(lhs, _lit);
```
ibnyusuf, line 82: @joe-hauns , the above displays the same inefficiency you spotted in `TermIndex`. Whatever solution you apply there, should be applied here as well.

joe-hauns: Sorted them all out.

joe-hauns:
I made some final changes, sorting out the inefficient calls to `getTermSort`. 
I got rid of *all* call sites of the function `getTermSort` (none of them were necessary, in most cases they were inefficent). 
Further i deprecated the function with a comment why, and what alternatives to use. Please have a look and lemme know what you think.

MichaelRawson:
Nice job @joe-hauns! I think we're running some more perf tests and then we should be good for merge.

MichaelRawson:
@joe-hauns - If I read @quickbeam123's message on Slack right, this is approved. :-) Can you fix the merge conflicts (sorry) and then we can finally get this in?

MichaelRawson:
@quickbeam123 - this one seems ready, objections? I know you said you'd have a glance over it.

quickbeam123:
Hi @joe-hauns, since you, through your work on this PR, became an expert on `Recycler`/`Recycled`, did you get some intuitions on when this is a good class to use and when not? (Maybe, we should have some guidelines to go into the comments next to the basic explanations.) (Maybe not all the original uses of the `Recycler` were such a good idea?)
In particular, I am thinking of:
- The pool of recycled instances is effectively leaked the moment it is clear there will be no demand for that particular class recycling in the future. From this perspective, shouldn't there be a requirement that we can reasonably expect the class to be always reused in the future (i.e., call site within a saturation loop)? Or, shouldn't there be a way to say "Dear recycled<my_type>, you are dismissed, please free all your pointers?"
- This simple design is made to work on a simple "per base type" basis (well, there does not seem to be a simple meaningful way around it). But then, do we expect people to reason along the lines like "OK, I am here, in my little preprocessing routine, making some nice use of, say, Recycled<Stack<unsigned>>. On a typical problem instance, this will be left with 5 stacks of capacity 1000 each. Someone in saturation is using Recycled<Stack<unsigned>> too, so this will get nicely reused." But maybe there in saturation, they will ever need just one such stack, and at most of size 25?
As an example, I looked at the `Recycled<DHSet<unsigned>> _used;` in `Rectify::Renaming` and I am not all sure this is a good place to use it. Assuming there being just one instance of `Rectify`, which will only need one instance of `Renaming`, we will simply remember the `DHSet<unsigned>` that got allocated through this forever, and never use it again (there is no other `Recycled<DHSet<unsigned>>` in the codebase, if I am not mistaken.

In an appeal to the principle: "Since this is being touch anyway, let's make it better (better documented, more foolproof, etc)", I hereby declare a discussion on this above points open!

MichaelRawson:
Let's discuss! But I'd tend towards "unless shown by profiling, just use automatic storage and be done with it" to avoid exactly these issues you mention. i.e.

```
void frobnicate() {
  Frobber frob1; // this
  static Frobber frob2; // but not this
  Recycled<Frobber> frob3; // or this
}
```

I am not sure of the performance characteristics of 2 versus 3. Since `Recycler` exists, I assume 2 was deemed insufficient at some point? `Recycled` also automatically handles the reset logic - which is good from a maintainability point of view but doesn't help with performance.

If there actually _is_ something that should be recycled somehow, but still has these issues, we can deal with it then.

easychair:
Is there a massive save or just 1 microsecond in a 1 minute-run?

A

joe-hauns:
> did you get some intuitions on when this is a good class to use and when not?
Tbf i did not really. What I mainly did is to replace all old call sites of `Recycler` and replace them by `Recycled`, expecting that whoever used it there knew why. 
Additionally the places where i put it I figured out by benchmarking and having a look at the runtime profile. And @easychair , yes in some cases there was a massive gain by using recycler. This was mostly for unification and the substitution tree iterators, which would be called *very* often during staturaiton and basically spend more than 90% of their time allocating fresh `Stack<TermList> todo` and similar things. 

> Dear recycled<my_type>, you are dismissed, please free all your pointers?"
You're raising a good point here @quickbeam123 . Especially, as you said when it comes to preprocessing. We could add a fuction `Recycler::cleanup` or so where it just deletes all the pointers that are currently held. The question is where exactly that should be done, and how we make sure we do it for exactly the types we want it to be done for.

```
void frobnicate() {
  Frobber frob1; // this
  static Frobber frob2; // but not this
  Recycled<Frobber> frob3; // or this
} 
```
Concerning @MichaelRawson 's example: (2) is definitely the fastest (this is what i saw from benchmarking), but it has two down-sites: firstly it is not re-entrant safe (imagine having some convoluted recursive call to the function and both calls start to act on the same `from2` stack), and it *always* leaks, and doesn't share the allocation with other functions. So I'd always prefer `Recycled` over `static` as it's just quite a lot safer, and easier to use, but still gives a big performance boost in cases where it's needed.

> unless shown by profiling, just use automatic storage and be done with it
By "use automatic storage" you mean (1), right? 
I'd agree with that, but maybe adding "if you know your code is hot during saturation" use (3).

Kernel/SortHelper.hpp
```diff
@@ -65,6 +65,7 @@ class SortHelper {
 
   static bool tryGetVariableSort(unsigned var, Formula* f, TermList& res);
   static TermList getVariableSort(TermList var, Term* t);
+  [[deprecated("This function is usually only used if we loose the information about the sort of a variable somewhere while over subterms. Recovering the information using this method iterating the literal/term again, is very inefficient and should be avoided. Make sure to use TermIterators that return TypedTermList instead, or raise a discussion on slack/github if you have a use case where this function is *really* needed.")]]
```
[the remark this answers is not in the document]
joe-hauns: I just checked and I'm not getting any warnings.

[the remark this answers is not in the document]
MichaelRawson: Does this suggest that the evil function can be deleted? If so I'm in favour of that.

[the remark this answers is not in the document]
joe-hauns: From my side that's fine. The reason why I did not just do that was to make merging easier for people relying on this function in waiting PRs. So whatever you are think is best!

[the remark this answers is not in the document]
MichaelRawson: Ah, good point. OK, then I'll schedule it for deletion some other time. :-)

MichaelRawson:
>By "use automatic storage" you mean (1), right?

Yes, I'm probably out of touch with what this is called in modern C++ lingo.

Generally I agree with @joe-hauns' assessment. If some allocation is _really_ hot (unification?) it might be worth using a class-level or compilation-unit `static` rather than any of the other options as it also avoids the first entry is-it-initialised check (which we now compile without compiler-inserted thread-safety locks, but it's still a non-zero cost). Hard to measure in something as big as Vampire though.

@quickbeam123 - I think this is ready to go? Discussions about recycling are a good idea but I think orthogonal to this.

quickbeam123:
OK, so why don't we proceed by you (@MichaelRawson or @joe-hauns) adding one more paragraph of comments to `Recycled`, explaining the intended use case (when the benefits should be the greatest) and the potential pitfalls (leaking the resource) and I will double check its use in `Rectify::Renaming` and possible eliminate it from there. 

Then I keep on checking the rest of the code ...

Indexing/ResultSubstitution.hpp
```diff
@@ -133,67 +133,12 @@ class ResultSubstitution
    */
   virtual bool isIdentityOnResultWhenQueryBound() {return false;}
 
-  virtual RobSubstitution* tryGetRobSubstitution() { return 0; }
```
quickbeam123, line 136: Could we please keep `tryGetRobSubstitution` as it was? I am using it in https://github.com/vprover/vampire/tree/dump_proof_details when generating `proof_extra` for Anja

quickbeam123:
Could we have a different naming scheme for the hpp files where `BottomUpEvaluation` specializations live?

In particular, my IDE gets confused about there being both
`Kernel/BottomUpEvaluation/TypedTermList.hpp`
`Kernel/TypedTermList.hpp`

Also `Kernel/BottomUpEvaluation/TermList.hpp` is only getting away, since we have not `Kernel/TermList.hpp`, but maybe the pattern should be fixed uniformly...

joe-hauns:
> Could we have a different naming scheme for the hpp files where `BottomUpEvaluation` specializations live?
> 
> In particular, my IDE gets confused about there being both `Kernel/BottomUpEvaluation/TypedTermList.hpp` `Kernel/TypedTermList.hpp`
> 
> Also `Kernel/BottomUpEvaluation/TermList.hpp` is only getting away, since we have not `Kernel/TermList.hpp`, but maybe the pattern should be fixed uniformly...
> 
> Alternatively, how about glueing `Kernel/TypedTermList.hpp` under `Kernel/Term.hpp`?

Fine with me. As long as it works with the `#include` dependencies.

joe-hauns:
Or what naming scheme do you have in mind?
Maybe `BottomUpEvaluation_TypedTermList.hpp` instead of `BottomUpEvaluation/TypedTermList.hpp`?

Indexing/ResultSubstitution.hpp
quickbeam123: Actually, if there is a principled solution under way, we don't need to rush it. (The proof_details branch may stay where it is---under current master---until we get around to merging `cleanup-result-substitution`.)

However, let me state what we need there. We don't care about getting a `RobSubstitution` per se. We care about being able to print the bindings (a part of the substitution) relevant, separately, for the query and the result term. Ideally, I would like `ResultSubstitution` to offer such functions. (although I can imagine going through the pain of collecting the variables from, e.g., the query, supplying them as arguments to a dummy term created just for the occasion, calling the respective apply function and reading of the result.) Any thoughts/ideas on this?

quickbeam123:
I must admit, thinking about it a bit more, the long version looks quite ugly.

@MichaelRawson, any suggestions? (Keeping things as they are with both `Kernel/BottomUpEvaluation/TypedTermList.hpp` and `Kernel/TypedTermList.hpp` counts as suggestion too.)

Thank you, by the way, for the nice explanation in `Recycler`!

MichaelRawson:
> any suggestions?

How about merging all of them into `BottomUpEvaluation`? Normally I'd argue that having more files is better (reduces compiler load if you don't pay for what you don't use) but there's probably not _that_ many recursive term-like things to BUE - currently exactly 3 - and they're all quite short implementations.

> Thank you, by the way, for the nice explanation in Recycler!

:-) Shame I goofed! Fixed now.

Indexing/ResultSubstitution.hpp
joe-hauns: Well so I think it's not an issue to implement this for other substitutions as well. All of them store all the bound variables in some sort of hashmap, anyways so we can easily iterate over that map and then use the relevant apply function as u say. As you need this only in proof outputing i guess there's not a big issue about performance, right?

But in general there's quite a lot of issues arount the `ResultSubstitution` interface, so i wanna completely delete it and get a new one that is more planned and less grown. But we can definitely introduce a function that allows for something like that easily in a new one.

joe-hauns:
> How about merging all of them into BottomUpEvaluation? Normally I'd argue that having more files is better (reduces compiler load if you don't pay for what you don't use) but there's probably not that many recursive term-like things to BUE - currently exactly 3 - and they're all quite short implementations.

This was how it was originally supposed to be but the issue is that this doesn't work out with `#include` dependencies. 
We could also name them `Kernel/BottomUpEvaluation/implTypedTermList.hpp` or something like that .. 
(As i introduced the original naming scheme i don't mind leaving it as is ofc, but if we can easily avoiding confusing IDEs that's good with me as well)

MichaelRawson:
>the issue is that this doesn't work out with #include dependencies

Right - you mentioned this but I wasn't smart enough this morning to figure out what you meant. Can you explain a bit more?

joe-hauns:
Wellll tbh i don't quite understand the exact issue, but I know as soon as i start doing that (what i tried out again just now), some things need to move files, imports need to shuffle ordre and such things. I don't have time rn to look into it properly, but I can try to untangle things another time if this is what we want to do. (and of course you can give it a try as well if u fancy ;) )

MichaelRawson:
I did it - making it work was not too bad I think. Revert if you hate it!

joe-hauns:
> I did it - making it work was not too bad I think. Revert if you hate it!

No objections :)

MichaelRawson:
CI hilariously uses the script from the current master, not the one I just pushed. I think it's just that the induction example fails - @joe-hauns, if you suspect something is up please look, otherwise I'll probably just nuke the failing check on master. We should only have really trivial things in the sanity-check script otherwise it becomes noise.

Indexing/ResultSubstitution.hpp
quickbeam123: Perfect, let's talk about the details later!

Indexing/ClauseCodeTree.cpp
```diff
@@ -258,23 +256,21 @@ void ClauseCodeTree::remove(Clause* cl)
...
+  firstsInBlocks->push(op);
   unsigned depth=0;
   for(;;) {
-    RemovingLiteralMatcher* rlm;
-    Recycler::get(rlm);
-    rlm->init(op, lInfos.array(), lInfos.size(), this, &firstsInBlocks);
-    rlms.push(rlm);
+    rlms->push(Recycled<RemovingLiteralMatcher, NoReset>());
+    RemovingLiteralMatcher* rlm = &*rlms->top();
+    rlm->init(op, lInfos.array(), lInfos.size(), this, &*firstsInBlocks);
```
quickbeam123, line 265: Around here, I would argue, things are getting pretty ugly. 

Not sure what to do about it, as the key gain overall was to REPLACE `Recycler` by `Recycled`, but could we give it a thought?

The nested pattern declared slightly above seems to be the root of this.

I am getting some headaches trying to figure out what would need to be done, if anything at all, to first get the `rlm` in some form, initialize it and only then push it to `rlms` (as in the original code).

(The `&*` is also confusing, but I am guessing this needed to get a reference, rather than a pointer, to whatever is stored inside the `Recycled` thing?)

Indexing/ClauseCodeTree.cpp
```diff
@@ -258,23 +256,21 @@ void ClauseCodeTree::remove(Clause* cl)
...
 
   iteration_restart:
     if(!rlm->next()) {
       if(depth==0) {
 	ASSERTION_VIOLATION;
 	INVALID_OPERATION("clause to be removed was not found");
       }
-      Recycler::release(rlms.pop());
```
quickbeam123, line 275: Why don't we need the `pop` anymore?

Indexing/ClauseCodeTree.cpp
```diff
@@ -590,7 +579,7 @@ Clause* ClauseCodeTree::ClauseMatcher::next(int& resolvedQueryLit)
   }
 
   for(;;) {
-    LiteralMatcher* lm=lms.top();
+    Recycled<LiteralMatcher, NoReset>& lm = lms.top();
```
quickbeam123, line 583: Is there an advantage of using `Recycled&` here. If we just want to use the object, the `&*` could be used, right?

Indexing/ClauseCodeTree.cpp
```diff
@@ -731,22 +720,20 @@ void ClauseCodeTree::ClauseMatcher::enterLiteral(CodeOp* entry, bool seekOnlySuc
     linfoCnt/=2;
   }
 
-  LiteralMatcher* lm;
-  Recycler::get(lm);
+  Recycled<LiteralMatcher, NoReset> lm;
   lm->init(tree, entry, lInfos.array(), linfoCnt, seekOnlySuccess);
-  lms.push(lm);
```
quickbeam123, line 737: This pattern I understand!

(also, it's reassuring that `lm` goes out of scope immediately after the move)

Indexing/ClauseCodeTree.cpp
joe-hauns: > (The &* is also confusing, but I am guessing this needed to get a reference, rather than a pointer, to whatever is stored inside the Recycled thing?)


Yes exactly. If we have `t: Recycled<T>` then `*t: T&` and `&*t: T*`. That's needed because of the signature of `RemovingLiteralMatcher::init`

I agreee that this part of the code is not very nice, but it was already not nice before I touched it. I just refactored it to keep the same behaviour. 

> I am getting some headaches trying to figure out what would need to be done, if anything at all, to first get the rlm in some form, initialize it and only then push it to rlms (as in the original code).

So you could do it like this for example.
```
    Recycled<RemovingLiteralMatcher, NoReset> rlm;
    rlm->init(...);
    rlms->push(std::move(rlm));
```
but then we cannot use the reference to rlm later in the code, as `Recycled` has the same ownership semantics as `unique_ptr`, that is, if it's moved the thing it's being moved out of is invalidated (to ensure it's not being double-freed and so.

Indexing/ClauseCodeTree.cpp
joe-hauns: hmmmm interesting .. this is indeed strange, and probably a bug. I'm very surprised that this didn't show up in our test runs .. is this code actually being executed in them? 🤔

Indexing/ClauseCodeTree.cpp
joe-hauns: Yes exactly! Doesn't make any difference here.

Indexing/ClauseCodeTree.cpp
MichaelRawson: Some of the code tree stuff is known to be dead code, so it could just not be. Jakob was thinking about reviving it, but I guess with his recent direction probably not.

MichaelRawson:
I looked into the CI failed induction problem. We can solve it with the same strategy that master uses, so no problems here.

Indexing/ClauseCodeTree.cpp
quickbeam123: OK, fixing...

PR #433, opened 2023-02-16
Fix show_fool option

ibnyusuf:
This PR should fix the `show_fool` option that had been broken by earlier higher-order work. Terms of Boolean sort are now shown as being of sort `$bool` if `show_fool` is on. Along the way, the following improvements have been made:

- Distinction between predicate and function types has been improved
- Commented out code from UIHelper has been removed
- Types for unary functions and predicates no longer add unnecessary brackets around the single argument type when printing

### DIFF
--- a/Kernel/OperatorType.hpp
+++ b/Kernel/OperatorType.hpp
@@ -94,6 +94,11 @@ private:
 
   static OperatorType* getTypeFromKey(OperatorKey* key, unsigned taArity);
 
+  static TermList getEmpty() { 
+    static TermList empty(0,false);
+    empty.makeEmpty();
+    return empty;
+  }
   //static const TermList PREDICATE_FLAG;
 
 public:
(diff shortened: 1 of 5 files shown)


Kernel/OperatorType.hpp
```diff
@@ -94,6 +94,11 @@ class OperatorType
 
   static OperatorType* getTypeFromKey(OperatorKey* key, unsigned taArity);
 
+  static TermList getEmpty() { 
+    static TermList empty(0,false);
```
quickbeam123, line 98: I find the static here a bit confusing, since we are returning by value. But it will work, so no big deal...

quickbeam123:
Also, could you please remind me: boolSort's typeConName() is "$o" or what do we print if showFOOL() is false?

Kernel/OperatorType.hpp
MichaelRawson: Yes - also I keep wanting a constructor for this in `TermList`, maybe it can go there? But also please kill the `static`, this `TermList empty; empty.MakeEmpty(); return empty;` should be inlined and optimised to "move bytes representing empty-term into some destination" which is what we want.

ibnyusuf: Right, good point! I knew I was doing something silly here.

ibnyusuf:
> Also, could you please remind me: boolSort's typeConName() is "$o" or what do we print if showFOOL() is false?

If `showFOOL()` is false we print `$o`

quickbeam123:
Cool! Thank you, Ahmed!

PR #436, opened 2023-03-22
added option --ignore_unrecognized_logic

joe-hauns:
I added an option that lets us attempt proof search even though we do not recognize a logic in an SMTLIB file.
Currently vampire just throws an error in that case. 
Motivation is that i for example got some benchmarks from `QF_UFDTNIA` and wanted to see what my branch does for them without adding all the code to properly support that logic.
I ran into issues like that quite often before and always just renamed the logic in all the smt files to something vampire knows. I think this is a nicer solution, and will also be nice for users who don't know vampire.
When vampire dies now because of an unrecognized logic it will also give the hint to use the option `--ignore_unrecoginized_logic on` to try proof serach anyways.

### DIFF
--- a/Lib/Exception.hpp
+++ b/Lib/Exception.hpp
@@ -214,8 +214,8 @@ class NotImplementedException
 
 #define VAMPIRE_EXCEPTION \
   throw Lib::Exception(__FILE__,__LINE__)
-#define USER_ERROR(msg) \
-  throw Lib::UserErrorException(msg)
+#define USER_ERROR(...) \
+  throw Lib::UserErrorException(__VA_ARGS__)
 #define INVALID_OPERATION(msg) \
   throw Lib::InvalidOperationException(msg)
 #define SYSTEM_FAIL(msg,err) \
--- a/Parse/SMTLIB2.cpp
+++ b/Parse/SMTLIB2.cpp
@@ -501,7 +501,11 @@ void SMTLIB2::readLogic(const vstring& logicStr)
   case SMT_UFBV:
     USER_ERROR("unsupported logic "+logicStr);
   default:
-    USER_ERROR("unrecognized logic "+logicStr);
+    if (env.options->ignoreUnrecognizedLogic()) {
+      break;
+    } else {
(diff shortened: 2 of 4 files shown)


MichaelRawson:
This would be great to have! I ran into this the other day when I wanted to run SMTCOMP schedules against a TPTP problem - which I agree is weird but I think legitimate.

This would leave the SMTLIB logic property blank/unitialised, right? There is at least one bug then, in `CASC/Schedules.cpp` where there's some switch/case statements without a default case. If you fix it or persuade me this isn't a bug I'm happy to merge, but let's think about some alternatives in the meantime:
1. If we don't specify a logic (or we don't recognise the specified logic), select the "strongest" (most features) logic that Vampire supports and issue a warning.
2. Allow specifying the SMTLIB logic via a flag, instead of specifying that we ignore it.

This is all based on the assumption that SMTLIB "logics" are just (fragments of) combinations of theories and don't carry semantics beyond that...which I think is true for now.

My preference would be for (1), but I'm happy to hear arguments one way or another.

selig:
Option 3. Add a logic ALL to the schedules where we ask spider for a schedule for smt problems ignoring logic. If no logic is given then default to this schedule. This is (effectively) what CVC5 does.

quickbeam123:
Yes, we need a "general purpose" (ALL) schedule from Spider to fill in the `default` path in the `smtcomp` schedule.

joe-hauns:
Yeah I'm also much in favor for giles' suggestion. 
But for now (until we have that schedule) we could just set the logic to `AUFDTNIRA` in this case? 
This is the strongest one we support so far right?

MichaelRawson:
>This is the strongest one we support so far right?

I believe so. Maybe we should nonetheless have an `ALL` value that is just treated the same way as `AUFDTNIRA`, since it's possible we support something stronger in future.

Arguably there is another option, but this will take some effort.
4. Ignore `(set-logic ...)` completely and use the existing property-scanning machinery to determine what the logic "actually" is. Maybe even re-do schedules based on this, like we do for TPTP, rather than dispatching based on the declared logic.

PR #438, opened 2023-03-27
distinct objects always have sort `$i` in TPTP

MichaelRawson:
Consider
```
fof(test, conjecture, "foo" != "bar").
```

Current master complains that we didn't set a type for "foo" (or "bar"). According to https://tptp.org/TPTP/TR/TPTPTR.shtml#FormulaeSection we should set this to `$i` regardless of the type context, and this is confirmed if you look at e.g. `SYN000^2`. Fix this.

### DIFF
--- a/Parse/TPTP.cpp
+++ b/Parse/TPTP.cpp
@@ -2953,6 +2953,10 @@ void TPTP::term()
       switch (tok.tag) {
         case T_STRING:
           number = env.signature->addStringConstant(tok.content);
+          // "distinct_object"s are _always_ of sort $i, even in typed contexts
+          env.signature->getFunction(number)->setType(
+            OperatorType::getConstantsType(AtomicSort::defaultSort())
+          );
           break;
         case T_INT:
           number = addIntegerConstant(tok.content,_overflow,_isFof);


Parse/TPTP.cpp
```diff
@@ -2953,6 +2953,10 @@ void TPTP::term()
       switch (tok.tag) {
         case T_STRING:
           number = env.signature->addStringConstant(tok.content);
+          // "distinct_object"s are _always_ of sort $i, even in typed contexts
+          env.signature->getFunction(number)->setType(
```
ibnyusuf, line 2957: Wouldn't it be more logical to set the type within the `addStringConstant` function, as there is never a case where we want a string constant without a type? Or would this have some unexpected implications for SMT?

MichaelRawson: I thought I replied to this earlier, but no sign of it in GitHub. Apologies if duplicate.

This is what I did initially, but on reflection it seemed that the signature method just adds a "string" constant (whatever that means), whereas if we want TPTP semantics we should imbue it at the TPTP call site. But I have no strong opinion and can just move it around if you prefer the other way. No unexpected implication either way for SMTLIB as far as I know.

ibnyusuf: This has already been merged, but for what it is worth, I think it is safer to set the type in `addStringConstant`. We don't want someone coming along and calling `addStringConstant` from some other part of the code, not realising that it doesn't set the type, and then blythely causing a crash again.

PR #439, opened 2023-03-28
kill off `env.property`, replace it with some "legacy" properties

MichaelRawson:
`env.property` contains the "current problem"'s properties. However, in Vampire it is possible to have more than one instance of a `Problem`, e.g. by preprocessing steps, recursive vampirism (`FMB/FunctionRelationshipInference`), or copying the problem for some reason (InstGen). This causes trouble, particularly with some use-after-free nastiness.

What should happen is that a `Problem` should own its own properties and be in charge of keeping them up to date. This is _almost_ what happens now, but it also has to keep `env.property` up-to-date.

Fix this. This comes in two commits:
1. Less controversial. Prevent use-after-free bugs by making `env.property` and `Problem::_property` a `SmartPtr<Property>` - i.e. reference counting. This is a legitimate case of shared ownership and unlikely to create reference cycles.
2. More controversial. There are only a few instances of `env.property` actually present "in the wild". Therefore, axe `env.property` and add a few properties directly to `env` for legacy reasons: mostly "is this problem polymorphic/higher-order?".

`Problem` now keeps its own properties. It still has to update the legacy properties, but this is much less onerous and these should slowly die off. It still keeps a reference-counted `Property` instance because when copying a problem we want to share the same properties. This is hazardous as we might mutate the same properties instance, but apparently this is an invariant Vampire keeps already.

I admit that the proper way of doing this is to weave instances of `Problem` throughout the code to reach the locations where we need `env.property`. However, I started doing this and the diff is enormous, touching old code that I don't understand. Let's do it more gently and we can try to kill off the legacy properties one-by-one.


MichaelRawson:
@ibnyusuf @selig @quickbeam123 - ping, you had opinions on this in #309.

ibnyusuf:
Looks good. Couple of comments:

- By "slowly die off" you mean that the code be re-engineered to thread the problem through to the desired point?
- Regarding the issue of copying problems whilst sharing properties, why not just copy the property as well?

MichaelRawson:
>By "slowly die off" you mean that the code be re-engineered to thread the problem through to the desired point?

Exactly, or in some cases not need it at all - FMB will apparently use the newer sort mechanism at some point, and some indices could just be always-polymorphic if it later turns out not to matter.

> Regarding the issue of copying problems whilst sharing properties, why not just copy the property as well?

Indeed, that could be done if desired (is it?). Then we wouldn't even need the reference counting. It's not completely trivial because `Property` doesn't have a copy constructor - but it absolutely could be done quickly if we want that.

MichaelRawson:
It wouldn't be technically hard to do so, but having thought a bit more about it since #309 I would have a few concerns:
- We do sometimes have more than one `Problem`, so having a global `env.problem` could be confusing/buggy - consider that we make a fresh `Problem` for some specific thing and then call a routine that somewhere queries `env.problem` for e.g. properties.
- A fair amount of existing code already does "the right thing" and passes a `Problem` around. Probably not entirely correctly, but not bad either. Would we want to convert this to use `env.problem`?
- Ideally I'd like to remove/reduce global state.

That said, I should think about whether I just broke the bugfix for #309.

MichaelRawson:
I think #309 is OK, `env.property` is now dead so it can't be used-after-free.

quickbeam123:
> We do sometimes have more than one Problem, so having a global env.problem could be confusing/buggy - consider that we make a fresh Problem for some specific thing and then call a routine that somewhere queries env.problem for e.g. properties.

but isn't there the same issue with all the proposed "legacy" properties in env?

I suspect it could be even worse, because you seem to be setting them `Problem::refreshProperty`. So if I create my own pet Problem and refresh its property, I will have reset these flags in the global state.

MichaelRawson:
Superseded by #449.

PR #441, opened 2023-05-02
Fixing symbol recording bug

ibnyusuf:
PR fixes a bug whereby new symbols are recorded as either function or predicate (even if it is a type constructor). There is already a fix for this on the `ahmed-new-hol` branch, but that branch does not look like it will be merged soon. In the mean time, the bug is affecting Konstantin as he attempts to use Vampire as a preprocessor for polymorphic problems.

### DIFF
--- a/Forwards.hpp
+++ b/Forwards.hpp
@@ -135,6 +135,8 @@ enum Color {
   COLOR_INVALID = 3u
 };
 
+enum SymbolType{FUNC, PRED, TYPE_CON};
+
 };
 
 namespace Indexing
--- a/Forwards.hpp
+++ b/Forwards.hpp
@@ -135,7 +135,7 @@ enum Color {
   COLOR_INVALID = 3u
 };
 
-enum SymbolType{FUNC, PRED, TYPE_CON};
+enum class SymbolType{FUNC, PRED, TYPE_CON};
 
 };
 
--- a/Kernel/InferenceStore.hpp
+++ b/Kernel/InferenceStore.hpp
@@ -73,7 +73,7 @@ public:
   };
 
   void recordSplittingNameLiteral(Unit* us, Literal* lit);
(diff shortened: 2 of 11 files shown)


Shell/Skolem.hpp
```diff
@@ -107,8 +107,8 @@ class Skolem
   DHMap<unsigned,TermList> _varSorts;
 
   // for some heuristic evaluations after we are done
-  Stack<unsigned> _introducedSkolemSyms;
-
+  Stack<pair<bool, unsigned>> _introducedSkolemSyms;
```
MichaelRawson, line 110: Will we ever introduce a Skolem predicate? I assume FOOL takes care of this.

Kernel/InferenceStore.hpp
```diff
@@ -101,7 +101,7 @@ class InferenceStore
 
 
   /** first is true for function symbols, second is symbol number */
```
MichaelRawson, line 103: Please update comment.

Forwards.hpp
```diff
@@ -135,6 +135,8 @@ enum Color {
   COLOR_INVALID = 3u
 };
 
+enum SymbolType{FUNC, PRED, TYPE_CON};
```
MichaelRawson, line 138: Could this be `enum class` without too much trouble?

MichaelRawson (approved):
Looks good, thanks! Minor comments inline but inclined to merge once checks complete.

Shell/Skolem.hpp
ibnyusuf: That's right

Kernel/InferenceStore.hpp
ibnyusuf: Done

Forwards.hpp
ibnyusuf: It can be, and making the change caught a bug right out of the box!

Issue #421, opened 2022-10-13
Skolem type in same block as term causes assertion violation.


MichaelRawson:
The input

```
tff(test, conjecture, ![A: $tType]: ![X: A]: X = X).
```

causes the following assertion violation:

```
Condition in file /home/michael/vampire/Shell/Skolem.cpp, line 564 violated:
first_pass || sym == last_sym+1
----- stack dump -----
Version : Vampire 4.7 (commit 38d54c2ae on 2022-10-12 12:12:18 +0100)
Control points passed: 35324
last control point:
          -Substitution::bind(int,Term*) (35324)
main
 vampireMode()
  doProving()
   getPreprocessedProblem
    Preprocess::preprocess
     Preprocess::preprocess3(Problem&)
      Preprocess::preprocess3(Unit*)
       Skolem::skolemise(Unit*)
        Skolem::skolemiseImpl(FormulaUnit*)
         Skolem::skolemise (Formula*)
----- end of stack dump -----
```

This is because in `Skolem.cpp` at the indicated line, we assert that fresh Skolem symbols have sequential indices. However, the Skolem for `A` is a type constructor, not a function symbol, so we fail here. @quickbeam123 - if you get time to look at this, is there an easy way to fix this, or is the technique doomed?

ibnyusuf:
I came across this issue on a branch I am working on, and changed the assertion to:

`ASS(!name_reuse || first_pass || sym == last_sym+1);`

which seemed to fix thigns. Not sure if this is valid though...

MichaelRawson:
> Could it be that the code is fine, but the assertion is invalid?

It's certainly broken _with_ name reuse on. Without it the code is fine (I think) but the assertion is too aggressive.

quickbeam123:
Good points! I am processing this now :)

quickbeam123:
OK, so this would be fine if the two quantifiers stayed in two distinct blocks. But flattening comes and merges the two blocks togeter. However, modifying flattening is not the right fix, since the user can already write them merged. Should we split blocks which mix standard sorts and `$tType`? Or should we simply write a more expensive version of symbol reuse in Skolem which does not use the "consecutiveness" trick? Or should we admit symbol reuse does not help that much but complicates the code considerably and remove it again?

By the way, I was surprised to see this
`tff(test, conjecture, ![X: A] : ![A: $tType] : X = X).`
parses (notice that `A`, when it first occurs, has not been declared yet.)
Actuall, even this parses:
[...]

PR #443, opened 2023-05-03
Cannot shuffle variables in quantifier list if polymorphic

ibnyusuf:
Polymorphism requires that type variables appear in the quantifier list befoere term variables, as the type of the term variables may refer to previously declared type variables.

Also added a couple of polymorphic problems to sanity check to hopefully reduce the number of breaking changes that we are seeing.


MichaelRawson (approved):
Sounds like a good fix - however, I wonder if this isn't too aggressive, killling some of the shuffling goodness - what do you think, @quickbeam123?

Would it be possible to shuffle the sort and term variables separately, or maybe even better have rectification split such dependent quantifiers?

quickbeam123:
As a quick fix, it looks like a reasonable way of avoiding a problem.

As an opportunity to learn a lesson and use it while striving for an overall better design, let's brainstorm:
- we already had https://github.com/vprover/vampire/issues/421
- there I wrote "But flattening comes and merges the two blocks togeter. However, modifying flattening is not the right fix, since the user can already write them merged."
- here we see one such merged one: `! [B: $tType,A: $tType,G: fun(A,B),F: fun(A,B)] :`
- would it be nice to live in a world where type variables and normal (first-order) quantifiers always live in different blocks? (Shouldn't we mandate this also for TPTP?)
- how important is the intuition that in a single block, quantification order does not matter
(intuitively, I find it quite important. The shuffling code relies on it)

How hard would it be, to make rectification do it? (Can we always have {all the types first} {then rest follows}?) Was I right that it's actually flattening that's happy to undo it? (Do we always apply them in a specific favourable order, for this to work?)

quickbeam123 (approved):

ibnyusuf:
I agree that keeping type variable quantifiers and term variable quantifiers separate is a good idea. Not sure about the best way to implement (make them syntactically different things? Add a check in flattening to make sure we don't undo this separation?)

joe-hauns (approved):
Sounds all good to me. Don't have too much of an opinion about what's happening before clausification as I don't have a proper understanding of the codebase pre-saturation. But looks and sounds all good :)

PR #444, opened 2023-05-05
remove name reuse

MichaelRawson:
See #157. Remove `Shell::NameReuse`, which aimed to de-duplicate Skolems and definitions up to some amount of redundancy. However, it has little experimental benefit and causes some bugs.

Let he who is without sin cast the first stone! I thought this would be neat and I even wrote a paper about it, but it is both less effective and much harder to integrate nicely than I initially hoped.


Shell/NewCNF.cpp
```diff
@@ -1000,53 +999,23 @@ Term* NewCNF::createSkolemTerm(unsigned var, VarSet* free, Formula *reuse_formul
...
 
   Term* res;
   bool isPredicate = (rangeSort == AtomicSort::boolSort());
   if (isPredicate) {
-    unsigned pred = reused_symbol;
-    if(!successfully_reused) {
-      pred = Skolem::addSkolemPredicate(arity, domainSorts.begin(), var);
-      if(name_reuse)
-        name_reuse->put(reuse_key, pred);
-      env.statistics->skolemFunctions++;
```
quickbeam123, line 1036: I guess the statistics code is factored out somewhere far-far-away?

Shell/Naming.cpp
```diff
@@ -1165,37 +1146,26 @@ Literal* Naming::getDefinitionLiteral(Formula* f, VList* freeVars) {
   }
 
   if(!_appify){
-    unsigned pred = reused_symbol;
-    if(!successfully_reused) {
-      pred = env.signature->addNamePredicate(arity);
-      env.statistics->formulaNames++;
```
quickbeam123, line 1171: The same with this one, I guess.

Shell/NewCNF.cpp
MichaelRawson: Yes, I moved it back to where it was originally - now at line 1097.

Shell/Naming.cpp
MichaelRawson: Yep, line 1220.

PR #449, opened 2023-05-24
only Problem in env and not "someone's property"

quickbeam123:


Kernel/Problem.cpp
```diff
@@ -98,7 +98,6 @@ void Problem::initValues()
   _mayHaveInequalityResolvableWithDeletion = true;
   _mayHaveXEqualsY = true;
   _propertyValid = false;
-  _property = env.property;
```
MichaelRawson, line 101: Is this correct? What did this do before?

Kernel/Problem.cpp
```diff
@@ -303,6 +301,7 @@ void Problem::readDetailsFromProperty() const
   _quantifiesOverPolymorphicVar = _property->quantifiesOverPolymorphicVar();
   _hasBoolVar = _property->hasBoolVar();
   _higherOrder = _property->higherOrder();
+  _hasNonDefaultSorts = _property->hasNonDefaultSorts();
```
MichaelRawson, line 304: Depending on the size of diff I'd be somewhat inclined to not lift these up to `Problem` - something else to keep synchronised. What do you think?

Kernel/Problem.cpp
```diff
@@ -474,14 +473,22 @@ bool Problem::quantifiesOverPolymorphicVar() const
   return _quantifiesOverPolymorphicVar.value();
 }
 
-bool Problem::higherOrder() const
+bool Problem::isHigherOrder() const
```
MichaelRawson, line 476: All of these should probably be inlined, but let's keep that for another PR.

Lib/Environment.hpp
```diff
@@ -93,11 +92,24 @@ class Environment
...
 
+  /**
+   * A global way of accessing "the problem vampire is working on", maily for checking its properties.
+   * Note that if in some special cases there is more than one Problem instance used at one time moment,
+   * one should know which is the main one and that one should be set/reset here.
+   *
+   * (In an ideal world, there would be no need for this function, as the correct Problem object would
+   * be explicitly passed to all the functions interested in knowning...)
+   */
+  Kernel::Problem* getMainProblem() { return _problem; }
```
MichaelRawson, line 103: I guess we have to have the pointer indirection as we can't copy/move a `Problem`?

MichaelRawson: Also, do we want the getter/setter or a bare object? I know this is considered best practice in some languages. C++ I'm not sure.

MichaelRawson (approved):
Seems neat to me! Closing my other PR in favour of this one.

Kernel/Problem.cpp
quickbeam123: Could it be you are confusing a deleted line for an added one?

Kernel/Problem.cpp
quickbeam123: It felt quite natural for this boolean one. But I already backtracked the lifts of "category" and "usesSorts" which I originally started as well..

Kernel/Problem.cpp
quickbeam123: Well, actually, you were right - so form of initialization was (as the tests have shown) necessary!

MichaelRawson:
@joe-hauns - expect you're busy right now but when you get back, could you take a look at the changes to unit tests in f8a3e48? Feels like there might be a better way to do this if we need a "main problem" for a unit test.

joe-hauns:
> @joe-hauns - expect you're busy right now but when you get back, could you take a look at the changes to unit tests in [f8a3e48](https://github.com/vprover/vampire/commit/f8a3e483bc4fc9c6ea6535a768217e96bebbd5a1)? Feels like there might be a better way to do this if we need a "main problem" for a unit test.

Sorry for the long reply time. 
In the long run I would like to have the whole unit testing code not depend on any global state like `env` and `env.options`, at all as that would allow for testing more things more easily. 
But there's a bit of a way to go for that and I think this PR looks great fine with the way things are right now!

PR #450, opened 2023-06-02
Clang 15 exploits our broken code; mitigate this

MichaelRawson:
As discussed internally, Clang/LLVM 15 can find a control path in which we would use an unitialised pointer - which would only happen in very obscure cases, but the optimiser then reshapes the code assuming that this cannot happen, breaking us quite severely. I freely admit I don't understand the exact mechanism, but this does fix the problem and it makes some kind of sense.


quickbeam123:
I am fine with immediate merge/rebase...

JakobR (approved):
Also works on Apple's clang now!

PR #451, opened 2023-06-05
fixes from `clang-tidy`

MichaelRawson:
Small fixes from the `clang-tidy` tool. Experiencing diminishing returns now, but it's nice to have a clean report. Comments inline to explain the changes.

### DIFF
--- a/FMB/FiniteModelBuilder.cpp
+++ b/FMB/FiniteModelBuilder.cpp
@@ -1673,7 +1673,7 @@ MainLoopResult FiniteModelBuilder::runImpl()
 #endif
     //TODO consider adding clauses directly to SAT solver in new interface?
     // pass clauses and assumption to SAT Solver
-    SATSolver::Status satResult;
+    SATSolver::Status satResult = SATSolver::UNKNOWN;
     {
       if (_opt.randomTraversals()) {
         TIME_TRACE(TimeTrace::SHUFFLING);
@@ -1683,7 +1683,6 @@ MainLoopResult FiniteModelBuilder::runImpl()
 
       _solver->addClausesIter(pvi(SATClauseStack::ConstIterator(_clausesToBeAdded)));
 
-      satResult = SATSolver::UNKNOWN;
       env.statistics->phase = Statistics::FMB_SOLVING;
 
       static SATLiteralStack assumptions(_distinctSortSizes.size());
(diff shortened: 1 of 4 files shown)


FMB/FiniteModelBuilder.cpp
```diff
@@ -1673,7 +1673,7 @@ MainLoopResult FiniteModelBuilder::runImpl()
 #endif
     //TODO consider adding clauses directly to SAT solver in new interface?
     // pass clauses and assumption to SAT Solver
-    SATSolver::Status satResult;
```
MichaelRawson, line 1676: Apparently `satResult` could be used unitialised. I don't think it can, but this doesn't hurt.

Kernel/InterpretedLiteralEvaluator.cpp
```diff
@@ -1172,6 +1172,7 @@ bool InterpretedLiteralEvaluator::balance(Literal* lit,Literal*& resLit,Stack<Li
     
     // find which arg of t2 is the non_constant bit, this is what we are unwrapping 
     TermList to_unwrap;
+    to_unwrap.makeEmpty();
```
MichaelRawson, line 1175: This actually looks like a bug. The default `TermList` constructor does _nothing_, but we need `to_unwrap` to be `isEmpty()` on first access. Possibly this "worked" by chance.

Kernel/SortHelper.cpp
```diff
@@ -361,8 +361,7 @@ void SortHelper::collectVariableSortsIter(CollectTask task, DHMap<unsigned,TermL
     
         unsigned position = 0;
         for (TermList* ts = term->args(); ts->isNonEmpty(); ts = ts->next()) {
-          CollectTask newTask;
```
MichaelRawson, line 364: The linter doesn't like that the enumeration `CollectTask::fncTag` is left unitialised by the constructor and filled in later. I'm not sure what the standard says about this, but in this case there's no need for it - we always know at `CollectTask()` what the `fncTag` will be.

Kernel/InterpretedLiteralEvaluator.cpp
quickbeam123: Should we maybe quickly ask @joe-hauns  about it?

Kernel/SortHelper.cpp
quickbeam123: Nice! This also looks like a neater design anyway!

Kernel/InterpretedLiteralEvaluator.cpp
joe-hauns: Yes looks like a bug. The code is not by me originally, so I can't say whether I overlooked something but it looks like you're right! 

What do you think of making the default constructor of `TermList` create an empty term instead of creating an uninitalized one? For me this has been a gotcha quite a few times.

MichaelRawson: I think that's sensible. Very likely this doesn't cause much performance overhead; I'll put it on TODO for later.

Issue #453, opened 2023-06-11
Polymorphism and Z3 not playing nicely


ibnyusuf:
As pointed out by Andrei on slack:

```
vampire_rel -awr 28 -aac none -amm sco -anc none -bd preordered -bs on -bsr on -drc off -fsr off -gsp on -gs on -gsaa from_current -inw on -nwc 1.5 -sas z3 -s -1003 -sos all -sac on -sp frequency -tha off -t 90.0 -uwa all --mode spider --test_id 1980632213 --normalize on --memory_limit 16384 --include /home/voronkov/TPTP /home/voronkov/TPTP/Problems/LCL/LCL749_5.p
Z3 exception:
Sorts |fun(X0,fun(X0,X0))| and X0 are incompatible
```

I guess this is unsurprising, because Z3 has no support for polymorphism. We need to decide how best to handle  the combination of AVATAR using Z3 as solver and polymorphism. Or should we only use minisat if we detect polymorphism?

selig:
Defaulting to minisat sounds like a good step whilst we find a more long-term solution. 

The alternative is the sat_fallback_for_smt option, which runs minisat alongside z3 and falls back to minisat if z3 ever fails... but if we know it's definitely going to fail then that doesn't sound very useful.

MichaelRawson:
I looked at this code not so long ago in #300 - there we partially handle it if the "polymorphism" is actually just an indexed family of monomorphic functions. This looks like it's more involved though.

Seems to me that the thing to do would be to keep the mechanism in #300 and either:
1. Ground all sort variables before handing to Z3.
2. Skip asserting clauses with sort variables to Z3. Then it's an under-approximation, but it is anyway.

ibnyusuf:
If I understand correctly, #300 was a fix for theory instantiation where we know that we are only passing ground things to Z3. Hence, we can safely treat function symbols as indexed families. 

When it comes to AVATAR, we pass non-ground things as well and this is where the issue arises.

ibnyusuf:
#455 should fix this. Once it is tested and merged, we can close this

ibnyusuf:
Just to note, that even with the fix #455 in place, we still receive an error when carrying out the run command. However, the error is totally different (see below). It is linked to the interaction between polymorphism and UWA. As the UWA implementation is set to totally change, I am not going to fix this now.

```
Version : Vampire 4.7 (commit 340dc86bd on 2023-06-23 14:21:36 +0100)
Control points passed: 2214116
[...]

PR #455, opened 2023-06-23
Fixing translation of polymorphic constants to Z3

ibnyusuf:
#300 fixed the translation of polymorphic terms with arity > 0. However, we treat constants differently and they were not being translated correctly. This PR should ensure that they are now translated correctly to Z3.

### DIFF
--- a/SAT/Z3Interfacing.cpp
+++ b/SAT/Z3Interfacing.cpp
@@ -854,10 +854,10 @@ struct ToZ3Expr
     bool isLit = trm->isLiteral();
 
     Signature::Symbol* symb;
-    SortId range_sort;
+    SortId rangeSort;
     if (isLit) {
       symb = env.signature->getPredicate(trm->functor());
-      range_sort = AtomicSort::boolSort();
+      rangeSort = AtomicSort::boolSort();
       // check for equality
       if( trm->functor()==0 || symb->equalityProxy()){
         ASS(trm->numTermArguments()==2);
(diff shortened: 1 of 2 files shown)


MichaelRawson:
The code seems fine, but I don't understand the fix.
> fixed the translation of polymorphic terms with arity > 0

Shouldn't polymorphic terms be of arity > 0 by definition? They need a sort variable to be polymorphic.

From #453
>If I understand correctly, https://github.com/vprover/vampire/pull/300 was a fix for theory instantiation where we know that we are only passing ground things to Z3. Hence, we can safely treat function symbols as indexed families.
>
>When it comes to AVATAR, we pass non-ground things as well and this is where the issue arises.

AVATAR should only pass ground things to Z3. If not (and apparently it doesn't), we need some logic to do the grounding to something reasonable. To start with I'd suggest a new uninterpreted sort and a new uninterpreted constant of that sort, although it's possible that we could do better. Then, the logic in #300 should produce a family of functions for each polymorphic function on-demand.

What am I missing?

ibnyusuf:
Please ignore the comment about non-ground and AVATAR. That was just a silly misunderstanding. I believe we only pass ground things to Z3 in AVATAR.

Indeed, polymorphic terms do have arity > 0. In an old version of this code, we treated something as a constant if `numTermArgs > 0`. However, that was fixed in master. I carry out the above "fix" for the same issue on a branch that didn't contain this fix and then ported it onto master without noticing the different condition.

I have now investigated the root cause of #300 more deeply and it is being cause by unification with abstraction code creating constraints between sorts incorrectly. As the whole UWA mechanism is set to change, I don't propose a fix now, but will leave the bug report open so that it can be revited once the new UWA implementation goes in. I'm closing this PR now.

MichaelRawson:
OK, thanks anyway @ibnyusuf - sounds like we got somewhere. :-)

PR #456, opened 2023-06-29
Special Functor Type-Safety

joe-hauns:
I made the special functor constants in `Term` an enum class. 
This has the advantage that in all switch statements where were we don't handle a variant we get a compile-time warning (instead of a runtime crash).

The reason i did this is because i ran into the following crash:

```
vampire --mode vampire --input_syntax tptp TPTP-v8.0.0/Problems/SET/SET017^1.p --decode lrs+10_1:1_sos=on:ss=included:st=1.2:urr=on:si=on:rawr=on:rtra=on_0

Condition in file /Users/jo/projects/vampire/master/Shell/Shuffling.cpp, line 295 violated:
false
Value of tl.toString() is: (^[X0 : $i, X3 : $i] : ((X0 = X3)))
----- stack dump -----
Version : Vampire 4.7 (commit 6d32a82dc on 2023-06-29 16:19:30 +0100)
Control points passed: 113907
last control point:
         -TermList::toString (113907)
main
 vampireMode()
  doProving()
   getPreprocessedProblem
    Preprocess::preprocess
     Shuffling::shuffle (Problem&)
      Shuffling::shuffle (UnitList*&)
       Shuffling::shuffle (Unit*
        Shuffling::shuffleIter
----- end of stack dump -----
```

Which is caused by the variant `Term::SpecialFunctor::LAMBDA` (former `Term::SF_LAMBDA`) not being handled in [this](https://github.com/vprover/vampire/blob/6d32a82dc2a003268785a73424efca941666ebf7/Shell/Shuffling.cpp#L270-L294) switch statement. 

I guess the either @ibnyusuf or @quickbeam123 should easily be able to fix this.

But refactoring to enums also showed a lot of other spots where some variants are not being handled. Maybe we can fix them together in this branch as it's a lot of places in the code base i never touched. 

In particular the other switch-cases are:
- `LET_TUPLE`
  - [x] [Shell/FOOLElimination.cpp line 506](https://github.com/vprover/vampire/blob/6d32a82dc2a003268785a73424efca941666ebf7/Shell/FOOLElimination.cpp#L506)

- `TUPLE`
  - [x] [Shell/FOOLElimination.cpp line 506](https://github.com/vprover/vampire/blob/6d32a82dc2a003268785a73424efca941666ebf7/Shell/FOOLElimination.cpp#L506)
  - [x] [Shell/NewCNF.cpp line 190](https://github.com/vprover/vampire/blob/6d32a82dc2a003268785a73424efca941666ebf7/Shell/NewCNF.cpp#L1190)
  - [x] [Shell/Property.cpp line 660](https://github.com/vprover/vampire/blob/6d32a82dc2a003268785a73424efca941666ebf7/Shell/Property.cpp#L660)
  - [x] [Shell/SubexpressionIterator.cpp line 120](https://github.com/vprover/vampire/blob/6d32a82dc2a003268785a73424efca941666ebf7/Shell/SubexpressionIterator.cpp#L120)

- `MATCH`:
  - [x] [Shell/Shuffling.cpp line 270](https://github.com/vprover/vampire/blob/6d32a82dc2a003268785a73424efca941666ebf7/Shell/Shuffling.cpp#L270)

- `LAMBDA`:
  - [x] [Shell/SymbolDefinitionInlining.cpp line 378](https://github.com/vprover/vampire/blob/6d32a82dc2a003268785a73424efca941666ebf7/Shell/SymbolDefinitionInlining.cpp#L378)
  - [x] [Shell/SymbolDefinitionInlining.cp line 76](https://github.com/vprover/vampire/blob/6d32a82dc2a003268785a73424efca941666ebf7/Shell/SymbolDefinitionInlining.cpp#L76)
  - [x] [Shell/SymbolOccurrenceReplacement.cp line 26](https://github.com/vprover/vampire/blob/6d32a82dc2a003268785a73424efca941666ebf7/Shell/SymbolOccurrenceReplacement.cpp#L26)
  - [x] [Kernel/FormulaTransformer.cpp line 101](https://github.com/vprover/vampire/blob/6d32a82dc2a003268785a73424efca941666ebf7/Kernel/FormulaTransformer.cpp#L101)
  - [x] [Kernel/SubstHelper.hpp line 288](https://github.com/vprover/vampire/blob/6d32a82dc2a003268785a73424efca941666ebf7/Kernel/SubstHelper.hpp#L288)
  - [x] [Shell/Flattening.cpp line 258](https://github.com/vprover/vampire/blob/6d32a82dc2a003268785a73424efca941666ebf7/Shell/Flattening.cpp#L258)
  - [x] [Kernel/TermTransformer.cpp line 151](https://github.com/vprover/vampire/blob/6d32a82dc2a003268785a73424efca941666ebf7/Kernel/TermTransformer.cpp#L151)
  - [x] [Shell/NNF.cpp line 258](https://github.com/vprover/vampire/blob/6d32a82dc2a003268785a73424efca941666ebf7/Shell/NNF.cpp#L258)
  - [x] [Shell/NewCNF.cpp# line 190](https://github.com/vprover/vampire/blob/6d32a82dc2a003268785a73424efca941666ebf7/Shell/NewCNF.cpp#L1190)
  - [x] [Shell/NewCNF.cpp line 340](https://github.com/vprover/vampire/blob/6d32a82dc2a003268785a73424efca941666ebf7/Shell/NewCNF.cpp#L340)
  - [x] [Shell/PredicateDefinition.cpp line 884](https://github.com/vprover/vampire/blob/6d32a82dc2a003268785a73424efca941666ebf7/Shell/PredicateDefinition.cpp#L884)
  - [x] [Shell/Shuffling.cpp line 270](https://github.com/vprover/vampire/blob/6d32a82dc2a003268785a73424efca941666ebf7/Shell/Shuffling.cpp#L270)
  - [x] [Shell/SimplifyFalseTrue.cpp line 342](https://github.com/vprover/vampire/blob/6d32a82dc2a003268785a73424efca941666ebf7/Shell/SimplifyFalseTrue.cpp#L342)
  - [x] [Shell/SineUtils.cp line 69](https://github.com/vprover/vampire/blob/6d32a82dc2a003268785a73424efca941666ebf7/Shell/SineUtils.cpp#L69)

### DIFF
--- a/Kernel/Term.hpp
+++ b/Kernel/Term.hpp
@@ -271,14 +271,26 @@ class Term
 {
 public:
   //special functor values
-  static const unsigned SF_ITE = 0xFFFFFFFF;
-  static const unsigned SF_LET = 0xFFFFFFFE;
-  static const unsigned SF_FORMULA = 0xFFFFFFFD;
-  static const unsigned SF_TUPLE = 0xFFFFFFFC;
-  static const unsigned SF_LET_TUPLE = 0xFFFFFFFB;
-  static const unsigned SF_LAMBDA = 0xFFFFFFFA;
-  static const unsigned SF_MATCH = 0xFFFFFFF9;
-  static const unsigned SPECIAL_FUNCTOR_LOWER_BOUND = 0xFFFFFFF9;
+  enum class SpecialFunctor {
+    ITE,
+    LET,
+    FORMULA,
+    TUPLE,
+    LET_TUPLE,
+    LAMBDA,
+    MATCH, // <- keep this one the last, or modify SPECIAL_FUNCTOR_LAST accordingly
+  };
+  static constexpr SpecialFunctor SPECIAL_FUNCTOR_LAST = SpecialFunctor::MATCH;
+
+  static constexpr unsigned SPECIAL_FUNCTOR_LOWER_BOUND  = 0xFFFFFFFF - unsigned(SPECIAL_FUNCTOR_LAST);
+  static SpecialFunctor toSpecialFunctor(unsigned f) {
+    ASS_GE(f, SPECIAL_FUNCTOR_LOWER_BOUND);
+    unsigned result = std::numeric_limits<unsigned>::max() - unsigned(f);
+    ASS_LE(result, unsigned(SPECIAL_FUNCTOR_LAST))
+    return SpecialFunctor(result);
+  }
+  static unsigned toNormalFunctor(SpecialFunctor f) 
+  { return std::numeric_limits<unsigned>::max() - unsigned(f); }
 
   class SpecialTermData
   {
@@ -322,43 +334,41 @@ public:
     /** Return pointer to the term to which this object is attached */
     const Term* getTerm() const { return reinterpret_cast<const Term*>(this+1); }
   public:
-    unsigned getType() const {
-      unsigned res = getTerm()->functor();
-      ASS_GE(res,SPECIAL_FUNCTOR_LOWER_BOUND);
-      return res;
-    }
-    Formula* getCondition() const { ASS_EQ(getType(), SF_ITE); return _iteData.condition; }
+    SpecialFunctor specialFunctor() const
+    { return getTerm()->specialFunctor(); }
+
+    Formula* getCondition() const { ASS_EQ(specialFunctor(), SpecialFunctor::ITE); return _iteData.condition; }
     unsigned getFunctor() const {
-      ASS_REP(getType() == SF_LET || getType() == SF_LET_TUPLE, getType());
-      return getType() == SF_LET ? _letData.functor : _letTupleData.functor;
+      ASS_REP(specialFunctor() == SpecialFunctor::LET || specialFunctor() == SpecialFunctor::LET_TUPLE, specialFunctor());
+      return specialFunctor() == SpecialFunctor::LET ? _letData.functor : _letTupleData.functor;
     }
-    VList* getLambdaVars() const { ASS_EQ(getType(), SF_LAMBDA); return _lambdaData._vars; }
-    SList* getLambdaVarSorts() const { ASS_EQ(getType(), SF_LAMBDA); return _lambdaData._sorts; }
-    TermList getLambdaExp() const { ASS_EQ(getType(), SF_LAMBDA); return _lambdaData.lambdaExp; }
-    VList* getVariables() const { ASS_EQ(getType(), SF_LET); return _letData.variables; }
+    VList* getLambdaVars() const { ASS_EQ(specialFunctor(), SpecialFunctor::LAMBDA); return _lambdaData._vars; }
+    SList* getLambdaVarSorts() const { ASS_EQ(specialFunctor(), SpecialFunctor::LAMBDA); return _lambdaData._sorts; }
+    TermList getLambdaExp() const { ASS_EQ(specialFunctor(), SpecialFunctor::LAMBDA); return _lambdaData.lambdaExp; }
+    VList* getVariables() const { ASS_EQ(specialFunctor(), SpecialFunctor::LET); return _letData.variables; }
     VList* getTupleSymbols() const { return _letTupleData.symbols; }
     TermList getBinding() const {
-      ASS_REP(getType() == SF_LET || getType() == SF_LET_TUPLE, getType());
-      return TermList(getType() == SF_LET ? _letData.binding : _letTupleData.binding);
+      ASS_REP(specialFunctor() == SpecialFunctor::LET || specialFunctor() == SpecialFunctor::LET_TUPLE, specialFunctor());
+      return TermList(specialFunctor() == SpecialFunctor::LET ? _letData.binding : _letTupleData.binding);
     }
-    TermList getLambdaExpSort() const { ASS_EQ(getType(), SF_LAMBDA); return _lambdaData.expSort; }
+    TermList getLambdaExpSort() const { ASS_EQ(specialFunctor(), SpecialFunctor::LAMBDA); return _lambdaData.expSort; }
     TermList getSort() const {
-      switch (getType()) {
-        case SF_ITE:
+      switch (specialFunctor()) {
+        case SpecialFunctor::ITE:
           return _iteData.sort;
-        case SF_LET:
+        case SpecialFunctor::LET:
           return _letData.sort;
-        case SF_LET_TUPLE:
+        case SpecialFunctor::LET_TUPLE:
           return _letTupleData.sort;
-        case SF_LAMBDA:
+        case SpecialFunctor::LAMBDA:
           return _lambdaData.sort;
-        case SF_MATCH:
+        case SpecialFunctor::MATCH:
           return _matchData.sort;
         default:
-          ASSERTION_VIOLATION_REP(getType());
+          ASSERTION_VIOLATION_REP(specialFunctor());
       }
     }
-    Formula* getFormula() const { ASS_EQ(getType(), SF_FORMULA); return _formulaData.formula; }
+    Formula* getFormula() const { ASS_EQ(specialFunctor(), SpecialFunctor::FORMULA); return _formulaData.formula; }
     Term* getTupleTerm() const { return _tupleData.term; }
     TermList getMatchedSort() const { return _matchData.matchedSort; }
   };
@@ -401,6 +411,9 @@ public:
   /** Function or predicate symbol of a term */
   const unsigned functor() const { return _functor; }
 
+
+  SpecialFunctor specialFunctor() const 
+  { return toSpecialFunctor(functor()); }
   vstring toString(bool topLevel = true) const;
   static vstring variableToString(unsigned var);
   static vstring variableToString(TermList var);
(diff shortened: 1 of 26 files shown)


Kernel/Term.hpp
```diff
@@ -271,14 +271,26 @@ class Term
...
+    LET,
+    FORMULA,
+    TUPLE,
+    LET_TUPLE,
+    LAMBDA,
+    MATCH, // <- keep this one the last, or modify SPECIAL_FUNCTOR_LAST accordingly
+  };
+  static constexpr SpecialFunctor SPECIAL_FUNCTOR_LAST = SpecialFunctor::MATCH;
+
+  static constexpr unsigned SPECIAL_FUNCTOR_LOWER_BOUND  = 0xFFFFFFFF - unsigned(SPECIAL_FUNCTOR_LAST);
```
MichaelRawson, line 285: I guess this is packed into a bitfield at some point? If so can the magic constant be pulled out as a definition like `FUNCTOR_BITS` so that `SPECIAL_FUNCTOR_LOWER_BOUND = (1 << FUNCTOR_BITS) - 1` or something?

MichaelRawson (commented):
Initial review looks good. Agreed, let's fix the other instances here - I also don't know them well but when I get to them (CADE currently) I'll work it out, this seems like the kind of change we want to encourage in the long run.

Kernel/Term.hpp
joe-hauns: I'm not quiiite sure what you mean here..
What we have right not (same as before) is that we just have `0..SPECIAL_FUNCTOR_LOWER_BOUND - 1` are normal functors, and the rest is reserved for the special ones. There's no bit-fields involved here.

Shell/Normalisation.cpp
```diff
@@ -450,25 +450,25 @@ Comparison Normalisation::compare(Term* t1, Term* t2)
   }
 
   if (t1->isSpecial() && t2->isSpecial()) {
-    comp = compare ((int)t1->getSpecialData()->getType(),
```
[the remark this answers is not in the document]
MichaelRawson: Might even work without the cast, but I forget the `enum class` cast rules. Thanks for explanation!

Kernel/Term.hpp
MichaelRawson: Sorry, I should have been clearer. Why `0xFFFFFFFF` - where does this number come from? It makes me concerned that this happens to work because you compiled it on something where the integers are the right size.

joe-hauns: Very true. I think this magic constant was put there by someone to represent the max of usize originally, and I copied it. I was gonna replace it by `std::numeric_limits<unsigned>::max()`, which I did in some places but forgot this one apparently.

joe-hauns: Fixed.

Shell/Normalisation.cpp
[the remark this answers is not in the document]
joe-hauns: Changed it to get rid of the overflowing cast. (`unsigned` cast was needed indeed)

MichaelRawson:
OK, I'm happy in principle - but we should fix these other things so I will make a point of working through them when I'm back from CADE next week. Great work, @joe-hauns!

quickbeam123:
This might take some time to do properly. 

Like treating LAMBDA in NewCNF would mean to make it HO-compatible. But of course we can explicitly assert unreachable in such places and move on.

By the way, where does MATCH come from? (Is there a paper about it?)

And when have you last time heard of TUPLE, let alone LET_TUPLE? Is it still Evgeny?

MichaelRawson:
>But of course we can explicitly assert unreachable in such places and move on.

Yes - I think this would be the way forward if it's a "research problem" to fix. Should this be `USER_ERROR` or `ASSERTION_VIOLATION`, in your view?

I expected `MATCH` to correspond to SMTLIB `match`. Tuples...well, TFX has them now (and tuple binding), so I guess they're worth keeping if we ever want to parse them.

quickbeam123:
It should be ASSERTION_VIOLATION inside NewCNF, for example. Ahmed is (or should be) making sure the USER_ERROR comes earlier in the pipeline.

Well, I guess I need to study what the SMTLIB match does :)

joe-hauns:
> Like treating LAMBDA in NewCNF would mean to make it HO-compatible. But of course we can explicitly assert unreachable in such places and move on.

Yeah I also think we don't have to solve all these research questions in this PR, but should just have a look which things are easy to fix, do that and put an `ASSERTION_VIOLATION`/`USER_ERROR` where things are not straight forward. I just didn't wanna just add assertions everywhere as I didn't know what the code was doing.

MichaelRawson:
Sounds good. I'll handle these this morning then - maybe `NOT_IMPLEMENTED` would be more semantic?

joe-hauns:
Yesss that sounds like a shout! Maybe adding a comment `/* open research question */` or where that's the case.

MichaelRawson:
@joe-hauns for `Shell/SubexpressionIterator` you commented out the case that handled `TUPLE` - but it seems reasonable to me. What do you know that I don't? :-)

joe-hauns:
> @joe-hauns for `Shell/SubexpressionIterator` you commented out the case that handled `TUPLE` - but it seems reasonable to me. What do you know that I don't? :-)

I only changed the [the first line](https://github.com/vprover/vampire/blob/6d32a82dc2a003268785a73424efca941666ebf7/Shell/SubexpressionIterator.cpp#L155):
replacing `SF_TUPLE` by `Term::SpecialFunctor::TUPLE`. It was commented out before. I had a look at the git blame and the comment was introduced by @ibnyusuf in [this commit](https://github.com/vprover/vampire/commit/8e0e1b2?diff=split).

MichaelRawson:
I did one of three things to handle the other cases:
1. Performance-critical case: `ASSERTION_VIOLATION` - hint to the compiler that this is unreachable. Will have "unfortunate" crash behaviour in release builds.
2. Not performance critical: `NOT_IMPLEMENTED` - raise an exception when we don't handle something.
3. Existing behaviour was to fall through, and this is not obviously wrong. Vigorously comment and maintain the fall-through behaviour explicitly.

@joe-hauns @quickbeam123 - can you give that commit a quick look over? Then I'll merge.

MichaelRawson:
> I only changed the [the first line](https://github.com/vprover/vampire/blob/6d32a82dc2a003268785a73424efca941666ebf7/Shell/SubexpressionIterator.cpp#L155): replacing `SF_TUPLE` by `Term::SpecialFunctor::TUPLE`. It was commented out before. I had a look at the git blame and the comment was introduced by @ibnyusuf in [this commit](https://github.com/vprover/vampire/commit/8e0e1b2?diff=split).

Oh, fun. Well, that commit looks like is was in the midst of serious hacking. Let's see what (if anything) breaks if we enable it again. Ping @ibnyusuf if you remember anything about this, but nbd if not.

Shell/PredicateDefinition.cpp
```diff
@@ -899,6 +899,13 @@ void PredicateDefinition::count (TermList ts,int add, Unit* unit)
         count(TermList(sd->getTupleTerm()), add, unit);
         break;
 
+      case Term::SpecialFunctor::LAMBDA:
```
joe-hauns, line 902: I don't know what the function count i supposed to do .. maybe a comment or a more self-explanatory name would be good if you know/managed to figure out what it does. Buuuut intuitively I think it should probably call `count` on the lambda variable as well, no? I think the fall-through we had before was not the correct behaviour but noone ever added the missing case. If we don't know what this is supposed to do I'd rather add an unimplemented and not a fall through.

Shell/SineUtils.cpp
```diff
@@ -85,6 +85,12 @@ void SineSymbolExtractor::addSymIds(Term* term, DHSet<SymId>& ids)
           addSymIds(sd->getTupleTerm(), ids);
           break;
         }
+        case Term::SpecialFunctor::LAMBDA:
+          // TODO we don't break existing fall-through behaviour for now,
+          // but we should handle this case:
+          // recursively addSymIds() for the lambda body
+          // recursively addSymIds() for the lambda sorts?
+          break;
```
joe-hauns, line 93: Same as before. I wouldn't go for a fall-through but either for recursive-calling on lambda sort + variable, or for `NOT_IMPLEMENTED`.

Shell/PredicateDefinition.cpp
MichaelRawson: OK - I've no idea, I was going for not breaking anything too much, but you're right - let's be told when we've not done it correctly. Can you replace these fallthroughs with `NOT_IMPLEMENTED` when you get a second please? Hacking on another branch atm.

easychair:
I have no confidence in newcnf (I don't mean the concept, I mean the
current implementation). I have not used it in Spider.

A

Shell/PredicateDefinition.cpp
MichaelRawson: Disregard request, got to it now. :-)

ibnyusuf:
> > I only changed the [the first line](https://github.com/vprover/vampire/blob/6d32a82dc2a003268785a73424efca941666ebf7/Shell/SubexpressionIterator.cpp#L155): replacing `SF_TUPLE` by `Term::SpecialFunctor::TUPLE`. It was commented out before. I had a look at the git blame and the comment was introduced by @ibnyusuf in [this commit](https://github.com/vprover/vampire/commit/8e0e1b2?diff=split).
> 
> Oh, fun. Well, that commit looks like is was in the midst of serious hacking. Let's see what (if anything) breaks if we enable it again. Ping @ibnyusuf if you remember anything about this, but nbd if not.

I have no idea why I commented out that code. Happy for it to be put back in.

MichaelRawson:
Everybody's happy, so this goes in. Thanks again @joe-hauns, great change!

PR #457, opened 2023-06-30
Fixes UWA polymorphism compability

joe-hauns:
Fixes the issue that UWA sometimes tries to introduce constraints between sorts, which is not allowed, and results in assertion violations.

An example is the following run:
```
./vampire TPTP-v8.0.0/Problems/LCL/LCL831_5.p --decode lrs+1011_1:1_fd=preordered:fsd=on:sos=on:thsq=on:thsqc=64:thsqd=32:uwa=ground:si=on:rawr=on:rtra=on_100
```


MichaelRawson:
@ibnyusuf - can you ack this? If so I'll push merge, seems sensible.

ibnyusuf:
I am not sure whether this will fix all the polymorphism UWA problems that we are seeing, but it is certainly an improvement on what we had before. So yes, let's get it in.

PR #458, opened 2023-07-05
Fixing equality weight computation

joe-hauns:
Fixes the following bug: 
In term sharing we compute the weight of the literal. For equalities this includes (since introducing polymorphism) that we add the weight of the equality sort (minus 1) to the literal. In our current implementation we double-add this weight for terms with two non-variable terms. This was confirmed to be a bug and not expected behaviour by @ibnyusuf . 
This mini-PR fixes this.


ibnyusuf (approved):
Looks good!

PR #459, opened 2023-07-12
Casc2023

quickbeam123:
Perhaps we could revert https://github.com/vprover/vampire/commit/5bcb424119cd91df85d7b42e4bbcadf5f6787a10, but we could also keep it. Otherwise, I think this is ready to go to master.


easychair:
The strategy part is not ready and will be improved. I would not
release it in the current form. I need 2-3 weeks to have better
schedules and (very important!) the CASC/portfolio mode should be
changed for architectures where instructions cannot be used.

A

PR #470, opened 2023-08-11
Proposition/question: more efficient use of unbound variables in RobSubstitution

mezpusz:
Due to the current wasteful use of variables in unifications via `_nextUnboundAvailable` in `RobSubstitution`, there is an unnecessary amount of variants for most of the terms during proof search. This places a huge burden on term sharing, for example.

My proposition is to backtrack the `_nextUnboundAvailable` value whenever we can, which can be done very easily. Initial tests suggests that this is beneficial, running `-sa discount -t 10` across TPTP resulted in 35 more solved problems (no lost problems) with the overall runtime going down from 138000 to 137000 seconds and the overall memory from 3430000 to 2250000 (!) MB. Of course, more elaborate tests would be needed. 

If the current behaviour is intentional for some reason, please let me know, there is the chance that I'm oblivious to something more subtle here. Is any code relying on what exact variables are there? Or if the backtracking itself is wasteful, I'm open to suggestions how else to do it.

### DIFF
--- a/Kernel/RobSubstitution.cpp
+++ b/Kernel/RobSubstitution.cpp
@@ -167,8 +167,13 @@ RobSubstitution::TermSpec RobSubstitution::deref(VarSpec v) const
     bool found=_bank.find(v,binding);
     if(!found) {
       binding.index=UNBOUND_INDEX;
+      unsigned nuaVal = _nextUnboundAvailable;
       binding.term.makeVar(_nextUnboundAvailable++);
-      const_cast<RobSubstitution&>(*this).bind(v,binding);
+      RobSubstitution* self = const_cast<RobSubstitution*>(this);
+      self->bind(v,binding);
+      if(self->bdIsRecording()) {
+        self->bdAdd(new NextUnboundVariableBacktrackObject(self, nuaVal));
+      }
       return binding;
     } else if(binding.index==UNBOUND_INDEX || binding.term.isTerm()
               || binding.term.isVSpecialVar()) {
--- a/Kernel/RobSubstitution.hpp
+++ b/Kernel/RobSubstitution.hpp
@@ -259,6 +259,19 @@ private:
   friend std::ostream& operator<<(std::ostream& out, RobSubstitution const& self)
   { return out << self._bank; }
 
+  class NextUnboundVariableBacktrackObject
+  : public BacktrackObject
+  {
+  public:
+    NextUnboundVariableBacktrackObject(RobSubstitution* subst, unsigned v) : _subst(subst), _v(v) {}
+    void backtrack() { _subst->_nextUnboundAvailable = _v; }
+    CLASS_NAME(RobSubstitution::NextUnboundVariableBacktrackObject);
+    USE_ALLOCATOR(NextUnboundVariableBacktrackObject);
+  private:
+    RobSubstitution* _subst;
+    unsigned _v;
+  };
+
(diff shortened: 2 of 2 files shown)


quickbeam123:
Shuffled instruction counting approves (TPTP FOF problems in 10000Mi):

```
master6801_dis10
Expect 8738.459135131345 +/- 11.240444977152796
reuseUV6802_dis10
Expect 8752.395242287354 +/- 11.210493971620222
```

although when it concerns memory, smaller footprints often have an even larger impact on reducing wall-clock time than instructions.

MichaelRawson:
That's an astonishing find, great work! I'm not aware of any reason it's done this way, and I agree with @quickbeam123 about the performance implication. I looked at the patch and it seems very reasonable.

@quickbeam123 - I'd be inclined to merge if you didn't see anything concerning in your runs. If it breaks something, I guess we'll find out soon enough. What do you think?

quickbeam123:
Yes, I agree this is a great find and don't see a problem with merging right now.

ibnyusuf:
A much needed improvement. Wanted to point out, that we can do this even more efficiently by simply resetting `_nextUnboundAvailable` every time we reach a leaf of the substitution tree. There is no need to back track, as we should never be applying a substituion from somewhere in teh middle of the tree. Not sure whether we will see much performance gain from this.

mezpusz:
> A much needed improvement. Wanted to point out, that we can do this even more efficiently by simply resetting `_nextUnboundAvailable` every time we reach a leaf of the substitution tree. There is no need to back track, as we should never be applying a substituion from somewhere in the middle of the tree. Not sure whether we will see much performance gain from this.

I'm not too up-to-date on the details of the current implementation, so I didn't want to touch it any further but if that is so, it sounds like a better solution! I guess these single increment backtracks allocated on the heap can amount to a bit of inefficiency.

MichaelRawson:
>Not sure whether we will see much performance gain from this.

Seems like we could (and the heap-allocated backtrack objects seem dubious...is it heap-allocated just to do dynamic despatch?) but it might be quite tricky to isolate uses of RobSubstitution within substitution trees from uses without, i.e. "just unify these terms please". Thanks for the tip anyhow, I'll keep it in mind.

joe-hauns:
From Ahmed:
> A much needed improvement. Wanted to point out, that we can do this even more efficiently by simply resetting _nextUnboundAvailable every time we reach a leaf of the substitution tree. There is no need to back track, as we should never be applying a substituion from somewhere in the middle of the tree. Not sure whether we will see much performance gain from this.

Whenever we reach a leave in the tree and only then unbound variables are being introduced, and then non of them will be in the substitution already. So the counter should alyways be zero when we reach a leave and on backtracking (leaving the leaf) we should reset it to zero. 

Concretely this means after the current whenever we enter a leave and start using a substitution, we will deref inside of `RobSubstitution::apply`, introducing backtrack objects `NextUnboundVariableBacktrackObject(self, 0)`, `NextUnboundVariableBacktrackObject(self, 1)`, .... , and then resets all of them at once when we leave the leaf. This is quite a lot of memory and work we waste. 
Related to this there is the issue that we retract only one `UNBOUND_VARIABLE` binding at a time, when backtracking which we could do much faster I'm pretty sure. 
I'll think about how to improve upon both of these things.

Anyways this PR is a very great find!! Thanks @mezpusz

PR #471, opened 2023-08-14
Remove using std and use std explicitly

kharus:
@MichaelRawson
This PR removes all `using namespace std;` from hpp files.
And prepends `std::` explicitly.

I end up using sed file as I had to checkout and re-apply changes few times.
I thought I attached resulting sed file here.
```sed
s/(<<\s*)endl/\1std::endl/g
s/printStack\(cout\)/printStack\(std::cout\)/g
s/cry\(cout\)/cry\(std::cout\)/g
s/cry\(cerr\)/cry\(std::cerr\)/g
s/([^:])cout(\s*<<)/\1std::cout\2/g
s/([([:blank:]])ostream/\1std::ostream/g
s/^ostream/std::ostream/g
s/([([:blank:],])istream/\1std::istream/g
s/(\S<)istream/\1std::istream/g
s/^istream/std::istream/g
s/([([:blank:],=])pair</\1std::pair</g
s/(\S<)pair</\1std::pair</g
s/([^:])make_pair(\(.+,\s*\w+)/\1std::make_pair\2/g
s/(,)make_pair(\(.+,\s*\w+)/\1std::make_pair\2/g
s/push\(make_pair/push\(std::make_pair/g
s/(<<\s*)flush/\1std::flush/g
s/([^:])cerr(\s*<<)/\1std::cerr\2/g
s/outputAllPremises\(cerr/outputAllPremises\(std::cerr/g
s/([[:blank:]<])ifstream/\1std::ifstream/g
s/([[:blank:]<])ofstream/\1std::ofstream/g
s/([^:])swap(\(\s*[^A-Z].+,)/\1std::swap\2/g
s/([^:])max(\(\s*[[:lower:]_1(].*,)/\1std::max\2/g
s/static int std::max/static int max/g
s/([^:])min(\(\s*[^A-Z].*,)/\1std::min\2/g
s/static int std::min/static int min/g
s/([[:blank:]<(])unique_ptr/\1std::unique_ptr/g
s/([^:>])get</\1std::get</g
s/([^:])numeric_limits</\1std::numeric_limits</g
s/([^:])initializer_list</\1std::initializer_list</g
s/([^:])make_unique</\1std::make_unique</g
s/([^:])make_tuple\(/\1std::make_tuple\(/g
s/([^:])setprecision/\1std::setprecision/g
s/(<<\s*)fixed/\1std::fixed/g
s/([^:])string values_header/\1std::string values_header/g
s/([^:v])string::npos/\1std::string::npos/g
s/([^:])shared_ptr</\1std::shared_ptr</g
s/([^:])max</\1std::max</g
s/Lib::make_unique</std::make_unique</g
s/([^:])streamsize/\1std::streamsize/g
s/([([:blank:]])unordered_set</\1std::unordered_set</g
s/([^:])lexicographical_compare\(/\1std::lexicographical_compare\(/g
s/([^:_])sort(\(.*begin)/\1std::sort\2/g
s/([^:_])list</\1std::list</g
s/(<<\s*)setw/\1std::setw/g
s/(<<\s*)left\b/\1std::left/g
s/(<<\s*)right\b/\1std::right/g
s/Size add = std::max/Size add = max/g
s/std::swap\(\*this, obj\)/swap\(\*this, obj\)/g
s/std::left_res/left_res/g
s/std::right_res/right_res/g
s/([^:])abs(\([^)]+\))/\1std::abs\2/g
```


MichaelRawson:
Thanks for the PR, @kharus! Always great to have new contributors to Vampire. I think this general direction is a good idea. Once we agree that this is what we want, I'll go through each line and review it before merge.

@quickbeam123 makes a good point that making this large a change does impact `git blame` somewhat. 

> How about we use the long names in the headers

Indeed - this is the main evil, once you do `using namespace ...` in a header it 'infects' anything you include.

>but still do using in the cpp files

Right, this is a good compromise position. How easy would this be for you, @kharus?

>Also in the headers there could be local using on a per class basis, right?

Sure, if desired to introduce an abbreviation. Like `using ClauseIterator = VirtualIterator<Clause *>`. But I'd probably prefer moving e.g. `using std::cout` to source files.

>I am not sure I ever want to keep typing std::cout << "My debug message" << std::endl;

Fair enough - would Joe's `DBG` macros work for you?

kharus:
> @quickbeam123 makes a good point that making this large a change does impact `git blame` somewhat.

Here goes my hope of applying `clang-format` :-)

> 
> > How about we use the long names in the headers
> 
> Indeed - this is the main evil, once you do `using namespace ...` in a header it 'infects' anything you include.
> 
> > but still do using in the cpp files
> 
> Right, this is a good compromise position. How easy would this be for you, @kharus?

Should be relatively easy to do.
- hpp files
I just pass through the previous pipeline.

- cpp files
some already have `using namespace std` for the rest of the which use std names I just prepend  `using namespace std` in using section.

How to do this git-wise?
As there will be a lot of changes. Is it worth trying to apply changes on the current branch? Shall I reset branch to master apply changes and force push it? Or close this PR and create another one?

kharus:
@quickbeam123 @MichaelRawson 
There are files like [NegativeExt.cpp](Inferences/NegativeExt.cpp)
```cpp
#if VDEBUG
#include <iostream>
using namespace std;
#endif
```
which import std in condition.
So during CI it will compile as DEBUG is enabled.
But generally it will fail as NegativeExt.cpp uses pair without qualification later in the code.
https://github.com/vprover/vampire/blob/a5000659a8067eddf5e7345978ac0faaf3cd4a7f/Inferences/NegativeExt.cpp#L81

kharus:
I made changes in a separate branch.
PR in my own fork https://github.com/kharus/vampire/pull/9

I can squash it and either reset current branch to it.
Or I can open a new PR.

MichaelRawson:
>Here goes my hope of applying clang-format :-)

Ours too! There seems to be some ways around this, but I think in the end it's probably not worth it - `git blame` is really helpful to work out whether the latest bug is 'oh yeah, I forgot this case two weeks ago' or 'this is 15 years old, how did it only break now?!'.

>I just added using std::pair in using section

Seems sensible to me.

>I made changes in a separate branch.

Amazing! That was fast. Your new branch looks good to me from a first glance, and it's a much smaller diff. Version control: however you want to do it is fine by me. If it were me I'd probably squash history on a branch with a single commit called something like 'remove "using namespace std" from headers' and open a new PR with that one, mentioning this one to link the discussion.

quickbeam123:
I don't think we generally mind more commits (as long as they are named in a reasonable way). So I wouldn't insist on squashing.  

On the other hand, there is a reasonable rule to have, namely that each commit should ideally compile. But since this breaks quite often, I wouldn't be so strict about it. It's just nice to keep it in mind ...

kharus:
Regarding `git blame` I had a quick check.
It is possible to ask `git blame` to ignore particular revisions with `--ignore-rev` and `--ignore-revs-file`.
Unfortunately there is not "standard" ignore-revs file like `.gitignore` and `.gitmodules`
But seems to be converging to `.git-blame-ignore-revs`
https://github.blog/changelog/2022-03-24-ignore-commits-in-the-blame-view-beta/
It can be explicitly set to any file by git config `blame.ignoreRevsFile`

This is for command-line. I did a quick check for editors and they seem to have support issues filed around 2020 might have been resolved.

Moreover I tried `Clion` and `VSCode` and they have more powerful (in my opinion) options
VSCode: Line history in Gitlens Inspect
Screenshot 2023-08-16 094201
CLion: Select line and Git->show history for selection
Screenshot 2023-08-16 094629

Admittedly all options require manipulations comparing with vanilla `git blame`.

Personally I glad this come up as I never knew about Show history for selection - I will be using it often now :-)

kharus:
I thought about Git and I think I'll close this PR and open another one with approach @quickbeam123 suggested.

As a side note my personal branches are always quite messy as I work on Home PC, Work laptop(arm/M2) and Work laptop(Intel). So if I need to move I just commit my work with "Partial work" and continue from another machine.
So I almost always squash to pr branch before submitting my work.

quickbeam123:
The next level trick I used a couple of time is https://stackoverflow.com/questions/4404444/how-do-i-git-blame-a-deleted-line Not sure it that has support in contemporary IDEs, but it's sometimes useful to know about...

There is obviously nothing against squashing or interactive rebasing of feature branches, if you have the energy and discipline for it, it's great!

MichaelRawson:
Closing in favour of #472.

PR #472, opened 2023-08-16
Remove using std in hpp files

kharus:
@MichaelRawson @quickbeam123 
This is a continuation of https://github.com/vprover/vampire/pull/471
Usually I would just apply suggested changes but because of large amount of diffs I decided to open a new PR instead.

Change in [Indexing/RequestedIndex.hpp](https://github.com/kharus/vampire/pull/9/files/0f1a3e4dddfa75a35c3e78e66ea7c12a5e27397f#diff-cd95f1b35669044294481a86471f58a890b166322eb6ebda2a8743f73da94950) is applied


Inferences/Superposition.cpp
```diff
@@ -178,7 +179,7 @@ bool Superposition::checkClauseColorCompatibility(Clause* eqClause, Clause* rwCl
   }
   if(getOptions().showBlocked()) {
     env.beginOutput();
-    env.out()<<"Blocked superposition of "<<eqClause->toString()<<" into "<<rwClause->toString()<<endl;
+    env.out()<<"Blocked superposition of "<<eqClause->toString()<<" into "<<rwClause->toString()<<std::endl;
```
quickbeam123, line 182: This is close to being just picky, but why the inconsistency here?

Minisat/core/Dimacs.h
```diff
@@ -43,7 +45,7 @@ static void readClause(B& in, Solver& S, vec<Lit>& lits) {
     for (;;){
         parsed_lit = parseInt(in);
         if (parsed_lit == 0) break;
-        var = abs(parsed_lit)-1;
+        var = std::abs(parsed_lit)-1;
```
quickbeam123, line 48: Another inconsistency with this one. Although, since Minisat has mostly been just "copy-pasted" from it's original source, simply not touching it, as long as it works, could also be a strategy. I also doubt we include `Dimacs.h` anyway, but that's not a point...

Inferences/Superposition.cpp
kharus: Superposition.cpp uses std in a bit clunky way.
It imports it in
```
#if VDEBUG
#include <iostream>
using namespace std;
#endif
```
But if VDEBUG is not turned on it fails.
I didn't want to have a file which has 2 `using namespace std;`
So I added `using std::pair;`
But then it still uses one `cout` outside of VDEBUG cc (which I think is a mistake) so for a moment I just added std:: to this cout.

Minisat/core/Dimacs.h
kharus: I had one test(tQuotientE.cpp) failing compiled with GCC if `abs` is without namespace qualification.
So I added `std::` to all abs to be safe.

Inferences/Superposition.cpp
quickbeam123: I wouldn't mind turning the conditional using into unconditional one. (And the printing uder showBlocked seems fine even in release). But this is all really just cosmetics.

Debug/TimeProfiling.cpp
```diff
@@ -57,7 +58,7 @@ TimeTrace::ScopedTimer::ScopedTimer(TimeTrace& trace, const char* name)
       .map([](auto& x) { return &*x; })
       .find([&](Node* n) { return n->name == name; })
       .unwrapOrElse([&]() { 
-          children.push(Lib::make_unique<Node>(name));
```
MichaelRawson, line 60: ...huh?! Any ideas how this compiled in the first place? Seems fine though.

quickbeam123:
A general question, though, about whether extra qualification is always the correct way of resolving thing (a question, because I only suspect there is a thing to discuss and hope someone knows better and will clarify):

Namely, with argument-dependent name lookup, sometimes the unqualified name might be the correct choice, right?
https://en.wikipedia.org/wiki/Argument-dependent_name_lookup

MichaelRawson:
>Namely, with argument-dependent name lookup, sometimes the unqualified name might be the correct choice, right?

Right - I'm going through all these very carefully, `std::swap` is particularly hazardous because of the copy-and-swap idiom.

I believe the way this works is that if you write e.g. `swap` and there is a "better" overload in scope the compiler uses that instead of `std::swap`, but defaults to `std::swap` if there is none.

MichaelRawson:
https://stackoverflow.com/a/5695855 - this answer was good for me.

Indexing/SubstitutionTree.hpp
```diff
@@ -86,7 +85,7 @@ class InstanceCntr {
   Cntr& _cntr;
 
   InstanceCntr& operator=(InstanceCntr&& other) 
-  { swap(other._cntr, _cntr); return *this; }
+  { std::swap(other._cntr, _cntr); return *this; }
```
MichaelRawson, line 88: Could you convert this and other occurrences to `using std::swap; swap(...)` please? Let me know if you need an explanation, I personally find this extremely unintuitive.

Kernel/Term.hpp
```diff
@@ -574,7 +574,7 @@ class Term
 
     TermList* ts1 = args();
     TermList* ts2 = ts1->next();
-    swap(ts1->_content, ts2->_content);
+    std::swap(ts1->_content, ts2->_content);
```
MichaelRawson, line 577: Same here.

Lib/DynamicHeap.hpp
```diff
@@ -213,7 +212,7 @@ class DynamicHeap {
   void swapInHeap1(size_t idxA, size_t idxB)
   {
     T* data1 = getData1();
-    swap(data1[idxA], data1[idxB]);
+    std::swap(data1[idxA], data1[idxB]);
```
MichaelRawson, line 215: Same here. I think this one might actually matter.

Lib/Option.hpp
```diff
@@ -242,7 +242,7 @@ class Option : OptionBase<A> {
   Option take() 
   {
     Option out;
-    swap(*this,out);
+    std::swap(*this,out);
```
MichaelRawson, line 245: Also here.

Lib/Stack.hpp
```diff
@@ -833,10 +833,10 @@ namespace std
 template<typename T>
 void swap(Lib::Stack<T>& s1, Lib::Stack<T>& s2)
 {
-  swap(s1._capacity, s2._capacity);
-  swap(s1._cursor, s2._cursor);
-  swap(s1._end, s2._end);
-  swap(s1._stack, s2._stack);
+  std::swap(s1._capacity, s2._capacity);
```
MichaelRawson, line 836: Also here.

Test/TestUtils.hpp
```diff
@@ -236,11 +236,11 @@ bool __permEq(L1& lhs, L2& rhs, Eq elemEq, DArray<unsigned>& perm, unsigned idx)
     return true;
   }
   for (unsigned i = idx; i < perm.size(); i++) {
-    swap(perm[i], perm[idx]);
+    std::swap(perm[i], perm[idx]);
```
MichaelRawson, line 239: Also here.

MichaelRawson (changes requested):
`std::swap` needs to be changed to bare, unqualified `swap` with `std::swap` in scope to enable ADL. Otherwise looks good to me!

Indexing/SubstitutionTree.hpp
kharus: Shall I put `using std::swap;` on top with other `using ` statements?

quickbeam123: I would suggest to do it only locally, since we are talking about headers!

quickbeam123: Also, keeping this together highlights it's an idiom and not something overlooked!

MichaelRawson (approved):

PR #479, opened 2023-08-17
remove the InstGen implementation

MichaelRawson:
See #157. Neither @easychair nor @quickbeam123 can demonstrate significant benefit to InstGen, and it is a non-trivial maintenance burden. In future it can of course be revived, but for now let's shelve it and focus on other things.


quickbeam123:
Looking good - fine to go from my side!

PR #480, opened 2023-08-17
Reset nextUnbound

ibnyusuf:
As discussed on Zulip, here is an alternative implementation that solves the `nextUnboundAvailable` issue without allocating any backtrack objects.


MichaelRawson (approved):
Looks good to me - have you run any testing for correctness/performance? If so I'd be inclined to merge.

ibnyusuf:
Still needs testing

quickbeam123:
Yes, this looks slightly better still. But even if not significantly better, it looks simpler and does not need heap. Fine with me!

quickbeam123 (approved):

PR #485, opened 2023-10-04
FOOL proofs TPTP compliant

quickbeam123:
Following the last email from Geoff this PR tries to make sure that
* definitions introduced during FoolElimination are called as such,
* they are not proof-wise dependent on the unit they are getting introduced for, but
* the unit the elimination outputs is proof-wise dependent on them.
* They are jointly called FOOL_XXX_DEFINITION (not ELIMINATION).
* They inform InferenceStore about the new symbol introduction (which makes the TPTP proof report that further to the user).
* The last aspect is added also to the TweeGoalTransformation code and to FunctionDefinitionIntroduction.

I also updated the MATCH elimination code in FoolElimination although that's probably never going to be (immediately) relevant to TPTP users.

### DIFF
--- a/Kernel/InferenceStore.cpp
+++ b/Kernel/InferenceStore.cpp
@@ -714,7 +714,8 @@ protected:
       vstring newSymbolInfo;
       if (hasNewSymbols(us)) {
         vstring newSymbOrigin;
-        if (rule == InferenceRule::FUNCTION_DEFINITION || rule == InferenceRule::FOOL_ITE_DEFINITION || rule == InferenceRule::FOOL_LET_DEFINITION) {
+        if (rule == InferenceRule::FUNCTION_DEFINITION ||
+          rule == InferenceRule::FOOL_ITE_DEFINITION || rule == InferenceRule::FOOL_LET_DEFINITION || rule == InferenceRule::FOOL_FORMULA_DEFINITION) {
           newSymbOrigin = "definition";
         } else {
           newSymbOrigin = "naming";
--- a/Kernel/InferenceStore.cpp
+++ b/Kernel/InferenceStore.cpp
@@ -713,7 +713,13 @@ protected:
     else if (!parents.hasNext()) {
       vstring newSymbolInfo;
       if (hasNewSymbols(us)) {
-	      newSymbolInfo = getNewSymbols("naming",us);
+        vstring newSymbOrigin;
+        if (rule == InferenceRule::FUNCTION_DEFINITION) {
+          newSymbOrigin = "definition";
+        } else {
+          newSymbOrigin = "naming";
+        }
+	      newSymbolInfo = getNewSymbols(newSymbOrigin,us);
       }
       inferenceStr="introduced("+tptpRuleName(rule)+",["+newSymbolInfo+"])";
     }
--- a/Kernel/InferenceStore.cpp
+++ b/Kernel/InferenceStore.cpp
@@ -652,7 +652,7 @@ protected:
       } else if (sym.first == SymbolType::PRED){
         symsStr << env.signature->predicateName(sym.second);
       } else {
-        symsStr << env.signature->typeConName(sym.second);       
+        symsStr << env.signature->typeConName(sym.second);
       }
       if (symIt.hasNext()) {
         symsStr << ',';
@@ -689,7 +689,6 @@ protected:
     default: ;
     }
 
-
     //get vstring representing the formula
 
     vstring formulaStr=getFormulaString(us);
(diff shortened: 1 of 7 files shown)


Shell/FOOLElimination.cpp
```diff
@@ -149,13 +150,11 @@ FormulaUnit* FOOLElimination::apply(FormulaUnit* unit) {
...
-   * MS/TODO: We should be presenting the new formula as following
-   * from the rectifiedUnit and the generated definitions
-   * (similarly to how this is done with Naming)
-   */
+  // add the master premise to the definitions and pass them to the inference object
+  UnitList::push(rectifiedUnit,_currentDefs);
   FormulaUnit* processedUnit = new FormulaUnit(processedFormula,
-      NonspecificInference1(InferenceRule::FOOL_ELIMINATION, rectifiedUnit));
+      NonspecificInferenceMany(InferenceRule::FOOL_ELIMINATION, _currentDefs));
+  _currentDefs = UnitList::empty();
```
MichaelRawson, line 158: Remind me - does `processedUnit` now own the previous value of `_currentDefs`, or are we leaking it here?

MichaelRawson (approved):
Looks fine to me! Impressively small change for quite a complex thing - most of the diff is killing extraneous whitespace, right?

What is `FOOL_FORMULA_DEFINITION` and how does it differ from `FOOL_ELIMINATION`? Is `FOOL_ELIMINATION` now dead?

Kernel/InferenceStore.cpp
```diff
@@ -700,29 +699,37 @@ struct InferenceStore::TPTPProofPrinter
...
+	      axiomName="unknown";
       }
       inferenceStr="file("+fileName+","+quoteAxiomName(axiomName)+")";
     }
     else if (!parents.hasNext()) {
       vstring newSymbolInfo;
       if (hasNewSymbols(us)) {
-	newSymbolInfo = getNewSymbols("naming",us);
+        vstring newSymbOrigin;
+        if (rule == InferenceRule::FUNCTION_DEFINITION ||
```
mezpusz, line 717: Would it make sense to introduce a checker function like `isFormulaTransformation`, `isSimplifyingInferenceRule`, etc? Although I'm not entirely convinced those cannot break if someone inserts a new value into `InferenceRule` into the wrong place... but introducing a new definition inference rule that is not added here might also break things, right? I'm actually not sure which one is better, what do you think?

Shell/FOOLElimination.cpp
mezpusz: According to my shallow investigation `Inference::destroy` takes care of it eventually, assuming the inference itself is ever destroyed.

quickbeam123:
Right, I decided to do some whitespace cleanup and reformatting on the units I was touching.

FOOL_ELIMINATION is not dead. But is used more consistently. What we used to do was this:
1. orig_formula
2. symbol_definition via, e.g., FOOL_ITE_ELIMINATION from 1
3. new_fromula via FOOL_ELIMINATION from 1
this does not make much sense, logic-wise, and also Geoff did not like it, we newly do
1. orig_formula
2. symbol_definition via, e.g., FOOL_ITE_DEFINITION from nothing (it's an axiom)
2. new_fromula via FOOL_ELIMINATION from 1 and 2

The formula case was even more broken, arguably, as it was using FOOL_ELIMINATION for both the definition and the new formula. Now the definition is called FOOL_FORMULA_DEFINITION and as with the others works in tandem with FOOL_ELIMINATION.

Shell/FOOLElimination.cpp
quickbeam123: It's potentially leaking (as many other things in vampire), but it's meant to be owned by the inference. 

Note that just before _currentDefs are passed to the inference, the original formula is added as well.

Kernel/InferenceStore.cpp
quickbeam123: It would definitely make sense to do something more elegant here. But I couldn't think of anything easy, so I just decided to enhance the already quite bad rule of "if it has no parents it must be naming".

However, not much can ever get broken here. It's really just about what's written in the inference info of the TPTP version of the proof. We will have more opportunities for making this straight when users like Geoff bug us in the future about things like this one. At the moment, I don't think there is a clear understanding of what should be written here for each inference, and I don't think there is much software out there that would actually care.

quickbeam123:
Thank you for having a look, by the way!

Kernel/InferenceStore.cpp
mezpusz: I agree, this particular condition probably stays like this until some major change in TPTP or SMTLIB, at which point we would do a more thorough sweep anyways.

mezpusz (approved):

MichaelRawson:
>Right, I decided to do some whitespace cleanup and reformatting on the units I was touching.

Great! Fine by me, to be clear.

I think this is ready to go - do we wait for @ibnyusuf or merge ASAP to close the feedback loop with Geoff?

ibnyusuf:
Looks fine to me. Just one question, we introduce definitions in quite a few places in the preprocessing pipeline. Do we do it correctly elsewhere? Both in the classical pipeline and in NewCNF? I've checked a few places myself and we do seem to do it correctly, but haven't had the opportunity to check thoroughly.

quickbeam123:
That's a good point, @ibnyusuf , I wasn't trying to be comprehensive on this front. (I also didn't look into the elimination of lambda's that' referenced from FoolElim.) I only did what was reasonably in sight of this change. Overall, I am happy to wait for the next complaint from Geoff :)

PR #487, opened 2023-10-09
iterative version of CNF::clausify

quickbeam123:
The original function left for a reference (now called `clausify_rec`).

On top of removing the danger of stack overflows, this also solves around 6 more TPTP problems in some 5s per problem, `-sa discount -awr 10` setting.

The downside is that it is harder to understand what's going on, but that's what the recursive function is still there for.

### DIFF
--- a/Shell/CNF.cpp
+++ b/Shell/CNF.cpp
@@ -74,6 +74,7 @@ void CNF::clausify (Unit* unit,Stack<Clause*>& stack)
  *
  * @since 27/12/2007 Manchester
  */
+/*
 void CNF::clausify_rec (Formula* f)
 {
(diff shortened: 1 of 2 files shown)


MichaelRawson (approved):
Nice change! I believe that the logic is identical and mechanically translated, no objections there. Would it be worth commenting out the original recursive function, or do you want it to continue compiling? It's likely to end up in the final binary unless the linker does some heavy lifting.

quickbeam123:
A good idea with the commenting! 

Yes, it's a mechanical translation. I tried to go even further and perhaps simplify things a bit, but the original is actually quite subtle, so I am not sure it can be easily simplified further.

PR #488, opened 2023-10-09
Implement assert-claim in smtlib2 (a vampiric extension of the format)

quickbeam123:
mimicking the TPTP `claim` role (actually, also a vampire-custom extension, I believe) and sharing a bit of the code

The benefit of claims is that they are lemmas that can be used, but need to be proven as well. Suggesting a good lemma can often be crucial for finding proofs quickly.

### DIFF
--- a/Parse/SMTLIB2.cpp
+++ b/Parse/SMTLIB2.cpp
@@ -290,6 +290,18 @@ void SMTLIB2::readBenchmark(LExprList* bench)
       continue;
     }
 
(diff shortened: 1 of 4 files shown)


MichaelRawson (approved):
Can I ask what the motivation for using this in SMT-LIB is? Otherwise fine, no problem with the code. And the extra consistency is nice! Go ahead and merge, I'm just curious.

quickbeam123:
Mainly inductive theorem proving, as there is no way to specify term algebras in TPTP.

PR #490, opened 2023-10-18
disable time profiling by default

MichaelRawson:
We have the new `TimeProfiling`, which is cool. However, we forgot to switch it off for the last competition, and it does apparently have a little overhead. Switch it off by default.

If @joe-hauns or others object, no problem, let's discuss - this just seemed the logical solution to me.


quickbeam123:
Was thinking a bit whether it would make sense to leave it in debug mode, but nobody cares about timings of the debug version... So yes!

PR #492, opened 2023-10-19
prevent readFromEncodedOptions from silently setting normalize to true

quickbeam123:
As @ibnyusuf noticed at some point, the function `Options::readFromEncodedOptions` silently sets _normalize to true. 

This means any vampire run using --decode will use normalize unless turned off explicitly again later.

This might have been a precaution against forgetting normalize for casc mode, but casc mode already does  
``env.options->setNormalize(true);``
in `vampire.cpp` which has the additional benefit that it applies normalization in the master process once (and not in every child).


MichaelRawson:
Makes sense to me.

PR #493, opened 2023-10-25
IsRenaming fix

mezpusz:
Until now, the `ResultSubstitution::isRenaming` function only checked if same variables are mapped to same variables in the renaming (which is always true), but not that the mapped variables are only used once.

Fortunately, this function is rarely ever called even when encompassment demodulation is on and the error did not result in incompleteness, only some demodulations not performed.

As a side note, how would people feel about enabling encompassment demodulation as default and getting rid of the messy original demodulation redundancy checks?

### DIFF
--- a/Indexing/ResultSubstitution.cpp
+++ b/Indexing/ResultSubstitution.cpp
@@ -72,12 +72,16 @@ ResultSubstitutionSP ResultSubstitution::fromSubstitution(RobSubstitution* s, in
  */
 bool ResultSubstitution::isRenamingOn(TermList t, bool result) 
 {
-  DHMap<TermList,TermList> renamingInMaking;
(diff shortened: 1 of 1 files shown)


quickbeam123:
I approve of the fix! Thank you, @mezpusz ! 

I will quickly check performance during today and then merge.

Why don't we start a new PR for the defaultification? This would mean updating the options (making encompass default means it does not need to be called "encompass" and can replace the current default). Then having a look and simplifying the two demodulation modules. It should also be updated in the samplers (I will do that). Regarding schedules, they don't seem to be using encompass, so they will just all get a better default (the new drc=on).

quickbeam123:
Master: 8849.7 -> fixed: 8858.1 (both dis10 under drc=encompass) good job!

Issue #376, opened 2022-06-01
Replace TTTT by BUE


quickbeam123:
Until recently, we had in vampire a class for transforming terms called `TermTransformerTransformTransformed` (TTTT), which we recently renamed to a more fitting `BottomUpTermTransformer`. It turned out that @joe-hauns has in the meantime implemented a more streamlined (and feature rich) version of the idea in `evaluateBottomUp` templated function. 

It would be nice to only use the later consistently everywhere in vampire, if possible. This is a "homework issue" by Martin ;)

PR #494, opened 2023-10-26
Fix hidden overloaded virtual function warning in AnswerExtractor

mezpusz:
Just renaming a function that apparently hides a virtual one, according to my clang. Also, I removed the `virtual` keyword since the function called is not virtual either, but I'm okay with leaving it if it is needed.

### DIFF
--- a/Inferences/ArrayTheoryISE.hpp
+++ b/Inferences/ArrayTheoryISE.hpp
@@ -37,7 +37,7 @@ public:
   ~ArrayTermTransformer();
   Literal* rewriteNegEqByExtensionality(Literal* l);
 protected:
-  TermList transformSubterm(TermList trm);
+  TermList transformSubterm(TermList trm) override;
 };
 
 class ArrayTheoryISE
--- a/Kernel/TermTransformer.cpp
+++ b/Kernel/TermTransformer.cpp
@@ -226,9 +226,17 @@ Term* TermTransformer::transformSpecial(Term* term)
 
 TermList TermTransformer::transform(TermList ts)
 {
-  if (ts.isVar()) {
-    return transformSubterm(ts);
+  // first let's try transforming ts directly
+  TermList transformed = transformSubterm(ts);
+  if (transformed != ts) {
+    // we did transform, so we are done
+    return transformed;
+  } else if (ts.isVar()) {
+    // we didn't transform, but it's a var (no way to recurse)
+    return ts;
   } else {
+    // try transform subterms
+    ASS(ts.isTerm());
     return TermList(transform(ts.term()));
   }
 }
--- a/Kernel/TermTransformer.cpp
+++ b/Kernel/TermTransformer.cpp
@@ -72,7 +72,7 @@ Term* TermTransformer::transform(Term* term)
       if(orig->isSort()){
         //For most applications we probably dont want to transform sorts.
         //However, we don't enforce that here, inheriting classes can decide
-        //for themselves        
+        //for themselves
         newTrm=AtomicSort::create(static_cast<AtomicSort*>(orig), argLst);
       } else {
         newTrm=Term::create(orig,argLst);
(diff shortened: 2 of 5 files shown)


MichaelRawson (approved):
Sensible. Will wait for @hzzv 's approval, but after that will merge.

quickbeam123:
This might be related to something I didn't really understand in the synthesis PR. 

There:
@hzzv made `TermList  TermTransformer::transform(TermList ts);` virtual, @MichaelRawson asked "Oh, this is fun. Did nobody override this before?!" and the reply was "Apparently not :)".

Do you guys remember what this was all about and how does it relate to @mezpusz's fix?

MichaelRawson:
So there's `TermTransformer::transform(TermList)`, which after the synthesis PR is dynamically-dispatched because `ConjectureSkolemReplacement` overloads it. This seems to be the intended behaviour, quoting from `TermTransformer.hpp`

```
 * The inheriting class implemets function transform(TermList)
 * and then the function transform(Literal*) uses it to transform subterms
 * of the given literal.
```

I'm not sure _how_ this worked before, hence my confusion in that review comment.

But here there is a name clash on `transform(Literal *)`, but I don't think the intention is to override it, just call the base class' version. So probably it should go all together?

MichaelRawson:
https://godbolt.org/z/9jr8qPMr5 - weirdly, name lookup seems to be broken with virtual and non-virtual functions with the same name. @mezpusz seems to have the smallest-diff fix, though in the long run we should probably rename `TermTransformer::transform(Literal *)`. I think `TermTransformer` is on the way out though (?), in favour of `BottomUpEvaluation`, see #376.

quickbeam123:
OK, so I changed this completely.
1) There is a different fix in AsnwerExtractor.hpp (I think a bit more c++ hardcore-ish, but it teaches a lesson.)
2) I fixed/added comments into TermTransformer.hpp fixing the typo there
3) streamlined `TermTransformer::transform(TermList ts)` (@MichaelRawson , could you please check whether you agree?)
4) went pedantic about using `override` 

The last question that I have is for @hzzv: Do you, by any chance, remember why you had to override `transform(TermList ts)` in `ConjectureSkolemReplacement`? I understand it does not the same thing as the base class's variant. But then again, I think it only needs to be called if there is FOOL in the input (There was no `TermList transform(TermList ts);` in `TermTransformer` until @aztek extended it for FOOL processing. I don't think the main function `TermTransformer::transform(Term* term)` only indirectly via `TermTransformer::transformSpecial`. Does it mean you had to mix ConjectureSkolemReplacement and FOOL?)

MichaelRawson:
>streamlined TermTransformer::transform(TermList ts) (@MichaelRawson , could you please check whether you agree?)

Looks good to me - shorter and pretty sure that should have zero/positive perf impact.

hzzv (approved):
Thanks for checking with me! It looks good to me (and I also confirmed it doesn't change the functionality for synthesis).

@quickbeam123 - yes, `ConjectureSkolemReplacement` creates FOOL terms. Its output is the program we're synthesizing, and can thus contain if-then-elses, where the conditions are formulas.

quickbeam123:
OK, but then I want to suggest we stick with Petra's version for the parent class too!

I just went through all the child classes and none of them (I believe), except ConjectureSkolemReplacement, is ever running on non-shared/special/fool terms. So it should not matter for those, and ConjectureSkolemReplacement will be relieved of its override.

Unless you protest, I will change things and simply this PR further.

quickbeam123:
@hzzv , @MichaelRawson , could you give this one extra check, please?

MichaelRawson:
I'm happy. Took a little while, but we got there in the end! This looks nice as a final solution. :-)

hzzv:
I agree with Michael! And yes, it works for synthesis :)

quickbeam123:
Thank you!

PR #495, opened 2023-10-26
remove old option randomization

quickbeam123:
Old option randomization code is (mostly) obsolete and replaced by the `--sample_strategy`, let's remove.

I am not 100% sure about this. Maybe @selig wants to step in? 

The new mechanism uses an external file in a domain-specific language. The language has more power for sampling (e.g., samplers for continuous domains). Also pulling the metadata out of the executable has its advantages (and some disadvantages). In any case, an external file means we can have as many samplers as we want (FOF, HOL, SMT, INDUCTION, SAT, ...) whereas the old mechanisms effort to cater for both SAT and UNS was already getting a bit messy.

The new mechanism doesn't (yet) have problem-property-dependent sampling (which the old one did), but that could be added as an extension.

### DIFF
--- a/Shell/Options.cpp
+++ b/Shell/Options.cpp
@@ -119,7 +119,6 @@ void Options::init()
                                         "preprocess",
                                         "preprocess2",
                                         "profile",
-                                        "random_strategy",
                                         "smtcomp",
                                         "spider",
                                         "tclausify",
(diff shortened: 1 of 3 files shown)


selig:
The old randomisation code is not currently fit for purpose (it has too many gaps). Given the existence of a new shiny thing, it's unlikely it will be made fit for purpose so it makes sense to remove it.

PR #496, opened 2023-10-26
Change hashes in Induction so that iteration orders are deterministic

mezpusz:
Tested with portfolio mode and Discount over UFDTLIA benchmarks (only structural induction, integer induction uses the same structures apart from the induction term set). Diff shows no change in the proofs over multiple runs, and there is no major diff compared to master.


mezpusz:
My bad, I remembered that `TermList` is properly hashed (wrong) but did not check where content was coming from. I'll change it back and add the same hash function.

quickbeam123:
Cool!

PR #497, opened 2023-10-27
Make encompassment demodulation the default demodulation mode

mezpusz:
As discussed, I'm opening this PR to move forward with encompassment demodulation and remove the old, more expensive and harder to maintain standard check.

My initial measurements show the following over the entirety of TPTP, with 10s timeout (uniques in parentheses):
- Discount: old 10865 (93), encompassment 10977 (205)
- Otter: old 10583 (114), encompassment 10794 (325)

### DIFF
--- a/Inferences/ForwardDemodulation.cpp
+++ b/Inferences/ForwardDemodulation.cpp
@@ -58,8 +58,7 @@ void ForwardDemodulation::attach(SaturationAlgorithm* salg)
 	  _salg->getIndexManager()->request(DEMODULATION_LHS_CODE_TREE) );
 
   _preorderedOnly = getOptions().forwardDemodulation()== Options::Demodulation::PREORDERED;
-  _redundancyCheck = getOptions().demodulationRedundancyCheck() != Options::DemodulationRedunancyCheck::OFF;
-  _encompassing = getOptions().demodulationRedundancyCheck() == Options::DemodulationRedunancyCheck::ENCOMPASS;
+  _redundancyCheck = getOptions().demodulationRedundancyCheck();
 }
 
 void ForwardDemodulation::detach()
@@ -103,14 +102,10 @@ bool ForwardDemodulationImpl<combinatorySupSupport>::perform(Clause* cl, Clause*
         continue;
       }
 
-
-      bool toplevelCheck = _redundancyCheck &&
-        lit->isEquality() && (trm==*lit->nthArgument(0) || trm==*lit->nthArgument(1));
-
-      // encompassing demodulation is always fine into negative literals or into non-units
-      if (_encompassing) {
-        toplevelCheck &= lit->isPositive() && (cLen == 1);
-      }
+      // encompassing demodulation is always fine into negative literals,
+      // non-units or non-top-level terms
+      bool toplevelCheck = _redundancyCheck && lit->isEquality() && lit->isPositive()
+        && (cLen == 1) && (trm==*lit->nthArgument(0) || trm==*lit->nthArgument(1));
 
       TermQueryResultIterator git=_index->getGeneralizations(trm, true);
       while(git.hasNext()) {
@@ -176,7 +171,7 @@ bool ForwardDemodulationImpl<combinatorySupSupport>::perform(Clause* cl, Clause*
         }
 
         // encompassing demodulation is fine when rewriting the smaller guy
-        if (toplevelCheck && _encompassing) {
+        if (toplevelCheck) {
           // this will only run at most once;
           // could have been factored out of the getGeneralizations loop,
           // but then it would run exactly once there
@@ -191,32 +186,9 @@ bool ForwardDemodulationImpl<combinatorySupSupport>::perform(Clause* cl, Clause*
           TermList other=EqHelper::getOtherEqualitySide(lit, trm);
           Ordering::Result tord=ordering.compare(rhsS, other);
           if(tord!=Ordering::LESS && tord!=Ordering::LESS_EQ) {
-            if (_encompassing) {
-              // last chance, if the matcher is not a renaming
-              if (qr.substitution->isRenamingOn(qr.term,true /* we talk of result term */)) {
-                continue; // under _encompassing, we know there are no other literals in cl
-              }
-            } else {
-              Literal* eqLitS=qr.substitution->applyToBoundResult(qr.literal);
-              bool isMax=true;
-              for(unsigned li2=0;li2<cLen;li2++) {
-                if(li==li2) {
-                  continue;
-                }
-                if(ordering.compare(eqLitS, (*cl)[li2])==Ordering::LESS) {
-                  isMax=false;
-                  break;
-                }
-              }
-              if(isMax) {
-                //RSTAT_CTR_INC("tlCheck prevented");
-                //The demodulation is this case which doesn't preserve completeness:
-                //s = t     s = t1 \/ C
-                //---------------------
-                //     t = t1 \/ C
-                //where t > t1 and s = t > C
-                continue;
-              }
+            // last chance, if the matcher is not a renaming
+            if (qr.substitution->isRenamingOn(qr.term,true /* we talk of result term */)) {
+              continue; // under toplevelCheck, we know there are no other literals in cl
             }
           }
         }
--- a/Shell/Options.cpp
+++ b/Shell/Options.cpp
@@ -1552,23 +1552,21 @@ void Options::init()
     _condensation.onlyUsefulWith(ProperSaturationAlgorithm());
     _condensation.setRandomChoices({"on","off","fast"});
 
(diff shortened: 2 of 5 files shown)


Inferences/ForwardDemodulation.cpp
```diff
@@ -175,48 +179,13 @@ bool ForwardDemodulationImpl<combinatorySupSupport>::perform(Clause* cl, Clause*
           continue;
         }
 
-        // encompassing demodulation is fine when rewriting the smaller guy
-        if (toplevelCheck && _encompassing) {
-          // this will only run at most once;
```
quickbeam123, line 180: @mezpusz , could you please just quickly comment on why you moved this condition? (Not saying it's a bad idea, just curious about the reason.)

mezpusz: Initially, I wanted to move everything that sets `toplevelCheck` into one line but that would have been messy. But still, setting it at the beginning seems clearer to me.

Not I get it why you ask: the ordering check is never run if there are no generalizations, which may save a bit of time, right? We can move it back, but then could we add a comment that it does not even run there necessarily, hence the place?

quickbeam123: We had a comment:

// this will only run at most once;
// could have been factored out of the getGeneralizations loop,
// but then it would run exactly once there

Still, happy to hear also @MichaelRawson. Not sure myself whether the benefit is super clear or simply worth the degradation of clarity.

mezpusz: Sorry, I completely missed the "exactly" and "at most" part of the comment!

Shell/Options.cpp
```diff
@@ -1502,16 +1502,14 @@ void Options::init()
     _condensation.tag(OptionTag::INFERENCES);
     _condensation.onlyUsefulWith(ProperSaturationAlgorithm());
 
-    _demodulationRedundancyCheck = ChoiceOptionValue<DemodulationRedunancyCheck>("demodulation_redundancy_check","drc",DemodulationRedunancyCheck::ON,{"off","encompass","on"});
+    _demodulationRedundancyCheck = BoolOptionValue("demodulation_redundancy_check","drc",true);
```
MichaelRawson, line 1505: Do we know whether becoming incomplete is ever a good idea, strategy-scheduling wise?

Inferences/ForwardDemodulation.cpp
```diff
@@ -58,8 +58,7 @@ void ForwardDemodulation::attach(SaturationAlgorithm* salg)
 	  _salg->getIndexManager()->request(DEMODULATION_LHS_CODE_TREE) );
 
   _preorderedOnly = getOptions().forwardDemodulation()== Options::Demodulation::PREORDERED;
-  _redundancyCheck = getOptions().demodulationRedundancyCheck() != Options::DemodulationRedunancyCheck::OFF;
-  _encompassing = getOptions().demodulationRedundancyCheck() == Options::DemodulationRedunancyCheck::ENCOMPASS;
+  _redundancyCheck = getOptions().demodulationRedundancyCheck();
```
MichaelRawson, line 61: At some future point this could just be a single call to `env.options->demodulationRedundancyCheck()`, I doubt it's worth keeping the state around. Not now, however.

MichaelRawson (approved):
Much nicer, great job @mezpusz - I defer to @quickbeam123 about whether we lose much with this.

Inferences/ForwardDemodulation.cpp
mezpusz: I'm not sure anymore. The code _does_ get more readable by this PR regardless of this change, maybe we should put it back for now?

quickbeam123: I vote for shifting back. The diff will become smaller and it will be easier to visually check things are not getting more buggy than before :)

Shell/Options.cpp
quickbeam123: It is definitely a good idea (to keep drc=false). That's why Andrei was not so impressed by this when we explained encompass it in Vienna.

Experimentally: if we were comparing "old with drc=on:" Union "old with drc=off" this would most likely beat "new drc-on" (which does the encompass). @mezpusz, if you have energy, such experiment could be revealing. (Just for Discount is enough.)

mezpusz: Do you mean `drc=off` would beat encompassment? Sure, I can run on more experiment. 

Btw, I never understood why demodulation on negative literals was sometimes disallowed with the old `drc=on`. Wouldn't an `s != t` be always bigger than `s = r` where `s` is the maximal term of the three, regardless of how `t` and `r` compare? Maybe this is why `drc=off` was preferred over `drc=on`? We used `drc=off` in induction schedules a lot because otherwise we couldn't rewrite the maximal term in a ground disequation in many cases.

Inferences/ForwardDemodulation.cpp
mezpusz: I agree. Updated.

Shell/Options.cpp
mezpusz: Using Discount, the uniques changed as follows:
| `drc` | `on` | `off` |
|--------|--------|--------|
| old | 91 | 244 |
| new | 44 | 85 |
which means that encompassment makes the gap between `off` and `on` a bit tighter, but there is still a gap.

quickbeam123: > Completeness:

I admit I don't remember the theory by heart, but there is again a comment (that's now getting deleted) about it. The problematic case seems to be:

s = t into s != t1 \\/ C

with s > t (we know always) plus t > t1 plus {s = t} > C. I guess with t > t1 we have {s = t} > {s != t1} (the polarity only comes in last). And standard redundancy asks that the reducer be smaller than the reduced...

> Maybe this is why drc=off was preferred over drc=on?

I think it's simpler. Rewriting things is good. The more (with drc=off) the merrier. (Also its more efficient not to check exotic, corner-case conditions.) Compromising completeness can hurt, though. At least sometimes, if only rarely. (That's why we had drc=on.)

Thus: Why not be too excited about encompassment? Because most of the problems that encompass solves and old-drc=on does not, will be covered by drc=off.

> Experiment:

Could you please reveal plain number of solved for every row at least for one of on/off?

mezpusz: > there is again a comment (that's now getting deleted) about it.

I mean that that comment (and the corresponding implementation) was maybe a bit too restrictive. Of course it depends on how we define > between clauses and literals.

> Rewriting things is good. The more (with drc=off), the merrier.

This also makes sense, and it's hard to tell when a particular equality needed for completeness is used at all.

Anyways, here are the number of solved problems:
Otter off: 9813
Otter old on: 9562
Otter new on (encompass): 9776

Discount off: 9992
Discount old on: 9840
Discount new on (encompass): 9953

mezpusz: This is an interesting point of view which I haven't considered yet. You're right, encompassment+off gives less variation then the other pair. I fail to see why there are so many uniques of "old on" compared to encompassment, but the way it skips demodulations that both `off` and `encompass` allow gives more variation. One more question is whether we care about completeness or not.

But the bottomline is, I am not so sure about this PR anymore.

quickbeam123: My part: I have around 1300 random strategies evaluated (with varying instr limits) over 9090 FOF problems from the latest TPTP. The drc values were sampled with the following frequencies: encompass:500,on:500,off:354. (The strategies in the set, however, are not independent samples, as there was pressure to solve unsolved problems or at least improve solutions times.) Now these strategies cover 7210 of 9090 (this includes sat results, but should be negligible). If we 

1) drop the drc=off strategies, we lose
41 solvable problems, for 1503 we will only get a worse best solution time, accumulated instr_lost: 2005432 of 7833518 
2) drop the drc=encompass strategies, we lose
54 solvable problems, for 1816 we will only get a worse best solution time, accumulated instr_lost: 2223449 of 7833518 
3) drop the default drc=on (which is a bit tricky to filter, because defaults are not mentioned in the strat string, and for strategies with fd=bd=off the default does not meaningfully count), we lose
42 of solvable problems, for 1497 we will only get a worse best solution time, accumulated instr_lost 2966861 of 7833518
 
I think the conclusion is, from the point of view of strategy scheduling, that we want to keep all three option values.

Still, I like to code simplification too :)

mezpusz: Honestly, I don't understand much of the data but I believe you that we need all of them.:) So, my proposal is that we keep all three of them, maybe switch the default to `encompass`, or just leave this as it is and come back to it after the Vampire developer meeting (hope you're coming too!).

quickbeam123: I agree we should keep this open for now. It serves as a good example of a tricky decision we often face during vampire development and leaving is as an item for a developer meeting agenda could be useful.

PR #498, opened 2023-10-30
turn off run-time type information

MichaelRawson:
We currently compile in support for [run-time type information](https://en.wikipedia.org/wiki/Run-time_type_information), but we don't use it - except as a runtime type-safety check in `Indexing::RequestedIndex<T>`, which is relatively lightly-used.

Therefore: turn off RTTI in the build system, profit from smaller binaries.


quickbeam123 (approved):
An excellent idea!

PR #499, opened 2023-10-31
remove old allocator macros

MichaelRawson:
In #469 I reworked `Allocator` quite a bit, but some older macros were left as future work. The future is now, in three commits:
1. Remove `CLASS_NAME`, a no-op.
2. Remove `BYPASSING_ALLOCATOR` and friends; also a no-op but tended to introduce surrounding scope historically, which can now be squished.
3. Remove `*ALLOC_UNKNOWN`: these were actually still used, but for no good reason. The only "semantic" change, please pay most attention to this one.

`*ALLOC_KNOWN` and `USE_ALLOCATOR` are even more future work, but these may have a performance impact and have to be done more carefully.

### DIFF
--- a/Minisat/mtl/Vec.h
+++ b/Minisat/mtl/Vec.h
@@ -33,7 +33,7 @@ OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWA
 #include "Minisat/mtl/IntTypes.h"
 #include "Minisat/mtl/XAlloc.h"
 
-#include "Lib/Allocator.hpp"
+#include "Lib/Timer.hpp"
 
 namespace Minisat {
 
@@ -106,9 +106,10 @@ void vec<T,_Size>::capacity(Size min_cap) {
     if (cap >= min_cap) return;
     Size add = max((min_cap - cap + 1) & ~1, ((cap >> 1) + 2) & ~1);   // NOTE: grow by approximately 3/2
     const Size size_max = std::numeric_limits<Size>::max();
-    if ( ((size_max <= std::numeric_limits<int>::max()) && (add > size_max - cap))
-    ||   (((data = (T*) REALLOC_UNKNOWN((void *)data, (cap += add) * sizeof(T),"Minisat::vec")) == NULL) && errno == ENOMEM) )
+    if ( ((size_max <= std::numeric_limits<int>::max()) && (add > size_max - cap)))
         throw OutOfMemoryException();
+
+    data = (T*) xrealloc((void *)data, (cap += add) * sizeof(T));
  }
 
 
@@ -133,7 +134,7 @@ void vec<T,_Size>::clear(bool dealloc) {
     if (data != NULL){
         for (Size i = 0; i < sz; i++) data[i].~T();
         sz = 0;
-        if (dealloc) DEALLOC_UNKNOWN(data,"Minisat::vec"), data = NULL, cap = 0; } }
+        if (dealloc) xfree(data), data = NULL, cap = 0; } }
 
 //=================================================================================================
 }
--- a/Minisat/mtl/XAlloc.h
+++ b/Minisat/mtl/XAlloc.h
@@ -29,7 +29,7 @@ OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWA
 #include <cerrno>
 #include <cstdlib>
 
-#include "Lib/Allocator.hpp"
+#include "Lib/Timer.hpp"
 
 namespace Minisat {
 
@@ -39,7 +39,8 @@ namespace Minisat {
 class OutOfMemoryException{};
 static inline void* xrealloc(void *ptr, size_t size)
 {
-    void* mem = REALLOC_UNKNOWN(ptr, size, "Minisat::xrealloc");
+    Lib::TimeoutProtector tp;
+    void* mem = std::realloc(ptr, size);
     if (mem == NULL && errno == ENOMEM){
         throw OutOfMemoryException();
     }else
@@ -48,7 +49,8 @@ static inline void* xrealloc(void *ptr, size_t size)
 
 static inline void xfree (void* ptr)
 {
-  DEALLOC_UNKNOWN(ptr,"Minisat::xrealloc");
+    Lib::TimeoutProtector tp;
+    std::free(ptr);
 }
 
 //=================================================================================================
--- a/SAT/Z3Interfacing.cpp
+++ b/SAT/Z3Interfacing.cpp
@@ -115,17 +115,12 @@ Z3Interfacing::Z3Interfacing(SAT2FO& s2f, bool showZ3, bool unsatCoresForAssumpt
   _config(),
   _context(_config),
   _solver(_context),
-  _model((STATEMENTS_TO_EXPRESSION(
-            BYPASSING_ALLOCATOR;
-            _solver.check();
-          ),
-         _solver.get_model())),
+  _model((STATEMENTS_TO_EXPRESSION(_solver.check();), _solver.get_model())),
   _assumptions(),
   _showZ3(showZ3),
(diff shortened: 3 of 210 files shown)


Minisat/mtl/Vec.h
```diff
@@ -106,9 +106,10 @@ void vec<T,_Size>::capacity(Size min_cap) {
     if (cap >= min_cap) return;
     Size add = max((min_cap - cap + 1) & ~1, ((cap >> 1) + 2) & ~1);   // NOTE: grow by approximately 3/2
     const Size size_max = std::numeric_limits<Size>::max();
-    if ( ((size_max <= std::numeric_limits<int>::max()) && (add > size_max - cap))
-    ||   (((data = (T*) REALLOC_UNKNOWN((void *)data, (cap += add) * sizeof(T),"Minisat::vec")) == NULL) && errno == ENOMEM) )
+    if ( ((size_max <= std::numeric_limits<int>::max()) && (add > size_max - cap)))
         throw OutOfMemoryException();
+
+    data = (T*) xrealloc((void *)data, (cap += add) * sizeof(T));
```
quickbeam123, line 112: I `xrealloc` expected to do the throwing corresponding to the second half of the original if-condition?

Minisat/mtl/XAlloc.h
```diff
@@ -39,7 +39,8 @@ namespace Minisat {
 class OutOfMemoryException{};
 static inline void* xrealloc(void *ptr, size_t size)
 {
-    void* mem = REALLOC_UNKNOWN(ptr, size, "Minisat::xrealloc");
+    Lib::TimeoutProtector tp;
+    void* mem = std::realloc(ptr, size);
```
quickbeam123, line 43: Why do we sometimes (e.g., in `Minisat/mtl/Vec.h`) use `xrealloc`/`xfree` and here `std::realloc`/`str::free`?

Minisat/mtl/XAlloc.h
```diff
@@ -48,7 +49,8 @@ static inline void* xrealloc(void *ptr, size_t size)
 
 static inline void xfree (void* ptr)
 {
-  DEALLOC_UNKNOWN(ptr,"Minisat::xrealloc");
+    Lib::TimeoutProtector tp;
```
quickbeam123, line 52: Kudos for adding the `TimeoutProtector`!

Minisat/mtl/XAlloc.h
MichaelRawson: So (I assume) all the `*ALLOC_UNKNOWN` calls in Minisat used to be calls to Minisat's `xrealloc` or `xfree`. Working on that assumption I reverted it - which might be bogus, but seems to be "reasonable enough". Here it's `std::realloc` because we are *in* `xrealloc`.

Minisat/mtl/Vec.h
MichaelRawson: Yes, it should.

SAT/Z3Interfacing.cpp
```diff
@@ -115,17 +115,12 @@ Z3Interfacing::Z3Interfacing(SAT2FO& s2f, bool showZ3, bool unsatCoresForAssumpt
   _config(),
   _context(_config),
   _solver(_context),
-  _model((STATEMENTS_TO_EXPRESSION(
-            BYPASSING_ALLOCATOR;
-            _solver.check();
-          ),
-         _solver.get_model())),
+  _model((STATEMENTS_TO_EXPRESSION(_solver.check();), _solver.get_model())),
```
quickbeam123, line 118: I think we now could go to
```
_model(_solver.check(),_solver.get_model()),
```
and delete the magical macro `STATEMENTS_TO_EXPRESSION`. What do you think?

MichaelRawson: Yes - I saw this and scratched my head a few times, decided I would come back to it when I was more certain. If you think it should go too, I'm happy.

Shell/Naming.cpp
```diff
@@ -317,14 +313,12 @@ Formula* Naming::apply_iter(Formula* top_f) {
...
+            delete[] gs;
           } else if (fs != f->args()) {
             f = new JunctionFormula(AND, fs);
           }
-          DEALLOC_UNKNOWN(sand.cls, "Naming::apply");
-          if (sand.negCls) {
-            DEALLOC_UNKNOWN(sand.negCls, "Naming::apply");
-          }
+          delete[] sand.cls;
+          delete[] sand.negCls;
```
quickbeam123, line 321: I assume delete already checks for 0 / not-0, right?

MichaelRawson: Right - standard apparently says that either (i) nothing will happen, or (ii) the deletion function will actually be called, but our deletion function (and that in the standard library) handle this case already. If not it's a bug there, not here.

Shell/Naming.cpp
```diff
@@ -145,12 +145,10 @@ Formula* Naming::apply_iter(Formula* top_f) {
       case AND: {
         FormulaList* fs = tas.f->args();
         unsigned length = FormulaList::length(fs);
-        void* mem = ALLOC_UNKNOWN(length * sizeof(int), "Naming::apply");
-        int* cls = array_new<int>(mem, length);
+        int *cls = new int[length]();
```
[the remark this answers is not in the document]
MichaelRawson: Default-initialises the allocated memory (which I think is zero for `T = int`). I'm not sure this is necessary, but it's the semantics of the previous code with `array_new` so I preserved it.

Minisat/mtl/Vec.h
quickbeam123: Actually, orig minisat does (https://github.com/niklasso/minisat/blame/master/minisat/mtl/Vec.h):

```
template<class T, class _Size>
void vec<T,_Size>::capacity(Size min_cap) {
    if (cap >= min_cap) return;
    Size add = max((min_cap - cap + 1) & ~1, ((cap >> 1) + 2) & ~1);   // NOTE: grow by approximately 3/2
    const Size size_max = std::numeric_limits<Size>::max();
    if ( ((size_max <= std::numeric_limits<int>::max()) && (add > size_max - cap))
    ||   (((data = (T*)::realloc(data, (cap += add) * sizeof(T))) == NULL) && errno == ENOMEM) )
        throw OutOfMemoryException();
 }
```

I wonder what `(T*)::realloc` even is. Some kind of "look in your namespace first"?

Minisat/mtl/XAlloc.h
quickbeam123: Oh yes, forgive me my stupidity.

SAT/Z3Interfacing.cpp
quickbeam123: Please!

Minisat/mtl/Vec.h
MichaelRawson: I think that's "call the global realloc and cast to `T *`". I'm not sure if there is a difference between `realloc` and `std::realloc`.

Shell/Naming.cpp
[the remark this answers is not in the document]
quickbeam123: Oh, this was a long rabbit hole you just made me research ;)

I think it's actually getting "value-initialized" (according to this: https://stackoverflow.com/questions/8106016/c-default-initialization-and-value-initialization-which-is-which-which-is-ca)

The last answer on https://stackoverflow.com/questions/39909411/what-is-the-default-value-of-an-array-in-c was the my first hit about the expression even making any sense. I didn't know you could get anything else for array new than "no initialization" for base types and "default constructor" (always) for custom types.

And I agree this aligns well with what was there before!

quickbeam123 (approved):

Shell/Naming.cpp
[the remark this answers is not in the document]
MichaelRawson: >I think it's actually getting "value-initialized"

Quite right - I was sloppy, I still don't really know which initialisation is which in C++. Certainly the "initialisation that zero-initialises POD and calls constructors".

PR #502, opened 2023-11-16
do not include congruence axioms for unused symbols

### LINKED ISSUE #501: Some useless congruence axioms by equality proxy under arithmetic inputs
Under ep=RSTC, Vampire may introduce congruence axioms for `$greater` and the likes (although they are already eliminated by the `InterpretedNormalizer` by then). 

The problem is that once parsed, predicates like $greater (and functions like binary minus) are part of the signature and the code in EqualityProxy(Mono) will iterate over them.

MichaelRawson:
Fixes #501 if correct.

I believe that if we make sure we refresh the property object, we can rely on `sym->usageCnt()` to detect unused symbols and avoid generating congruence axioms for them.

Not sure if the refresh is actually needed, seems to work without it - but better safe than sorry.

### DIFF
--- a/Shell/Preprocess.cpp
+++ b/Shell/Preprocess.cpp
@@ -393,14 +393,13 @@ void Preprocess::preprocess(Problem& prb)
      twee.apply(prb,(env.options->tweeGoalTransformation() == Options::TweeGoalTransformation::GROUND));
    }
 
-   // refresh symbol usage counts, can skip unused symbols for equality proxy
-   prb.getProperty();
-
    if (!prb.isHigherOrder() && _options.equalityProxy()!=Options::EqualityProxy::OFF && prb.mayHaveEquality()) {
      env.statistics->phase=Statistics::EQUALITY_PROXY;
      if (env.options->showPreprocessing())
        env.out() << "equality proxy" << std::endl;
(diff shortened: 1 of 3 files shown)


Shell/Preprocess.cpp
```diff
@@ -393,6 +393,9 @@ void Preprocess::preprocess(Problem& prb)
      twee.apply(prb,(env.options->tweeGoalTransformation() == Options::TweeGoalTransformation::GROUND));
    }
 
+   // refresh symbol usage counts, can skip unused symbols for equality proxy
+   prb.getProperty();
```
quickbeam123, line 397: How about we hide this below under the `if` so that we tax only `the ep!=off` branch?

MichaelRawson: Good idea, fixed.

MichaelRawson: Actually could even further optimise and do this only if `ep == RSTC`, because we only need to know for congruence axioms. Let me know if you care.

quickbeam123: I think this is fine, thank you!

quickbeam123 (approved):

PR #508, opened 2023-12-01
Refactoring and Extending Unification with Abstraction

joe-hauns:
Main motivation for this PR was implementing our LPAR23 paper "Refining Unification With Abstraction". In this paper we introduced a new algorithm for unification with abstraction that generalizes the original unification with abstraction (UWA) algorithm, and gives us more control over how and when to perform UWA.
For example in the old algorithm when unifying `f(a + x + y)` with `f(b + a)` we were only introduce a constraint `a + x + y != b + a`, while the new algorithm allows introducing the simplified constraint `x + y != b` straight away.
Another example is unifying `f(a + x + b)` with `f(b + a)`, where we now can straight derive the substitution `{ x -> 0 }`, instead of introducing the constraint `a + x + b != b + a`.
More examples motivation and explanation can be found in the paper.

In order to implement this new algorithm without compromising the performance of ordinary unification there are now two distinct unification algorithms: 
- `RobSubstition::unify` 
   classic Rob unification algorithm as we had before the first introduction of unification with abstraction
- `AbstractingUnifier::unify` 
  the new unification with abstraction algorithm that we introduced in our LPAR23 paper

Additionally to do unification with abstraction as introduced in the paper we need to be able to create `TermSpec`s in unification that contain variables of two different variable banks:
Say we want to unify
`x + f(y)    <- var bank 0`
and   
`f(y)        <- var bank 1`
Then an mgu is x -> -f(y/0) + f(y/1)
This cannot be directly represented in vampire as our TermSpec can only hold 1 variable bank per term, and not multiple per subterm. So what we want to do instead is introduce two 
new glue variables varibles G0, G1
```
G0 -> -f(y)/0
G1 ->  f(y)/1
```
and return as mgu
`{x -> G0 + G1, G0 -> -f(y)/0, G1 -> f(y)/1}`
For this in RobSubstition we introduced a new constant `GLUE_INDEX`, and helper functions `createTerm`, and `introGlueVariable`.
 
Further quite a lot of tidying/refactoring around substitution trees and how they interface with UWA as well as some library code (e.g. Metaiterators that is being used there) has been done.
The main changes are:
- Query Results:
  Whenevery we called unify on a substitution tree we used to get a SLQueryResult or a TermQueryResult which contains some data (TermList, Literal, Clause), as well as a unifier.
  The unifier used to always be a pair (Constraints, Substitution), no matter what query we used to make on the index. This means most of the time (e.g. when we didn't use unification with abstraction but normal unificatio, generalization, or instantiation we would get a nullpointer constraints). This means functions that deal with a {Term,Literal}QueryResult had to check with assertions whether the constraints are non-null and had callers of the functions had to know whether the functions can actually deal with constraints and so on. 
  I moved this to be dealt with the typesystem now.
  Now `TermQueryResult` is parametrized by the unifier:
  - `TQueryRes<AbstractingUnifier*>` for unification with abstraction
  - `TQueryRes<ResultSubstititionSP>` for instantiation and generalization
  Same for ``LiteralQueryResult`.
  In the future I also want to split `ResultSubstitution`  into `UnifyingSubstitution`, and `GeneralizingSubstitution`, as both of them have very different behaviour but use the same interface and just throw exceptions when the wrong function is called which is very hard to work with if you don't know the codebase well. Parametrizing The results will also allow for such a refactor.

  We have now different functions to query unfication with and without abstraction:
  ```
  VirtualIterator<          ResultSubstititionSP> getUnifications(TypedTermList lit, bool complementary, bool retrieveSubstitutions = true);
  VirtualIterator<TQueryRes<AbstractingUnifier*>> getUwa(TypedTermList t, Options::UnificationWithAbstraction uwa, bool fixedPointIteration);
  ```
- removal of various not needed fields and parameters in substitution tree that were associated with unification with abstraction but that are not needed anymore (e.g. useC that enabled making it possible to use unifiacion with abstraction with the related SubstitutionTree; we can now use UWA with and substitution tree hence we don't need the flag, rfSubs (this was related to very special variables which i got rid of all along), ...)

- shortened the RobSubstition code for `apply` and `getApplicationResultWeight` using BottomUpEvaluation instead of manual bottom up traversal. A performance bottlekneck that came from this change was how term sharing used to work:
  We create allocate term, then check if it already exists, and if so we deallocate it again and return the other one. 
  In order to get rid of this unnecessary allocation i added functionality to `Set` to check whether an element exists before actually allocating it the relevant function is `Set::rawFindOrInsert`, and is documented in the hpp file.

- replaced `VirtualIterator<void>` by `VirtualIterator<tuple<>>`, in `LookaheadLiteralSelector` as this allows for it being usable in combination with `IterTraits`. I do not remember if/why this was needed and can change it back if anyone opposes for some reason.

- extended and optimized general functionality Coproduct.hpp & Option.hpp

- improved Recycler

- added macros to automatically implement Hash, Equals, and Comparison operators for custom types to `Lib/Reflection.hpp`

- added unit tests to `UnitTests/tUnificationWithAbstraction.cpp` (this is where +1000 lines comes from which makes the PR seem like it's adding loads of lines of code). The test in there also test UWA in the context of substition trees


quickbeam123:
Some randomized testing with discount10 on TPTP:
```
Master: Expect 8746.846343722325 +/- 11.23557282358998
Branch: Expect 8763.348315603816 +/- 11.218682373124999
```
So some nice performance improvement!

Shell/Options.cpp
```diff
@@ -3356,6 +3365,11 @@ bool Options::complete(const Problem& prb) const
     return false;
   }
 
+  if (unificationWithAbstraction() != UnificationWithAbstraction::OFF) {
+    // can, depending on concrete UWA option and rules it is used in result in saturation, where using normal unification rules would still be applicable
```
quickbeam123, line 3369: I find this comment hard to understand!

Shell/Options.cpp
```diff
@@ -1738,6 +1744,9 @@ void Options::init()
     _maximumXXNarrows.addProblemConstraint(hasHigherOrder());    
     _maximumXXNarrows.tag(OptionTag::HIGHER_ORDER);
 
+    // TODO we have two ways of enabling function extensionality abstraction atm:
```
quickbeam123, line 1747: A TODO? "We should sort this out before merging into master." ;)

Shell/Options.cpp
```diff
@@ -1172,6 +1172,12 @@ void Options::init()
     _unificationWithAbstraction.tag(OptionTag::THEORIES);
     _lookup.insert(&_unificationWithAbstraction);
 
+    _unificationWithAbstractionFixedPointIteration = BoolOptionValue("unification_with_abstraction_postpro","uwa_fpi",
+                                     false);
+    _unificationWithAbstractionFixedPointIteration.description="TODO";
+    _unificationWithAbstractionFixedPointIteration.tag(OptionTag::INFERENCES);
```
quickbeam123, line 1178: Descriptions are also nice!

Shell/Rectify.cpp
```diff
@@ -319,73 +314,57 @@ Literal* Rectify::rectify (Literal* l)
...
-  // literal not changed
-  m->destroy();
-  return l;
 } // Rectify::rectify (Literal*)
 
 /**
  * Rectify a list of terms @b from to the list of terms @b to.
  * Return true if the list has changed.
  * @since 24/03/2008 Torrevieja
  */
```
quickbeam123, line 345: Here the generalization makes the old comment outdated.

MichaelRawson (approved):
Really nice. I'm trusting you with the new tests, the substitution tree stuff and anything in UWA - but the rest looks great to me. Especially the iterators and BUE seem to be "coming of age". :-) I think there's a few things that could be fixed but generally I approve.

Indexing/ClauseCodeTree.hpp
```diff
@@ -97,6 +97,7 @@ class ClauseCodeTree : public CodeTree
   {
     void init(ClauseCodeTree* tree_, Clause* query_, bool sres_);
     void reset();
+    bool keepRecycled() const { return lInfos.keepRecycled(); }
```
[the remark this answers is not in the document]
MichaelRawson: Ahah - later I found it's part of `Recycled`. I assume this checks whether some object actually has any resources worth keeping.

MichaelRawson:
Merge conflicts are probably due to my crude `USE_ALLOCATOR` deletion spree. Thinking about it I should have waited with that until this was in. Oh well - it's probably quite easy to fix @joe-hauns, if I've deleted a `USE_ALLOCATOR` it can probably go on this branch too unless you have some reason to believe it's worthwhile.

Shell/Options.cpp
joe-hauns: updated it. Better now?

Shell/Options.cpp
joe-hauns: Yesss absolutely. I'm not quite sure what to do about it:
The old way was the option `-fe`, and now it is its own `-uwa` option, which is better I think. So we could get rid of `-fe abstraction`, but some of the schedules might rely on the old option, right? What do you think?

Indexing/CodeTreeInterfaces.hpp
```diff
@@ -47,32 +47,16 @@ class CodeTreeTIS : public TermIndexingStructure
 
   TermQueryResultIterator getGeneralizations(TypedTermList t, bool retrieveSubstitutions = true);
   bool generalizationExists(TermList t);
+  // TODO: get rid of NOT_IMPLEMENTED
```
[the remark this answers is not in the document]
joe-hauns: Exactly. I think we should have not one index trait that has the functions `getUwa` and `getUnification` but two interfaces `UwaIndex` and `UnificationIndex` (and `GenIndex` etc) or something like that. It's not 100% thought out but this is what I was thinking about.

Indexing/IndexManager.hpp
```diff
@@ -98,6 +99,8 @@ class IndexManager
   DHMap<IndexType,Entry> _store;
 
   Index* create(IndexType t);
+  Shell::Options::UnificationWithAbstraction _uwa;
+  bool _uwaFixedPointIteration;
```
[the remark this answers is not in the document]
joe-hauns: We cannot quite use that. For `_uwaFixedPointIteration` we can, but not for `_uwa` as we have this interplay between the options `-fe abstraction` vs `-uwa function_extensionality` and then additionally the constraint that function extensionality abstaction only being enabled when the problem is higher order. Soooo I'm not quite sure what to do. I think the big question is, do we care about the current higher order schedules, or do we drop them anyways because there's ahmed's better higher order branch, and we don't maintain our higher order stuff until that get's merge and we replace the master stuff in favor of ahmed's new things. @ibnyusuf @quickbeam123 , what do you think?

Shell/Options.cpp
joe-hauns: (@MichaelRawson @ibnyusuf Have a look at this comment as well for context)

Indexing/LiteralIndex.cpp
```diff
@@ -41,17 +42,6 @@ SLQueryResultIterator LiteralIndex::getAll()
...
-	  bool complementary, bool retrieveSubstitutions)
-{
-  return _is->getUnifications(lit, complementary, retrieveSubstitutions);
-}
-
-SLQueryResultIterator LiteralIndex::getUnificationsWithConstraints(Literal* lit,
-          bool complementary, bool retrieveSubstitutions)
-{
-  return _is->getUnificationsWithConstraints(lit, complementary, retrieveSubstitutions);
-}
```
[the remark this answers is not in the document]
joe-hauns: it's been moved to the header (because templated) and the new function name is `getUwa`

Indexing/LiteralSubstitutionTree.cpp
```diff
@@ -28,67 +31,35 @@
...
 : _trees(env.signature->predicates() * 2)
-, _useC(useC)
 { }
 
 SLQueryResultIterator LiteralSubstitutionTree::getUnifications(Literal* lit, bool complementary, bool retrieveSubstitutions)
-{ return getResultIterator<UnificationsIterator>(lit, complementary, retrieveSubstitutions, /* constraints */ false); }
-
-SLQueryResultIterator LiteralSubstitutionTree::getUnificationsWithConstraints(Literal* lit, bool complementary, bool retrieveSubstitutions)
-{ return getResultIterator<UnificationsIterator>(lit, complementary, retrieveSubstitutions, /* constraints */ true); }
+{ return pvi(getResultIterator<SubstitutionTree::Iterator<RetrievalAlgorithms::RobUnification>>(lit, complementary, retrieveSubstitutions)); }
```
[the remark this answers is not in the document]
joe-hauns: Exactly

Kernel/RobSubstitution.cpp
```diff
@@ -116,320 +120,217 @@ bool RobSubstitution::isUnbound(VarSpec v) const
...
   //Aux terms don't contain special variables, ergo
   //should be shared.
   //ASS(!b.term.isTerm() || b.index!=AUX_INDEX || b.term.term()->shared());
   ASS_NEQ(v.index, UNBOUND_INDEX);
 
-  if(bdIsRecording()) {
-    bdAdd(new BindingBacktrackObject(this, v));
-  }
-  _bank.set(v,b);
+  bind(_bindings, v, std::move(b));
```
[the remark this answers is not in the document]
joe-hauns: it actually is. But it wasn't at some point on the refactoring journey. I changed this.

Kernel/RobSubstitution.cpp
```diff
@@ -578,184 +496,43 @@ Literal* RobSubstitution::apply(Literal* lit, int index) const
...
-    }
-    terms.push(t);
-    termRefVars.push(vs);
-
-    toDo.push(t->args());
-    toDoIndex.push(ts.index);
-  }
-  ASS(toDo.isEmpty() && toDoIndex.isEmpty() && terms.isEmpty() && argSizes.length()==1);
-  return argSizes.pop();
+  return BottomUpEvaluation<AutoDerefTermSpec, size_t>()
```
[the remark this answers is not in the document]
joe-hauns: You mean builder patterns for different things, or `BottomUpEvaluation` in different places?

Kernel/RobSubstitution.hpp
```diff
@@ -17,55 +17,331 @@
...
+  VarSpec(unsigned var, int index) : var(var), index(index) {}
+
+  friend std::ostream& operator<<(std::ostream& out, VarSpec const& self);
+
+  /** number of variable */
+  unsigned var;
+  /** index of variable bank */
+  int index;
+
+  auto asTuple() const { return std::tie(var, index); }
```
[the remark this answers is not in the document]
joe-hauns: `std::tie` doesn't copy but creates references, so if you'd call `std::tie(_myStack, _myInt)` it will be of type `std::tuple<Stack<T>&, int&>` instead of copying the stack.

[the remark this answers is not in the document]
joe-hauns: I just always use `std::tie` for the `asTuple` functions so it's future proof if someone adds a field that is not trivially copy constructible.

Kernel/RobSubstitution.hpp
```diff
@@ -17,55 +17,331 @@
...
 
+
+const int GLUE_INDEX=-3;
+const int SPECIAL_INDEX=-2;
+const int UNBOUND_INDEX=-1;
 namespace Kernel
 {
+struct VarSpec
+{
+  VarSpec() {}
```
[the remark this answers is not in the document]
joe-hauns: Alas not because `DHMap::Entry` calls default constructors. I thought about changing that because it's often quite annoying but i wasn't arsed (yet) because it's quite some effort.

joe-hauns:
> Merge conflicts are probably due to my crude `USE_ALLOCATOR` deletion spree. Thinking about it I should have waited with that until this was in. Oh well - it's probably quite easy to fix @joe-hauns, if I've deleted a `USE_ALLOCATOR` it can probably go on this branch too unless you have some reason to believe it's worthwhile.

From my side all `USE_ALLOCATOR` calls can go. I just put them there to satisfy the old allocator and not get any assertion violations because of that.

quickbeam123:
I tried to merge master in, to start a new round of tests (some bugs got fixed in master), but the merge felt tricky. Could you please do the merge when you have time?

Kernel/RobSubstitution.cpp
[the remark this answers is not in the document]
MichaelRawson: The latter - seems like there's lots of manual stack emulation that essentially builds up a term. I had no particular examples in mind, I was just pleased. :smile:

MichaelRawson:
>From my side all USE_ALLOCATOR calls can go. I just put them there to satisfy the old allocator and not get any assertion violations because of that.

OK - my last deletion round was just the "obviously useless" ones - if you have any objects that are small and frequently allocated you might want to keep it for now otherwise you might dent performance on this branch without meaning to. Of course, the rest can go!

Indexing/IndexManager.hpp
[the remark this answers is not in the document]
ibnyusuf: Higher-order support on `ahmed-new-hol` is an immense improvement on what we have in current master. However, it also conflicts with a lot of what is in this PR, so it will be a large task to get it in on top of this.

My thinking is, @joe-hauns , if you feel confidant that you will be able to merge in the new higher-order branch down th eline then it is fine to break current higher-order support and modify the compiler to reject all higher-order problems. This would be temporary until `ahmed-new-hol` is merged.

On the other hand, if you are not sure whether you will get around to merging `ahmed-new-hol`, then breaking current HOL support on `master ` would leave Vampire without higher-order support on master indefinitely.

I will be able to assist with merging in `ahmed-new-hol` if you go down that route.

Shell/Options.cpp
ibnyusuf: Please see my comment below with regards to breaking higher-order support on current master.

Indexing/IndexManager.hpp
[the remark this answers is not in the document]
joe-hauns: So I think that branch we already talked about at some point this summer, right? I think the plan was then to reimplement it as it diverged too much from master, right? So my plan was & is to take on this work (and of course work with you on that @ibnyusuf ). To that i committed to in the dev meeting the other week in vienna. There we said we'd keep the current branch `ahmed-new-hol` (assuming this is what entered last years CASC) as the go-to higher order branch until i did the merging/reimplementing.

So if this is the plan anyways we do not need to bother about the old HOL things that are in this PR too much, as we are going to replace them by other stuff and have a better HOL branch as well.  (Of course I'm not suggesting to just kick everything out but I mean not bothering about little changes in which options are set and so on).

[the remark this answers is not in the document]
ibnyusuf: Great, in that case I'd say break current higher-order support in master to your heart's content! 

The safest thing to do would be to add a few lines of code to the parser to reject higher-order problems perhaps with a message to say that higher-order support will be returning in the near future. I don't know that we have any users of higher-order Vampire, but in case anyone does try it, that will be a lot easier to understand than failing in a subtle, difficult to understand manner.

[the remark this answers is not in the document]
MichaelRawson: @hetzenmat may be interested on working on Vampire's HOL support part-time: I'll try to organise a meeting in the last week of January with Andrei so we can discuss how to work on this.

SAT/Z3Interfacing.cpp
```diff
@@ -49,19 +49,19 @@ struct BottomUpChildIter<z3::expr>
   z3::expr _self;
 
   /** constructs an iterator over the children of the current node */
-  BottomUpChildIter(z3::expr a) : _idx(0), _self(a) {}
+  BottomUpChildIter(z3::expr a, EmptyContext = EmptyContext()) : _idx(0), _self(a) {}
```
quickbeam123, line 52: Hi Joe, coming back to reading the diff for this PR.

Here, I am getting quite baffled. What is `EmptyContext` and why does a partial specialization of `BottomUpChildIter` need one in its constructor signature (and elsewhere), while also immediately ignoring it (and at the same time supplying a default argument for it)?

This might not be an issue of Z3Interfacing, but perhaps of some (arguably) missing documentation in `BottomUpEvaluation.hpp`? I am sure you hade done some cool work there, but if future vampire developers are to benefit from it, we need more "self-explanatory-dness" or perhaps more comments somewhere. (Some comments I sure might have missed - feel free to point me where appropriate ;))

Lib/Stack.hpp
```diff
@@ -124,6 +129,45 @@ class Stack
     std::swap(*this,s);
   }
 
+/**
```
quickbeam123, line 132: This function is cool! :)

Lib/Stack.hpp
```diff
@@ -190,7 +234,7 @@ class Stack
   template<class It>
   static Stack fromIterator(It it) {
     Stack out;
-    out.loadFromIterator(it);
+    out.moveFromIterator(std::move(it));
```
quickbeam123, line 237: Was it OK to change the semantics (as I assume this did) from copy to move for all the callers? 

Would it make sense to call this convenience function also moveFromIterator to stress this?

Lib/Stack.hpp
MichaelRawson: I agree! But...thinking more about it, won't the second stack (on destruction or resize) try to de-allocate a pointer halfway into an allocated region?

Lib/Stack.hpp
```diff
@@ -304,22 +373,7 @@ class Stack
    * @since 11/03/2006 Bellevue
    */
   inline
-  void push(const C& elem)
-  {
```
quickbeam123, line 308: Just curious: did you realize neither reference nor rvalue reference make sense in this context?

Lib/Stack.hpp
quickbeam123: A good point! 

Even on destruction, every list tries to deallocate.

And these will pass a wrong number of elements to `DEALLOC_KNOWN` because neither will know the original `_capacity`?

Lib/Sort.hpp
```diff
@@ -26,21 +26,44 @@
 
 namespace Lib {
 
+template<class T>
+struct HasCompareFunction 
+{ 
+  template<class U> 
+  static constexpr decltype(std::declval<U const&>().compare(std::declval<U const&>()), bool()) check(int) { return true; }
```
quickbeam123, line 33: Sheer sorcery!

SAT/Z3Interfacing.cpp
joe-hauns: There's actually some documentation missing. I just pushed an update!

Lib/Stack.hpp
joe-hauns: > I agree! But...thinking more about it, won't the second stack (on destruction or resize) try to de-allocate a pointer halfway into an allocated region?

Yesss exactly! Sounds like this is not allowed then...? I wasn't aware of that. I think the function is not in use actually. I wrote it at some point coz i needed it and kept it coz it seems useful but we can remove it again I think.

Lib/Stack.hpp
joe-hauns: I actually just realized that we don't need the moveFromIterator function, because if the iterator returns a non-reference value the expression `it.next()` will be a rvalue reference anyways so the `std::move` is not needed except if the iterator returns a lvalue reference in which the move is actually not what I wanted to do. So this is a kind-of-bug and i just changed it.

Lib/Stack.hpp
joe-hauns: Yes exactly. So if we take  a regular value `C` instead of a reference, then it's up to the caller whether they wanna move or copy into the function, while if we use `C const&` we force the caller to copy and if we use `C&&` we force the caller to move.

Lib/Sort.hpp
joe-hauns: it is a real pain indeed. I added some documentation for usage.

Lib/Hash.hpp
```diff
@@ -215,6 +237,17 @@ class DefaultHash
     return hash;
   }
 
+  template<class Iter>
+  static unsigned hashIter(
+      Iter iter,
+      unsigned hash = FNV32_OFFSET_BASIS
+      ) {
+    while (iter.hasNext()) {
+      hash = (hash ^ iter.next()) * FNV32_PRIME;
```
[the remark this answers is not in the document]
quickbeam123: Is this still an open thing?

Shell/Options.cpp
quickbeam123: Let me try to understand better: is this TODO only about redundancy (having two ways of achieving the same) where the one way is new and the other is a now deprecated-ish HO code (superseded by the ahmed-new-hol branch)?

joe-hauns: Yess exactly. So the function extensionality abstraction used to be done in a different way before. it was a special mechanism using `VSpecialVar`s and such, but now instead it's being done within the UWA framework. This is why before we had a separate option to turn it on, while now it is just a UWA option variant.  In order to keep compability with old schedules i kept the `-fe abstraction` option.

Kernel/Term.hpp
```diff
@@ -171,7 +176,35 @@ class TermList {
   /** make the term empty (so that isEmpty() returns true) */
   inline void makeEmpty()
   { _content = FUN; }
+  inline static TermList empty()
+  { TermList out; out.makeEmpty(); return out; }
   /** make the term into a reference */
```
quickbeam123, line 181: The comment above is now misplaced!

Also, could you please add a bit of comment explaining what `Top` is for?

Kernel/Term.hpp
```diff
@@ -966,21 +1016,54 @@ class Literal
...
+    if (functor != lit->functor() || polarity != lit->polarity()) return false;
+
+    if (commutative) {
+      ASS_EQ(arity, 2)
+      ASS(rightArgOrder(getArg(0), getArg(1)))
+
+      if (someIf(lit->isTwoVarEquality(), [&](){ return lit->twoVarEqSort(); }) != twoVarEqSort) {
+        return false;
+      }
+      return std::make_tuple(*lit->nthArgument(0), *lit->nthArgument(1)) == std::make_tuple(getArg(0), getArg(1));
```
quickbeam123, line 1036: What is the magic about make_tuple here? Is this a fancy way of saying 

`(*lit->nthArgument(0) == getArg(0)) && (*lit->nthArgument(1) == getArg(1))
`
or is more going on here?

Also, can we be sure that `rightArgOrder` also holds for `lit`?

Kernel/Term.cpp
```diff
@@ -104,6 +109,9 @@ void Term::destroyNonShared()
   }
 }
 
+bool TermList::ground() const
+{ return isTerm() && term()->ground(); }
+
```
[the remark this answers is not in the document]
quickbeam123: Should we move it, then?

Kernel/Term.cpp
```diff
@@ -866,26 +870,31 @@ Term* Term::create(unsigned function, unsigned arity, const TermList* args)
...
+  bool share = range(0, arity).all([&](auto i) { return argSafeToShare(args[i]); });
 
-  const TermList* curArg = args;
-  const TermList* argStopper = args+arity;
-  while (curArg!=argStopper) {
-    *ss = *curArg;
-    --ss;
-    if (!curArg->isSafe()) {
-      share = false;
+  auto allocTerm = [&]() {
```
quickbeam123, line 875: Do you, guys, know whether capturing word-sized creatures (ints, pointers) in lambdas by reference makes a difference, performance-wise to passing-by value?

joe-hauns: i don't know for sure but i would be surprised if the compiler creates an actual reference in a case like this. Here the lambda should be completely inlined as it's just called in the `Set::rawFindOrInsert`. But in general (say if the closure was stored somewhere and the reference is only used later) it might make a difference i reckon. But that's all just what I think should be going on so not 100% sure.

Kernel/LookaheadLiteralSelector.cpp
```diff
@@ -46,7 +47,7 @@ using namespace Saturation;
  */
 struct LookaheadLiteralSelector::GenIteratorIterator
 {
-  DECL_ELEMENT_TYPE(VirtualIterator<void>);
+  DECL_ELEMENT_TYPE(VirtualIterator<std::tuple<>>);
```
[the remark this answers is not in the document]
quickbeam123: I think it's some form of an abstraction where the lookahead selection iterates through the indexes like if it were performing inferences on one of the yet to be selected literals and is only interested in COUNTING how many hits it would get. (While the iterators in their original role actually generate various things - substitutions, clauses. And there are just being forgotten here.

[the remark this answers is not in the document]
MichaelRawson: I'm still not clear why it's now empty-tuple rather than void, but makes sense to me and we have bigger fish to fry.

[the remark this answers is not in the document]
quickbeam123: I think it's simply because you can't write `return void` or `return something` that would type as `void`. Or something like that ;)

Kernel/RobSubstitution.hpp
```diff
@@ -17,55 +17,353 @@
...
-  bool unifyArgs(Term* t1,int index1, Term* t2, int index2, MismatchHandler* hndlr=0);
+  bool unifyArgs(Term* t1,int index1, Term* t2, int index2);
   bool matchArgs(Term* base,int baseIndex, Term* instance, int instanceIndex);
 
   void denormalize(const Renaming& normalizer, int normalIndex, int denormalizedIndex);
-  bool isUnbound(unsigned var, int index) const
+  bool isUnbound(VarSpec v) const;
+
+  /** introduces a new "glue" variable, and binds it to forTerm. 
+   * Glue variables live in their own namespace that is only used within this rob substitution. They are used only to represent subterms of other terms in the substitution.
```
quickbeam123, line 316: Nice!

Indexing/TermIndex.cpp
```diff
@@ -62,10 +48,9 @@ void SuperpositionSubtermIndex::handleClause(Clause* c, bool adding)
                                               : EqHelper::getSubtermIterator(lit,_ord);
     while (rsti.hasNext()) {
       auto tt = TypedTermList(rsti.next());
-      ((TermSubstitutionTree*)_is)->handle(tt, lit, c, adding);
+      ((TermSubstitutionTree*)&*_is)->handle(tt, lit, c, adding);
```
quickbeam123, line 51: Couldn't the `&*` be dropped?

Indexing/TermSharing.cpp
```diff
@@ -197,26 +180,20 @@ Term* TermSharing::insert(Term* t)
...
     }
-  }
-  else {
-    t->destroy();
-  }
-  return s;
 } // TermSharing::insert
 
-AtomicSort* TermSharing::insert(AtomicSort* sort)
+ // TODO update documentation
```
quickbeam123, line 185: Could this TODO be now too? :)

Indexing/TermSharing.cpp
```diff
@@ -245,22 +222,18 @@ AtomicSort* TermSharing::insert(AtomicSort* sort)
...
-  }
-  else {
-    sort->destroy();
-  }
-  return s;
 } // TermSharing::insert
 
 /**
+ * TODO update documentation
  * Insert a new literal in the index and return the result.
```
quickbeam123, line 229: And here...

Indexing/TermSharing.cpp
```diff
@@ -330,18 +291,13 @@ Literal* TermSharing::insert(Literal* t)
...
-  return s;
 } // TermSharing::insert
 
 /**
  * Insert a new literal in the index and return the result.
  * @since 28/12/2007 Manchester
  */
-Literal* TermSharing::insertVariableEquality(Literal* t, TermList sort)
+void TermSharing::computeAndSetSharedVarEqData(Literal* t, TermList sort)
 {
```
quickbeam123, line 301: And perhaps also here.

Inferences/BinaryResolution.cpp
```diff
@@ -51,15 +53,16 @@ using namespace Kernel;
 using namespace Indexing;
 using namespace Saturation;
 
+// TODO remove after refactor
+using BRQueryRes = LQueryRes<AbstractingUnifier*>;
```
quickbeam123, line 57: Isn't this now?

Inferences/BinaryResolution.cpp
```diff
@@ -177,71 +182,25 @@ Clause* BinaryResolution::generateClause(Clause* queryCl, Literal* queryLit, SLQ
     }
   }
 
-  bool synthesis = (env.options->questionAnswering() == Options::QuestionAnsweringMode::SYNTHESIS);
-  Literal* cAnsLit = synthesis ? queryCl->getAnswerLiteral() : nullptr;
```
quickbeam123, line 181: What will happen to QuestionAnsweringMode? Is that dealt with in some other way?

Indexing/TermIndex.cpp
joe-hauns: No because _is is a `std::unique_ptr<T>` and not a `T*`  So the `*_is` is of type `T&` and `&*_is` is of type `T*`. It looks a bit odd but u need that pattern quite some times with smart pointers.

Inferences/Superposition.cpp
```diff
@@ -522,58 +470,25 @@ Clause* Superposition::performSuperposition(
...
-        return 0;
+        res->push(currAfter);
       }
-
-      (*res)[next] = constraint;
-      next++;   
     }
   }
 
-  if (bothHaveAnsLit) {
```
quickbeam123, line 559: I guess also here, in Superposition, we have lost synthesis/question answering?

Inferences/BinaryResolution.cpp
joe-hauns: okayyy sorry it seems like i deleted this accidentally during merging at some point. I didn't ever see this bit of code conciously before (as it's rather new compared to my changes). I'll look into it now.

joe-hauns: Okay i fixed it in  BR & Superposition.  @hzzv  could you have a quick look whether this looks right?

Kernel/Term.hpp
joe-hauns: Yeah it's just a fancy way of saying that!
You're right we can't be sure but it's an implicit assumption i made. I added an assertion.

Inferences/BinaryResolution.cpp
hzzv: BR & Superposition work now, thanks!
However, your branch crashes for me on this input: https://github.com/vprover/vampire_benchmarks/blob/master/synthesis/non-recursive/smt2/group-ex3-square_commutativity.smt2
with `--decode lrs+10_3:1_av=off:ep=off:lcm=reverse:qa=synthesis:sp=reverse_arity:urr=on_300`.

The issue seems to be somewhere in URR:
```
vampireMode()
??:?
Saturation::ProvingHelper::runVampireSaturation(Kernel::Problem&, Shell::Options const&)
??:?
Saturation::ProvingHelper::runVampireSaturationImpl(Kernel::Problem&, Shell::Options const&)
??:?
Kernel::MainLoop::run()
??:?
Saturation::SaturationAlgorithm::runImpl()
??:?
Saturation::SaturationAlgorithm::activate(Kernel::Clause*)
??:?
Inferences::CompositeSGI::generateSimplify(Kernel::Clause*)
??:?
Inferences::CompositeGIE::generateClauses(Kernel::Clause*)
??:?
Inferences::URResolution::generateClauses(Kernel::Clause*)
??:?
Inferences::URResolution::processAndGetClauses(Inferences::URResolution::Item*, unsigned int, Lib::List<Kernel::Clause*>*&)
??:?
Inferences::URResolution::processLiteral(Lib::List<Inferences::URResolution::Item*>*&, unsigned int)
??:?
??
??:0
Lib::handleSignal(int)
??:?
Debug::Tracer::printStack(std::ostream&)
??:?
```

joe-hauns: Great thanks :)

[fixed it](https://github.com/vprover/vampire/pull/508/commits/9ecb46d597fbba1793ce890d6adc84d5269763c7)). I'm pretty sure this bug was already present in master.

Kernel/Term.cpp
MichaelRawson: Agree with Joe, resolving.

quickbeam123: Yes, sure. Here I actually just wanted to learn what you think :)

MichaelRawson: Ah - you were perfectly clear, I was too terse in my previous comment! Aside: something that bit me before is capturing variables by *dangling* reference is particularly easy - be careful if you decide to capture-by-reference.

Indexing/SubstitutionTree.cpp
```diff
@@ -50,14 +51,9 @@ using namespace Indexing;
  * Initialise the substitution tree.
  * @since 16/08/2008 flight Sydney-San Francisco
  */
-SubstitutionTree::SubstitutionTree(bool useC, bool rfSubs)
+SubstitutionTree::SubstitutionTree()
```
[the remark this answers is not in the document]
joe-hauns: I'm not quite sure what you mean... There's still the fields `_nextVar` and `_root` that need to be initialized in the constructor.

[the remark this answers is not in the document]
joe-hauns: ahh i see.. you mean writing
```
class SubstitutionTree {
...
    int _nextVar = 0;
    Node* _root = nullptr;
...
};
```
I wasn't aware you could do that in c++. I thought this was a java-only thing. 
I don't care we can do it either way but i never saw it like this around the codebase. But I reckon I'm quite in favor of using it if we can actually.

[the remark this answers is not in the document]
joe-hauns: changed it

Indexing/SubstitutionTree.hpp
```diff
@@ -52,1420 +56,1461 @@
...
-     */
-    virtual void makeEmpty() { term.makeEmpty(); }
-    static void split(Node** pnode, TermList* where, int var);
+      /** term at this node */
+      TermList _term;
+#define CACHE_FUNCTOR 0
+#if CACHE_FUNCTOR
+      /** if _term is a Term* we cache the functor so we are faster than calling _term->term(); */
+      unsigned _functor;
+#endif // CACHE_FUNCTOR
```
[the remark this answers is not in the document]
joe-hauns: yessss. I actually think to remember it helped in a lot of cases but I don't know why i turned it off in the end. 
Shall we run a comparison experiment with `CACHE_FUNCTOR=1` vs `CACHE_FUNCTOR=0` ? What do you think  @quickbeam123 ?

Inferences/Superposition.cpp
```diff
@@ -522,58 +470,25 @@ Clause* Superposition::performSuperposition(
...
+  if(needsToFulfilWeightLimit && !passiveClauseContainer->fulfilsWeightLimit(weight, numPositiveLiteralsLowerBound, inf)) {
     RSTAT_CTR_INC("superpositions skipped for weight limit after the clause was built");
     env.statistics->discardedNonRedundantClauses++;
-    construction_fail:
-    res->destroy();
-    return 0;
+    return nullptr;
   }
 
   //TODO update AYB
```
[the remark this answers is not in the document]
joe-hauns: tbh i'm not sure what the update was supposed to be. But we're definitely now recording function extionality abstraction as uwa, so i reckon that's fine...?

Kernel/BottomUpEvaluation.hpp
```diff
@@ -97,87 +92,276 @@ struct BottomUpChildIter
...
+   *    // (optional)
+   *    // some term sutructures (like AutoDerefTermSpec in RobSubstition) need a context to be traversed
+   *    // with BottomUpChildIter. this context needs to be passed here.
+   *    // .context(AutoDerefTermSpec::Context { .subs = this, })
+   *
+   *    // applys the evaluation
+   *    .apply(someTerm)
+   *    ;
+   * ```
+   */
```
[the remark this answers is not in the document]
joe-hauns: hmmm not at all. I'd love to know it though haha

Kernel/Term.cpp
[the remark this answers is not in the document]
joe-hauns: That's actually not possible because `class Term` is declared after `class TermList`, so we can't call `Term::ground` in `TermList::ground` ....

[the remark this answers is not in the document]
joe-hauns: we could ofc add an 
```
inline bool TermList::ground() const
{ return isTerm() && term()->ground(); }
```
to the header if you think it will impact performance.

Lib/Hash.hpp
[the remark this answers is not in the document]
joe-hauns: ahhh right. Tbh i didn't really care how exactly we hash. We can refactor this as you said to remove `HashUtils::combine` but I think this will be a bit of effort so I'm not so keen on it right now. So I'd say let's defer this to a separate PR?

Indexing/SubstitutionTree.cpp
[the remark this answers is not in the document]
quickbeam123: I think it's a relatively recent thing. With older standards the fields need to static? and const? or something. 

Also a bit hesitant to change the coding style, but since the initialization is with the declaration, I can see how it could bite us and it does feel like an improvement (in this way things are done).

Indexing/SubstitutionTree.hpp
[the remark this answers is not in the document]
quickbeam123: Yes, we could! Not sure if I will find time for this until the IJCAR deadline, but if the two weeks don't matter than let's do this too!

Kernel/Term.cpp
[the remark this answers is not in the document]
quickbeam123: Will leave this to @MichaelRawson to decide :)

Kernel/BottomUpEvaluation.hpp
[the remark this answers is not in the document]
MichaelRawson: No probs, me either!

Kernel/Term.cpp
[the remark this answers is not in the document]
MichaelRawson: It probably will, but I'm happy to move this in a future PR if we've decided performance so far is acceptable.

Inferences/Superposition.cpp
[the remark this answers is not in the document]
MichaelRawson: Fine by me.

Indexing/SubstitutionTree.hpp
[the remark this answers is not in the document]
MichaelRawson: OK - I think you're happy with performance as-is, let's merge this and remember to play with `CACHE_FUNCTOR` later? I'd be a little bit surprised given it's "only" an indirect load for the functor from a `Term *` but happy either way.

Indexing/SubstitutionTree.cpp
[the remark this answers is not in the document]
MichaelRawson: Right - I wasn't aware until recently either, but I think it's been available since C++11. Generally in favour, especially if we can remove a constructor altogether. Thanks for the change!

PR #510, opened 2023-12-08
do not consider {inequality,general} splitting as naming a formula

### LINKED ISSUE #509: Smart ep=RSTC skips congruences for introduced symbols for which it shouldn't?
We saturate (and think we are complete) on the following unsat problem via:

./vampire_dbg_master_6968 --decode dis+21_1:10_rp=on:ep=RSTC:ins=1_0 Problems/REL/REL022+2.p -p off
% SZS status CounterSatisfiable for REL022+2
% # SZS output start Saturation.
% # SZS output end Saturation.
Condition in file Shell/UIHelper.cpp, line 486 violated:
!s_expecting_unsat
----- stack dump -----
Version : Vampire 4.8 (commit ef21e5860 on 2023-12-01 08:47:07 +0000)
_start
__libc_start_main
???
main
vampireMode()
Shell::UIHelper::outputResult(std::ostream&)
Debug::Assertion::violated(char const, int, char const)
Debug::Tracer::printStack(std::ostream&)
----- end of stack dump -----

Maybe we are saving too much on the congruence axioms in equality_proxy? Maybe inequality splitting does not mark its new symbol properly? Could this also happen with e.g. GeneralSplitting? How to prevent this from happening in the future when an uninitiated person introduced a new symbol?

MichaelRawson:
Should fix #509 - the problem was that I over-eagerly removed stuff from congruence axioms with `ep=RSTC`. Mea culpa etc, I have eaten my humble pie twice now.

This one is quite subtle. The argument for skipping introduced symbols goes that you could do equality proxy completely before clausification, on the input problem - so you don't need to include introduced names because they aren't in the original problem. *However* we don't actually do equality proxy before clausification, and inequality splitting happens before equality proxy, changing the "input problem" in such a way that we need to include its definition predicates. I'm not sure about general splitting, but perhaps a similar problem exists so I've also included them just to be sure.

There are, I think, two solutions. This one - include congruence axioms for stuff we're not 100% sure about. Or, perhaps better - do the equality-proxy transform before clausification.

### DIFF
--- a/Kernel/Signature.hpp
+++ b/Kernel/Signature.hpp
@@ -429,7 +429,9 @@ class Signature
   unsigned addSkolemTypeCon(unsigned arity,const char* suffix = 0);
   unsigned addFreshPredicate(unsigned arity, const char* prefix, const char* suffix = 0);
   unsigned addSkolemPredicate(unsigned arity,const char* suffix = 0);
+  // add a predicate naming a sub-formula - if you are not sure, use `addFreshPredicate`
   unsigned addNamePredicate(unsigned arity);
+  // same as `addNamePredicate`, but produces a function of Boolean sort
   unsigned addNameFunction(unsigned arity);
   void addEquality();
   unsigned getApp();
--- a/Shell/GeneralSplitting.cpp
+++ b/Shell/GeneralSplitting.cpp
@@ -227,7 +227,7 @@ bool GeneralSplitting::apply(Clause*& cl, UnitList*& resultStack)
   }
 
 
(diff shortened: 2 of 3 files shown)


easychair:
If clausification introduces new symbols, we cannot do EP before
clausification.

A

MichaelRawson:
>If clausification introduces new symbols, we cannot do EP before clausification.

New *equations* we have to watch out for. New symbols can be fine; for instance, it's OK to skip Skolems.

Anyhow, @quickbeam123 prefers the manually-marked solution, closing.

PR #513, opened 2023-12-12
i386: fix the major problems at least

MichaelRawson:
We are alerted in #512 that Vampire is broken on 32-bit platforms, which we knew, but weren't aware that anyone cared. We currently reject 32-bit systems at compile-time in `Lib/Portability.hpp`.

However, it doesn't actually seem that hard to get Vampire running on 32-bit systems. The biggest problem is that `sizeof(size_t)` is now 32, and we assume `sizeof(TermList) == sizeof(size_t)` to be at least 64 to get the bits we want into it - but this can be fixed by saying that `TermList` should be exactly 64 bits instead of `sizeof(size_t)`.

This patch produces a 32-bit Vampire that runs plausibly on my system. Since we have users who want 32-bit support, I propose that we relax our 64-bit-only policy until we end up in a situation where we really can't support 32-bit systems any more.

@barracuda156 - you might want to test on this branch. I don't guarantee this fixes PowerPC, or even that we don't have serious bugs even on i386, but it should be much less crashy.

### DIFF
--- a/Indexing/LiteralSubstitutionTree.cpp
+++ b/Indexing/LiteralSubstitutionTree.cpp
@@ -55,7 +55,7 @@ SLQueryResultIterator LiteralSubstitutionTree::getVariants(Literal* query, bool 
 SLQueryResultIterator LiteralSubstitutionTree::getAll()
 {
   return pvi(
-        iterTraits(getRangeIterator((unsigned long)0, _trees.size()))
+        iterTraits(getRangeIterator((size_t)0, _trees.size()))
          .flatMap([this](auto i) { return LeafIterator(&_trees[i]); })
          .flatMap([](Leaf* l) { return l->allChildren(); })
          .map([](const LeafData& ld) { return SLQueryResult(ld.literal, ld.clause); })
--- a/Lib/Portability.hpp
+++ b/Lib/Portability.hpp
@@ -19,12 +19,6 @@ static_assert(
     "Vampire assumes that there are 8 bits in a `char`"
 );
 
-static_assert(
-    sizeof(void *) == 8,
-    "Vampire assumes that the size of a pointer is 8 bytes for efficiency reasons. "
-    "This may be fixed/relaxed in future, but for the moment expect problems if running on other architectures."
-);
-
 // enable warnings for unused results
 // C++17: replace this with [[nodiscard]]
 #ifdef __GNUC__
--- a/SAT/SATClause.hpp
+++ b/SAT/SATClause.hpp
@@ -51,6 +51,7 @@ public:
   void setInference(SATInference* val);
 
   void* operator new(size_t,unsigned length);
+  void operator delete(void *);
 
   /**
    * Return the (reference to) the nth literal
--- a/Kernel/MLMatcher.cpp
+++ b/Kernel/MLMatcher.cpp
@@ -108,10 +108,10 @@ bool createLiteralBindings(Literal* baseLit, LiteralList const* alts, Clause* in
 	altBindingPtrs++;
 	altBindingData+=numVars;
 	if(resolvedLit) {
-	  ::new (altBindingData++) TermList((uint64_t)0);
+	  (altBindingData++)->setContent(0);
 	} else {
           // add index of the literal in instance clause at the end of the binding sequence
-	  ::new (altBindingData++) TermList((uint64_t)instCl->getLiteralPosition(alit));
+	  (altBindingData++)->setContent(instCl->getLiteralPosition(alit));
 	}
       }
       if(MatchingUtils::matchReversedArgs(baseLit, alit)) {
@@ -121,10 +121,10 @@ bool createLiteralBindings(Literal* baseLit, LiteralList const* alts, Clause* in
 	altBindingPtrs++;
 	altBindingData+=numVars;
 	if(resolvedLit) {
-	  ::new (altBindingData++) TermList((uint64_t)0);
+	  (altBindingData++)->setContent(0);
 	} else {
           // add index of the literal in instance clause at the end of the binding sequence
-	  ::new (altBindingData++) TermList((uint64_t)instCl->getLiteralPosition(alit));
+	  (altBindingData++)->setContent(instCl->getLiteralPosition(alit));
 	}
       }
 
@@ -138,10 +138,10 @@ bool createLiteralBindings(Literal* baseLit, LiteralList const* alts, Clause* in
       altBindingPtrs++;
       altBindingData+=numVars;
       if(resolvedLit) {
-        ::new (altBindingData++) TermList((uint64_t)0);
+        (altBindingData++)->setContent(0);
       } else {
         // add index of the literal in instance clause at the end of the binding sequence
-        ::new (altBindingData++) TermList((uint64_t)instCl->getLiteralPosition(alit));
+        (altBindingData++)->setContent((uint64_t)instCl->getLiteralPosition(alit));
       }
     }
   }
@@ -154,7 +154,7 @@ bool createLiteralBindings(Literal* baseLit, LiteralList const* alts, Clause* in
       *altBindingPtrs=altBindingData;
       altBindingPtrs++;
       altBindingData+=numVars;
-      ::new (altBindingData++) TermList((uint64_t)1);
+      (altBindingData++)->setContent(1);
     }
     if(baseLit->isEquality() && MatchingUtils::matchReversedArgs(baseLit, resolvedLit)) {
       ArrayStoringBinder binder(altBindingData, variablePositions);
@@ -162,7 +162,7 @@ bool createLiteralBindings(Literal* baseLit, LiteralList const* alts, Clause* in
       *altBindingPtrs=altBindingData;
       altBindingPtrs++;
       altBindingData+=numVars;
-      ::new (altBindingData++) TermList((uint64_t)1);
+      (altBindingData++)->setContent(1);
     }
 
   }
--- a/Kernel/MLMatcher.cpp
+++ b/Kernel/MLMatcher.cpp
@@ -108,10 +108,10 @@ bool createLiteralBindings(Literal* baseLit, LiteralList const* alts, Clause* in
 	altBindingPtrs++;
 	altBindingData+=numVars;
 	if(resolvedLit) {
-	  ::new (altBindingData++) TermList((size_t)0);
+	  ::new (altBindingData++) TermList((uint64_t)0);
 	} else {
           // add index of the literal in instance clause at the end of the binding sequence
-	  ::new (altBindingData++) TermList((size_t)instCl->getLiteralPosition(alit));
+	  ::new (altBindingData++) TermList((uint64_t)instCl->getLiteralPosition(alit));
 	}
       }
       if(MatchingUtils::matchReversedArgs(baseLit, alit)) {
@@ -121,10 +121,10 @@ bool createLiteralBindings(Literal* baseLit, LiteralList const* alts, Clause* in
 	altBindingPtrs++;
 	altBindingData+=numVars;
 	if(resolvedLit) {
-	  ::new (altBindingData++) TermList((size_t)0);
+	  ::new (altBindingData++) TermList((uint64_t)0);
 	} else {
           // add index of the literal in instance clause at the end of the binding sequence
-	  ::new (altBindingData++) TermList((size_t)instCl->getLiteralPosition(alit));
+	  ::new (altBindingData++) TermList((uint64_t)instCl->getLiteralPosition(alit));
 	}
       }
 
@@ -138,10 +138,10 @@ bool createLiteralBindings(Literal* baseLit, LiteralList const* alts, Clause* in
       altBindingPtrs++;
       altBindingData+=numVars;
       if(resolvedLit) {
-        ::new (altBindingData++) TermList((size_t)0);
+        ::new (altBindingData++) TermList((uint64_t)0);
       } else {
         // add index of the literal in instance clause at the end of the binding sequence
-        ::new (altBindingData++) TermList((size_t)instCl->getLiteralPosition(alit));
+        ::new (altBindingData++) TermList((uint64_t)instCl->getLiteralPosition(alit));
       }
     }
   }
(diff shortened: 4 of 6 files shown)


Indexing/LiteralSubstitutionTree.cpp
```diff
@@ -55,7 +55,7 @@ SLQueryResultIterator LiteralSubstitutionTree::getVariants(Literal* query, bool
 SLQueryResultIterator LiteralSubstitutionTree::getAll()
 {
   return pvi(
-        iterTraits(getRangeIterator((unsigned long)0, _trees.size()))
+        iterTraits(getRangeIterator((size_t)0, _trees.size()))
```
MichaelRawson, line 58: This should always have been `size_t` to match `_trees.size()`

SAT/SATClause.hpp
```diff
@@ -51,6 +51,7 @@ class SATClause
   void setInference(SATInference* val);
 
   void* operator new(size_t,unsigned length);
+  void operator delete(void *);
```
MichaelRawson, line 54: This is weird; 32-bit Clang and GCC complains that this function doesn't exist for generating exception-cleanup code, but never actually calls it and I don't see why it would be needed. If we ever actually need an implementation, it should just call `DEALLOC_KNOWN(this, length)`.

MichaelRawson:
Generally, `TermList` is hazardous and could use an overhaul. But that will be a big job.

SAT/SATClause.hpp
barracuda156: Complains in a sense of warnings or errors? I did not have any compile-time errors besides failing static asserts. (But produced binary was broken.)

MichaelRawson: Errors, surprisingly: https://godbolt.org/z/9sb819xT8 - try with/without `-m32`.

barracuda156:
@MichaelRawson Building from 3c2909cbbe6df7edc610110355cf99db45fca79d fails for me with:
```
:info:build FAILED: CMakeFiles/obj.dir/Kernel/MLMatcher.cpp.o 
:info:build /opt/local/bin/g++-mp-13 -DCHECK_LEAKS=0 -DVDEBUG=1 -DVTIME_PROFILING=0 -DVZ3=1 -I/opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_math_vampire/vampire/work/vampire-3c2909cbbe6df7edc610110355cf99db45fca79d -isystem /opt/local/include -pipe -mcpu=native -mtune=native -I/opt/local/include -D_GLIBCXX_USE_CXX11_ABI=0 -g -arch ppc -mmacosx-version-min=10.6 -Wall -fno-threadsafe-statics -fno-rtti -fsized-deallocation -std=c++14 -MD -MT CMakeFiles/obj.dir/Kernel/MLMatcher.cpp.o -MF CMakeFiles/obj.dir/Kernel/MLMatcher.cpp.o.d -o CMakeFiles/obj.dir/Kernel/MLMatcher.cpp.o -c /opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_math_vampire/vampire/work/vampire-3c2909cbbe6df7edc610110355cf99db45fca79d/Kernel/MLMatcher.cpp
:info:build /opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_math_vampire/vampire/work/vampire-3c2909cbbe6df7edc610110355cf99db45fca79d/Kernel/MLMatcher.cpp: In function 'bool {anonymous}::createLiteralBindings(Kernel::Literal*, const Kernel::LiteralList*, Kernel::Clause*, Kernel::Literal*, unsigned int*&, Kernel::TermList**&, Kernel::TermList*&)':
:info:build /opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_math_vampire/vampire/work/vampire-3c2909cbbe6df7edc610110355cf99db45fca79d/Kernel/MLMatcher.cpp:111:54: error: call of overloaded 'TermList(size_t)' is ambiguous
:info:build   111 |           ::new (altBindingData++) TermList((size_t)0);
:info:build       |                                                      ^
:info:build In file included from /opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_math_vampire/vampire/work/vampire-3c2909cbbe6df7edc610110355cf99db45fca79d/Debug/TimeProfiling.hpp:19,
:info:build                  from /opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_math_vampire/vampire/work/vampire-3c2909cbbe6df7edc610110355cf99db45fca79d/Lib/Metaiterators.hpp:29,
:info:build                  from /opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_math_vampire/vampire/work/vampire-3c2909cbbe6df7edc610110355cf99db45fca79d/Lib/BinaryHeap.hpp:26,
:info:build                  from /opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_math_vampire/vampire/work/vampire-3c2909cbbe6df7edc610110355cf99db45fca79d/Kernel/MLMatcher.cpp:17:
:info:build /opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_math_vampire/vampire/work/vampire-3c2909cbbe6df7edc610110355cf99db45fca79d/Kernel/Term.hpp:100:12: note: candidate: 'Kernel::TermList::TermList(Kernel::Term*)'
:info:build   100 |   explicit TermList(Term* t) : _content(0) {
:info:build       |            ^~~~~~~~
:info:build /opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_math_vampire/vampire/work/vampire-3c2909cbbe6df7edc610110355cf99db45fca79d/Kernel/Term.hpp:98:12: note: candidate: 'Kernel::TermList::TermList(uint64_t)'
:info:build    98 |   explicit TermList(uint64_t data) : _content(data) {}
:info:build       |            ^~~~~~~~
:info:build /opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_math_vampire/vampire/work/vampire-3c2909cbbe6df7edc610110355cf99db45fca79d/Kernel/Term.hpp:91:7: note: candidate: 'constexpr Kernel::TermList::TermList(const Kernel::TermList&)'
:info:build    91 | class TermList {
:info:build       |       ^~~~~~~~
:info:build /opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_math_vampire/vampire/work/vampire-3c2909cbbe6df7edc610110355cf99db45fca79d/Kernel/Term.hpp:91:7: note: candidate: 'constexpr Kernel::TermList::TermList(Kernel::TermList&&)'
:info:build /opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_math_vampire/vampire/work/vampire-3c2909cbbe6df7edc610110355cf99db45fca79d/Kernel/MLMatcher.cpp:124:54: error: call of overloaded 'TermList(size_t)' is ambiguous
:info:build   124 |           ::new (altBindingData++) TermList((size_t)0);
:info:build       |                                                      ^
:info:build /opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_math_vampire/vampire/work/vampire-3c2909cbbe6df7edc610110355cf99db45fca79d/Kernel/Term.hpp:100:12: note: candidate: 'Kernel::TermList::TermList(Kernel::Term*)'
:info:build   100 |   explicit TermList(Term* t) : _content(0) {
:info:build       |            ^~~~~~~~
:info:build /opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_math_vampire/vampire/work/vampire-3c2909cbbe6df7edc610110355cf99db45fca79d/Kernel/Term.hpp:98:12: note: candidate: 'Kernel::TermList::TermList(uint64_t)'
:info:build    98 |   explicit TermList(uint64_t data) : _content(data) {}
:info:build       |            ^~~~~~~~
:info:build /opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_math_vampire/vampire/work/vampire-3c2909cbbe6df7edc610110355cf99db45fca79d/Kernel/Term.hpp:91:7: note: candidate: 'constexpr Kernel::TermList::TermList(const Kernel::TermList&)'
:info:build    91 | class TermList {
:info:build       |       ^~~~~~~~
:info:build /opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_math_vampire/vampire/work/vampire-3c2909cbbe6df7edc610110355cf99db45fca79d/Kernel/Term.hpp:91:7: note: candidate: 'constexpr Kernel::TermList::TermList(Kernel::TermList&&)'
:info:build /opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_math_vampire/vampire/work/vampire-3c2909cbbe6df7edc610110355cf99db45fca79d/Kernel/MLMatcher.cpp:141:52: error: call of overloaded 'TermList(size_t)' is ambiguous
:info:build   141 |         ::new (altBindingData++) TermList((size_t)0);
:info:build       |                                                    ^
:info:build /opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_math_vampire/vampire/work/vampire-3c2909cbbe6df7edc610110355cf99db45fca79d/Kernel/Term.hpp:100:12: note: candidate: 'Kernel::TermList::TermList(Kernel::Term*)'
:info:build   100 |   explicit TermList(Term* t) : _content(0) {
:info:build       |            ^~~~~~~~
:info:build /opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_math_vampire/vampire/work/vampire-3c2909cbbe6df7edc610110355cf99db45fca79d/Kernel/Term.hpp:98:12: note: candidate: 'Kernel::TermList::TermList(uint64_t)'
:info:build    98 |   explicit TermList(uint64_t data) : _content(data) {}
:info:build       |            ^~~~~~~~
:info:build /opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_math_vampire/vampire/work/vampire-3c2909cbbe6df7edc610110355cf99db45fca79d/Kernel/Term.hpp:91:7: note: candidate: 'constexpr Kernel::TermList::TermList(const Kernel::TermList&)'
:info:build    91 | class TermList {
:info:build       |       ^~~~~~~~
```

MichaelRawson:
@barracuda156 - a question I should already have asked - is your PowerPC machine 32- or 64- bit?

I don't get these compilation errors, but it looks like changing casts of this form from `(size_t)0` to `(uint64_t)0` would work. Could you try it and let me know? As it happens both overloads do the same thing in this case.

barracuda156:
@MichaelRawson Machine is a sense of a hardware (G5 / ppc970) supports both, but 10.6 only supports `ppc`.
(In principle I could test for `ppc64` too, but not at the moment, since I have no functional set-up on 10.5.8.)

barracuda156:
P. S. Please allow me until tomorrow, I am away from the fast PowerMac now. Will update with results once back to it.

MichaelRawson:
> P. S. Please allow me until tomorrow, I am away from the fast PowerMac now. Will update with results once back to it.

You must be new here. ;-) No hurry!

barracuda156:
@MichaelRawson Rebuilt from your branch, replacing `(size_t)` with `(uint64_t)`  in `MLMatcher.cpp`, but the binary remains broken on launch:

```
(gdb) run
Starting program: /opt/local/bin/vampire 
Reading symbols for shared libraries +++++.... done

Program received signal EXC_BAD_ACCESS, Could not access memory.
Reason: KERN_PROTECTION_FAILURE at address: 0x00000008
0x0002dc14 in Kernel::Term::functor ()
(gdb) where
#0  0x0002dc14 in Kernel::Term::functor ()
#1  0x00038ebc in Kernel::Term::isSpecial ()
#2  0x00059a00 in Kernel::Term::toString ()
#3  0x0005993c in Kernel::TermList::toString ()
#4  0x00053e3c in Kernel::OperatorType::toString ()
#5  0x000498e4 in Kernel::Signature::addInterpretedPredicate ()
#6  0x0004959c in Kernel::Signature::addEquality ()
#7  0x00007014 in Lib::Environment::Environment ()
#8  0x000073cc in __static_initialization_and_destruction_0 ()
#9  0x0000741c in _GLOBAL__sub_I_Environment.cpp ()
#10 0x8fe15d60 in __dyld__ZN16ImageLoaderMachO16doInitializationERKN11ImageLoader11LinkContextE ()
#11 0x8fe0f70c in __dyld__ZN11ImageLoader23recursiveInitializationERKNS_11LinkContextEj ()
#12 0x8fe0f804 in __dyld__ZN11ImageLoader15runInitializersERKNS_11LinkContextE ()
#13 0x8fe02708 in __dyld__ZN4dyld24initializeMainExecutableEv ()
#14 0x8fe08690 in __dyld__ZN4dyld5_mainEPK11mach_headermiPPKcS5_S5_ ()
#15 0x8fe017cc in __dyld__ZN13dyldbootstrap5startEPK11mach_headeriPPKcl ()
#16 0x8fe01064 in __dyld__dyld_start ()
```

barracuda156:
Two issues to have in mind for `ppc`: it is Big endian and it has 4-byte bool (and spinlock). If anything of these matters and is not accounted for in the code.

MichaelRawson:
Moved conversation about PowerPC back to the original issue. This PR should nonetheless fix 32-bit Intel.

barracuda156:
I will try to test it on i386 in a few days.

Could we also fix those type mismatches, as per your suggestion? As long as that works across other archs, it is desirable: while it is not enough for ppc, it is needed at least to fix the compilation.

MichaelRawson:
>I will try to test it on i386 in a few days.

Thanks - I can actually test this locally on a 64-bit machine, but if you have actually 32-bit silicon that'd be great.

>Could we also fix those type mismatches, as per your suggestion? As long as that works across other archs, it is desirable: while it is not enough for ppc, it is needed at least to fix the compilation.

Good point. I'll update this.

quickbeam123 (approved):

quickbeam123:
Approved. Modulo the question about the removed static assert...

Lib/Portability.hpp
```diff
@@ -19,12 +19,6 @@ static_assert(
     "Vampire assumes that there are 8 bits in a `char`"
 );
 
-static_assert(
```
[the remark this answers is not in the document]
MichaelRawson: Since `TermList` *contains* a pointer, I would say it would be trivially-satisfied. Of course, if you want it for the maintenance value, I'm happy to do it.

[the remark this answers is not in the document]
quickbeam123: No, stupid me. Taking it back :)

barracuda156:
Thank you! And I will try with `ppc` build tonight and update in the issue.

PR #514, opened 2023-12-14
actually remove `_id` from Term, fix goof with TERM_DIST_VAR_UNKNOWN

MichaelRawson:
In #337 there was two separate mishaps.
1. @quickbeam123 realised that `_id` could be moved from `Term` itself to its "info" `TermList`, but I failed to actually remove it from `Term` so there's been some waste bytes knocking around in `Term`. Aside: we might actually want to move it back if we can fit more useful stuff into the 32 bits in `TermList`.
2. I meant to actually compute `(2^21) - 1` to work out the maximum value of `TERM_DIST_VAR_UNKNOWN`, but instead wrote it down literally - where the `^` is interpreted as XOR. This means that since then `TERM_DIST_VAR_UNKNOWN` has been ...22. :facepalm:

### DIFF
--- a/Kernel/Term.hpp
+++ b/Kernel/Term.hpp
@@ -40,8 +40,8 @@
 
 // the number of bits used for "TermList::_info::distinctVars"
 #define TERM_DIST_VAR_BITS 21
-// maximum number that fits in a 21-bit integer - update if TERM_DIST_VAR_BITS changes
-#define TERM_DIST_VAR_UNKNOWN 0x1FFFFF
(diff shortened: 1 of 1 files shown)


quickbeam123 (approved):

quickbeam123:
Let's be the tough guys and just fix the expression (proposal incoming), what do you think?

By the way, isn't it suspicious we didn't see this crashing in any way? (Or could some heuristics have been broken?)

Otherwise, nice!

quickbeam123:
Yes, and between parentheses. Just fixed and force-pushed.

PR #515, opened 2023-12-24
fix a bug in DefinitionIntroduction

quickbeam123:
- since handleClause is not guaranteed to be always called after generateClauses (it's listening on Passive!), the same definitions could be returned more than once
- in the worst case, a definition would have been deleted already (redundant), which could then lead to a segfault

### DIFF
--- a/Inferences/DefinitionIntroduction.hpp
+++ b/Inferences/DefinitionIntroduction.hpp
@@ -26,21 +26,16 @@ public:
   void attach(SaturationAlgorithm *salg) override {
     GeneratingInferenceEngine::attach(salg);
     attachContainer(salg->getPassiveClauseContainer());
-    reset = false;
   }
 
   void handleClause(Clause *cl, bool adding) override {
-    if(reset)
-      _definitions.reset();
-
-    reset = false;
-    if(adding)
+    if(adding) {
       process(cl);
+    }
   }
 
   ClauseIterator generateClauses(Clause *cl) override {
-    reset = true;
-    return pvi(decltype(_definitions)::Iterator(_definitions));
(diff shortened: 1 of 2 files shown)


Inferences/DefinitionIntroduction.hpp
```diff
@@ -26,21 +26,16 @@ class DefinitionIntroduction: public GeneratingInferenceEngine, public Index {
...
-    if(adding)
+    if(adding) {
       process(cl);
+    }
   }
 
   ClauseIterator generateClauses(Clause *cl) override {
-    reset = true;
-    return pvi(decltype(_definitions)::Iterator(_definitions));
+    return pvi(ownedArrayishIterator(std::move(_definitions)));
```
MichaelRawson, line 38: Does `std::move` do what you want here? I think it's only supposed to leave the moved-from object in some unspecified and possibly-bad state safe for destruction - but maybe we're OK here.

PS I'm glad you know the right `ownerArrayishIterator` magic!

MichaelRawson (approved):
Very happy with this! Sorry, I'm not sure I intended for this to be a permanent feature at all - code could be better and I've been meaning to improve the algorithm, which is also questionable.

Inferences/DefinitionIntroduction.hpp
quickbeam123: This is what Joe does elsewhere, e.g. `CompositeSGI::generateSimplify` (the `move` on `Stack` is done via swap with essentially zeros, so leaves the thing in a sound state).

quickbeam123:
It's part of master, it's being tested! ;)

Issue #516, opened 2024-01-02
Revive `gcov`


inpefess:
After trying to evaluate test coverage of the project, I realised that, historically, there was `gcov` support (flags in the old Makefile: https://github.com/vprover/vampire/blob/1dd0ea6a4eb21e5a58b67560763a6630dfdeb446/Makefile#L29 and even a script to generate reports from collected coverage data: https://github.com/vprover/vampire/blob/1dd0ea6a4eb21e5a58b67560763a6630dfdeb446/scripts/dogcov.sh).
In the CMake build, `gcov` seems to be gone, and reviving it poses some problems. Namely, it doesn't write coverage data because of a known `gcov` and `_exit` interplay problem (https://github.com/kr/ct/issues/12).
It seems that `_exit` (instead of more graceful `exit`) appeared in unit tests only to run them in Cygwin (https://github.com/vprover/vampire/commit/6279e2c8f19f50cdee281f5fd4985798769ecb06 with an explanation in https://github.com/vprover/vampire/wiki/Cygwin).

I propose adding `gcov` flags to CMake (in `Debug` mode) and updating the existing report-generation script.

I also propose to make coverage a Linux-only feature (by adding some `#ifdef CYGWIN` lines) since one can't have `exit` and `gcov` both working properly in Cygwin.

MichaelRawson:
Relevant, from #517:

>[gcov] helped me identify functions not covered by other tests (Set.hpp already had a decent coverage) or specific cases missed by more general ones. I can create a PR for [this issue] based on my current pipeline and existing Giles' script.

---

>I propose adding gcov flags to CMake (in Debug mode) and updating the existing report-generation script.

If you are willing and it can be done mostly orthogonally to the rest of the build, I'd be very happy to merge it. I suspect others might use it too. I'd been intending to cleanup `scripts/`, it's mostly ancient. Having some green wood in there would be healthy.

>In the CMake build, gcov seems to be gone, and reviving it poses some problems.

Indeed - I'm not sure anyone used it for a long time. In the meantime the CMake build sprung into existence.

>It seems that _exit (instead of more graceful exit) appeared in unit tests only to run them in Cygwin

I'm not sure that anyone cares much about this case - maybe you do? If not, we could go back to `exit` and not bother with the Cygwin conditional-compile.

inpefess:
> If you are willing and it can be done mostly orthogonally to the rest of the build, I'd be very happy to merge it.

I'll keep changes minimal and only for `Debug` section.

[...]

PR #517, opened 2024-01-02
Add Set unit tests

inpefess:
`Set.hpp` has relatively high coverage, but not 100%. I also added `DummyHash` to cover cases when calls to `hash` return the same value (or value with a reserved meaning in current implementation). One shouldn't use `DummyHash` anywhere but for unit-tests.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -742,6 +742,7 @@ set(UNIT_TESTS
     UnitTests/tIterator.cpp
     UnitTests/tOption.cpp
     UnitTests/tStack.cpp
+    UnitTests/tSet.cpp
     )
 source_group(unit_tests FILES ${UNIT_TESTS})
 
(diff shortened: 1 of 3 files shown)


MichaelRawson:
Hey, thanks! Always happy to have more unit tests.

Did you use `gcov` to determine the coverage, looking at #516 ? Was it useful?

inpefess:
> Hey, thanks! Always happy to have more unit tests.

Great! I'll continue adding tests for `Lib` classes then.

> Did you use `gcov` to determine the coverage, looking at #516 ? Was it useful?

Yes, exactly. It helped me identify functions not covered by other tests (`Set.hpp` already had a decent coverage) or specific cases missed by more general ones. I can create a PR for #516 based on my current pipeline and existing Giles' script.

MichaelRawson:
>Great! I'll continue adding tests for `Lib` classes then.

Fantastic. Fair warning that I intend in the mid-term to improve (remove) some of the functionality in `Lib/`, but I suspect that your unit tests will be nonetheless useful in this endeavour. Besides, I have a looooong Vampire to-do list, so it won't be very soon. :-)

PR #518, opened 2024-01-02
Add coverage

### LINKED ISSUE #516: Revive `gcov`

inpefess:
Closes #516
Breaks unit-tests support for Cygwin.
Coverage data collection takes time, but if one can run the tests in parallel, it takes under one minute (`tInterpretedFunctions.cpp` is the longest-running).


CMakeLists.txt
```diff
@@ -881,6 +881,9 @@ set(UNIT_TEST_OBJ   )
 set(UNIT_TEST_CASES )
 if (COMPILE_TESTS) 
   include(CTest)
+  set(CMAKE_CXX_FLAGS ${CMAKE_CXX_FLAGS} --coverage)
```
MichaelRawson, line 884: Should this be guarded for e.g. GCC/Clang? Or perhaps CMake has some abstraction for this?

MichaelRawson (approved):
I'm generally happy, thanks for the fiddly work reviving coverage!

@joe-hauns @RobCoutel - I believe you are the main unit testers. Is this useful for you, and do you mind the increased timeout? If it's a problem we could think about how to make the tests run faster or putting coverage behind a flag.

CMakeLists.txt
joe-hauns: Soo if enabling this breaks unittesting support for Cygwin, I think it should be guarded in some way.  I'm everything but a cmake expert tho

MichaelRawson: My thinking is that nobody uses Cygwin for development, only occasionally as a "this really has to work on Windows" last-ditch effort, and therefore they don't want/need unit tests. Is this true?

joe-hauns: Well noone who's a core vampire dev right now but there might be future contributors who want..? I think if we can just add a `if (compiler !=  Cygwin)` guard before that line and everything's fine we might was well do it, right? Or is it more complicated than that?

MichaelRawson: See discussion in #516 - we currently call `_Exit` somewhere because Cygwin didn't like `exit` in 2014 (!), but `_Exit` breaks coverage for some reason. `_Exit` is kind of evil anyway and I'm not sure it breaks Cygwin any more - @inpefess can you confirm or deny?

inpefess: Let me check how it works now with `exit` on Cygwin

inpefess:
It seems that `master` doesn't build with tests enabled now on Cygwin anyway: https://github.com/vprover/vampire/issues/519

My branch builds on Cygwin (not in Debug mode) to no surprise since I've changed `_exit` only in unit-test related code.

For gcc/clang (@MichaelRawson asked earlier) --- they both have `--coverage` flag, I'm testing locally wih `gcc`, and CI seems to use `clang`.

CMakeLists.txt
MichaelRawson: If you can modify this call to `set()` to something like

```cmake
if(CMAKE_CXX_COMPILER_ID STREQUAL GNU OR CMAKE_CXX_COMPILER_ID MATCHES Clang$)
  set(CMAKE_CXX_FLAGS ${CMAKE_CXX_FLAGS} --coverage)
endif()
```

so that we don't break other compilers in principle, I'm happy to merge. Cygwin may well be broken, but it apparently already was so let's take one step at a time!

inpefess: @MichaelRawson done

PR #522, opened 2024-02-12
Small demodulation improvement

mezpusz:
Equations such as the following
f(x,f(y,z)) = f(z,f(y,x))
give two identical ways to demodulate, but we index both variants.

Since these equations are by definition incomparable as well, it can hurt
the performance of demodulation a bit as such demodulators are tried twice
in the unsuccessful case.

Experiments show a slight improvement over TPTP-v8.1.2 with `-sa discount -t 10`:
avg time: 5.13053s (old), 5.11622s (new)
solved: 10852 (old), 10865 (new)

### DIFF
--- a/Kernel/EqHelper.cpp
+++ b/Kernel/EqHelper.cpp
@@ -21,6 +21,7 @@
 #include "TermIterators.hpp"
 #include "ApplicativeHelper.hpp"
 #include "Lib/Metaiterators.hpp"
+#include "Matcher.hpp"
 
 #include "EqHelper.hpp"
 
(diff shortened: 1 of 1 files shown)


MichaelRawson (approved):
Nice. Seems like a slam-dunk to me, @quickbeam123.

>give two identical ways to demodulate, but we index both variants

Are there any other cases where this idea happens? I couldn't come up with any single-equation cases off the top of my head.

easychair:
I am suspicious, results are too good because I cannot imagine

  1. That many problems contain such demodulators
  2. Even if they do, 13 more problems would be solved
  3. With 0.02s average improvement we solve 13 more problems. What is avg,
by the way? average over what?

It would be good if someone checks what is going on.

A

selig:
For a `-sa discount -t 10` experiment, an increase of 15 problems feels like some could fall into standard variability. 

I'd be tempted to repeat to see how much falls into variability and whether the same 15 are gained then look at those to confirm they come from this option.

However, the code change seems a net positive without the results; do less work = good.

mezpusz:
Yes, I agree that the experiment was far from thorough, I mainly wanted to see that we don't lose proofs and that the proofs are the same (although I think there could be small changes in the proof search due to a code tree or substitution tree being organized in a slightly different way with and without a certain demodulator being indexed).

There are hundreds of benchmarks with e.g. AC-axioms where this makes a difference, but overall in the 25257 benchmarks this averages out.

Let me know if more details are needed.

quickbeam123:
I will run my own experiments anyway, will let you know!

quickbeam123:
I see around 5 problem improvement on average under shuffling:

8764.7 -> 8769.6

with discount10 and 1000 Mi instruction limit.

Although I still didn't properly work out statistics, I would guess the probability that this makes things worse is actually quite low! ;)

quickbeam123 (approved):

PR #526, opened 2024-02-29
Minor improvement in instruction limiting, randomization, and strategy sampling

quickbeam123:
- Making sure random seed is set consistently for both vampireMode and portfolioMode
- one last call to `Timer::updateInstructionCount();` in Statistics (to report a more realistic value on successful termination; otherwise it would be the one from the last timer tick)
- don't warn in `Options.cpp` for using `_simulatedInstructionLimit` (it's actually useful to have it turned on during strategy sampling where we don't know up front if we will end up with lrs or not)
- `atotf` is compatible with `sil` now
- don't randomize `termId` until a seed is set to non-default, e.g., not until after parsing

### DIFF
--- a/Kernel/Term.cpp
+++ b/Kernel/Term.cpp
@@ -37,7 +37,9 @@ constexpr unsigned Term::SPECIAL_FUNCTOR_LOWER_BOUND;
 
 void Term::setId(unsigned id)
 {
-  if (env.options->randomTraversals()) {
+  if (env.options->randomTraversals() &&
+      Random::seed() != 1) { // not until a proper seed has been set (i.e. after parsing!)
+      // (cf ProvingHelper::runVampire and getPreprocessedProblem in vampire.cpp)
     id += Random::getInteger(1 << 12) << 20; // the twelve most significant bits are randomized
   }
    _args[0]._info.id = id;
--- a/Saturation/Splitter.cpp
+++ b/Saturation/Splitter.cpp
@@ -675,9 +675,17 @@ void Splitter::init(SaturationAlgorithm* sa)
 #endif
 
   if (opts.splittingAvatimer() < 1.0) {
-    _stopSplittingAtTime = opts.splittingAvatimer() * opts.timeLimitInDeciseconds() * 100;
+    unsigned timeLimit = opts.simulatedTimeLimit(); // is also stored in deciseconds
+    if (timeLimit == 0) {
+      timeLimit = opts.timeLimitInDeciseconds();
+    }
+    _stopSplittingAtTime = opts.splittingAvatimer() * timeLimit * 100;
 #if VAMPIRE_PERF_EXISTS
-    _stopSplittingAtInst = opts.splittingAvatimer() * opts.instructionLimit();
+    unsigned instrLimit = opts.simulatedInstructionLimit();
+    if (instrLimit == 0) {
+      instrLimit = opts.instructionLimit();
+    }
+    _stopSplittingAtInst = opts.splittingAvatimer() * instrLimit;
 #endif
   } else {
     _stopSplittingAtTime = 0;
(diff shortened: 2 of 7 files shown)


Kernel/Term.cpp
```diff
@@ -37,7 +37,9 @@ constexpr unsigned Term::SPECIAL_FUNCTOR_LOWER_BOUND;
 
 void Term::setId(unsigned id)
 {
-  if (env.options->randomTraversals()) {
+  if (env.options->randomTraversals() &&
+      Random::seed() != 1) { // not until a proper seed has been set (i.e. after parsing!)
```
MichaelRawson, line 41: This seems wrong - if we don't set a random seed or set it to 1, we won't randomise `id`. Perhaps we could abuse `env.statistics->phase` here?

Kernel/Term.cpp
```diff
@@ -37,7 +37,9 @@ constexpr unsigned Term::SPECIAL_FUNCTOR_LOWER_BOUND;
...
+      // (cf ProvingHelper::runVampire and getPreprocessedProblem in vampire.cpp)
     id += Random::getInteger(1 << 12) << 20; // the twelve most significant bits are randomized
```
MichaelRawson, line 43: Also, could you remind me why this doesn't lead to collisions? 2^20 doesn't seem so large a space.

Saturation/Splitter.cpp
```diff
@@ -675,9 +675,17 @@ void Splitter::init(SaturationAlgorithm* sa)
...
   if (opts.splittingAvatimer() < 1.0) {
-    _stopSplittingAtTime = opts.splittingAvatimer() * opts.timeLimitInDeciseconds() * 100;
-#if VAMPIRE_PERF_EXISTS
-    _stopSplittingAtInst = opts.splittingAvatimer() * opts.instructionLimit();
+    unsigned timeLimit = opts.simulatedTimeLimit(); // is also stored in deciseconds
+    if (timeLimit == 0) {
+      timeLimit = opts.timeLimitInDeciseconds();
+    }
+    _stopSplittingAtTime = opts.splittingAvatimer() * timeLimit * 100;
+#ifdef VAMPIRE_PERF_EXISTS
```
MichaelRawson, line 683: `#if VAMPIRE_PERF_EXISTS`, it's always defined.

MichaelRawson (changes requested):
Generally happy! Just a few small things you might want to look at.

Kernel/Term.cpp
quickbeam123: I think I tested it. The point it we now only set Options::randomSeed -> Random::seed() after parsing. And randomizing termIds during parsing is evil, because then it's hard to sync up the portfolio master->worker setting with the vampire mode (plain single-strategy) setting. Setting random seed to 1 and wanting randomized termId's is something I can live without. (A random random seed is almost never equal to 1.) On the other hand, relying on phase means I will need to make sure nobody messes around with phase. And it's dealt with in so many places that I don't trust it atm.

Kernel/Term.cpp
quickbeam123: It will lead to a collision. But what are the changes this will happen much earlier than after assigning an id to almost all the 2^32 terms (after which we get an overflow even without `randomTraversals`)?

Note that this's is `id +=`. I wouldn't mind using xor. But the 32 bits are NOT split into a random part and an incremented part.

Saturation/Splitter.cpp
quickbeam123: Thanks for noticing this!

Kernel/Term.cpp
MichaelRawson: OK. Might try and set some kind of unreasonable default value at some point so we can detect it, but low priority. Thanks for the explanation!

MichaelRawson (approved):

PR #531, opened 2024-03-04
add guard for HOL for now until support makes it to mainline

MichaelRawson:
As widely requested, until HOL support makes it into mainline we should reject HOL problems before starting any work.

The precise point of intervention is up for debate. I chose "after the problem is loaded but before preprocessing" because it's arguably simplest for a temporary fix, and I think we at least *parse* the majority of HOL. We should also report `SZS status Inappropriate`, but as this is temporary I'm inclined not to.

### DIFF
--- a/Shell/UIHelper.cpp
+++ b/Shell/UIHelper.cpp
@@ -356,6 +356,13 @@ Problem* UIHelper::getInputProblem(const Options& opts)
   }
 
   Problem* res = new Problem(units);
+  if(res->isHigherOrder())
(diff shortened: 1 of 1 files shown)


quickbeam123:
Is higher-orderness already correctly recognized, just after the Problem constructor?

MichaelRawson:
It seems to be in testing, but you're right - I'll check.

quickbeam123:
It's fine on my side. Let me merge and reconcile the interactive mode with this.

PR #532, opened 2024-03-05
Storing Custom Data in SubstitutionTree and friends

joe-hauns:
This PR refactors substitution trees (and the respective Index interfaces) so we can store arbitrary data in the leaves.

A substitution tree can be thought of a special map `Map<Key,Value>` with `Key` being something we can decompose into substitutions, and then perform unification with (i.e. currently in vampire we have `Key in { Literal*, TypedTermList}`).
Up to this PR `Value` used to be fixed: a tuple `(TermList, Literal*, Clause*)` (called `LeafData`). 
This makes sense for some use cases (i.e. some inference rules), but not for all of them. This lead to the fact that e.g. in many inference rules the data `(TermList(t), nullptr, nullptr)` was being inserted which is a waste of memory, and leads to bugs as we always have to check whether there are null values present or not. Further always having this fixed `Value` data was rather unflexible, as we cannot store more (or different) information than `(TermList, Literal*, Clause*)` to compute an inference. An examples where more/different data needs to be stored are 
- the Fourier Motzkin Rule from the ALASCA caluclus (PR coming some time after this)
- the data `TermList _extra` that has been added for some higher-order inferences
- if i remember correctly there was also the need for something like this from @mezpusz in his work on induction.

This PR introduces a type parameter `LeafData` to `Substitution`, `Index` and friends which makes the changes described above possible, and at the same time simplifies the SubstitutionTree interface quite a bit.
The default old leaf data class is still exists as `Indexing::TermLiteralClause` for term indices and `Indexing::LiteralClause` for Literal indices. Further I introduced `Indexing::TermWithValue<V>` which can be used if you want to have a term as a key, and some other value associated with it. 
Additionally there is `Indexing::TermWithoutValue` if you want to consider the SubstitutionTree as a set and not a map.


Indexing/ClauseVariantIndex.hpp
```diff
@@ -51,6 +51,7 @@ class ClauseVariantIndex
 
 class SubstitutionTreeClauseVariantIndex : public ClauseVariantIndex
 {
+  using LiteralSubstitutionTree = Indexing::LiteralSubstitutionTree<LiteralClause>;
```
MichaelRawson, line 54: Not the point of the change, but remind me why a *clause* variant index keeps a literal in it?

Indexing/Index.hpp
```diff
@@ -19,102 +19,139 @@
...
 using namespace Lib;
 using namespace Saturation;
 
-
-/**
- * Class of objects which contain results of single literal queries.
- */
-template<class Unifier>
-struct LQueryRes
+struct LiteralClause
```
MichaelRawson, line 43: Could this be a POD struct with no constructor, etc?

Indexing/ClauseVariantIndex.hpp
joe-hauns: Well I never touched it, except for doing refactoring that should make everything do what it used to do, but looking into it briefly just now revealed that it only uses a substitution tree to match the "main literal" (`getMainLiteral`) of a clause and then does matches the rest of the literals on a per-result basis. I think this miiiight be a performance bottlekneck in some cases but I can't tell how much this really happens in practice.

Indexing/Index.hpp
joe-hauns: Yess we can change that. I only used it during refactoring to be able to have the non-null assertions there.

joe-hauns: done.

Lib/SkipList.hpp
```diff
@@ -60,8 +61,8 @@ class SkipList
   inline
   void insert(Value val)
   {
-    Value* pval = insertPosition(val);
-    *pval = val;
+    void* pval = insertPositionRaw(val);
+    new(pval) Value(std::move(val));
```
MichaelRawson, line 65: Nice. This should show up at least a bit in perf testing.

Indexing/SubstitutionTree.hpp
```diff
@@ -108,200 +108,193 @@ using namespace Kernel;
...
-      LeafData(Clause* cls, Literal* literal, TermList term)
-      : clause(cls), literal(literal), term(term), sort(TermList::empty()), extraTerm(TermList::empty()) {}
+  SubstitutionTree(SubstitutionTree const&) = delete;
+  SubstitutionTree& operator=(SubstitutionTree const& other) = delete;
+  static void swap(SubstitutionTree& self, SubstitutionTree& other) {
+    std::swap(self._nextVar, other._nextVar);
+    std::swap(self._root,    other._root);
+  }
+  SubstitutionTree& operator=(SubstitutionTree && other) { swap(*this,other); return *this; }
+  SubstitutionTree(SubstitutionTree&& other) : SubstitutionTree(0) { swap(*this, other); }
```
MichaelRawson, line 149: Copy-and-swap idiom but only for move and move-assign, right?

Indexing/SubstitutionTree.hpp
```diff
@@ -108,200 +108,193 @@ using namespace Kernel;
...
+  typedef VirtualIterator<LeafData*> LDIterator;
 
-    template<class I, class TermOrLit, class... Args> 
-    inline auto iterator(TermOrLit query, bool retrieveSubstitutions, bool reversed, Args... args)
-    { return ifElseIter(_root == nullptr, 
-        [&](){ return EmptyIter<ELEMENT_TYPE(I)>{}; },
-        [&](){ return iterPointer(std::make_unique<I>(this, _root, query, retrieveSubstitutions, reversed, std::move(args)...)); }); }
+  template<class Unifier>
+  struct QueryResult {
+    LeafData const* data;
```
MichaelRawson, line 193: If we're redesigning the query interface anyway, is it feasible/worth making this `const LeafData &data`?

MichaelRawson (approved):
Not sure why some of comments "leaked"...but whatever. Looks great to me, thanks @joe-hauns! These are some really nice changes.

@quickbeam123 - this looks like a huge change, but much of it is mechanical.

Indexing/LiteralSubstitutionTree.hpp
```diff
@@ -128,9 +151,12 @@ class LiteralSubstitutionTree
 private:
   SubstitutionTree& getTree(Literal* lit, bool complementary)
   {
-    auto idx = complementary ? lit->header() : lit->complementaryHeader();
+    auto findNegative = complementary ? lit->isPositive() : lit->isNegative();
+    auto idx = toIdx(lit->functor(), findNegative);
     while (idx >= _trees.size()) {
-      _trees.push(SubstitutionTree());
+      auto f = idxToFunctor(_trees.size());
```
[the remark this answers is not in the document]
joe-hauns: This was a minor optimization, but i don't think it makes a difference in runtime so i reverted to the old constructor as it's simpler to use.

Indexing/SubstitutionTree_FastGen.cpp
```diff
@@ -19,20 +19,20 @@
...
-#include "SubstitutionTree.hpp"
-
 namespace Indexing
 {
 
 
 
-const unsigned SubstitutionTree::GenMatcher::BACKTRACK_SEPARATOR;
+template<class LeafData_>
+const unsigned SubstitutionTree<LeafData_>::GenMatcher::BACKTRACK_SEPARATOR;
```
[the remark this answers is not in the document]
joe-hauns: tbh i don't know what this does. I just added the template mechanically so it will compiled. My intuition is the following:
During generalization we keep a stack of var -> term assignments. We need to be able to backtrack these, a bunch at a time. The special constant `BACKTRACK_SEPARATOR` helps us to mark which bunches we want to backtrack at once. But I don't understand what's going on in `*_Fast{Inst,Gen}` in details so I'd rather not add a semi-correct comment ;) 

We need the templating here as `SubstitutionTree` is not a `class` but a `template<class> class`. Otherwise we cannot specify the path. And you're right that this is surely being inlined ways.

Kernel/BestLiteralSelector.hpp
```diff
@@ -151,8 +151,7 @@ class CompleteBestLiteralSelector
     //literals will be selected.
     bool allSelected=false;
 
-    if(isNegativeForSelection(litArr[0]) && 
-      (!combSup || canBeSelected(litArr[0], &maxTermHeads))) {
+    if(isNegativeForSelection(litArr[0])) {
```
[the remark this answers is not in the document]
joe-hauns: Exactly. Only equivalent refactoring here :)

Kernel/Term.cpp
```diff
@@ -1709,6 +1709,11 @@ bool Kernel::operator<(const TermList& lhs, const TermList& rhs)
   if (lhs.isTerm()) {
     ASS(rhs.isTerm())
     return lhs.term()->getId() < rhs.term()->getId();
+  } else if (lhs.isEmpty() || rhs.isEmpty()) {
+    auto cmp = lhs.isEmpty() - rhs.isEmpty();
+    if (cmp != 0) return cmp < 0;
+    else return false;
+
```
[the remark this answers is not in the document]
joe-hauns: Yes that's right. I didn't care about how it's being ordered but just that it works with empty terms as well now.

Indexing/SubstitutionTree.hpp
joe-hauns: Yes. But could we use it for anything else as well?

MichaelRawson: In principle also for copy/copy-assign, but I don't think we want to. Just checking this.

Indexing/SubstitutionTree.hpp
MichaelRawson: No objection, as long as it's considered rather than oversight.

Indexing/SubstitutionTree_FastGen.cpp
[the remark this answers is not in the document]
MichaelRawson: Sounds sensible to me. Thanks!

Inferences/Induction.cpp
```diff
@@ -441,7 +443,7 @@ void InductionClauseIterator::processLiteral(Clause* premise, Literal* lit)
           // add formula with default bound
           if (_opt.integerInductionDefaultBound()) {
             InductionFormulaIndex::Entry* e = nullptr;
-            static TermQueryResult defaultBound(TermList(theory->representConstant(IntegerConstantType(0))), nullptr, nullptr, ResultSubstitutionSP());
+            static Bound defaultBound = Bound(DefaultBound{ .term = TypedTermList(theory->representConstant(IntegerConstantType(0))) });
```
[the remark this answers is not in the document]
hzzv: The induction part looks good to me!

MichaelRawson:
I think merging @mezpusz's branch broke this one, sorry @joe-hauns - are we ready to go otherwise?

joe-hauns:
merged master again! Everthing's clear from my side.

quickbeam123:
Hey hey, just warming up on this:
```
In file included from SAT/MinisatInterfacing.cpp:15:
In file included from SAT/MinisatInterfacing.hpp:17:
In file included from SAT/SATSolver.hpp:18:
In file included from SAT/SATLiteral.hpp:21:
In file included from ./Shell/Options.hpp:43:
In file included from ./Forwards.hpp:18:
In file included from ./Lib/VString.hpp:26:
In file included from ./Lib/STLAllocator.hpp:27:
In file included from ./Lib/Allocator.hpp:26:
./Debug/Assertion.hpp:269:36: error: call to function 'operator<<' that is neither visible in the template definition nor found by argument-dependent lookup
              << val1Str << " == " << val1 << "\n"
                                   ^
./Kernel/Term.hpp:376:37: note: in instantiation of function template specialization 'Debug::Assertion::violatedEquality<Kernel::Term::SpecialFunctor, Kernel::Term::SpecialFunctor>' requested here
    Formula* getCondition() const { ASS_EQ(specialFunctor(), SpecialFunctor::ITE); return _iteData.condition; }
                                    ^
./Debug/Assertion.hpp:87:25: note: expanded from macro 'ASS_EQ'
      Debug::Assertion::violatedEquality(__FILE__, __LINE__, #VAL1, #VAL2, VAL1, VAL2); \
                        ^
./Kernel/Term.hpp:1159:15: note: 'operator<<' should be declared prior to the call site or in namespace 'Kernel'
std::ostream& operator<<(std::ostream& out, Kernel::Term::SpecialFunctor const& self);
```

joe-hauns:
How are you compiling this? I'm not getting the warning using clang, or gcc, in either release or debug mode.

quickbeam123:
Sorry, a false alarm. Something must have been not up to date in my repo.

Indexing/Index.hpp
```diff
@@ -19,102 +19,124 @@
...
 {
-  TQueryRes() {}
-  TQueryRes(TermList t, Literal* l, Clause* c, Unifier unifier)
-  : term(t), literal(l), clause(c), unifier(std::move(unifier)) {}
-
-  TermList term;
-  Literal* literal;
-  Clause* clause;
   Unifier unifier;
+  Data const* data;
```
quickbeam123, line 121: Would it make sense for this to be a const reference instead? (Just curious.)

Indexing/Index.hpp
```diff
@@ -19,102 +19,124 @@
...
+
+  QueryRes() {}
+  QueryRes(Unifier unifier, Data const* data) 
+    : unifier(std::move(unifier))
+    , data(std::move(data)) {}
```
quickbeam123, line 126: Does it make sense to `std::move` a pointer? (Again, not 100% sure, but it looks strange to me.)

Kernel/ClauseQueue.hpp
```diff
@@ -46,9 +46,7 @@ class ClauseQueue
   /** True if the queue is empty */
   bool isEmpty() const
   { return _left->nodes[0] == 0; }
-#if VDEBUG
```
quickbeam123, line 49: I don't really mind, but why not keep ClauseQueue printing to Debug mode?

Kernel/Renaming.cpp
```diff
@@ -77,7 +78,19 @@ bool Renaming::identity() const
...
 /**
- * Make the renaming normalize variables of term or literal @c t
+ * Make the renaming normalize variables of literal @c t
+ */
+void Renaming::normalizeVariables(const Literal* t)
+{
+  normalizeVariables((const Term*) t);
+  if (t->isEquality()) {
+    normalizeVariables(SortHelper::getEqualityArgumentSort(t));
+  }
```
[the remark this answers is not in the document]
quickbeam123: Wouldn't it make sense to only do this for `isTwoVarEquality` equalities?

Indexing/Index.hpp
quickbeam123: I am also thinking of actually communicating this by value. (Where the value becomes a pointer if Data is a pointer.) This would mean it would only be efficient, if user's data is large, to require it to be heap allocated, before it's inserted into the index. Did you consider this option?

Indexing/Index.hpp
joe-hauns: No you're right, it doesn't make a difference. I just do always write a move in constructors just as sort of best-practice without thinking about the types.

Kernel/Renaming.cpp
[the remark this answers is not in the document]
bhayat-quantinuum: Yes, I think this should only be for two var equality literals. That way, we don't need a call `SortHelper::getEqualityArgumentSort` and can remove the new include as well.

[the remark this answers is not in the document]
joe-hauns: I'm not quite sure why you mean this is not sufficient, but I'm also not opposed. Why do you think so?

Indexing/Index.hpp
joe-hauns: I think it's best to keep it being a pointer here. So basically whenever we return a `QueryRes`, the data in this query result is stored somewhere in the leave of a substitution tree which is itself already heap allocated. So we don't have to access (dereference + copy) the data anywhere (not even to the heap) if we don't use it. Usually the data will consist of a term + some other things, which will be bigger in terms of memory than `Data const*`.

In terms of using a `Data const&` instead: Using templates together with references is quite a pain (implicit copies and other odd things that can occur rather unexpectedly). So this is why i opted for a `const*` instead here.

quickbeam123: Thanks for the explanations! I am glad to hear I am not the only one who finds working with references challenging ;)

Inferences/ForwardDemodulation.cpp
```diff
@@ -151,19 +150,16 @@ bool ForwardDemodulationImpl<combinatorySupSupport>::perform(Clause* cl, Clause*
         } else {
           rhsS = subs->applyToBoundResult(rhs);
         }
-        if(resultTermIsVar){
```
quickbeam123, line 154: This looks like an interesting change. Do you remember, why we don't need this line anymore?

Indexing/LiteralIndex.hpp
```diff
@@ -25,121 +25,123 @@
...
   friend std::ostream& operator<<(std::ostream& out,                 LiteralIndex const& self) { return out << *self._is; }
   friend std::ostream& operator<<(std::ostream& out, OutputMultiline<LiteralIndex>const& self) { return out << multiline(*self.self._is, self.indent); }
 
 protected:
-  LiteralIndex(LiteralIndexingStructure* is) : _is(is) {}
+  LiteralIndex(LiteralIndexingStructure<Data>* is) : _is(is) {}
 
-  void handleLiteral(Literal* lit, Clause* cl, bool add);
+  void handle(Data data, bool add)
+  { _is->handle(std::move(data), add); }
```
quickbeam123, line 59: Could one make it explicit, with `via handle(Data&& data, bool add)` that data is going to be moved? Or was the intent to copy data to `LiteralIndex::handle` but move them to `_is->handle`?)

Inferences/ForwardDemodulation.cpp
joe-hauns: Tbh I'm not sure whether I did this intentionally. I reverted it now anyways.

Indexing/LiteralIndex.hpp
joe-hauns: Yes exactly this is the reason why we're using the tie here. We can't do this in `LiteralClause` because we can't return a reference to the result of `clause->number()` and `literal->getId()` (as they are only live during the call of `asTuple`), thus we need to use `make_tuple` in this case. In any case we want as tuple just to return some tuple of which evey element can be ordered using `std::less`, hashed, ..., and what exact the type tuple elements are doesn't matter.

Indexing/Index.hpp
```diff
@@ -19,102 +19,124 @@
...
+
+  TermWithValue(TypedTermList term, Value v)
+    : term(term)
+    , value(std::move(v))
+  {}
+
+  TypedTermList const& key() const { return term; }
+
+  auto asTuple() const
+  { return std::tie(term, value); }
```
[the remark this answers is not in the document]
joe-hauns: Yes exactly this is the reason why we're using the tie here. We can't do this in LiteralClause because we can't return a reference to the result of clause->number() and literal->getId() (as they are only live during the call of asTuple), thus we need to use make_tuple in this case. In any case we want as tuple just to return some tuple of which evey element can be ordered using std::less, hashed, ..., and what exact the type tuple elements are doesn't matter.

Indexing/LiteralIndex.hpp
quickbeam123: I was just guessing that since you currently only use `handle` with temporaries, we might get a tiny more efficient code with `&&`. Or is this a place where the compiler can save the copy for us, @MichaelRawson ?

joe-hauns: I think in terms of efficiency it should be just the same as long as we didn't implement the constructor `Data(Data&&)` in some unexpected way. For me the design-question would just be whether we want to force the caller to explicitly call a copy constructor or not.

MichaelRawson: Temporaries are probably OK, I think this is more a matter of taste as Joe suggests. I expect most are now guaranteed to be elided since C++17, see the example at the bottom of https://en.cppreference.com/w/cpp/language/copy_elision

As a rule I'd probably take everything by value if you need to own the object (unless you semantically wanted an rvalue for some reason), allowing either a copy or move. This may be shaped by Rust habits more than C++ considerations.

PR #533, opened 2024-03-06
put LTB and dependencies on hiatus

MichaelRawson:
See #157. @quickbeam123's recent interactive mode removes the need for LTB from a user perspective. LTB is also on hiatus at CASC.

My view is that if LTB comes back and we need to restore it in the future, we can pay the maintenance cost then. Therefore: remove LTB and some of its dependencies from Vampire. `Lib::FreshnessGuard` is apparently unrelated but also unused.


Lib/Environment.cpp
```diff
@@ -174,36 +163,7 @@ bool Environment::haveOutput()
 ostream& Environment::out()
 {
   ASS(_outputDepth);
```
MichaelRawson, line 165: All of these `env.out` related functions are now almost trivial. Could they go altogether in a subsequent PR?

MichaelRawson:
Also remove `tTwoVampires`. This has been of questionable value for a while but now it mostly tests stuff that no longer exists.

Lib/Environment.cpp
quickbeam123: Yes, let's remove `env.out`. It's not been used consistently throughout anyway (and was doing the same as `std::cout <<`, I think).

quickbeam123 (approved):

PR #534, opened 2024-03-12
remove env.out()

MichaelRawson:
See #157 and #533. `env.out()` and friends no longer do much, so they can be replaced with `std::cout` mechanically.

### DIFF
--- a/CASC/PortfolioMode.cpp
+++ b/CASC/PortfolioMode.cpp
@@ -45,15 +45,17 @@
 
 #include "PortfolioMode.hpp"
 
-using namespace std;
 using namespace Lib;
 using namespace CASC;
 using Lib::Sys::Multiprocessing;
+using std::cout;
+using std::cerr;
+using std::endl;
 
 PortfolioMode::PortfolioMode(Problem* problem) : _prb(problem), _slowness(env.options->slowness()), _syncSemaphore(2) {
   unsigned cores = System::getNumberOfCores();
   cores = cores < 1 ? 1 : cores;
-  _numWorkers = min(cores, env.options->multicore());
+  _numWorkers = std::min(cores, env.options->multicore());
   if(!_numWorkers)
   {
     _numWorkers = cores >= 8 ? cores - 2 : cores;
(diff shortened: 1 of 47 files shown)


CASC/PortfolioMode.cpp
```diff
@@ -95,32 +95,30 @@ bool PortfolioMode::perform(Problem* problem)
   }
 
   if (outputAllowed()) {
-    env.beginOutput();
     if (resValue) {
-      addCommentSignForSZS(env.out());
-      env.out()<<"Success in time "<<Timer::msToSecondsString(env.timer->elapsedMilliseconds())<<endl;
+      addCommentSignForSZS(std::cout);
+      std::cout<<"Success in time "<<Timer::msToSecondsString(env.timer->elapsedMilliseconds())<<endl;
     }
```
quickbeam123, line 101: What do you think about `std::cout` vs just `cout` (and having the right namespace/using directive always active)? This is minor, but since we are already touching XXX files and it's just about a simple text replacing? 

I personally prefer the lazy (shorter) version.

And here it is particularly strange, as the fully qualified use is followed up by a short `endl`.

(Moreover, we could discuss whether `addCommentSignForSZS` shouldn't simply hardwire `std::cout` and be parameterless.)

MichaelRawson: Could be either, matter of taste etc. Two mild points against unqualified `cout`:
1. This suggests that `using namespace std;` is somewhere floating around. This is true in Vampire but it shouldn't be. `using std::cout;` is OK but I think I'd prefer it scoped relatively close to use.
2. Generally I'd explicitly qualify `std::` so that when I actually want ADL magic it stands out.

But I'm not fussy, either works for me.

`addCommentSignForSZS` - absolutely. Although that's also used inconsistently so maybe that could/should go too.

quickbeam123: OK, let's prefer the fully qualified then.

With `addCommentSignForSZS` there seems to be at least one place in the code where we don't want to use `cout`, so let's keep it as is for now at least...

CASC/PortfolioMode.cpp
```diff
@@ -45,15 +45,17 @@
 
 #include "PortfolioMode.hpp"
 
-using namespace std;
 using namespace Lib;
 using namespace CASC;
 using Lib::Sys::Multiprocessing;
+using std::cout;
+using std::cerr;
+using std::endl;
```
quickbeam123, line 54: What do you think, @MichaelRawson , about this particular file now, after my change?

quickbeam123: Funnily, `move` worked without the full qualification even after I removed `using namespace std;`.

MichaelRawson: Commit message probably a little harsh! But looks good to me.

MichaelRawson: > Funnily, `move` worked without the full qualification even after I removed `using namespace std;`.

Could be! I assume this is ADL at work.

Lib/Environment.cpp
```diff
@@ -119,51 +114,6 @@ int Environment::remainingTime() const
...
-/**
- * Release the output stream
- */
-void Environment::endOutput()
-{
-  ASS_G(_outputDepth,0);
-
-  _outputDepth--;
-  if(_outputDepth==0) {
-    cout.flush();
```
quickbeam123, line 143: Shell we theoretically be missing some flushings now?

quickbeam123 (approved):

Lib/Environment.cpp
MichaelRawson: Theoretically. In practice most call sites use `std::endl` somewhere which will cause an equivalent flush. Also not sure the OS is required to obey a runtime flush request, so it may never have done what we wanted.

PR #539, opened 2024-03-20
Demodulation cosmetics

quickbeam123:
1) getDemodulationLHSIterator does not need to be given the options object (we already know at the sites where we use it, whether we are doing FWD or BWD)
2) call SortHelper::getEqualityArgumentSort exactly once in withEqualitySort (and not potentially twice); but don't use on empty iterators

These are almost trivial changes. Just something I noticed before the breakfast...

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -394,6 +394,7 @@ set(VAMPIRE_INFERENCE_SOURCES
     Inferences/BackwardSubsumptionResolution.cpp
     Inferences/BinaryResolution.cpp
     Inferences/Condensation.cpp
+    Inferences/DemodulationHelper.cpp
     Inferences/DistinctEqualitySimplifier.cpp
     Inferences/EqualityFactoring.cpp
     Inferences/EqualityResolution.cpp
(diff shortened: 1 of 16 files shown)


MichaelRawson (approved):
Agreed, cosmetically better and we should generally reduce use of `env`.

quickbeam123:
Strictly better for discount10:

```
Sort by SAT
1006 ['problemsSTD_master7392_dis10_i10000.pkl']
1006 ['problemsSTD_mdc7396_dis10_i10000.pkl']

Sort by UNS
7763 ['problemsSTD_master7392_dis10_i10000.pkl']
7766 ['problemsSTD_mdc7396_dis10_i10000.pkl']

Total 1006 / 7766

Greedy cover (UNS):
# problemsSTD_mdc7396_dis10_i10000.pkl contributes 7766 total 7766
Total 7766
```

joe-hauns (commented):
Looks all good to me! Loving the code deduplication using `DemodulationHelper`!

joe-hauns (approved):

Issue #462, opened 2023-07-18
Reduce reliance on platform-specific APIs, improve Windows support


MichaelRawson:
Vampire should not need a great deal of platform-specific code. This is a tracking issue to remove as much as possible, and make what we absolutely need as cross-platform as we can.

We currently use platform-specific code:
- [ ] for launching and monitoring processes in portfolio mode
- [ ] for monitoring elapsed time
- [ ] for monitoring used instructions
- [ ] for semaphores used to synchronise portfolio processes
- [ ] for our custom memory allocator
- [ ] because we forgot to remove it

I'm currently looking at whether we can remove the memory allocator and the semaphores. Watch this space!

MichaelRawson:
Using something akin to UNIX `fork()` is not really negotiable - we want to keep the existing Vampire state, notably the parsed problem, when we start a strategy - but it [doesn't exist on Windows](https://stackoverflow.com/questions/985281/what-is-the-closest-thing-windows-has-to-fork). We will have to rely on Cygwin for this on Windows, and hope that the various incompatibilities don't bite us.

MichaelRawson:
Allocator is dead, long live Allocator! #469 removed the platform-specific code for allocation.

MichaelRawson:
No more semaphores, no more semaphore support requirement.

MichaelRawson:
Replacing global `operator new` portably to limit memory usage was so painful in the end (#592) that we gave up and call POSIX `setrlimit` in #594. Ho hum. Better POSIX than Linux.

PR #540, opened 2024-03-20
file-based portfolio synchronisation

MichaelRawson:
See #462. We currently synchronise child Vampires and their proof printing by UNIX semaphores, which are quite complex beasts. However, C11 (and therefore C++17) can synchronise on a file by passing an "x"clusive flag to `fopen`: if the file exists, `fopen` will fail. Note that the standard only says "to the extent the platform supports exclusive access", but it seems to work quite well in practice.

Therefore:
1. We determine a path that should be written to. This is either what the user requested or a temporary file.
2. We delete it if it exists!
3. The first child to *not* find that file creates it and writes its proof there.
4. Other children find the file and do not print a proof.
5. The parent Vampire knows the path and reads from it once all children have finished (one way or another).

I think this is simpler and certainly more portable, but of course it needs lots of testing: it would be a silly way to lose problems.


MichaelRawson:
To explain the signal-handling change: currently we handle many different signals with very similar but slightly different code. This looks more like organic growth over time rather than something intentionally complex, so I've streamlined it. There are now only two broad classes of signal we handle:
1. Exit now, silently: `SIGHUP` (portfolio parent died), `SIGINT` (Ctrl-C), `SIGTERM` (polite termination), and `SIGXCPU` (we're being warned that we've used an external resource limit and are about to be killed). `SIGXCPU` has moved from the other class, I don't think we want to special-case this.
2. Crash of some kind: everything else, including `SIGQUIT` as apparently that's an abnormal halt intended to be used for debugging. Here we print a stack trace, indicate something to Spider, etc.

quickbeam123:
Since you already polished the signals, could you please give me back the feature we used to have, that Vampire reports statistics (if asked) and termination reason on CTRL+C?

Lib/System.cpp
```diff
@@ -115,13 +115,16 @@ void handleSignal (int sigNum)
     TERMINAL_SIGNAL_HANDLED = true;
     Shell::reportSpiderFail();
     if(Shell::outputAllowed(true)) {
-      std::cout << getpid() << " Aborted by signal " << signalToString(sigNum);
+      Shell::addCommentSignForSZS(std::cout);
+      std::cout << "Aborted by signal " << signalToString(sigNum);
```
MichaelRawson, line 119: Do we perhaps still want the PID to work out which process has received the signal? But no strong feelings, just wondered if this was an oversight.

MichaelRawson: Ahah, my mistake. Thanks!

Issue #384, opened 2022-06-09
Possible issue in substitution trees - check whether _nextVar really needs to keep growing forever


quickbeam123:
(The problem being we would eventually run out `_nextVar`s - which actually happened before, in an actual run, when `SPEC_UPPER_BOUND` was "only" `10000000`, now it's about 50times larger.)

From Michael:

"We only "run out" of special variables because substitution trees do something a bit surprising with `_nextVar` to achieve freshness."

MichaelRawson:
@ibnyusuf @joe-hauns - I *believe* this can now be closed and `SPEC_UPPER_BOUND` can be deleted. We no longer have "very special vars" so all special variables are just the right amount of special. What do you think?

joe-hauns:
Yesss I think you're right. I think i just didn't spot that when i removed very special variables.

MichaelRawson:
Thanks! I'll add it to the default-constructor PR.

PR #541, opened 2024-03-22
have the TermList default constructor zero-initialise

MichaelRawson:
Somewhat trivial, but in discussion with @quickbeam123 have `TermList` zero-initialise its content. In (almost?) all cases this should be immediately overwritten, but it means that when there is a bug, we will get a hard null-dereference crash rather than...something else.

### DIFF
--- a/Indexing/TermSharing.cpp
+++ b/Indexing/TermSharing.cpp
@@ -144,6 +144,7 @@ void TermSharing::computeAndSetSharedTermData(Term* t)
       }
       else 
       {
+        ASS(tt->isTerm());
         ASS_REP(tt->term()->shared(), tt->term()->toString());
         
         Term* r = tt->term();
--- a/Kernel/Term.hpp
+++ b/Kernel/Term.hpp
@@ -101,7 +101,8 @@ class TermList {
 public:
   // divide by 4 because of the tag, by 2 to split the space evenly
   static const unsigned SPEC_UPPER_BOUND = (UINT_MAX / 4) / 2;
-  TermList() : _content(0) {};
+  /* default constructor, satisfying isEmpty() */
+  TermList() : _content(FUN) {}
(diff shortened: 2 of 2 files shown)


easychair:
I might be not up-to-date with the current code, but ... if the default
initialization is ever used, would this not break term sharing? Or maybe in
the future, if we decide that the default initialization can be legally
used?

Andrei

MichaelRawson:
@joe-hauns - I had the same idea, but it seemed to me that an empty TermList should be a zero-initialised TermList. This doesn't work for some reason and crashes a lot - so I should figure out why, if two people want this.

@easychair - this is a good point. In principle it should not break term sharing, because we should never insert a Term containing an empty TermList into term sharing. But it would be smart to have some checks for this.

Either way - on hold for now until I look into it more, probably next week at this point.

easychair:
There is a school of thought that says one should initialize memory which
we are not supposed to access by rubbish that would almost certainly
guarantee that accessing it will break the application.

null (0) is not rubbish

Of course, initializing by rubbish (even constant rubbish) may take a few
microseconds and leave Vampire in second place in CASC :)

Cheers, Andrei

MichaelRawson:
> Could we indstead of zero-initializing it initializing it in such a way that `Term::isEmpty` returns true?

Done. :-) I've also added an assertion to TermSharing to make sure we don't inadvertently break that.

I did briefly try to have two categories of "empty" TermLists: the current one and a zero TermList, but Vampire uses both "invalid TermList" and "off the end of an argument list" somewhat blurrily so that wasn't workable.

MichaelRawson:
See #384 - we no longer have "very special" variables, so `SPEC_UPPER_BOUND` is pointless. Also remove this here as I'm passing. Checking whether something is a special variable should now be very slightly faster.

PR #543, opened 2024-04-10
Code tree search structs refactor + lazily expanded flat terms

mezpusz:
Two main changes in this pull request:
- `CodeTree::SearchStruct` is refactored avoiding duplication of subclasses. Raw array containing value-target pairs is replaced with STL containers for better and easier handling.
- Added new tag `FUN_UNEXPANDED` to `FlatTerm` in which case only the tag, term pointer and number of subentries is filled out first. A call on `FlatTerm::Entry::expand` fills out the same entries for the arguments. The point is that we expand tha flat term lazily which in some cases would be a bottleneck when the `FlatTerm` is not even used.

Additionally, added some basic printing for code trees.

### DIFF
--- a/Indexing/ClauseCodeTree.cpp
+++ b/Indexing/ClauseCodeTree.cpp
@@ -312,7 +312,7 @@ bool ClauseCodeTree::removeOneOfAlternatives(CodeOp* op, Clause* cl, Stack<CodeO
 {
   unsigned initDepth=firstsInBlocks->size();
 
-  while(!op->isSuccess() || op->getSuccessResult()!=cl) {
+  while(!op->isSuccess() || op->getSuccessResult<Clause>()!=cl) {
     op=op->alternative();
     if(!op) {
       firstsInBlocks->truncate(initDepth);
@@ -574,7 +574,7 @@ Clause* ClauseCodeTree::ClauseMatcher::next(int& resolvedQueryLit)
       }
     }
     else if(lm->op->isSuccess()) {
-      Clause* candidate=static_cast<Clause*>(lm->op->getSuccessResult());
+      Clause* candidate=lm->op->getSuccessResult<Clause>();
       RSTAT_MCTR_INC("candidates", lms.size()-1);
       if(checkCandidate(candidate, resolvedQueryLit)) {
 	RSTAT_MCTR_INC("candidates (success)", lms.size()-1);
--- a/Indexing/ClauseCodeTree.cpp
+++ b/Indexing/ClauseCodeTree.cpp
@@ -74,6 +74,7 @@ void ClauseCodeTree::insert(Clause* cl)
   cctx.init();
 
   for(unsigned i=0;i<clen;i++) {
+    cctx.nextLit();
     compileTerm(lits[i], code, cctx, true);
   }
   code.push(CodeOp::getSuccess(cl));
@@ -185,7 +186,7 @@ void ClauseCodeTree::matchCode(CodeStack& code, CodeOp* startOp, size_t& matched
       if(treeOp->isSearchStruct()) {
 	SearchStruct* ss=treeOp->getSearchStruct();
 	CodeOp** toPtr;
-	if(ss->getTargetOpPtr(code[i], toPtr) && *toPtr) {
+	if(ss->getTargetOpPtr<false>(code[i], toPtr) && *toPtr) {
 	  treeOp=*toPtr;
 	  continue;
 	}
--- a/Indexing/CodeTree.cpp
+++ b/Indexing/CodeTree.cpp
@@ -35,6 +35,9 @@
 namespace Indexing
 {
 
+#define GET_CONTAINING_OBJECT(ContainingClass,MemberField,object) \
+  reinterpret_cast<ContainingClass*>(reinterpret_cast<size_t>(object)-offsetof(ContainingClass,MemberField))
+
 using namespace std;
 using namespace Lib;
 using namespace Kernel;
@@ -71,7 +74,7 @@ CodeTree::LitInfo CodeTree::LitInfo::getOpposite(const LitInfo& li)
   ft->changeLiteralPolarity();
 #if GROUND_TERM_CHECK
   ASS_EQ((*ft)[1].tag(), FlatTerm::FUN_TERM_PTR);
-  (*ft)[1]._ptr=Literal::oppositeLiteral(static_cast<Literal*>((*ft)[1].ptr()));
+  (*ft)[1]._ptr=Literal::complementaryLiteral(static_cast<Literal*>((*ft)[1].ptr()));
 #endif
 
   LitInfo res=li;
@@ -127,7 +130,7 @@ void CodeTree::MatchInfo::init(ILStruct* ils, unsigned liIndex_, DArray<TermList
 }
 
 
-CodeTree::ILStruct::ILStruct(Literal* lit, unsigned varCnt, Stack<unsigned>& gvnStack)
+CodeTree::ILStruct::ILStruct(const Literal* lit, unsigned varCnt, Stack<unsigned>& gvnStack)
 : varCnt(varCnt), sortedGlobalVarNumbers(0), globalVarPermutation(0), timestamp(0)
 {
   ASS_EQ(matches.size(), 0); //we don't want any uninitialized pointers in the array
@@ -321,7 +324,7 @@ CodeTree::CodeOp CodeTree::CodeOp::getTermOp(InstructionSuffix i, unsigned num)
   return res;
 }
 
-CodeTree::CodeOp CodeTree::CodeOp::getGroundTermCheck(Term* trm)
+CodeTree::CodeOp CodeTree::CodeOp::getGroundTermCheck(const Term* trm)
 {
   ASS(trm->ground());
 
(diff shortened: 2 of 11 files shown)


Indexing/CodeTree.cpp
```diff
@@ -357,109 +360,137 @@ bool CodeTree::CodeOp::equalsForOpMatching(const CodeOp& o) const
...
 
+template<bool doInsert>
 bool CodeTree::SearchStruct::getTargetOpPtr(const CodeOp& insertedOp, CodeOp**& tgt)
 {
   switch(kind) {
   case FN_STRUCT:
     if(!insertedOp.isCheckFun()) { return false; }
-    tgt=&static_cast<FnSearchStruct*>(this)->targetOp(insertedOp.arg());
+    tgt=&static_cast<FnSearchStruct*>(this)->targetOp<doInsert>(insertedOp.arg());
     return true;
```
mezpusz, line 451: I'm not sure returning anything here makes sense as we only check the result in debug mode where the violation below would kick in, otherwise it's simply undefined behaviour

Indexing/CodeTree.hpp
```diff
@@ -109,7 +109,7 @@ class CodeTree
    */
   struct ILStruct
   {
-    ILStruct(Literal* lit, unsigned varCnt, Stack<unsigned>& gvnStack);
+    ILStruct(const Literal* lit, unsigned varCnt, Stack<unsigned>& gvnStack);
```
MichaelRawson, line 112: I don't object to this at all. But aren't (almost) all term/literal pointers morally `const`? What caused you to make this change? Some compiler lint? Just curious.

Indexing/CodeTree.hpp
```diff
@@ -254,65 +256,70 @@ class CodeTree
...
     };
 
+    /**
+     * Actual code op for this search structure. This construction
+     * implies that search structure operations cannot be stored
+     * in @b CodeBlock containers.
+     */
     CodeOp landingOp;
     Kind kind;
+    vvector<CodeOp*> targets;
```
MichaelRawson, line 294: Could this be `vvector<CodeOp>`?

Indexing/CodeTree.hpp
```diff
@@ -254,65 +256,70 @@ class CodeTree
...
   };
 
-  struct FixedSearchStruct
+  template<SearchStruct::Kind k>
+  struct SearchStructImpl
   : public SearchStruct
   {
-    FixedSearchStruct(Kind kind, size_t length);
-    ~FixedSearchStruct();
+    USE_ALLOCATOR(SearchStructImpl);
```
MichaelRawson, line 304: Did this give a performance win? If not, no need for it any more.

Indexing/CodeTree.hpp
```diff
@@ -254,65 +256,70 @@ class CodeTree
     CodeOp* _alternative;
   };
 
+  /**
+   * A search structure that collects alternatives of the same
+   * kind (either function symbols or ground terms) and offers
+   * more efficient searching and insertion over them.
+   */
```
MichaelRawson, line 263: Could you say something about *what* the data structure is? IIRC it was something non-obvious.

Indexing/CodeTree.cpp
```diff
@@ -35,6 +35,9 @@
 namespace Indexing
 {
 
+#define GET_CONTAINING_OBJECT(ContainingClass,MemberField,object) \
+  reinterpret_cast<ContainingClass*>(reinterpret_cast<size_t>(object)-offsetof(ContainingClass,MemberField))
```
MichaelRawson, line 39: This looks dubious to me as I don't think `size_t` is guaranteed to be big enough. What I think you want to do it cast to `char *` and then do the arithmetic on that.

Indexing/CodeTree.cpp
```diff
@@ -357,109 +360,137 @@ bool CodeTree::CodeOp::equalsForOpMatching(const CodeOp& o) const
...
-CodeTree::SearchStruct::SearchStruct(Kind kind)
+CodeTree::SearchStruct::SearchStruct(Kind kind, size_t length)
 : kind(kind)
 {
   landingOp.setAlternative(0);
   landingOp.setLongInstr(SEARCH_STRUCT);
-}
+  ASS(length);
 
-void CodeTree::SearchStruct::destroy()
```
MichaelRawson, line 381: Not needed any more?

Indexing/CodeTree.cpp
MichaelRawson: If you can put the assertion "inside" somehow, do it. Otherwise I think it's fine, it indicates *something* and the compiler will happily inline/discard unused values.

Indexing/CodeTree.cpp
```diff
@@ -627,45 +597,36 @@ CodeTree::~CodeTree()
 CodeTree::CodeBlock* CodeTree::firstOpToCodeBlock(CodeOp* op)
 {
   ASS(!op->isSearchStruct());
-
-  //the following line gives warning for not being according
-  //to the standard, so we have to work around
```
MichaelRawson, line 632: :grimacing: Glad that you cleaned this up.

MichaelRawson (approved):
Fine by me, see comments. As usual I trust you with the details of what you're actually doing but it looks generally sensible and there's less hold-on-to-your-hat pointer manouevring going on, so I'm happy about that!

Indexing/ClauseCodeTree.cpp
```diff
@@ -74,6 +74,7 @@ void ClauseCodeTree::insert(Clause* cl)
   cctx.init();
 
   for(unsigned i=0;i<clen;i++) {
+    cctx.nextLit();
```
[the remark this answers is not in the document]
mezpusz: I moved it out for a different change in a sequel branch, but I think it fits better here.

Indexing/CodeTree.hpp
mezpusz: There was a change that enforced this. I'm not sure the change is in the branch anymore, but I thought I'd leave it here if it compiles.

Indexing/CodeTree.cpp
mezpusz: Good catch, looks like I slipped here. Probably should reinstate this altogether.

mezpusz: I think introducing virtual destructors for this does not worth it, right?

Indexing/CodeTree.hpp
mezpusz: Doable, but upon removal we would have to shrink the vector to get rid of the non-existing entry, now we simply have null.

Indexing/CodeTree.hpp
mezpusz: Not sure, I will remove it.

Kernel/FlatTerm.cpp
```diff
@@ -163,5 +190,33 @@ void FlatTerm::swapCommutativePredicateArguments()
...
+void FlatTerm::Entry::expand()
+{
+  if (tag()==FUN) {
+    return;
+  }
+  ASS(tag()==FUN_UNEXPANDED);
+  ASS(this[1].tag()==FUN_TERM_PTR);
+  ASS(this[2].tag()==FUN_RIGHT_OFS);
+  Term* t = this[1].ptr();
+  size_t p = FlatTerm::functionEntryCount;
```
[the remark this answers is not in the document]
mezpusz: You mean just changing the identifier, not to a macro?

Kernel/FlatTerm.hpp
```diff
@@ -54,14 +63,20 @@ class FlatTerm
...
+     * arguments with FUN_UNEXPANDED values. Otherwise does nothing.
+     */
+    void expand();
 
     union {
       Term* _ptr;
       struct {
-	unsigned tag : 2;
-	unsigned number : 30;
+	unsigned tag : 4;
```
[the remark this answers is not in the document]
MichaelRawson: Sure! If you're willing to do it *and* to factor it out, even better - take a look at what is done in `TermList`. `BitUtils` sounds like a sensible place to me - probably the bit-twiddling templated functions can be moved there.

Kernel/FlatTerm.cpp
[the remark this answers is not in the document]
MichaelRawson: Exactly - it just looked something mutable initially.

Indexing/CodeTree.cpp
MichaelRawson: Depends - I think you just `delete` whatever instead, right? As long as you get the deletion logic right, I'm fine with it.

Virtual destructors: only if you need different destruction logic per-class. I think here it's just deleting the memory it occupied.

Indexing/CodeTree.hpp
MichaelRawson: Ah - OK. It could be worth doing some kind of "tombstone" logic, setting a flag in `CodeOp`? Doing the double-indirect possibly costs you a bit. But we can leave that for a future merge.

Indexing/CodeTree.hpp
MichaelRawson: Thanks. I'm also not sure how Vampire's allocator feels about deleting a derived class via a pointer of base type so treat that kind of thing with caution. At some point `USE_ALLOCATOR` should go altogether and be retained only for a few select classes.

Kernel/FlatTerm.hpp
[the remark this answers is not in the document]
mezpusz: Regarding the `alignas`, should it be done in `Term.hpp`?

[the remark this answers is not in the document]
MichaelRawson: Yes, something like `class alignas(8) Term { ... };`. We *probably* get this "naturally", but it would be good to enforce it.

Indexing/CodeTree.hpp
mezpusz: You mean sorted array and a binary search method? I'm not sure what else is there.

MichaelRawson: Ah - OK, that worked in the end? Great stuff, ignore me then. I thought there was some complicated triangular tree thing going on implicitly that I was too stupid to see.

Kernel/FlatTerm.hpp
```diff
@@ -55,38 +55,56 @@ class FlatTerm
...
-	unsigned number : 28;
-      } _info;
-    };
+    uint64_t _content;
+
+    static constexpr unsigned
+      TAG_BITS_START = 0,
+      TAG_BITS_END = TAG_BITS_START + 3,
+      NUMBER_BITS_START = TAG_BITS_END,
+      NUMBER_BITS_END = NUMBER_BITS_START + 27,
```
MichaelRawson, line 78: You can have up to 64 - 3 = 61 bits for `number` here, if you want. 27 is a bit random, I might go for 29 (29 + 3 = 32), 32, or 61. Performance for any of these will be ballpark the same but you might shave a few cycles off here or there with one choice or another.

Kernel/FlatTerm.hpp
```diff
@@ -55,38 +55,56 @@ class FlatTerm
...
+      TERM_BITS_END = CHAR_BIT * sizeof(Term *);
+
+    // various properties we want to check
+    static_assert(TAG_BITS_START == 0, "tag must be the least significant bits");
+    static_assert(TERM_BITS_START == 0, "term must be the least significant bits");
+    static_assert(sizeof(void *) <= sizeof(uint64_t), "must be able to fit a pointer into a 64-bit integer");
+    static_assert(FUN_UNEXPANDED < 8, "must be able to squash orderings into 3 bits");
+
+  // getters and setters
+#define GET_AND_SET(type, name, Name, NAME) \
```
MichaelRawson, line 89: Just to say the `GET_AND_SET` macro is not actually necessary. But if you like it, feel free.

Lib/BitUtils.hpp
```diff
@@ -70,6 +70,45 @@ class BitUtils
...
+  // get the bits of `_content` between `lower` and `upper`
+  template<unsigned lower, unsigned upper, class T>
+  static uint64_t getBits(const T& t) {
+    auto mask = bitmask64<lower, upper>();
+    return (t._content & mask) >> lower;
+  }
+
+  // set the bits of `_content` between `lower` and `upper` to corresponding bits of `data`
+  template<unsigned lower, unsigned upper, class T>
+  static void setBits(T& t, uint64_t data) {
```
MichaelRawson, line 97: I guess these should be called `getBits64` and `setBits64`, but I'll fix it eventually.

Indexing/CodeTree.hpp
```diff
@@ -188,65 +191,76 @@ class CodeTree
...
+      INSTRUCTION_BITS_START = 0,
+      INSTRUCTION_BITS_END = INSTRUCTION_BITS_START + 3,
+      ARG_BITS_START = INSTRUCTION_BITS_END,
+      ARG_BITS_END = CHAR_BIT * sizeof(uint64_t),
+      DATA_BITS_START = 0,
+      DATA_BITS_END = CHAR_BIT * sizeof(void *);
+
+    static_assert(sizeof(void *) <= sizeof(uint64_t), "must be able to fit a pointer into a 64-bit integer");
+    static_assert(SEARCH_STRUCT < 8, "must be able to squash instructions into 3 bits");
+    static_assert(alignof(Term) == 8);
```
MichaelRawson, line 240: Strictly speaking `alignof(Term)` should be a *multiple* of 8. But this is probably fine, and we'll get a nice compile-time failure if it isn't.

MichaelRawson:
Thanks for the fixes/improvements, @mezpusz! I'm happy apart from the `GET_CONTAINING_OBJECT` macro when you want to take a look, @quickbeam123.

Indexing/CodeTree.cpp
mezpusz: You're right, this is also implemented on [Linux](https://www.linuxjournal.com/files/linuxjournal.com/linuxjournal/articles/067/6717/6717s2.html). The main problem is with `char*` we need to `const_cast` as well, because it is called with both const and non-const arguments. Let me know which one do you think is better.

MichaelRawson: Two macros, one for each case? Could also then be `static_cast`, I think.

quickbeam123:
No new assertion violations and slightly improves performance. I think it's ready to go!

MichaelRawson:
OK! I'll wait to see how/if @mezpusz wants to fix `GET_CONTAINING_OBJECT` and then we can merge, I think.

Issue #542, opened 2024-04-05
Miscellanious Crashes


joe-hauns:
Doing some randomized testing I ran into a couple of miscellanious crashes. Many of them are related to parsing. In some cases we just do not support what's being parsed, in some cases it's an actual crash. I decided I'll make an issue here so we can dicuss all of these crashes, and decide how to fix them, or how to exit more gracefully than crashing:

# Bug 1: 
Files:
    - ITP/ITP236_4.p
    - ITP/ITP285_4.p
    - ITP/ITP266_4.p
    - ITP/ITP235_4.p
    - ITP/ITP220_4.p
Options: none
Issue: some crash in the TPTP parser

Example:
```
❯ ./cmake-build/vampire-master-Debug/clang/vampire TPTP-v8.2.0/Problems/ITP/ITP285_4.p
% Running in auto input_syntax mode. Trying TPTP
Condition in file /Users/jo/projects/vampire/master/Kernel/SortHelper.cpp, line 95 violated:
!subst.isEmpty() || (result.isTerm() && (result.term()->isSuper() || result.term()->ground()))
----- stack dump -----
Version : Vampire 4.8 (commit 23d7e9c3b on 2024-03-22 08:41:08 +0000)
call stack: 0x182bce0e0 0x10055eb04 0x100b7545c 0x100b74f14 0x100b74344 0x100b9ccc4 0x100b9dce4 0x100ba0de4 0x100bb053c 0x100bb1084 0x1006c74f0 0x1006c7244 0x10056c99c 0x10056ec44
(use '--traceback on' to get a human-readable stack trace)
----- end of stack dump -----
```

# Bug 2
Files:
- SWW/SWW508_5.p

Options: `-fda 4`

Issue: Crash in function definition introduction

Example:
```
❮ ./cmake-build/vampire-master-Debug/clang/vampire  TPTP-v8.2.0/Problems/SWW/SWW508_5.p -fdi 4

% Running in auto input_syntax mode. Trying TPTP
Condition env.signature->functionArity(function) == arity in file /Users/jo/projects/vampire/master/Kernel/Term.cpp, line 833 was violated, as:
env.signature->functionArity(function) == 2
arity == 1
----- stack dump -----
Version : Vampire 4.8 (commit 23d7e9c3b on 2024-03-22 08:41:08 +0000)
[...]

PR #544, opened 2024-04-10
avoid sortedness errors with polymorphically let-bound constants

MichaelRawson:
Fixes bug 1 of #542 by providing a fast path for constants in `getResultSort` that also does not assert a condition that isn't true for let-bound constants.

A TFX1 `$let` might capture sort variables. Consider e.g. polymorphic lists and the following axiom.

```
tff(bug, axiom, ![A: $tType]: $let(x: list(A), x := nil(A), e)).
```

In the expression `e`, `x: list(A)`. But this is a non-polymorphic constant (both term and type arity 0) with a variable in its result sort, which `SortHelper` doesn't like. It seems to be eliminated just fine in later FOOLing around.

### DIFF
--- a/Kernel/SortHelper.cpp
+++ b/Kernel/SortHelper.cpp
@@ -92,7 +92,27 @@ TermList SortHelper::getResultSort(const Term* t)
   getTypeSub(t, subst);
   Signature::Symbol* sym = env.signature->getFunction(t->functor());
   TermList result = sym->fnType()->result();
-  ASS(!subst.isEmpty()  || (result.isTerm() && (result.term()->isSuper() || result.term()->ground())));  
+
+  /*
+  either
+  1. the substitution is non-empty
+  2. the result is ground (or $tType)
+  3. t is let-bound: consider the following TFF1
+
+      % polymorphic constant
+      tff(c_type, type, c: !>[A: $tType]: A).
+      % some polymorphic predicate, not important
+      tff(p_type, type, p: !>[A: $tType]: A > $o).
+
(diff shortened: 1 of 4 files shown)


Kernel/SortHelper.cpp
```diff
@@ -88,10 +88,19 @@ TermList SortHelper::getResultSort(const Term* t)
...
 
-  Substitution subst;
-  getTypeSub(t, subst);
   Signature::Symbol* sym = env.signature->getFunction(t->functor());
   TermList result = sym->fnType()->result();
+
+  // fast path for constants
+  // also avoids assertion failures below with $let-bound constants,
+  // which have arity 0 but their sort may contain sort variables (!)
+  if(!t->arity()) {
```
mezpusz, line 97: According to [this](https://ceur-ws.org/Vol-2162/paper-07.pdf) paper, let binders can be non-constant as well. Not saying I have a better idea for now, just that in general we should expect this in other cases as well.

MichaelRawson: Good catch. By some miracle we actually seem to parse and process these beasts correctly (which then triggers the old assertion again). They could even have type arguments. :-O

Proposal: flag let-bound symbols and exempt them from such assertions. This was my original idea but it seemed too heavy-handed for just constants.

MichaelRawson:
A completely different solution, should be ready for re-review. Thanks @mezpusz !

mezpusz (approved):
Looks fine to me, thanks for taking care of this!

quickbeam123 (approved):

PR #547, opened 2024-04-19
Ordering check with applied terms + optimised unidirectional comparison for LPO and KBO

mezpusz:
This PR refactors functionality of KBO and LPO such that two terms can be compared modulo some substitution without creating the result of the application. Some notes:

- The overhead for regular terms on which no substitution has to be applied should be minimal.
- The new functionality is only used in the main check for demodulation, but it would be possible to use it elsewhere (e.g. in superposition, binary resolution, demodulation helper, subsumption demodulation, etc.)
- `KBOForEPR` has been removed as the use case in my opinion is not that prevalent and the code does not give a major improvement over `KBO`.
- `SKIKBO` has not been adapted, but I did not remove it due to some functionaility being used elsewhere.

Additionally, a unidirectional check has been implemented for KBO and exposed for LPO. This is far from super optimal, but it's a bit more efficient than the bidirectional check when we don't need the latter.

I did some optimisations here and there in other parts too, see comments below.


Inferences/ForwardDemodulation.cpp
```diff
@@ -116,45 +131,26 @@ bool ForwardDemodulationImpl<combinatorySupSupport>::perform(Clause* cl, Clause*
...
+        auto lhs = qr.data->term;
+
         // TODO:
         // to deal with polymorphic matching
         // Ideally, we would like to extend the substitution
         // returned by the index to carry out the sort match.
         // However, ForwardDemodulation uses a CodeTree as its
         // indexing mechanism, and it is not clear how to extend
         // the substitution returned by a code tree.
-        Recycled<RobSubstitution> subst;
```
mezpusz, line 126: Turns out this uses a lot (up to 5%) of the runtime. Does anyone have any idea what causes it? Maybe called too many times?

quickbeam123:
Will look forward to doing some performance checks with this one!

Kernel/SubstHelper.hpp
```diff
@@ -26,6 +26,41 @@ namespace Kernel {
...
+      aboveVar(aboveVar && t.isVar() ? false : aboveVar), applicator(applicator) {}
+
+  AppliedTerm(TermList t, const AppliedTerm& parent)
+    : term(parent.aboveVar && t.isVar() ? (*parent.applicator)(t.var()) : t),
+      aboveVar(parent.aboveVar && t.isVar() ? false : parent.aboveVar), applicator(parent.applicator) {}
+
+  /**
+   * Only allow comparisons if we can guarantee that the terms are the same.
+   */
+  bool operator==(const AppliedTerm& other) const {
```
mezpusz, line 59: Call this something else.

Inferences/BackwardDemodulation.cpp
```diff
@@ -119,41 +126,27 @@ struct BackwardDemodulation::ResultFn
...
     TermList lhs=arg.first;
+    TermList rhs=EqHelper::getOtherEqualitySide(_eqLit, lhs);
 
     // AYB there used to be a check here to ensure that the sorts
     // matched. This is no longer necessary, as sort matching / unification
     // is handled directly within the tree
 
-    TermList rhs=EqHelper::getOtherEqualitySide(_eqLit, lhs);
+    auto subs = qr.unifier;
+    ASS(subs->isIdentityOnResultWhenQueryBound());
```
MichaelRawson, line 136: So this is always true now? Was it always so?

Inferences/ForwardDemodulation.cpp
MichaelRawson: `Recycled<T>` has to do some reset logic and lookup some recycled object, whereas a static will just do some initialisation logic (at least with `-fno-threadsafe-statics`, which we have). So `Recycled<T>` will be slower, but may re-use more objects.

Inferences/ForwardDemodulation.cpp
```diff
@@ -53,6 +53,21 @@ using namespace Kernel;
...
+    auto res = subst->applyToBoundResult(v);
+    if constexpr (applyVSubst) {
+      return vSubst.apply(res, 0);
+    } else {
+      return res;
+    }
+  }
+  ResultSubstitution* subst;
+  const RobSubstitution& vSubst;
+};
```
MichaelRawson, line 69: I think we could save some allocated bytes in the `!applyVSubst` case by not having the `vSubst` field at all. Lmk if you want help with the template stuff.

Inferences/ForwardDemodulation.hpp
```diff
@@ -36,6 +36,8 @@ class ForwardDemodulation
   void detach() override;
   bool perform(Clause* cl, Clause*& replacement, ClauseIterator& premises) override = 0;
 protected:
+  template<bool> struct Applicator;
+
```
MichaelRawson, line 40: Why is this necessary? I think the `Applicator` is only used inside `ForwardDemodulation.cpp`.

Inferences/ForwardDemodulation.cpp
```diff
@@ -167,9 +163,26 @@ bool ForwardDemodulationImpl<combinatorySupSupport>::perform(Clause* cl, Clause*
           }
         }
   #endif
-        if(!preordered && (_preorderedOnly || ordering.compare(trm,rhsS)!=Ordering::GREATER) ) {
+
+        auto subs = qr.unifier;
+        ASS(subs->isIdentityOnQueryWhenResultBound());
+
+        ScopedPtr<SubstApplicator> appl;
```
MichaelRawson, line 170: Does this need to be a heap allocation at all, actually? I could imagine e.g.

```c++
Applicator<true> varSubst;
Applicator<false> notVarSubst;
SubstApplicator *appl = lhs.isVar() ? &varSubst : &notVarSubst;
```

(in which case, disregard previous comment about saving bytes)

Kernel/KBO.hpp
```diff
@@ -155,13 +155,20 @@ class KBO
 
   using PrecedenceOrdering::compare;
   Result compare(TermList tl1, TermList tl2) const override;
+
+  Result compare(AppliedTerm&& t1, AppliedTerm&& t2) const override;
+  bool isGreater(AppliedTerm&& t1, AppliedTerm&& t2) const override;
```
MichaelRawson, line 160: Reminder to go with pass-by-value.

Kernel/LPO.cpp
```diff
@@ -79,35 +79,44 @@ Ordering::Result LPO::comparePrecedences(Term* t1, Term* t2) const
 } // LPO::comparePrecedences
 
 Ordering::Result LPO::compare(TermList tl1, TermList tl2) const
+{
+  return compare(AppliedTerm(tl1),AppliedTerm(tl2));
```
MichaelRawson, line 83: This might not be completely zero-cost as you'll have to do some null checks in the `AppliedTerm` version. But it's probably fine - if it shows up as a hit in profiling we can probably do some template magic to smooth it out.

Kernel/LPO.hpp
```diff
@@ -43,21 +45,24 @@ class LPO
   ~LPO() override = default;
 
   using PrecedenceOrdering::compare;
-  [[nodiscard]] Result compare(TermList tl1, TermList tl2) const override;
+  Result compare(TermList tl1, TermList tl2) const override;
```
MichaelRawson, line 48: Could these be `[[nodiscard]]` again?

Kernel/Ordering.hpp
```diff
@@ -103,6 +108,8 @@ class Ordering
 
   Result compareEqualities(Literal* eq1, Literal* eq2) const;
 
+  static bool containsVar(const AppliedTerm& s, TermList var);
```
MichaelRawson, line 111: `unsigned var` so you can't pass a `Term *`?

Kernel/SubstHelper.hpp
```diff
@@ -26,6 +26,41 @@ namespace Kernel {
...
+};
+
+/**
+ * Term with a substitution applied to it lazily.
+ */
+struct AppliedTerm
+{
+  TermList term;
+  bool aboveVar;
+  const SubstApplicator* applicator;
```
MichaelRawson, line 41: Could we eliminate `aboveVar` and check whether `applicator == nullptr` instead?

Kernel/Term.hpp
```diff
@@ -905,6 +915,8 @@ class alignas(8) Term
   unsigned _isTwoVarEquality : 1;
   /** Weight of the symbol */
   unsigned _weight;
+  /** Cached weight of the term for KBO, otherwise -1 and invalid. */
+  int _kboWeight;
```
MichaelRawson, line 919: Could a comment be made on the difference between `weight` and `kboWeight`? I always find this confusing.

Lib/DHMap.hpp
```diff
@@ -363,7 +363,6 @@ class DHMap
    */
   bool getValuePtr(Key key, Val*& pval)
   {
-    ensureExpanded();
```
MichaelRawson, line 366: Is this intended?

Kernel/KBO.cpp
```diff
@@ -93,47 +102,36 @@ class KBO::State
  * traverse(Term*,int) for both terms/literals in case their
  * top functors are different.)
  */
-Ordering::Result KBO::State::result(Term* t1, Term* t2)
+Ordering::Result KBO::State::result(const AppliedTerm& t1, const AppliedTerm& t2)
 {
```
MichaelRawson, line 106: This seems like a case where it might be worth having `Applied<Term *>` versus `Applied<TermList>`.

Kernel/KBO.cpp
```diff
@@ -93,47 +102,36 @@ class KBO::State
...
-  //KBO::compare methods.
-  //- or literals t1 and t2 are equal but for their polarity. But such
-  //literals should never occur in a clause that would exist long enough
-  //to get to ordering --- it should be deleted by tautology deletion.
-  ASS_NEQ(res, EQUAL);
   return res;
 }
 
-Ordering::Result KBO::State::innerResult(TermList tl1, TermList tl2)
+Ordering::Result KBO::State::innerResult(const AppliedTerm& tl1, const AppliedTerm& tl2)
```
MichaelRawson, line 132: Could these be downgraded to just `TermList`? This doesn't seem to use `applicator` at all.

MichaelRawson (approved):
Nice! Subject to some questions I'm happy.

Inferences/BackwardDemodulation.cpp
mezpusz: Currently we only use `CodeTreeTIS` in `ForwardDemodulation` (resp. `SubstitutionTree` in `BackwardDemodulation`) for which this (respectively `isIdentityOnQueryWhenResultBound()`) holds. And `ResultSubstitution::isRenaming` also assumes this, so this makes the code more consistent IMO. We can put them back if the indices change (which I don't see happening soon).

Inferences/ForwardDemodulation.cpp
mezpusz: Good idea! I'll try to make it work..

Kernel/LPO.cpp
mezpusz: Added a separate constructor for this case, hopefully it solves it.

Kernel/LPO.hpp
mezpusz: I'm a bit confused because I get `-Wunused-result` warnings even without `[[nodiscard]]`. Do we need this across all compilers?

Kernel/SubstHelper.hpp
mezpusz: Sadly, no as `applicator==nullptr` implies `aboveVar==false` but the converse does not hold.

Kernel/Ordering.hpp
mezpusz: This would mean having three `TermList` ctors inside the function, but sure. I just thought it's easier if we add an assertion.

Kernel/KBO.cpp
MichaelRawson: Your call - just an observation.

Kernel/LPO.hpp
MichaelRawson: Ahah - `Result` itself is now `[[nodiscard]]`, so these are useless. Thanks!

Kernel/Ordering.hpp
MichaelRawson: Up to you.

MichaelRawson:
Thanks for changes! Fine to merge for me, imagine @quickbeam123 would like to check some perf.

Inferences/ForwardDemodulation.cpp
mezpusz: Okay, so can we have this for now? Or should I revert?

MichaelRawson: Yes, of course - I was just explaining the rationale, I think your change is pragmatic.

Inferences/ForwardDemodulation.cpp
MichaelRawson: This is no longer heap-allocated thanks to your changes, so we don't need to worry. Thanks for trying!

MichaelRawson: For reference, I'd probably have done

```c++
struct Bar;
struct Baz;

template<bool b>
struct Foo;

template<>
struct Foo<false> {
	Foo(Bar *bar) : bar(bar) {}
	Bar *bar;
};

template<>
struct Foo<true> {
	Foo(Baz &baz) : baz(baz) {}
	Baz &baz;
};
```

...but I don't say this is any better - and now it doesn't matter anyway!

Kernel/SubstHelper.hpp
```diff
@@ -26,6 +26,45 @@ namespace Kernel {
...
+  virtual TermList operator()(unsigned v) const = 0;
+};
+
+/**
+ * Term with a substitution applied to it lazily.
+ */
+struct AppliedTerm
+{
+  TermList term;
+  bool aboveVar;
```
joe-hauns, line 40: Could you maybe document the intuition of `aboveVar`? Maybe an example would be helpful.

Kernel/SubstHelper.hpp
```diff
@@ -26,6 +26,45 @@ namespace Kernel {
...
+
+  AppliedTerm(TermList t, AppliedTerm parent)
+    : term(parent.aboveVar && t.isVar() ? (*parent.applicator)(t.var()) : t),
+      aboveVar(parent.aboveVar && t.isVar() ? false : parent.aboveVar), applicator(parent.applicator) {}
+
+  /**
+   * Only allow comparisons if we can guarantee that the terms are the same.
+   */
+  bool operator==(AppliedTerm) const = delete;
+  bool operator!=(AppliedTerm) const = delete;
```
joe-hauns, line 61: Out of interest: What is this needed for?
I guess something with subtyping ...? I don't really use subtyping so I'm not sure what it's supposed to prevent.

Kernel/Term.hpp
```diff
@@ -903,8 +913,11 @@ class alignas(8) Term
   unsigned _hasInterpretedConstants : 1;
   /** If true, the object is an equality literal between two variables */
   unsigned _isTwoVarEquality : 1;
-  /** Weight of the symbol */
+  /** Weight of the symbol, i.e. sum of symbol and variable occurrences. */
   unsigned _weight;
+  /** Cached weight of the term for KBO, otherwise -1 and invalid. Note that
+   * KBO symbol weights are not necessarily 1, so this can differ from @b _weight. */
+  int _kboWeight;
```
joe-hauns, line 920: So `_kboWeight` is pre-computed on a per-term version. But it's actually not only dependent on the term but also on the KBO instance that computes the weight (i.e. it depends on which weight function that KBO uses). Thus having one value per term implicitly assumes that we always only have exactly one KBO instances, which is an assumption that could be broken in the future. I'm thinking for example multiple vampire instances running on one machine using the same memory for terms & so but using different weight functions, etc. Sooo it would maybe be good to add an assertion here that makes sure that we are only using one KBO instance to set the weight and only set the weight once.

I'm thinking adding a field `void* _kboInstance` that is initialized with nullpoint and needs to be set with `void setKboWeight(int weight, void* kbo)` (using `void*` so noone has the idea of calling that kbo instance), and when we call `setKboWeight` we add `ASS(_kboInstance == nullptr)`, and we also change `getKboWeight()` to `getKboWeight(void* kbo)` which checks `_kboInsatnce == kbo` before returning the weight. What are you thinking about that?
(obviously hiding all that safety code behing `#if VDEBUG`)

joe-hauns: Another case where we could have multiple KBO instances is unit testing ofc.

UnitTests/tKBO.cpp
```diff
@@ -376,3 +379,210 @@ TEST_FUN(kbo_test23) {
   ASS_EQ(ord.compare(f(alpha), g(beta)), Ordering::Result::INCOMPARABLE)
 }
 
+// isGreater tests
+
+bool isGreaterSymmetric(const KBO& ord, TermList t1, TermList t2) {
+  return ord.isGreater(AppliedTerm(t1),AppliedTerm(t2))
+    && !ord.isGreater(AppliedTerm(t2),AppliedTerm(t1));
```
joe-hauns, line 386: Maybe add 
```
&& ord.compare(t1, t2) == GREATER
&& ord.compare(t2, t1) == LESS
```

UnitTests/tKBO.cpp
```diff
@@ -376,3 +379,210 @@ TEST_FUN(kbo_test23) {
...
+TEST_FUN(kbo_isGreater_test04) {
+  DECL_DEFAULT_VARS
+  DECL_SORT(srt)
+  DECL_FUNC (f, {srt}, srt)
+  DECL_FUNC (g, {srt}, srt)
+
+  auto ord = kbo(weights(make_pair(f, 10u)), weights());
+
+  ASS(!ord.isGreater(AppliedTerm(f(x)), AppliedTerm(g(g(g(g(g(y))))))));
+  ASS(!ord.isGreater(AppliedTerm(g(g(g(g(g(y)))))), AppliedTerm(f(x))));
```
joe-hauns, line 434: Also maybe add 
```
ord.compare(...) == INCOMPARABLE
```

Kernel/Ordering.cpp
```diff
@@ -289,6 +272,35 @@ Ordering::Result PrecedenceOrdering::compare(Literal* l1, Literal* l2) const
...
+
+Ordering::Result Ordering::compare(AppliedTerm lhs, AppliedTerm rhs) const
+{
+  NOT_IMPLEMENTED;
+}
+
+bool Ordering::isGreater(AppliedTerm lhs, AppliedTerm rhs) const
+{
+  NOT_IMPLEMENTED;
+}
```
joe-hauns, line 302: I don't like the pattern of adding virtual functions and then adding throwing errors when functions are not overridden, as this means that people who use classes can't rely on the typesystem and subtyping but need to know which instance of an abstract class is being used where and in which ways this instance is being used. 

I'd suggest instead having 
```
class AppliedTermOrdering : public Ordering {
  Result compare(AppliedTerm lhs, AppliedTerm rhs) const = 0;
  ...
};
```

And implement this interface for `KBO`, but not for `LPO`.
What do you think? @MichaelRawson @mezpusz

Kernel/Ordering.cpp
```diff
@@ -289,6 +272,35 @@ Ordering::Result PrecedenceOrdering::compare(Literal* l1, Literal* l2) const
...
+  }
+  VariableIterator vit(s.term.term());
+  while (vit.hasNext()) {
+    auto v = vit.next();
+    if ((*s.applicator)(v.var()).containsSubterm(var)) {
+      return true;
+    }
+  }
+  return false;
+}
```
joe-hauns, line 292: Is there a reason why this is an instance function of `Ordering` and not of `AppliedTerm`? Feels more intuitive to me, but maybe I'm missing something.

Kernel/Term.hpp
quickbeam123: Good point! 

But note that we already cache orientations of equalities within the literals and I think there, there is some code that looks into env and only if the ordering we are computing the orientation with is the same as some main one, are we allowed to use / fill the cache. Couldn't that be a solution here as well?

joe-hauns: Good shout! That's the very same thing and I'd also thought about using `env.ordering`
I'd still opt for having an additional `Ordering*` parameter as that insures two things: People will always need to have access to the ordering wrt when they call the respective function (thus think about that things are ordering-dependent), and also it makes things more future-proof: If we decide we do want the option to have multiple orderings supported at once, we could have a `static DHMap<...> Term::_kboWeights` behind `setKboWeight` and so. 
But I don't think this is strictly necessary, it's just how i'd do it. I'm fine with either as long as there is some assertion that catches the potential future bugs :)

Kernel/Ordering.cpp
MichaelRawson: Fine by me - I also don't like `NOT_IMPLEMENTED` "abstract" methods, but it seems to be prevalent so I was willing to let this slide. Since LPO will probably have this in the future if it doesn't already (I got a bit confused with what @mezpusz does here and what he plans to do) we could also have this be an abstract method in `Ordering`, and have LPO `NOT_IMPLEMENT` it for now.

Inferences/ForwardDemodulation.cpp
```diff
@@ -167,9 +163,23 @@ bool ForwardDemodulationImpl<combinatorySupSupport>::perform(Clause* cl, Clause*
           }
         }
   #endif
-        if(!preordered && (_preorderedOnly || ordering.compare(trm,rhsS)!=Ordering::GREATER) ) {
+
+        auto subs = qr.unifier;
+        ASS(subs->isIdentityOnQueryWhenResultBound());
+
+        Applicator<true> varSubst(subs.ptr(), subst);
+        Applicator<false> notVarSubst(subs.ptr(), subst);
```
joe-hauns, line 171: This is quite a general remark I'd love to see in the general vampire coding style: Could we add comments for boolean arguments when their semantics are not clear from the context?
In this case for example have 
```
        Applicator</* applyVsubst */ true> varSubst(subs.ptr(), subst);
        Applicator</* applyVsubst */ false> notVarSubst(subs.ptr(), subst);
```
Which makes the code more readable in many cases.

Inferences/ForwardDemodulation.cpp
joe-hauns: I don't quite know whether we really need templates in this case. The types don't have a lot of common behaviour depending on the boolean template argument. I'd just use 
```
struct Applicator1 : SubstApplicator {
  Applicator1(ResultSubstitution* subst) : subst(subst) {}
  TermList operator()(unsigned v) const override {
    return subst->applyToBoundResult(v);
  }
  ResultSubstitution* subst;
};

struct Applicator2 : SubstApplicator {
  Applicator2(ResultSubstitution* subst, const RobSubstitution& vSubst) : subst(subst), vSubst(vSubst) {}
  TermList operator()(unsigned v) const override {
    return vSubst.apply(subst->applyToBoundResult(v), 0);
  }
  ResultSubstitution* subst;
  const RobSubstitution& vSubst;
};
```

I think this would make clearer what's going on. 
(Obviously we'd need some better names than `Applicator1` and `Applicator2` but I don't quite know what they're being used for hence no better names from my side)

Inferences/ForwardDemodulation.cpp
```diff
@@ -167,9 +163,23 @@ bool ForwardDemodulationImpl<combinatorySupSupport>::perform(Clause* cl, Clause*
           }
         }
   #endif
-        if(!preordered && (_preorderedOnly || ordering.compare(trm,rhsS)!=Ordering::GREATER) ) {
+
+        auto subs = qr.unifier;
+        ASS(subs->isIdentityOnQueryWhenResultBound());
+
+        Applicator<true> varSubst(subs.ptr(), subst);
```
joe-hauns, line 170: Could we maybe give `subs` and `subst` different names indicating their use cases better? 
Like `eqSortSubs` and `unifier` or something similar?

Inferences/ForwardDemodulation.cpp
```diff
@@ -167,9 +163,23 @@ bool ForwardDemodulationImpl<combinatorySupSupport>::perform(Clause* cl, Clause*
...
+        Applicator<false> notVarSubst(subs.ptr(), subst);
+        auto appl = lhs.isVar() ? (SubstApplicator*)&varSubst : (SubstApplicator*)&notVarSubst;
+
+        if (!preordered && (_preorderedOnly || !ordering.isGreater(AppliedTerm(trm),AppliedTerm(rhs,appl,true)))) {
+          // if (ordering.compare(AppliedTerm(trm),AppliedTerm(rhs,&appl,true))==Ordering::GREATER) {
+          //   USER_ERROR("is greater " + trm.toString() + " " + subs->applyToBoundResult(rhs).toString());
```
joe-hauns, line 176: Could you get rid of the commented out code, or do you really wanna keep it for the future?

Inferences/ForwardDemodulation.cpp
MichaelRawson: An excellent point - I thought they were used in some templated function, but they're not.

Kernel/KBO.cpp
```diff
@@ -182,106 +181,206 @@ void KBO::State::recordVariable(unsigned var, int coef)
...
+      _weightDiff += _kbo.symbolWeight(t.term.term()) * coef;
+      recState.push(State{ t, 0 });
     }
-    ts=stack.pop();
   }
 }
 
-void KBO::State::traverse(Term* t1, Term* t2)
+template<bool unidirectional>
+Ordering::Result KBO::State::traverseLex(AppliedTerm tl1, AppliedTerm tl2)
```
joe-hauns, line 235: I don't quite like the look of this function. In every part of the function it branches on `unidirectional` and there is huge chunks of code in every conditional, and the behaviour shared between the two cases `undirectional = true` and `unidirectional = false` do not have a lot in common. This makes the function even longer than it was before (and in my view it was already too long before this refactor), and thus even harder to understand what was going on. I'd suggest to have two different functions 
`traverseLexBidir` and `traverseLexUnidir` or so. 
If you really want to make use of code sharing between the two cases I'd suggest factoring out functions with appropriate names that describe what's going on in the case splits.

Kernel/KBO.cpp
```diff
@@ -57,11 +59,18 @@ class KBO::State
...
 
-  void traverse(Term* t1, Term* t2);
-  void traverse(TermList tl,int coefficient);
-  Result result(Term* t1, Term* t2);
-private:
-  void recordVariable(unsigned var, int coef);
+  template<bool unidirectional>
+  Result traverseLex(AppliedTerm t1, AppliedTerm t2);
+  template<bool unidirectional>
+  Result traverseNonLex(AppliedTerm t1, AppliedTerm t2);
```
joe-hauns, line 65: Maybe add a comment like: `traverseLex` is the "normal" traversing as in the paper <title>, and `traverseNonLex` is an optimized version of `traverseLex` that gives the same results as `traverseLex` but does not compute the result `_lexResult` and can be used in cases where we only care about the weight difference and the variable result, but not the lex result. 
(I don't think what I wrote here is quite accurate but a correct comment in the same spirit would be sweet )

joe-hauns (commented):
I went through all the files and added some coments. There's quite some simplifications/refactors in this PR I like and there's a couple of things I think that could be improved (comments above). 
I'm quite curious about the performance improvements! Looks like you put quite a lot of thought into technical details how to optimize the ordering checks :)

Kernel/Ordering.cpp
joe-hauns: Actually I just saw that LPO has a (sort-of) default-implementation. So I think the function could just be made purely virtual and we could leave away the `NOT_IMPLEMENTED`.

Kernel/SubstHelper.hpp
mezpusz: This is done so that the default generated operators are not used accidentally, thinking that they do a proper (and expensive) deep equals check when they in fact don't. We use the more suggestively named operator below.

Kernel/SubstHelper.hpp
mezpusz: I'll add an example.

UnitTests/tKBO.cpp
mezpusz: This is done deliberately to check `isGreater` and that in case of true the opposite relation does not hold.

UnitTests/tKBO.cpp
mezpusz: Same here, we're checking `isGreater`.

Kernel/Ordering.cpp
mezpusz: Good point, I added this here because `KBO` and `LPO` both use it but now that you mention it that would be more intuitive.

Inferences/ForwardDemodulation.cpp
mezpusz: Thanks, I left it there accidentally.

Kernel/Ordering.cpp
MichaelRawson: Pure-virtual in `Ordering`, `NOT_IMPLEMENTED` in the to-be-axed SKIKBO?

Inferences/ForwardDemodulation.cpp
mezpusz: I added untemplated classes, see new commit.

Inferences/ForwardDemodulation.cpp
mezpusz: Good suggestion! I added two specialized functions, see new commit.

Kernel/SubstHelper.hpp
joe-hauns: Ah right i see! Yeah i like having the explicit shallow equals operation!

mezpusz:
@joe-hauns Thanks for the review! I tried to address your comments, but let me know if there is something I misunderstood, or is still not satisfactory.

Inferences/ForwardDemodulation.cpp
joe-hauns: Did you overlook this one? I can't see any diff here.

joe-hauns (approved):
Looks all great to me! Thanks for the revision. from my side all good to go :)

quickbeam123:
Is this stable/ready for one quick round of my checks?

PR #549, opened 2024-05-03
Avoid demodulator renormalization

mezpusz:
Another small optimization for forward demodulation.

Code trees index terms normalized to increase the number of shared instructions. Hence, the substitution we get from matching is for the normalized term and not the original indexed term, hence we need to renormalize to get a substitution for the original term.

If done during each query, this renormalization can consume up to a third (!) of the runtime in cases when demodulation fails a lot (mainly due to the later ordering checks).

Instead, we can normalize the term to be indexed before indexing it and tell the code tree that no renormalization is needed. I added a new leaf data `DemodulatorData` where we can cache all kinds of relevant bits for a particular orientation of a demodulator; and I left the code for the original leaf data intact so this shouldn't cause problems for application where code trees are used for different purposes.

There are a couple of discussion points:
- Do we use type traits anywhere? I added a type trait for normalized indexed data, but not sure it fits what we have elsewhere.
- Where should the normalization happen? It is now done by the caller and checked in the `DemodulatorData` ctor, so that no accidental "overnormalization" happens by an ignorant caller.
- Should we check preorderedness/normalizedness/other demodulator properties? I placed some in `DemodulatorData` ctor, which should be cheap or we could skip them altogether.


Indexing/TermIndex.cpp
```diff
@@ -123,8 +123,20 @@ void DemodulationLHSIndex::handleClause(Clause* c, bool adding)
   Literal* lit=(*c)[0];
   auto lhsi = EqHelper::getDemodulationLHSIterator(lit,
                     _opt.forwardDemodulation()== Options::Demodulation::PREORDERED, _ord);
+  bool preordered = EqHelper::isPreorderedForDemodulation(lit, _ord);
```
quickbeam123, line 126: `EqHelper::getDemodulationLHSIterator` already checks the ordering. It feels a bit wasteful to immediately afterwards ask a question we have just known the answer to.

mezpusz: Yes, good point! How about returning a pair from `EqHelper::getDemodulationLHSIterator` with the iterator and preorderedness?

quickbeam123: Sounds good! And you could try assigning the pair into a std::tie to avoid the confusing first/second...

quickbeam123 (approved):

Indexing/CodeTreeInterfaces.cpp
```diff
@@ -114,66 +119,69 @@ class CodeTreeTIS::ResultIterator
...
+  Data* _found;
   bool _finished;
   CodeTreeTIS* _tree;
-  Recycled<TermCodeTree::TermMatcher> _matcher;
+  Recycled<typename TermCodeTree<Data>::TermMatcher> _matcher;
 };
 
-void CodeTreeTIS::_insert(TypedTermList t, Literal* lit, Clause* cls)
+template<class Data>
+void CodeTreeTIS<Data>::handle(Data data, bool insert)
```
MichaelRawson, line 157: Consider lifting this to the header file for tiny perf wins, I expect `insert` is often a constant at the call site.

Indexing/IndexManager.hpp
```diff
@@ -47,7 +47,6 @@ enum IndexType {
 
   DEMODULATION_SUBTERM_SUBST_TREE,
   DEMODULATION_LHS_CODE_TREE,
-  DEMODULATION_LHS_SUBST_TREE,
```
MichaelRawson, line 50: No more `DEMODULATION_LHS_SUBST_TREE`? I'm happy if so, just wondering why it died.

Indexing/TermCodeTree.cpp
```diff
@@ -64,39 +66,40 @@ void TermCodeTree::insert(TermInfo* ti)
...
+    dptr=rtm.op->template getSuccessResult<Data>();
+    if (*dptr==data) {
       break;
     }
   }
   
   rtm.op->makeFail();
   
-  delete rti;
+  delete dptr;
```
MichaelRawson, line 102: Could we initialise `dptr` to `nullptr`? This would prevent us deleting garbage if we fail to find something and allow us to `ASS(dptr)` here.

MichaelRawson (approved):
This is really nice and makes the substitution/code tree interface more uniform, love it.

Indexing/CodeTreeInterfaces.cpp
mezpusz: Forgive my ignorance, but what exactly happens when I lift this to the header? Can it be inlined?

Indexing/IndexManager.hpp
mezpusz: Since for this particular case we assert `ResultSubstitution::isIdentityOnQueryWhenResultBound()`, so implicitly asserting the use of `DEMODULATION_LHS_CODE_TREE`, I figured it's better to remove it. Otherwise, something like this can exist with substitution trees as well if needed, just not explicitly for demodulation.

Indexing/CodeTreeInterfaces.cpp
MichaelRawson: Exactly, other translation units have access to the definition so they can choose to inline it if they like.

PR #552, opened 2024-05-08
Runtime specialized unidirectional ordering checks for KBO and LPO

mezpusz:
Implements runtime specialized KBO and LPO mainly to be used by forward demodulation, which is still very hot (in terms of runtime) when we try to demodulate with incomparable equations. A similar idea was published in [this paper](https://link.springer.com/chapter/10.1007/978-3-540-25984-8_3) for KBO. As far as I know this is the first runtime specialized LPO.

The implementation is documented in the code, so I won't go into details here. The idea is that for a demodulator `s = t`, we precompile the comparison `s > t` so that when performing a comparison `s\theta > t\theta` under a substitution `\theta`, only variable-variable, variable-term or term-variable comparisons need to be performed (plus some simplified weight/variable checks for KBO).

I ran a cross-check using the original comparisons against the new ones with the entire TPTP using Otter and Discount, 20 seconds each, and there wasn't any difference between the two, so we get a fairly high confidence that the runtime specialized versions behave the same as the regular ones.

Note that this implementation is not final, just an initial one for CASC, I would like to iron out potential bottlenecks later. The results are promising, mostly improving for both KBO and LPO, although the latter is more dramatic (as the original LPO implementation is quadratic and we get a bigger boost by shaving off precomputable subcomparisons). But to avoid potential problems in the general option space, let's keep this feature behind a flag for now.

P.S. We could squash the branch commits as they contain some of the recent PRs' content as well.


Kernel/LPOComparator.cpp
```diff
@@ -0,0 +1,491 @@
...
+  static DHMap<std::pair<TermList,TermList>,pair<Stack<Instruction>,Instruction::BranchTag>*> _cache;
+
+  pair<Stack<Instruction>,BranchTag>** ptr;
+  // We have a local cache for subresults to avoid too much computation.
+  if (!_cache.getValuePtr(make_pair(tl1,tl2),ptr)) {
+    return *ptr;
+  }
+  // Allocate on heap so that cache reallocation
+  // won't affect partial results.
+  auto res = new pair(Stack<Instruction>(), BranchTag::T_JUMP);
```
MichaelRawson, line 282: This is probably quite hot and creates a double-indirection which it might be possible to avoid - for now I think you are ironing out performance later.

Kernel/KBOComparator.hpp
```diff
@@ -0,0 +1,129 @@
...
+      ins._setSecondUint(v2);
+      return ins;
+    }
+    static Instruction uintInt(InstructionTag t, unsigned v1, int v2) {
+      Instruction ins;
+      ins._setTag(t);
+      ins._setFirstUint(v1);
+      ins._setCoeff(v2);
+      return ins;
+    }
```
MichaelRawson, line 96: What do these two functions do? Why are they different?

MichaelRawson (approved):
Does the job!

Shell/Options.cpp
```diff
@@ -1541,6 +1541,16 @@ void Options::init()
     _demodulationRedundancyCheck.onlyUsefulWith(Or(_forwardDemodulation.is(notEqual(Demodulation::OFF)),_backwardDemodulation.is(notEqual(Demodulation::OFF))));
     _demodulationRedundancyCheck.addProblemConstraint(hasEquality());
 
+    _demodulationPrecompiledComparison = BoolOptionValue("demodulation_precompiled_comparison","dpc",false);
+    _demodulationPrecompiledComparison.description=
+       "Precompiles ordering constraints on unorientable demodulators which results in less overhead when actually comparing."
+       "Only works with KBO currently.";
```
quickbeam123, line 1547: According to the comments of the PR, isn't this outdated? (I.e., works also for LPO?)

Kernel/Ordering.hpp
```diff
@@ -69,6 +81,10 @@ class Ordering
    * under some substitutions captured by @b AppliedTerm. */
   virtual bool isGreater(AppliedTerm t1, AppliedTerm t2) const = 0;
 
+  /** Optimised function used for checking that @b lhs is greater than @b rhs,
+   * under substitution represented by @b applicator. */
+  virtual bool isGreater(TermList lhs, TermList rhs, const SubstApplicator* applicator, OrderingComparatorUP& comparator) const = 0;
```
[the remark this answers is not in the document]
mezpusz: The implementation creates the object if needed and then calls a `const` member function. So if I'm not mistaken it would be exactly the same, except the interface would incorrectly claim that we don't modify the object. Or is it  more complicated than this?

Kernel/LPOComparator.hpp
```diff
@@ -0,0 +1,123 @@
...
+ * that has quadratic time complexity @see LPO::lpo.
+ */
+class LPOComparator
+: public OrderingComparator
+{
+public:
+  LPOComparator(const LPO& lpo) : _lpo(lpo), _instructions(), _res(Instruction::BranchTag::T_JUMP) {}
+
+  /** This function performs the runtime specialization and creates a comparator object. */
+  static LPOComparator* create(TermList tl1, TermList tl2, const LPO& lpo);
```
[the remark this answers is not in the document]
mezpusz: Right, but do you mean we should plainly do the preprocessing in the ctor? Not that it does anything illegal which shouldn't be done in a ctor. (Although now I realize that the entire `Induction` inference code is done in the ctor, so this should be familiar :-))

Kernel/LPOComparator.cpp
mezpusz: Yes, this is stupid. I think we could just move a temporary result into the cache at the very end.

Kernel/KBOComparator.hpp
mezpusz: The second one is used in the specific case when the second argument can be also negative. It could be removed if there is some way to cast a negative integer "bitwise" to an unsigned. Is there such a thing?

Shell/Options.cpp
mezpusz: Yes, thanks for catching it!

Kernel/LPOComparator.hpp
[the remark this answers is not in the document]
MichaelRawson: Right - if sensible + possible I think it would make sense.

I thought you would need move semantics for the `unique_ptr` construction later but `make_unique` or something might do the right thing, and maybe this is move-safe anyhow.

Kernel/Ordering.hpp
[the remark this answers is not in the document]
MichaelRawson: Ahah. Right, the *pointer* is mutable, but the object pointed-to is not (morally). Then either ignore this or consider `std::unique_ptr<const OrderingComparator>`.

Kernel/KBOComparator.hpp
MichaelRawson: ...probably, but that sounds like one of those "spot the undefined behaviour" exercises waiting to happen! OK - a comment would be good if you're making another commit, but in any case fine.

Kernel/Ordering.hpp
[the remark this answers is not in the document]
mezpusz: That's a good idea, it didn't occur to me first.

Kernel/KBOComparator.hpp
mezpusz: Looking at the recently added `setBits` and `getBits`, should we be worried about breaking on some compilers? According to [this](https://stackoverflow.com/questions/9656219/how-can-uint64-t-store-a-negative-number-isnt-it-meant-to-be-unsigned), if `uint64_t` does not fit the assigned value, the compiler copies bitwise, but this is not guaranteed. Anyways, if you feel confident that this does not break, we can remove the constructor as well.

Kernel/LPOComparator.cpp
mezpusz: Turns out the main problem is that handle multiple subresults as pointers, and these could change during reallocation. An alternative evil is to copy the subresults to temporary objects which seems even less efficient. What do you think?

MichaelRawson: In a future version of this I might have (i) a hashmap to indices and (ii) an array of these objects - this is still an indirection, of course, but it's a bit more spatial and one can't forget to delete things.

Or, it seems from a quick skim that the results are "finished" and their stacks of instructions won't change after `createHelper` returns. If this is true, we could return a `Instruction *` - length - `BranchTag` triple from `createHelper` and have the cache alone own the `const Stack<Instruction>`.

But for now this works!

quickbeam123:
OK, so this again helps by a measurable bit! (A quick debug mode check under way...)

And let's say I don't want to read the code of the new Comparator classes, being happy that the improvement is hidden behind an option. How much, do you think, will this extension complicate the life of future modificators of the ordering code? Or can we at least say, "If we won't want to maintain this, we can easily disconnect it again."?

quickbeam123 (approved):

MichaelRawson:
>How much, do you think, will this extension complicate the life of future modificators of the ordering code?

I think this would be OK, the optimised version can be replaced with a call to the slower version.

Kernel/KBOComparator.hpp
MichaelRawson: I think the implementation of those functions is fine (although maybe we should statically check that the type is reasonable), but casting `int` to `uint64_t` and back is probably dodgy. What behaviour do you intend with this? Will the coefficient ever be negative?

This could probably be avoided altogether by using a 32-bit bitfield and then a 32-bit regular union, but then we'd have to parameterise `BitUtils`...which is probably fine, I just didn't do it yet.

mezpusz: Yes, in fact we only have a weight check instruction if there is at least one negative value somewhere, otherwise the balance would be trivially OK.

How about a `reinterpret_cast` like [here](https://stackoverflow.com/questions/30048135/efficient-way-to-bit-copy-a-signed-integer-to-an-unsigned-integer)?

mezpusz:
> How much, do you think, will this extension complicate the life of future modificators of the ordering code? Or can we at least say, "If we won't want to maintain this, we can easily disconnect it again."?

Yes, as @MichaelRawson said, we can just fall back to calling `Ordering::compare` or `Ordering::isGreater` directly. My (probably biased) opinion is that the code is not that complicated, so maintaining it wouldn't be a problem either for someone who also maintains `KBO` and `LPO`. And as the recent bug indicates, there is the benefit that we have a different implementation to cross-check with!

Kernel/KBOComparator.hpp
MichaelRawson: Sounds sensible - I think at least this communicates the intent well, and there are definitely other sign-conversion maybe-bugs in Vampire. I think this is the last blocker, so please push and I'll merge. :-)

PR #554, opened 2024-05-16
remove `MemoryLeak` and `CHECK_LEAKS`, can now use e.g. Valgrind

MichaelRawson:
We had some homebrew leak-checking machinery, but I don't think it was used recently. After Allocator's recent diet, you can now use regular tools like Valgrind to find leaks, should anybody actually be doing that.


MichaelRawson:
Tag #157.

Indexing/TermSharing.cpp
```diff
@@ -45,28 +45,6 @@ TermSharing::TermSharing()
...
-  while (ls.hasNext()) {
-    ls.next()->destroy();
-  }
-  Set<AtomicSort*,TermSharing>::Iterator ss(_sorts);
-  while (ss.hasNext()) {
-    ss.next()->destroy();
-  }
-#endif
-}
-
```
quickbeam123, line 69: How about we keep CHECK_LEAKS and, e.g., this bit? Otherwise, the valgrinder will need to write it again?

MichaelRawson: Sure!

PR #558, opened 2024-05-16
parser fixes, thanks to Joe

MichaelRawson:
See #542. Fix some of the crashes reported there.
1. When creating a tuple sort, we should call `AtomicSort::create`, not `Term::create`. I think it was probably fine data-wise, but we fail an assertion.
2. We catch `USER_ERRORS` and wrap them as `PARSE_ERRORS` with a line number. I'm not sure how useful this is and it means we can't distinguish between "we parsed this but can't do anything with it" (e.g. HOL + theories), and "syntax error". Don't do that.
3. Some `PARSE_ERRORS` should be `USER_ERRORS`.

### DIFF
--- a/Kernel/Term.cpp
+++ b/Kernel/Term.cpp
@@ -1191,7 +1191,7 @@ TermList AtomicSort::arraySort(TermList indexSort, TermList innerSort)
 }
 
 TermList AtomicSort::tupleSort(unsigned arity, TermList* sorts)
-{ return TermList(Term::create(env.signature->getTupleConstructor(arity), arity, sorts)); }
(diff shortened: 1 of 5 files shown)


quickbeam123:
Could we somehow propagate the `lineNumber()` even with the `USER_ERROR`s?

quickbeam123:
The value is in being able to say "You are doing something stupid <AROUND HERE>." which is more informative than "You are doing something stupid in this 5 MB file of yours." 

Not saying the current way is the best solution, but could we have this PR without 2. and think about it a bit more?

MichaelRawson:
Done.

Issue #557, opened 2024-05-16
Bad saturation on `SYO500_8.008`


MichaelRawson:
Strategy can be minimised to `-drc encompass -foolp on -kws precedence -spb units` (Otter or DISCOUNT also work for reproducibility). Playing around with the options gets you a proof (!).  I guess that some combination of these options is subtly incomplete, but we don't know about it yet.

quickbeam123:
Funnily, finite model finding finds a proof of size 3 and then crashes. But that's probably a different bug:
```./vampire Problems/SYO/SYO500_8.008.p -sa fmb -foolp on 
% Running in auto input_syntax mode. Trying TPTP
Detected minimum model sizes of [1]
Detected maximum model sizes of [34]
TRYING [1]
TRYING [2]
TRYING [3]
Finite Model Found!
% SZS status CounterSatisfiable for SYO500_8.008
Condition n < _size in file ./Lib/DArray.hpp, line 105 was violated, as:
n == 1
_size == 0
```

quickbeam123:
"Works" equally well on the smaller `Problems/SYO/SYO500_8.002.p`

quickbeam123:
The problem is likely that
```
[SA] forward reduce: 3. ($true != $false) [fool axiom]
      replaced by 166. ($true != bG6) <- (~3) [forward demodulation 3,161]
     using 161. ($false = bG6) <- (~3) [trivial inequality removal 160]
```
showing that `$false` is not the smallest constant. 

`show_ordering` reveals that it is the smallest in the precedence
```
% function precedences, smallest symbols first (line format: `<name> <arity>`) 
% ===== begin of function precedences ===== 
% $$false 0
% $$true 0
% bG0 0
% bG1 0
...
```
but not regarding the symbol weights
```
% Weights of function (line format: `<name> <arity> <weight>`)
% ===== begin of function weights ===== 
% bG0 0 1
% bG1 0 2
% bG2 0 3
% bG3 0 4
% bG4 0 5
% bG5 0 6
% bG6 0 7
% bG7 0 8
% $$false 0 9
% $$true 0 10
...
```

Shall a KBO expert fix this? ;)

MichaelRawson:
>Shall a KBO expert fix this? ;)

Is it a bird? Is it a plane? It's @mezpusz with his mighty convenient KBO superpowers (and recent patch series).

quickbeam123:
[...]

PR #559, opened 2024-05-16
set kbo weights of fool constants to 1 no matter the weight scheme

quickbeam123:
Fixes https://github.com/vprover/vampire/issues/557

### DIFF
--- a/Kernel/KBO.cpp
+++ b/Kernel/KBO.cpp
@@ -1214,6 +1214,13 @@ bool KboSpecialWeights<PredSigTraits>::tryGetWeight(unsigned functor, unsigned& 
 
 bool KboSpecialWeights<FuncSigTraits>::tryGetWeight(unsigned functor, unsigned& weight) const
 {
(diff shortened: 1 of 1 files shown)


MichaelRawson (approved):
Seems sensible.

joe-hauns:
Why/in which rule do we need the property that `$false` is least? Can we just make saturation incomplete when the fool constant weight is not 1 and the respective options are enabled?

quickbeam123:
In FOOL paramodulation, I believe. What you suggest is also a possibility...

joe-hauns:
I see. I missed the crutial option `-foolp`... 
I think we still need to make the saturation incomplete if a user sets `_foolConstant` to something different to `1` using the weight setting file, right?

mezpusz (approved):

quickbeam123:
I tried not to follow the path for making _foolConstant settable from a file, but maybe it's possible somehow?

In any case, I think the file input is so rare that working on a robust way of figuring out completeness for user-specified weight scheme and precedence might not be worth the effort. Unless it's already there and a one-liner would fix it?

PR #566, opened 2024-05-31
Streamline question answering

quickbeam123:
Many small things to improve user experience with question answering. 

The new-old mode is newly invoked via `-qa plain` (previously, `-qa answer_literal`) and is mostly separate in functionality from `-qa synthesis` although code sharing, where available, was kept in place.

`-qa plain` is now compatible with AVATAR.

Disjunctive answers (with don't care variables, whose skope is clarified using a "∀") are supported.

If a skolem appears in the answer, the user is told where it came from.

Forall-exists questions are supported. Using a trick from synthesis, the forall block gets skolemized, but the skolems are recorded and replaced by the original variables back in the answer. `-qa plain` makes the user aware of these by starting the answer with a "λ" and binding these variables.

The original variable names from the question are kept from the parser and used to "pretty print" the answer.

Some non-trivial changes to properly track "derived from goal" (regardless of question answering), which are used to decide on "Theorem" vs "ContradictoryAxioms" and trigger some soundness checks in debug mode.

It is possible to specify what an answer should not look like, in order to get a different one. This can be used (through iterative vampire invocations) to get multiple answers. (Doing this in one run would be technically more involved and wouldn't be as reliable anyway.)

Formulas of Inference == INPUT have their unit input type shown as well (in parenthesis):
```
1. ? [X0] : p(X0) [input(axiom)]
3. ? [X0] : p(X0) [input(conjecture)]
```
So that the `negated_conjecture` inference which follows does not come out of the blue for the user.

Maybe there is something I forgot?


Kernel/Term.cpp
```diff
@@ -776,7 +774,7 @@ const vstring& Literal::predicateName() const
 
 
 bool Literal::isAnswerLiteral() const {
-  return env.signature->getPredicate(functor())->answerPredicate();
+  return isNegative() && env.signature->getPredicate(functor())->answerPredicate();
```
MichaelRawson, line 777: Are answer literals *always* negative? When @hzzv gets to it she might want to check this against her synthesis code.

Kernel/Unit.cpp
```diff
@@ -153,20 +155,17 @@ Formula* Unit::getFormula()
 vstring Unit::inferenceAsString() const
 {
 #if 1
```
MichaelRawson, line 157: Could this `#if 1` die now?

Kernel/Unit.hpp
```diff
@@ -84,17 +84,26 @@ class Unit
   /** see isTheoryAxiom in Inference.cpp */
   bool isTheoryAxiom() const { return _inference.isTheoryAxiom(); }
 
-  /** return true if there is an input node in the deriviation  */
-  bool derivedFromInput() const;
-  /** return true if there is a node in the derivation that is derivedFromGoal
-      this is a current workaround for the fact that GS doesn't always set the
-      correct inputType **/
-  bool derivedFromGoalCheck() const;
-
```
MichaelRawson, line 93: Does this functionality exist elsewhere now?

Lib/StringUtils.cpp
```diff
@@ -45,6 +45,17 @@ vstring StringUtils::replaceChar(vstring str, char src, char target)
   return vstring(buf.array(), len);
 }
 
+void StringUtils::replaceAll(vstring& where, const vstring& from, const vstring& to)
```
MichaelRawson, line 48: Might be worth warning about quadratic runtime in a comment.

Shell/UIHelper.cpp
```diff
@@ -490,12 +507,24 @@ void UIHelper::outputResult(ostream& out)
       ofstream latexOut(env.options->latexOutput().c_str());
 
       LaTeX formatter;
-      latexOut << formatter.refutationToString(env.statistics->refutation);
+      latexOut << formatter.refutationToString(refutation);
     }
 
-    ASS(!s_expecting_sat);
+    // the following two sanity checks are performed only after the proof printing, so we can also have a look at the proof, when we get a report back
```
MichaelRawson, line 513: Ahah. This is our `SZS dubious` opportunity!

Shell/UIHelper.hpp
```diff
@@ -42,8 +42,7 @@ class UIHelper {
     UnitList::FIFO _units;
     SMTLIBLogic _smtLibLogic;
     bool _hasConjecture;
-    bool _proofHasConjecture;
-    LoadedPiece() : _smtLibLogic(SMT_UNDEFINED), _hasConjecture(false), _proofHasConjecture(true) {}
+    LoadedPiece() : _smtLibLogic(SMT_UNDEFINED), _hasConjecture(false) {}
```
MichaelRawson, line 45: Could this just be C++11 assignments to the fields?

Shell/Options.hpp
```diff
@@ -490,10 +490,9 @@ class Options
   };
 
   enum class QuestionAnsweringMode : unsigned int {
-    ANSWER_LITERAL = 0,
-    FROM_PROOF = 1,
-    SYNTHESIS = 2,
-    OFF = 3
+    PLAIN = 0,
```
MichaelRawson, line 493: I think `FROM_PROOF` is no more, right? Did the unused `FROM_PROOF` code get removed already?

Inferences/InvalidAnswerLiteralRemovals.cpp
```diff
@@ -0,0 +1,63 @@
...
+    if (lit->isAnswerLiteral() && !lit->computableOrVar())
+      return nullptr;
+  }
+  return cl;
+}
+
+UndesiredAnswerLiteralRemoval::UndesiredAnswerLiteralRemoval(const vstring& avoidThese)
+{
+  // TODO: catch parsing exceptions and complain properly
+  _avoiders = Parse::TPTP::parseClauseFromString(avoidThese);
```
MichaelRawson, line 37: I have to ask...what's going on here, and why are we parsing clauses?

MichaelRawson (approved):
Generally very happy, this looks like a big improvement! Just some quick questions.

Kernel/InferenceStore.cpp
```diff
@@ -287,81 +235,21 @@ struct InferenceStore::ProofPrinter
...
+      UnitIterator parents= cs->getParents();
+
+      out << Int::toString(cs->number()) << ". ";
       FormulaUnit* fu=static_cast<FormulaUnit*>(cs);
       if (env.colorUsed && fu->inheritedColor() != COLOR_INVALID) {
         out << " IC" << fu->inheritedColor() << " ";
       }
       out << fu->formula()->toString() << ' ';
 
-      out <<"["<<Kernel::ruleName(rule);
```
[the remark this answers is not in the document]
quickbeam123: I don't think so. `Inference::name()` is implemented using `Kernel::ruleName`...

Kernel/InferenceStore.cpp
```diff
@@ -533,10 +421,11 @@ struct InferenceStore::TPTPProofPrinter
     switch(rule) {
     case InferenceRule::INPUT:
       if (origin==UnitInputType::CONJECTURE) {
-	return "conjecture";
-      }
-      else {
-	return "axiom";
+	      return "conjecture";
```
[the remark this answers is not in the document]
quickbeam123: I see it too, here at Github, but seems OK locally ...

Kernel/InferenceStore.cpp
```diff
@@ -287,81 +235,21 @@ struct InferenceStore::ProofPrinter
...
-      out << cl->toString() << vstring("\n");
-    }
-    else {
-      out << _is->getUnitIdStr(cs) << ". ";
+      out << cl->toString() << "\n";
+    } else {
+      InferenceRule rule = cs->inference().rule();
+      UnitIterator parents= cs->getParents();
+
+      out << Int::toString(cs->number()) << ". ";
```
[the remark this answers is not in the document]
quickbeam123: Indeed!

Kernel/InferenceStore.cpp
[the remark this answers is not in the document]
MichaelRawson: Is there a hard tab at the first character of the line, perhaps?

Kernel/Term.cpp
quickbeam123: A good point to ask Petra to check.

The reason I added this is that, while normally the added answer literal really occurs only negatively (except in the answer literal resolver, the unit clause which helps closing the proof, but never enters proof search explicitly) I saw it positive under AVATAR (through add_complementary) and the clause were it occurred was not meant for the answer literal resolver to resolve against and finish the proof.

Lib/StringUtils.cpp
quickbeam123: The next replacement can only happens after a position where the previous ended. Is it still quadratic?

Although if you do `a^100,a,b^100`, you will indeed get an answer `b^10000`. But isn't that what you expect from a replaceAll?

MichaelRawson: So from my quick reading I'd expect `where.replace` to have linear-ish time, and so the thing has complexity `O(mn)` where `m` is the length of `where` and `n` is the number of occurrences of `from`. But `n` is very likely small in practice so probably not a big deal, you're right.

Shell/UIHelper.hpp
quickbeam123: Done!

Shell/Options.hpp
quickbeam123: I think there was no code. Just plans ...

Inferences/InvalidAnswerLiteralRemovals.cpp
quickbeam123: You can say: `./vampire -qa plain -qaat "ans0(sK1)|ans0(f(X))"` and we will skip proofs that would lead to answers matching `ans0(sK1)` or `ans0(f(X))`. It's quite advantageous to do this from the command line and not from a tptp  file (we can wait until the skolem makes sense) and it's quite advantageous to (ab)use the TPTP parser.

MichaelRawson: Ahah. Neat! I failed to do the mental wiring-up, thanks.

Kernel/Unit.cpp
quickbeam123: Dead.

Kernel/Unit.hpp
quickbeam123: Yes, done in one pass during 
```
bool seenInputInference = refutation->minimizeAncestorsAndUpdateSelectedStats();
```
in UIHelper.


`refutation->derivedFromGoal()` is correctly updated during the minimization.

Kernel/TermIterators.cpp
```diff
@@ -439,22 +439,35 @@ Term* NonVariableNonTypeIterator::next()
...
+  unsigned taArity;
   unsigned arity;
-
-  if(t->isLiteral() && static_cast<Literal*>(t)->isEquality()){
+  if (t->isSpecial()) {
+    // This is a very incomplete support for special terms (which normally get eliminated during preprocessing).
+    // This satisfies the a use in AnswerLiteralManager (the synthesis version) where $ite-s may have creeped in.
+    // (We don't mind being iteration-incomplete there so skip the $ite-condition,
+    // which is a formula and would make things much more complicated here.)
+    // TODO decide: is it worth extending this properly (as usually, we won't encounter specail terms here)?
```
hzzv, line 450: Nit: "special", not "specail"

Parse/TPTP.hpp
```diff
@@ -852,6 +860,18 @@ class TPTP
   /** This field stores names of input units if the
    * output_axiom_names option is enabled */
   static DHMap<unsigned, vstring> _axiomNames;
+
+  /**
+   * During question parsing, we store the mapping from int variables
```
hzzv, line 865: Do we expect this to be used only for TPTP and not for SMT-LIB inputs?

Kernel/Term.cpp
hzzv: In synthesis the answer literals should always be negative.
If you're adding some logic that also allows them in positive context, maybe we could also add an assert somewhere checking that if `QuestionAnsweringMode::SYNTHESIS` is on, the answer literals are negative? (Maybe in `UncomputableAnswerLiteralRemoval::simplify`.)

Kernel/TermIterators.cpp
hzzv: ...also actually: was the TermIterator used on special terms? I thought we only introduce the special if-then-elses at the very end, when constructing the program, and that we don't do anything with them except printing. (I can be wrong, will have more time to look at it tomorrow.)

quickbeam123: I think I did this because I saw synthesis break here after some (unifying) changes by me.

Parse/TPTP.hpp
quickbeam123: I was only coding this up with the TPTP users in mind. Also, not that this only kicks in (for performance reasons) only for the tptp role "question". 

Not sure what the analogue would be in SMT-LIB, but I am definitely not against it.

hzzv: For synthesis, we add the answer literals for any conjecture in the form `(assert-not ([forall (...) (] exists ... [)] ))`.
The `assert-not` is somewhat non-standard (in addition to us also Z3 supports it, but not cvc5 for example), but we've been using it for conjectures for some time already.

hzzv (approved):
Overall looks good! I tried it with the recursion-free synthesis examples, nothing breaks (though we lost one, but it seems it's just random proof divergence).

MichaelRawson:
Thank you, @hzzv!

Do we want to postpone merging until we cut a post-CASC release? If so I can hurry along the release.

Issue #565, opened 2024-05-30
Vampire crashing when printing a finite model:


quickbeam123:
```
./vampire_dbg_master_7630 -sa fmb Problems/ITP/ITP383_10.p 
% Running in auto input_syntax mode. Trying TPTP
TRYING [1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1]
Finite Model Found!
% SZS status Satisfiable for ITP383_10
Condition in file ./Lib/DHMap.hpp, line 193 violated:
e
----- stack dump -----
Version : Vampire 4.8 (commit d71672a8c on 2024-05-20 12:15:43 +0200)
call stack: 0x7fff2063ef3d 0x100737119 0x100735960 0x1007348de 0x100731fd3 0x10045ac7d 0x10045af04 0x1000f2a5b 0x100698da3 0x10069cb7a 0x100385403 0x100020858 0x1000234a5
(use '--traceback on' to get a human-readable stack trace)
----- end of stack dump -----
Martins-MacBook-Pro-4:vampire mbassms6$ ./vampire_dbg_master_7630 -sa fmb Problems/ITP/ITP383_10.p --traceback on
% Running in auto input_syntax mode. Trying TPTP
TRYING [1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1]
Finite Model Found!
% SZS status Satisfiable for ITP383_10
Condition in file ./Lib/DHMap.hpp, line 193 violated:
e
----- stack dump -----
Version : Vampire 4.8 (commit d71672a8c on 2024-05-20 12:15:43 +0200)
call stack: 0x6 0x7fff2063ef3d 0x100737119 0x100735960 0x1007348de 0x100731fd3 0x10045ac7d 0x10045af04 0x1000f2a5b 0x100698da3 0x10069cb7a 0x100385403 0x100020858 0x1000234a5
invoking atos(1) ...
0x6
0x7fff2063ef3d
main (in vampire_dbg_master_7630) (vampire.cpp:795)
dispatchByMode(Kernel::Problem*) (in vampire_dbg_master_7630) (vampire.cpp:560)
vampireMode(Kernel::Problem*) (in vampire_dbg_master_7630) (vampire.cpp:366)
doProving(Kernel::Problem*) (in vampire_dbg_master_7630) (vampire.cpp:158)
[...]

Issue #461, opened 2023-07-18
Crash on model output from FMB - sorts?


MichaelRawson:
Reported by Geoff from CASC-29.

If you run current master with `--decode fmb+10_1_bce=on:fmbas=function:fmbsr=1.2:fde=unused:nm=0_846 ITP383_10.p` we get a crash during model printing [here](https://github.com/vprover/vampire/blob/fbb735283c8b498620f97fa9ad22e2267be2d2a2/FMB/FiniteModelBuilder.cpp#L2297) where the item doesn't exist in the map. Debug build is fine.

MichaelRawson:
Original report was
> I see that Vampire didn't manage to output a model for three problems in
> the TFN division ...
>     ITP383_10 - https://www.tptp.org/CASC/29/WWWFiles/Results/TFN/Vampire-SAT---4.8/ITP383_10
>     ITP391_1  - https://www.tptp.org/CASC/29/WWWFiles/Results/TFN/Vampire-SAT---4.8/ITP391_1
>     ITP389_1  - https://www.tptp.org/CASC/29/WWWFiles/Results/TFN/Vampire-SAT---4.8/ITP389_1
> 
> It looks like it established that a finite model exists, but could not 
> get it out, right? Any comment (for the CASC report)?

PR #568, opened 2024-06-10
CASC changes

MichaelRawson:
We made some last-minute changes for CASC, as usual. Let's put these here for general inspection and choose what to merge eventually.


Shell/Statistics.hpp
```diff
@@ -221,6 +229,7 @@ class Statistics
   unsigned equationalTautologies;
   /** number of forward subsumed clauses */
   unsigned forwardSubsumed;
+  unsigned forwardIbuSubsumed;
```
MichaelRawson, line 232: Comment please!

FMB/FiniteModelBuilder.cpp
```diff
@@ -2172,6 +2172,7 @@ void FiniteModelBuilder::onModelFound()
   //}
 
   //Evaluate removed propositions and predicates
+#if 0
```
MichaelRawson, line 2175: Remember to debug this properly: do we have an issue for this?

Indexing/LiteralIndex.cpp
```diff
@@ -107,7 +107,11 @@ void UnitClauseLiteralIndex::handleClause(Clause* c, bool adding)
 {
   if(c->length()==1) {
     TIME_TRACE("unit clause index maintenance");
-    
+
+    if (_skipNonequational && !(*c)[0]->isEquality()) {
```
MichaelRawson, line 111: This seems like something other indices might use, right? Did e.g. LHS indices get updated?

Kernel/Clause.cpp
```diff
@@ -75,7 +77,8 @@ Clause::Clause(unsigned length,const Inference& inf)
     _reductionTimestamp(0),
     _literalPositions(0),
     _numActiveSplits(0),
-    _auxTimestamp(0)
+    _auxTimestamp(0),
+    _supData(0)
 {
   // MS: TODO: not sure if this belongs here and whether EXTENSIONALITY_AXIOM input types ever appear anywhere (as a vampire-extension TPTP formula role)
```
MichaelRawson, line 83: Since this is here: I think `EXTENSIONALITY_AXIOM` might be completely useless and I've never seen it in TPTP input: was it perhaps used for experiments wrt "Extensional Crisis and Proving Identity"? If so it can probably go.

Lib/Allocator.cpp
```diff
@@ -33,18 +33,27 @@ void Lib::setMemoryLimit(size_t limit) { LIMIT = limit; }
...
 // TODO does not support get_new_handler/set_new_handler as we don't use it, but we could
-void *operator new(size_t size) {
+void *operator new(size_t size, std::align_val_t align) {
   if(ALLOCATED + size > LIMIT)
     throw std::bad_alloc();
   ALLOCATED += size;
   {
     Lib::TimeoutProtector tp;
-    if(void *ptr = std::malloc(size))
+    if(void *ptr = std::aligned_alloc(static_cast<size_t>(align), size))
```
MichaelRawson, line 42: TODO this needs to be fixed on Mac. `size` needs to be a multiple of `align`, apparently.

Lib/Allocator.cpp
```diff
@@ -33,18 +33,27 @@ void Lib::setMemoryLimit(size_t limit) { LIMIT = limit; }
...
   }
   throw std::bad_alloc();
 }
 
+// align-less operator new
+// forwards to the aligned version with alignof(max_align_t)
+void *operator new(size_t size) {
+  return operator new(
+    size,
+    static_cast<std::align_val_t>(alignof(std::max_align_t))
```
MichaelRawson, line 53: Maybe this is why we get so many `align=16` requests to the aligned version?

quickbeam123:
`-fibus` and `-fsoe`, while being used in the UEQ schedule are candidates for immediate kick-out. So `_skipNonequational` in `UnitClauseLiteralIndex::handleClause` or `forwardIbuSubsumed` in `Statistics` might not need any polishing.

FMB/FiniteModelBuilder.cpp
quickbeam123: We have two:
https://github.com/vprover/vampire/issues/461
https://github.com/vprover/vampire/issues/565

Indexing/LiteralIndex.cpp
quickbeam123: LHS is equational only by design, isn't it?

Lib/Allocator.cpp
quickbeam123: Because we say we want everything aligned "maximally"? (I thought you wrote this code, @MichaelRawson ;))

Lib/Allocator.cpp
quickbeam123: (And it might not like being called with `align==1`.)

Kernel/Clause.cpp
quickbeam123: The input type EXTENSIONALITY_AXIOM could go. It was out custom tptp extension, IIRC.

The actual trick (which may be fed ext. axioms also through "fishing") is still doing something an appears in strategies...

Indexing/LiteralIndex.cpp
MichaelRawson: Right - but if indices can inherit from this and get the right behaviour that would be a nice pattern. Looking at it maybe it doesn't link up the way I think it does, sorry.

PR #572, opened 2024-06-14
remove `splPrefix`, not necessary

MichaelRawson:
There's a static `splPrefix` floating around in `Splitter`, which is mirrored as `splitPrefix` in `InferenceStore`. It doesn't seem to do much and does not conform with the [convention](https://tptp.cs.miami.edu/TSTP/NewSymbolNames.html).

Just compute this thing on-demand and conform to the convention.

### DIFF
--- a/Kernel/InferenceStore.cpp
+++ b/Kernel/InferenceStore.cpp
@@ -512,7 +512,6 @@ struct InferenceStore::TPTPProofPrinter
 {
   TPTPProofPrinter(ostream& out, InferenceStore* is)
   : ProofPrinter(out, is) {
-    splitPrefix = Saturation::Splitter::splPrefix; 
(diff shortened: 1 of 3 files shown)


MichaelRawson:
It's "namespaced" by putting `_split` on the end. So unless we have other names of the form `spN_split`, no.

quickbeam123:
Is there are chance these names could mix with any other term material? I am thinking of the evil user coming with a TPTP problem with a predicate `sp5_split` (perhaps by coincidence) in the signature.

MichaelRawson:
You are quite right - I thought it was just "already broken" and this wouldn't fix it (but would be considerably shorter) but I think `addFreshFunction` does some linear search, which surprised me.

Let's close this and I'll keep this corner in mind when I get to look at some of my other gripes with `Signature`.

PR #575, opened 2024-07-08
Defaults impemenations for optimized Ordering methods

joe-hauns:
Added default implementations for new optimized Kernel::Ordering methods.


MichaelRawson (approved):
Looks great to me. I like the 'optionally-optimised methods' pattern used here.

mezpusz (approved):

PR #577, opened 2024-07-10
replace signal-based Timer with a ticking thread

MichaelRawson:
See #157, #462. Our current timer arranges for the process to be signalled with `SIGALRM` every millisecond, which is OK but not very portable and requires our code to be signal-safe, a concept which seems to be poorly specified/understood compared to thread safety.

This replaces the mechanism with an auxiliary thread per process that wakes up, checks resource limits, updates instruction counts, then sleeps for a millisecond or so. On process exit, this thread unceremoniously dies.

I don't have good intuition for how this performs. In principle this should be equal or better than the old mechanism. On my machine I get comparable numbers of generated clauses with a 10-second timeout, but the new implementation times out much more precisely.

Also:
- remove some code that was only needed for the old implementation
- simplify the `Timer` API
- remove some manual "did we time out yet" code from random places - AFAIK this has not been useful since we implemented external timeout logic

Some of the code needs close attention: I will flag the tricky bits with GitHub comments.


CASC/PortfolioMode.cpp
```diff
@@ -632,12 +632,7 @@ void PortfolioMode::runSlice(Options& strategyOpt)
 {
   System::registerForSIGHUPOnParentDeath();
   UIHelper::portfolioParent=false;
-
-  env.timer->reset();
-  env.timer->start();
-
-  Timer::resetInstructionMeasuring();
-  Timer::setLimitEnforcement(true);
+  Timer::reinitialise();
```
MichaelRawson, line 635: This must be called here as we are in a new process. The parent process' timer thread will keep ticking, but it won't kill us or update our instruction counts.

CASC/PortfolioMode.cpp
```diff
@@ -632,12 +632,7 @@ void PortfolioMode::runSlice(Options& strategyOpt)
 {
   System::registerForSIGHUPOnParentDeath();
   UIHelper::portfolioParent=false;
-
-  env.timer->reset();
-  env.timer->start();
-
-  Timer::resetInstructionMeasuring();
-  Timer::setLimitEnforcement(true);
```
MichaelRawson, line 640: Now unnecessary.

FMB/FiniteModelBuilder.cpp
```diff
@@ -1601,9 +1601,6 @@ MainLoopResult FiniteModelBuilder::runImpl()
       }
       cout << "]" << endl;
     }
-    Timer::syncClock();
-    if(env.timeLimitReached()){ return MainLoopResult(Statistics::TIME_LIMIT); }
-
```
MichaelRawson, line 1606: Lots of code like this. I don't like to remove code that I don't understand, but as far as I can tell this was pointless once we had signal-based timeouts.

Lib/Allocator.cpp
```diff
@@ -42,7 +44,6 @@ void *operator new(size_t size, std::align_val_t align_val) {
     throw std::bad_alloc();
   ALLOCATED += size;
   {
-    Lib::TimeoutProtector tp;
```
MichaelRawson, line 45: We don't need to protect calls to the system allocator any more as it's supposed to be thread-safe (but was *not* signal-safe).

Lib/Portability.hpp
```diff
@@ -19,4 +19,12 @@ static_assert(
     "Vampire assumes that there are 8 bits in a `char`"
 );
 
+#define VAMPIRE_PERF_EXISTS 0
```
MichaelRawson, line 22: Move this preprocessor here so you don't have to include `Timer.hpp` just for this.

vampire.cpp
```diff
@@ -784,6 +779,9 @@ int main(int argc, char* argv[])
       exit(0);
     }
 
+    // can only happen after reading options as it relies on `env.options`
+    Timer::reinitialise();
+
```
MichaelRawson, line 784: This has been lifted a little earlier during startup as I don't see why we wouldn't want to timeout during parsing. Happy to hear objections.

quickbeam123: We do want during parsing. Are you sure we didn't? But we don't want to time out during the interactive mode. Could you please check that's fine?

MichaelRawson: Do I have it right that you determined this already and also fixed it in e583c89?

quickbeam123: Yes, we definitely had a bug since the interactive mode and making your PR compatible with the fix is probably already enough...

CASC/PortfolioMode.cpp
quickbeam123: But do we then inherit the parents "o'clock" and "instr count"? Because they should start from 0...

Saturation/SaturationAlgorithm.cpp
```diff
@@ -733,8 +725,8 @@ void SaturationAlgorithm::init()
     _symEl->init(this);
   }
 
-  _startTime = env.timer->elapsedMilliseconds();
-  _startInstrs = env.timer->elapsedMegaInstructions();
+  _startTime=Timer::elapsedMilliseconds();
+  _startInstrs=Timer::elapsedMegaInstructions();
```
quickbeam123, line 729: Maybe these could now go too?

quickbeam123: Oh, I didn't realized it's for LRS. 

Hmm, but while I see `_startTime` there, `_startInstrs` seem to be missing (although by analogy they should probably appear on line 113 there). What do you think?

MichaelRawson: OK, so `_startInstrs` is definitely not used. Now that I look I'm not sure that what LRS does with `_startTime` makes sense either, probably you don't care when you started saturating so much as when the Vampire run started. Should we get rid of `_startTime` as well and just use the elapsed time for LRS, or do you have some counter-argument?

quickbeam123: I think it's important that we don't start LRS-ing before we've made some number of iterations around the saturation loop, but that's happening on the line `    if (processed<=10) {`. 

The `_startTime` and `_startInstrs` are there to implement `opt.lrsFirstTimeCheck()` and there it's probably even more intuitive to count the time from the start of things (i.e. not from the start of saturation). Moreover, the current solution for time is confusing: e.g., lrsFirstTimeCheck=25% and half of the time is spent parsing and preprocessing, then the first check will only kick in at around 75% allotted time (because it ignores startup, but only estimates the %time spent from the beginning). Long story - I think the can both go and we can implement `lrsFirstTimeCheck` along the lines of the instr version, which does not need `_startInstrs`.

PR #578, opened 2024-07-11
Removing instantiation sat algorithm from schedules

### LINKED ISSUE #576: bad test id errors
Running vampire with the portfolio `snake_tptp_uns` all over smtlib and tptp gives quite some user errors of the shape `bad test id ins+<strategy-string>`. This is due to the removal of the `ins` saturation algorithm, which is still present in the schedule. 
What do you think we should do about this? Should we just remove all these strategies from the schedules?

joe-hauns:
fixes #576


MichaelRawson:
Low chance of breaking something so I'll just merge, @quickbeam123.

This is somewhat my fault for not doing this when I removed InstGen. Sorry!

PR #504, opened 2023-11-24
use addr2line(1) for stacktraces

MichaelRawson:
Something to chew on over the weekend. @hzzv brought to my attention that the stack trace mechanism could be better: it doesn't include file/line information where available, and it's kinda flaky with static builds for some reason I don't fully understand.

The current solution calls a system function to obtain a backtrace in the form of an array of pointers, then jumps through system hoops to map them to symbols. Instead, I propose that we shell out to `addr2line(1)`, which is very often installed on POSIXy machines. If we have it, great. If not, we're crashing anyway so we don't lose much, and we print out the addresses so that we can figure it out offline. This also rids us of a library dependency and some build magic. What do you think?

PR #579, opened 2024-07-15
invoke a debugger instead of going via addr2line and friends

MichaelRawson:
In #412 I (perhaps foolishly, in hindsight) removed the `CALL` mechanism with some code that found stack frames, then asked the system to resolve them to function names. This had a lot of benefits, but it was a bit flaky and had to be improved in #504 so that we shell out to `addr2line`.

This worked, but in itself was also flaky in the presence of [ASLR](https://en.wikipedia.org/wiki/Address_space_layout_randomization), which is becoming harder to avoid. On ARM Macs, for example, it seems you can't switch it off easily.

So: since we already shell out to a program, why not use something designed for this? Replace the whole mess with invocations of `gdb` or `lldb` in their batch modes to get a backtrace. This should (hopefully) be the last attempt in this sorry saga.

Things still to check:
- Does this work for @quickbeam123 , the perennial tricky customer?
- Does it work on StarExec? (possibly of reduced importance given StarExec's impending decommissioning)
- Does it interact well with #577?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,7 +5,6 @@ cmake_minimum_required (VERSION 3.14.0)
 project(Vampire)
 
 include(CheckIPOSupported)
-include(CheckPIESupported)
 
 # require the compiler to use C++17
 set(CMAKE_CXX_STANDARD 17)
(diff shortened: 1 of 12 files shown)


MichaelRawson:
No dice on StarExec. Not sure if we care.

quickbeam123:
Will try to investigate a bit more, but currently, this kills my terminal with 
[terminal.txt](https://github.com/user-attachments/files/16264026/terminal.txt)

MichaelRawson:
...your actual terminal emulator dies, right? I have no idea what to do about that, sorry. If you redirect output to a file does it perhaps not crash the terminal?

PR #580, opened 2024-07-18
Demodulation redundancy check cannot be turned off

mezpusz:
During the latest refactor, I forgot to actually use the flag which checks `drc=off`.

I wonder what effects this had on our CASC schedules.

### DIFF
--- a/Inferences/DemodulationHelper.cpp
+++ b/Inferences/DemodulationHelper.cpp
@@ -34,6 +34,10 @@ DemodulationHelper::DemodulationHelper(const Options& opts, const Ordering* ord)
 
 bool DemodulationHelper::redundancyCheckNeededForPremise(Clause* rwCl, Literal* rwLit, TermList rwTerm) const
 {
+  if (!_redundancyCheck) {
+    return false;
+  }
+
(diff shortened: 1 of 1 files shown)


quickbeam123:
The CASC FOF schedules were using strategies developed with a previous version in which drc=off still worker, by I reevaluated them with the then current vampire, so the schedule construction "knew" drc=off does not matter. These schedules will now be slightly off.

On the other hand, the UEQ strategies were discovered with the buggy vampire and I so if you check, they never try to set drc=off as it is a non-default value and had been observed not to make a difference. So the UEQ strategies will not be affected at all.

quickbeam123 (approved):

MichaelRawson (approved):
Provided we think `drc=off` is still useful, I'm happy.

mezpusz:
> Provided we think `drc=off` is still useful, I'm happy.

https://github.com/vprover/vampire/pull/497 Maybe something relevant. I tried to convince Martin that either `drc=on` or `drc=off` has to go, but they seem to have a better "coverage" together than encompassment plus one or the other.

PR #582, opened 2024-07-22
GREATER_EQ and LESS_EQ have to go

mezpusz:
I have finally got annoyed enough to remove the two enum values that annoyed many Vampire developers before me (based on comments in the code).

There were probably earlier `Ordering` implementations which used these values but now nothing returns them on the term level, and literal-level comparisons like `Ordering_Equality` only used them when they were returned from the term level. Removing them gives a cleaner code and will hopefully allow more optimisations down the road.

I know only of `ArgumentOrderVals` which relies on the exact layout of `Ordering`, please let me know if there are more of these somewhere.

### DIFF
--- a/Kernel/Ordering.hpp
+++ b/Kernel/Ordering.hpp
@@ -62,10 +62,8 @@ public:
   enum [[nodiscard]] Result {
     GREATER=1,
     LESS=2,
-    GREATER_EQ=3,
-    LESS_EQ=4,
-    EQUAL=5,
-    INCOMPARABLE=6
+    EQUAL=3,
+    INCOMPARABLE=4
   };
 
   Ordering();
@@ -95,7 +93,7 @@ public:
 
   virtual void show(std::ostream& out) const = 0;
 
-  static bool isGorGEorE(Result r) { return (r == GREATER || r == GREATER_EQ || r == EQUAL); }
+  static bool isGorE(Result r) { return (r == GREATER || r == EQUAL); }
 
   void removeNonMaximal(LiteralList*& lits) const;
 
@@ -107,12 +105,8 @@ public:
     switch(r) {
     case GREATER:
       return LESS;
-    case GREATER_EQ:
-      return LESS_EQ;
     case LESS:
       return GREATER;
-    case LESS_EQ:
-      return GREATER_EQ;
     case EQUAL:
     case INCOMPARABLE:
       return r;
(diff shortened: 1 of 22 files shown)


Kernel/Ordering.hpp
```diff
@@ -95,7 +93,7 @@ class Ordering
 
   virtual void show(std::ostream& out) const = 0;
 
-  static bool isGorGEorE(Result r) { return (r == GREATER || r == GREATER_EQ || r == EQUAL); }
+  static bool isGorE(Result r) { return (r == GREATER || r == EQUAL); }
```
mezpusz, line 96: I think this could be called `isGreaterOrEqual` for clarity but I'm okay with this too.

Kernel/Ordering_Equality.cpp
```diff
@@ -45,21 +45,11 @@ class Ordering::EqCmp
   }
 
   Result compare_s1Gt1(TermList s1,TermList s2,TermList t1,TermList t2) const;
```
mezpusz, line 47: Maybe this entire class could be just `static` methods, I think having an object of this class lost its value some refactors before. Having class members `s1,s2,t1,t2` is especially confusing since then we use arguments with the same name in almost all methods. What do you think?

Kernel/Ordering.hpp
quickbeam123: I vote for the longer name, if that's not too much work...

Kernel/Ordering_Equality.cpp
quickbeam123: Yes, please! 

The `s1,s2,t1,t2;` members can become local in `compareEqualities` and that's it, right?

mezpusz: Yep, that should be it.

quickbeam123 (approved):

Kernel/Term.hpp
```diff
@@ -88,10 +88,8 @@ enum ArgumentOrderVals {
   AO_UNKNOWN=0,
   AO_GREATER=1,
   AO_LESS=2,
-  AO_GREATER_EQ=3,
-  AO_LESS_EQ=4,
-  AO_EQUAL=5,
-  AO_INCOMPARABLE=6,
+  AO_EQUAL=3,
+  AO_INCOMPARABLE=4,
```
MichaelRawson, line 92: It's slightly annoying that this still needs 5 values. Otherwise we could save *one whole bit* (gasp) in terms.

Are things only `AO_EQUAL` syntactically? If so I could imagine `AO_EQUAL_OR_UNKNOWN` and a fast equality check to differentiate `AO_UNKNOWN` and `AO_EQUAL`. But this is premature optimisation.

MichaelRawson (approved):

Kernel/Term.hpp
MichaelRawson: I think so. But let's keep that for another PR - I'll put it on my list to investigate 'eventually' unless you get inspired. :-)

PR #583, opened 2024-08-01
Implement conditional redundancy handler

mezpusz:
First checkpoint in a series of improving what was known (briefly) as `InstanceRedundancyHandler` (cf. `irc` flag).
- Renamed to `ConditionalRedundancyHandler` and added more general functionality.
- Accordingly added options and statistics to reflect the current features.
- Refactored binary resolution and superposition code a bit, which contained some unused stuff.
- Removed the incompleteness flag in `Options.cpp` as I believe I have a proof of completeness.

Some motivation statistics by running a single strategy (discount/otter) for 10s:

| baseline | crc | croc | crac | crlc |
| discount | 9985(0) | 10013(124) | 10001(147) | 10041(163) | 9979(124) |
| otter | 9721(0) | 9764(110) | 9770(142) | 9772(130) | 9745(101) |
| baseline | cr[oa]c | cr[ol]c | cr[al]c | cr[oal]c |
| discount | 9985(0) | 10018(172) | 9964(135) | 10009(163) | 9988(169) |
| otter | 9721(0) | 9785(163) | 9766(143) | 9771(131) | 9780(161) |

(cr[oal]c means `-crc on -croc on -crac on -crlc on`)

More info/documentation/evaluation comes later as this is work in progress.


Inferences/EqualityFactoring.cpp
```diff
@@ -92,8 +92,8 @@ struct EqualityFactoring::FactorablePairsFn
 
 struct EqualityFactoring::ResultFn
 {
-  ResultFn(EqualityFactoring& self, Clause* cl, bool afterCheck, bool instanceRedundancyCheck, Ordering& ordering, bool fixedPointIteration)
-      : _self(self), _cl(cl), _cLen(cl->length()), _afterCheck(afterCheck), _instanceRedundancyCheck(instanceRedundancyCheck), _ordering(ordering), _fixedPointIteration(fixedPointIteration) {}
+  ResultFn(EqualityFactoring& self, Clause* cl, bool afterCheck, const ConditionalRedundancyHandler* condRedHandler, Ordering& ordering, bool fixedPointIteration)
```
MichaelRawson, line 95: Could this be a reference, or is it nullable? Similarly elsewhere.

Inferences/EqualityFactoring.cpp
```diff
@@ -163,9 +163,7 @@ struct EqualityFactoring::ResultFn
       }
     }
 
-    if (_instanceRedundancyCheck &&
-      !InstanceRedundancyHandler::handleReductiveUnaryInference(_cl, &absUnif.subs(), &_ordering))
-    {
+    if (!_condRedHandler->handleReductiveUnaryInference(_cl, &absUnif.subs())) {
```
MichaelRawson, line 166: ...and then if it's nullable, don't you need a check here?

Inferences/FunctionDefinitionRewriting.cpp
```diff
@@ -64,7 +64,8 @@ Clause* performRewriting(
 
   Applicator appl(subst.ptr());
 
-  if (helper && !helper->isPremiseRedundant(rwClause,rwLit,rwTerm,tgtTermS,eqLHS,&appl)) {
+  Ordering::Result temp;
```
MichaelRawson, line 67: Perhaps it would be worth adding an overload for the "we don't care about the ordering result" case.

Lib/SharedSet.hpp
```diff
@@ -430,7 +430,7 @@ class SharedSet {
   }
   static unsigned hash(const T* arr, size_t len)
   {
-    static_assert(std::is_arithmetic<T>::value, "T must be safely hashable");
+    // static_assert(std::is_arithmetic<T>::value, "T must be safely hashable");
```
MichaelRawson, line 433: Uh-oh. Did this fail?

Saturation/SaturationAlgorithm.cpp
```diff
@@ -208,7 +209,7 @@ SaturationAlgorithm::SaturationAlgorithm(Problem& prb, const Options& opt)
     _fwSimplifiers(0), _simplifiers(0), _bwSimplifiers(0), _splitter(0),
     _consFinder(0), _labelFinder(0), _symEl(0), _answerLiteralManager(0),
     _instantiation(0), _fnDefHandler(prb.getFunctionDefinitionHandler()),
-    _generatedClauseCount(0),
+    _conditionalRedundancyHandler(), _generatedClauseCount(0),
```
MichaelRawson, line 212: Pointless initialiser, I think.

Saturation/Splitter.cpp
```diff
@@ -1497,6 +1497,16 @@ void Splitter::onClauseReduction(Clause* cl, ClauseIterator premises, Clause* re
   }
 }
 
+void Splitter::addConditionalReduction(SplitSet* splits, ConditionalRedundancyEntry* e)
```
MichaelRawson, line 1500: Not to split hairs, but "conditionally reduced" already has a meaning in AVATAR. This is a different meaning, right?

Shell/ConditionalRedundancyHandler.hpp
```diff
@@ -0,0 +1,127 @@
...
+  unsigned refcnt = 1;
+
+  void deactivate() {
+    ASS(!splits->isEmpty());
+    active = false;
+  }
+
+  void obtain() {
+    refcnt++;
+  }
```
MichaelRawson, line 54: Would it be worth wrapping this in e.g. a `std::shared_ptr` or at least RAIIng this?

Shell/ConditionalRedundancyHandler.cpp
```diff
@@ -0,0 +1,564 @@
...
+ * Vampire. It is protected by applicable
+ * copyright laws.
+ *
+ * This source code is distributed under the licence found here
+ * https://vprover.github.io/license.html
+ * and in the source directory
+ */
+/**
+ * @file ConditionalRedundancyHandler.cpp
+ * Implements class ConditionalRedundancyHandler.
```
MichaelRawson, line 12: I would be considerably happier if there were a (possibly brief) explanation of what's going on here at an intuitive level.

MichaelRawson (changes requested):
Cool! Nice results and the AVATAR integration seems much less painful than expected. Generally happy and only asking for minor changes - as usual I'll trust that your new code does what it says on the tin.

Lib/SharedSet.hpp
mezpusz: Ah yes. A pointer should also be safely hashable, right? Then I can add a disjunct with `std::is_pointer`.

Shell/ConditionalRedundancyHandler.hpp
mezpusz: Code trees need a raw pointer, so at least there a shared pointer cannot be used (unless we heap allocate a shared pointer). Otherwise, what exactly you had in mind for RAIIing?

mezpusz:
> the AVATAR integration seems much less painful than expected.

We could still try to extend it to unfrozen inferences later. But this version seemed like a good starting point. :)

Saturation/Splitter.cpp
mezpusz: Then maybe just `addConditionalRedundancyEntry`? The concepts are indeed very similar, not just in name. 😃

Lib/SharedSet.hpp
MichaelRawson: Yes please!

Shell/ConditionalRedundancyHandler.hpp
MichaelRawson: Ah, the raw pointer thing makes sense then. The `obtain`/`release` stuff just looked like a `std::shared_ptr`.

MichaelRawson (approved):

Shell/ConditionalRedundancyHandler.hpp
mezpusz: As an afterthought, if you think a heap-allocated `shared_ptr` is not weird, we can use that instead to get rid of this code.

MichaelRawson: I don't think it's too weird, but let's keep it as-is for now. Later when I am feeling brave I can go looking for refcounting patterns and see if things can be improved.

PR #584, opened 2024-08-02
Fixes several model printing bugs (and more)

quickbeam123:
Quite a few changes in FiniteModelBuilder and around with the ultimate aim of Vampire being able to correctly model-check all finite models it generates. 

This PR achieves this goal for inputs in CNF and under `fde=none:erd=off:updr=off:bce=off` - (This means we verify models we output, but so far only for strategies which don't manipulate the signature or eliminate things.)

The next step will be to support undoing these elimination operations and making the model work for the original input (and to do the corresponding operations also for saturations, where we will simply dump the definitions of the eliminated symbols to document how one could recover a model, if they had one.)

### DIFF
--- a/FMB/FiniteModelBuilder.hpp
+++ b/FMB/FiniteModelBuilder.hpp
@@ -135,7 +135,7 @@ private:
   // are used to give the interpretation of the function/predicate if a model is found
   DHMap<unsigned,Literal*> _deletedFunctions;
   DHMap<unsigned,Unit*> _deletedPredicates;
-  DHMap<unsigned,Unit*> _partiallyDeletedPredicates; 
+  DHMap<unsigned,Unit*> _partiallyDeletedPredicates;
   DHMap<unsigned,bool> _trivialPredicates;
   // if del_f[i] (resp del_p[i]) is true then that function (resp predicate) should be ignored
   DArray<unsigned> del_f;
--- a/FMB/FiniteModelBuilder.hpp
+++ b/FMB/FiniteModelBuilder.hpp
@@ -131,8 +131,8 @@ private:
   ScopedPtr<SATSolverWithAssumptions> _solver;
 
   // if del_f[i] (resp del_p[i]) is true then that function (resp predicate) should be ignored
-  DArray<unsigned> del_f;
-  DArray<unsigned> del_p;
+  DArray<bool> del_f;
+  DArray<bool> del_p;
 
   // Store monotonicity_info (see Monotonicity::check) for every sort detected (or made) monotonic
   DHMap<unsigned,DArray<signed char>*> _monotonic_vampire_sorts;
--- a/FMB/FiniteModelBuilder.hpp
+++ b/FMB/FiniteModelBuilder.hpp
@@ -136,6 +136,8 @@ private:
 
   // Store monotonicity_info (see Monotonicity::check) for every sort detected (or made) monotonic
   DHMap<unsigned,DArray<signed char>*> _monotonic_vampire_sorts;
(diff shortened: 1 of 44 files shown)


SAT/SATSolver.hpp
```diff
@@ -24,13 +24,13 @@ namespace SAT {
...
 public:
-  enum VarAssignment {
+  enum class VarAssignment {
     TRUE,
     FALSE,
     DONT_CARE,  // to represent partial models
     NOT_KNOWN
   };
-  enum Status {
+  enum class Status {
```
MichaelRawson, line 33: Nice. Very happy about this!

Kernel/Problem.hpp
```diff
@@ -191,7 +191,7 @@ class Problem {
   UnitList* _units;
   DHMap<unsigned,Literal*> _deletedFunctions;
   DHMap<unsigned,Unit*> _deletedPredicates;
-  DHMap<unsigned,Unit*> _partiallyDeletedPredicates; 
+  DHMap<unsigned,Unit*> _partiallyDeletedPredicates;
```
MichaelRawson, line 194: Since I am reading this anyway and I suspect you understand this better...when is a predicate only partially deleted? Could there be a comment somewhere?

FMB/SortInference.hpp
```diff
@@ -121,18 +122,15 @@ class SortInference {
...
-  DArray<unsigned> _del_f;
-  DArray<unsigned> _del_p;
-  Stack<DHSet<unsigned>*> _equiv_v_sorts;
-  IntUnionFind _equiv_vs;
+  const DArray<bool>& _del_f;
+  const DArray<bool>& _del_p;
 
   Stack<std::pair<unsigned,unsigned>>& _sort_constraints;
-
+  DHMap<unsigned,DArray<signed char>*>& _monotonic_vampire_sorts;
```
MichaelRawson, line 133: Could these fields please have a comment as elsewhere?

MichaelRawson (approved):
As usual I trust your new code does what you hope it does! But I don't see any immediate problems and I do see big type safety and complexity improvements.

MichaelRawson:
I've mostly just asked for some explanatory comments...once you've done what you are willing to do go ahead and merge.

FMB/FiniteModelBuilder.hpp
```diff
@@ -130,16 +130,14 @@ class FiniteModelBuilder : public MainLoop {
...
   // if del_f[i] (resp del_p[i]) is true then that function (resp predicate) should be ignored
-  DArray<unsigned> del_f;
-  DArray<unsigned> del_p;
+  DArray<bool> del_f;
+  DArray<bool> del_p;
+
+  // Store monotonicity_info (see Monotonicity::check) for every sort detected (or made) monotonic
+  DHMap<unsigned,DArray<signed char>*> _monotonic_vampire_sorts;
+  Stack<unsigned> _sortFunctions; // sort functions to remember - need to be eliminated from the model in the end
+  Stack<unsigned> _sortPredicates; // sort predicates to remember - need to be eliminated from the model in the end
```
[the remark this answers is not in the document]
quickbeam123: In the sense of https://smallbone.se/papers/sort-it-out-with-monotonicity.pdf (page 3)

Kernel/Problem.hpp
quickbeam123: It's when UPDR decides only one impliction of an <=> is needed. 

Note that this will all get revised in a subsequent PR. (If you remember the big `#if 0` section we put into FiniteModelBuilder before CASC, I made it even bigger here, so these (partially)deleted creatures currently don't play any role at all.)

PR #585, opened 2024-08-04
Remove VarManager

quickbeam123:
- was only ever used in a trivial way
- also simplified the Skolemization interface a bit


MichaelRawson (approved):
Tag #157. Nice!

PR #587, opened 2024-08-08
remove AWClauseContainer

quickbeam123:
this was probably used by InstGen only; 

not the same thing as the still-needed (of course) AWPassiveClauseContainer as such


MichaelRawson (approved):
Tag #157.

Issue #592, opened 2024-09-03
Aligned allocation is broken: `error: 'aligned_alloc' is not a member of 'std'; did you mean 'aligned_union'?`


barracuda156:
@MichaelRawson We got a new problem here, could you please take a look?

```
/opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_math_vampire/vampire/work/vampire-4.9/Lib/Allocator.cpp:46:25: error: 'aligned_alloc' is not a member of 'std'; did you mean 'aligned_union'?
   46 |     if(void *ptr = std::aligned_alloc(align, size))
      |                         ^~~~~~~~~~~~~
      |                         aligned_union
```

`aligned_alloc` is not provided by the compiler, it is provided or not by the OS. Since MacOS does not have it on most versions, this fails.
It should use `posix_memalign` as a fallback or `malloc`.

Also broken in the master:
```
FAILED: CMakeFiles/obj.dir/Lib/Allocator.cpp.o 
/opt/local/bin/ccache /opt/local/bin/g++-mp-14 -DCHECK_LEAKS=0 -DVDEBUG=0 -DVTIME_PROFILING=0 -DVZ3=0 -I/opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_math_vampire/vampire/work/vampire-2f897a03725680d11eefb03b1a2e7022158e7f7a -pipe -DNDEBUG -I/opt/local/include -D_GLIBCXX_USE_CXX11_ABI=0 -O3 -DNDEBUG -std=c++17 -arch ppc -mmacosx-version-min=10.6 -Wall -fno-threadsafe-statics -fno-rtti -fsized-deallocation -MD -MT CMakeFiles/obj.dir/Lib/Allocator.cpp.o -MF CMakeFiles/obj.dir/Lib/Allocator.cpp.o.d -o CMakeFiles/obj.dir/Lib/Allocator.cpp.o -c /opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_math_vampire/vampire/work/vampire-2f897a03725680d11eefb03b1a2e7022158e7f7a/Lib/Allocator.cpp
/opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_math_vampire/vampire/work/vampire-2f897a03725680d11eefb03b1a2e7022158e7f7a/Lib/Allocator.cpp: In function 'void* operator new(size_t, std::align_val_t)':
[...]

PR #593, opened 2024-09-03
Allocator.cpp: fix for macOS where aligned_alloc may not be supported

barracuda156:
Fixes: https://github.com/vprover/vampire/issues/592

@MichaelRawson Could you please review this?


MichaelRawson (approved):
Thanks! I plan a follow-up PR to remove this code altogether, but in the meantime this is a good fix. :-)

PR #594, opened 2024-09-03
use rlimit to enforce memory limits rather than global allocators

MichaelRawson:
As discussed with @quickbeam123 and related to #592. The current standard-compliant solution to memory limiting seems a bit fragile and under-reports memory use in some seemingly-random cases. Use POSIX `setrlimit` with `RLIMIT_DATA` to cause `ENOMEM` when we exceed the specified limit.

### DIFF
--- a/Saturation/ProvingHelper.cpp
+++ b/Saturation/ProvingHelper.cpp
@@ -57,9 +57,6 @@ using namespace Shell;
   catch(const std::bad_alloc &) {
     env.statistics->terminationReason=Statistics::MEMORY_LIMIT;
     env.statistics->refutation=0;
-    size_t limit=Lib::getMemoryLimit();
-    //add extra 1 MB to allow proper termination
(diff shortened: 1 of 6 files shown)


Saturation/ProvingHelper.cpp
```diff
@@ -57,9 +57,6 @@ using namespace Shell;
   catch(const std::bad_alloc &) {
     env.statistics->terminationReason=Statistics::MEMORY_LIMIT;
     env.statistics->refutation=0;
-    size_t limit=Lib::getMemoryLimit();
-    //add extra 1 MB to allow proper termination
-    Lib::setMemoryLimit(limit+1000000);
```
MichaelRawson, line 62: This doesn't work because `setrlimit` cannot increase the limit once it is lowered. *However*, it doesn't seem to be actually necessary as we free some things while the exception is in flight.

PR #595, opened 2024-09-03
remove Z3/pthreads workarounds from CMakeLists

MichaelRawson:
@mezpusz reports that we broke his build because we use `<thread>` now.
1. Link to the thread-support library in the CMake build.
2. Remove some cruft from CMake that worked around some static-link/pthread stuff. @quickbeam123 and I are not sure this is still necessary, if we get complaints we can put it back.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -809,6 +809,7 @@ if(CMAKE_BUILD_TYPE STREQUAL Debug)
   add_compile_definitions(VDEBUG=1)
 elseif(CMAKE_BUILD_TYPE STREQUAL Release)
   add_compile_definitions(VDEBUG=0)
+  add_compile_definitions(NDEBUG)
 endif()
 
 # enable for time profiling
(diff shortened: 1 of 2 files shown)


MichaelRawson:
(I also noticed that we don't set `NDEBUG` in release builds, which we should)

mezpusz:
I can confirm that now the CMake build works on the problematic machine, thanks for the change! However, the `Makefile` at root is still broken with undefined reference to `pthread_create`. Do you think there is an easy fix there?

MichaelRawson:
@quickbeam123 is probably the guy to ask, but we can probably copy the generated flags from the CMake build. Do you use the Makefile?

mezpusz:
> Do you use the Makefile?

I don't, but if it's there someone will use it. :smirk: I'm okay with deleting it in the long run too.

quickbeam123:
Given the other PR, maybe `-fsized-deallocation` could go from CMake too?

PR #596, opened 2024-09-05
AO_EQUAL is never used

MichaelRawson:
The argument-order value `AO_EQUAL` is never used. This means we can save a bit in `TermList`!


MichaelRawson:
Lies! It's cast from `Ordering`'s `Result`. Will push a comment.

PR #599, opened 2024-09-09
Refactor ordering comparators

mezpusz:
This PR fixes some annoying things in `OrderingComparator` and subclasses:
- `KBOComparator` uses info from `KBO::State` to duplicate less effort and code.
- `OrderingComparator` is created but not called by `Ordering` anymore, this allows more flexibility.
- `OrderingComparator` has now a default implementation.
- `LPOComparator` has been simplified and now it is created lazily on the fly, as eager creation of the entire tree was a bottleneck in some cases. This is not justified for `KBOComparator` yet, so I did not bother doing it now.


quickbeam123:
TPTP performance (`-i 2000`) seems OK:

Sort by UNS
7341 ['problemsSTD_master8465_lpo_i2000.pkl']
7344 ['problemsSTD_marton8471_lpo_i2000.pkl']
7510 ['problemsSTD_master8465_i2000.pkl']
7514 ['problemsSTD_marton8471_i2000.pkl']

(Of course, 4 problems are typically less than the variance of the used `lrs`, but is seems good enough.)

quickbeam123:
And here the `dpc` version of this:

Sort by UNS
7355 ['problemsSTD_master8465_lpo_dpc_i2000.pkl']
7360 ['problemsSTD_marton8471_lpo_dpc_i2000.pkl']
7509 ['problemsSTD_master8465_dpc_i2000.pkl']
7511 ['problemsSTD_marton8471_dpc_i2000.pkl']

Kernel/KBO.cpp
```diff
@@ -955,14 +879,9 @@ bool KBO::isGreater(AppliedTerm lhs, AppliedTerm rhs) const
   return isGreaterOrEq(lhs,rhs)==GREATER;
 }
 
-bool KBO::isGreater(TermList lhs, TermList rhs, const SubstApplicator* applicator, OrderingComparatorUP& comparator) const
+OrderingComparatorUP KBO::createComparator(TermList lhs, TermList rhs) const
```
MichaelRawson, line 882: This could be worth putting in the header file depending on how hot it is.

Kernel/KBO.hpp
```diff
@@ -186,6 +185,83 @@ class KBO
...
+   * (https://doi.org/10.1007/s10817-006-9031-4)
+   * @since 30/04/2008 flight Brussels-Tel Aviv
+   */
+  class State
+  {
+  public:
+    /** Initialise the state */
+    State(KBO* kbo)
+      : _kbo(*kbo)
+    {}
```
MichaelRawson, line 200: Would it make sense to call `init` from the constructor (or equivalent code)? As it is this innocent-looking aggregate may be uninitialised if you don't call `init` explicitly. But I don't have a strong view.

Kernel/KBOComparator.hpp
```diff
@@ -113,7 +111,7 @@ class KBOComparator
   private:
     uint64_t _content;
   };
-  const KBO& _kbo;
+  bool _ready;
```
MichaelRawson, line 114: `bool _ready = false;` and remove from constructor?

Kernel/KBOComparator.cpp
```diff
@@ -20,12 +20,20 @@ using namespace std;
 using namespace Lib;
 using namespace Shell;
 
-KBOComparator::KBOComparator(TermList tl1, TermList tl2, const KBO& kbo)
-  : _kbo(kbo), _instructions()
+KBOComparator::KBOComparator(TermList lhs, TermList rhs, const KBO& kbo)
+  : OrderingComparator(lhs, rhs, kbo), _ready(false), _instructions()
```
MichaelRawson, line 24: `instructions()` pointless I think.

Kernel/KBOComparator.cpp
```diff
@@ -81,26 +89,34 @@ KBOComparator::KBOComparator(TermList tl1, TermList tl2, const KBO& kbo)
...
     // weight and variable balances first
-    DHMap<unsigned,int> vars;
-    int w = 0;
-    countSymbols(kbo, vars, w, lhs, 1);
-    countSymbols(kbo, vars, w, rhs, -1);
 
     // we only care about the non-zero weights and counts
     bool varInbalance = false;
-    DHMap<unsigned,int>::Iterator vit(vars);
+    auto state = kbo._state;
```
MichaelRawson, line 95: Given the following shenanigans maybe we should add a static `std::is_trivially_copyable` assertion to `KBO::State`.

Kernel/KBOComparator.cpp
```diff
@@ -154,8 +170,15 @@ KBOComparator::KBOComparator(TermList tl1, TermList tl2, const KBO& kbo)
   }
 }
 
-bool KBOComparator::check(const SubstApplicator* applicator) const
+bool KBOComparator::check(const SubstApplicator* applicator)
 {
+  if (!_ready) {
```
MichaelRawson, line 175: Why do we do this lazy-loading thing instead of calling it in the constructor? Is it sometimes unused?

Kernel/LPOComparator.cpp
```diff
@@ -52,438 +48,193 @@ bool unify(TermList tl1, TermList tl2, TermList& orig1, TermList& orig2)
...
+      }
+      default:
+        ASSERTION_VIOLATION;
       }
-      break;
     }
-    default:
-      ASSERTION_VIOLATION;
+loop_end:
+    node->release();
```
MichaelRawson, line 219: This looks RAIIish (i.e. should be a smart pointer of some kind)...is it, or did I miss the path through the loop where you don't release the `node`?

Kernel/LPOComparator.hpp
```diff
@@ -20,101 +20,91 @@ namespace Kernel {
...
+      return *this;
+    }
+  };
+
+  struct Node {
+    Node(TermList lhs, TermList rhs)
+      : refcnt(0), lhs(lhs), rhs(rhs), eqBranch(BranchTag::T_NOT_GREATER), gtBranch(BranchTag::T_GREATER), incBranch(BranchTag::T_NOT_GREATER) {}
+
+    void acquire() {
+      refcnt++;
```
MichaelRawson, line 68: This looks suspiciously like `std::shared_ptr`.

MichaelRawson (approved):
Nice. I left some comments but they're more suggestions, rather than "fix this". When you have addressed as many as you are willing to (which could be 0) let me know and I'll merge.

Kernel/KBO.hpp
mezpusz: Otherwise some parts could remain non-zero?

Kernel/KBOComparator.cpp
mezpusz: Note that `kbo._state` is a pointer, so hopefully nothing is copied?

Kernel/LPOComparator.cpp
mezpusz: We could actually just put it into a `Branch` (or whatever the RAII container ends up being, e.g. `shared_ptr`). The point here is that the original container is disrupted so `node` could be deleted while we're processing it.

Kernel/KBOComparator.cpp
mezpusz: Yes, it might be never used. For example, we create the object when inserting the demodulator into the LHS index, but only initialize it when we try to demodulate with it. Of course, the initialization itself could be also done lazily, i.e. only processing the comparator to the point we actually use it, I just couldn't get around to it this time and it didn't seem like a too hot spot.

Kernel/KBO.cpp
mezpusz: I tried doing it but some incomprehensible error came up about some cyclic inclusion. I don't think it's that hot though.

Kernel/KBOComparator.cpp
MichaelRawson: Ah, right - so maybe we should make it immovable or something. :thinking:  But not important!

Kernel/LPOComparator.hpp
```diff
@@ -39,43 +39,24 @@ class LPOComparator
...
 
-  struct Node;
+  class Node;
 
   struct Branch {
     BranchTag tag;
-    Node* n;
+    std::shared_ptr<Node> n;
 
     explicit Branch(BranchTag t) : tag(t), n(nullptr) { ASS(t==BranchTag::T_GREATER || t==BranchTag::T_NOT_GREATER); }
```
MichaelRawson, line 48: I think `n(nullptr)` is now the default.

Kernel/LPOComparator.hpp
```diff
@@ -39,43 +39,24 @@ class LPOComparator
     T_UNKNOWN,
   };
 
-  struct Node;
+  class Node;
 
   struct Branch {
     BranchTag tag;
-    Node* n;
+    std::shared_ptr<Node> n;
```
MichaelRawson, line 46: I only said it looked like a `shared_ptr`! Did you check for overhead? I think `shared_ptr<T>` has both a strong and weak count and maybe has to use `size_t` for counts rather than `unsigned`. If the overhead is unacceptable we could also look at `Lib::SmartPtr`.

But the code is much nicer. :innocent:

Kernel/KBO.hpp
MichaelRawson: Right. But might be worth it depending on the usage pattern.

Kernel/LPOComparator.hpp
mezpusz: I agree that it's much nicer! I assume there is just a little bit of overhead during `extend` calls. Maybe @quickbeam123 can do one more test with lpo? 🙂

Kernel/LPOComparator.hpp
mezpusz: I would leave it here to be explicit about storing null in this case.

Kernel/LPOComparator.hpp
MichaelRawson: Perhaps this is the usual [`std::make_shared` footgun](https://stackoverflow.com/questions/20895648/difference-in-make-shared-and-normal-shared-ptr-in-c?rq=3) biting us?

mezpusz: Maybe I was a bit too fast to change to `std::shared_ptr`. I can try the `Lib::SharedPtr` variant.

mezpusz: Sorry for doing the lazy thing. I've just changed it to `Lib::SmartPtr`, but only did some local tests which show no difference in performance. @quickbeam123 can we try again with this version?

MichaelRawson: Suspicion: both `std::shared_ptr` and `Lib::SmartPtr` have to allocate an additional control block somewhere, whereas you did this intrusively originally. If you can be bothered, `std::make_shared` ought to remove that overhead, but if not, the original was (is) fine by me, it just looked smartpointery.

quickbeam123:
```
7344 ['problemsSTD_marton8472_lpo_dpc_i2000_2.pkl']
7352 ['problemsSTD_marton8472_lpo_dpc_i2000.pkl']
7354 ['problemsSTD_marton8473_lpo_dpc_i2000_2.pkl']
7355 ['problemsSTD_master8465_lpo_dpc_i2000.pkl']
7357 ['problemsSTD_marton8471_lpo_dpc_i2000_2.pkl']
7360 ['problemsSTD_marton8471_lpo_dpc_i2000.pkl']
7362 ['problemsSTD_marton8473_lpo_dpc_i2000.pkl']
```

intrude ~ 7358.5
Smart ~ 7358.0
shared ~ 7348.0

So I'd say let's keep Smart and that's it.

mezpusz: @quickbeam123 Thanks for the comparison!

@MichaelRawson sorry for not taking the advice! I think we could have a `SmartPtr` removal PR at some point if this is really the case. Otherwise, I would vote for doing something "smart" as the original code looked terrible and it looks like it doesn't hurt much.

MichaelRawson: I'm happy! Glad the nice thing worked out somehow. :smile:

PR #600, opened 2024-09-09
Fix conditional redundancy handler literal constraints with encompassment demodulation incompleteness issue

mezpusz:
We assume in `DemodulationHelper` that the rewriting clause is unit, which resulted in false redundancies inside `ConditionalRedundancyHandler` where this assumption is not true with `crlc=on`.

Instead of overcomplicating original code and potentially slowing down demodulation, I duplicated the relevant part in `ConditionalRedundancyHandler`.

Tests indicate no further incompleteness so far, efficiency is comparable to previous version.

### DIFF
--- a/Shell/ConditionalRedundancyHandler.cpp
+++ b/Shell/ConditionalRedundancyHandler.cpp
@@ -18,6 +18,8 @@
 #include "Indexing/CodeTreeInterfaces.hpp"
 #include "Indexing/ResultSubstitution.hpp"
 
+#include "Inferences/DemodulationHelper.hpp"
+
 #include "Statistics.hpp"
 
 using namespace std;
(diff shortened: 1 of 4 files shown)


Shell/ConditionalRedundancyHandler.cpp
```diff
@@ -606,4 +605,43 @@ bool ConditionalRedundancyHandlerImpl<enabled, ordC, avatarC, litC>::handleReduc
...
 }
 
+/**
+ * This function is similar to @b DemodulationHelper::isPremiseRedundant.
+ * However, here we do not assume that the rewriting equation is unit, which necessitates some additional checks.
+ */
+template<bool enabled, bool ordC, bool avatarC, bool litC>
+bool ConditionalRedundancyHandlerImpl<enabled, ordC, avatarC, litC>::isSuperpositionPremiseRedundant(
+  Clause* rwCl, Literal* rwLit, TermList rwTerm, TermList tgtTerm, Clause* eqCl, TermList eqLHS,
+  const SubstApplicator* eqApplicator, Ordering::Result& tord) const
```
MichaelRawson, line 615: This seems like it should produce linker errors. Why doesn't it?

MichaelRawson (approved):
Cool! Nice spot.

Shell/ConditionalRedundancyHandler.cpp
mezpusz: What do you mean? :smile:

MichaelRawson: So there's this template function at the end of the file with 2^4 possible instances (I think), but it doesn't seem to be explicitly instantiated. I think I figured it out: you call the instances you need around line 303, right?

mezpusz: Yep, on line 442 in particular. I'm guessing the explicit instantiations in the factory function above causes a cascade of instantiations of everything needed. 🪄

PR #601, opened 2024-09-15
Calling SharedPtr ctor with nullptr causes violation; call default ctor

mezpusz:


mezpusz:
Sorry @MichaelRawson, I should have listened!

MichaelRawson:
I'm equally culpable! Thinking about it you might get a performance win here too as you won't allocate a control block for `nullptr`.

mezpusz:
You're right, this should be superior in all possible ways! Thanks for the fast merge!

PR #605, opened 2024-09-18
Code tree subsumption

mezpusz:
Revival of code tree subsumption. The following fixes were necessary:
- Flat term size calculation was off for polymorphic equalities.
- `LIT_END` was not set properly in the code tree for `ILStruct` objects.
- Added equality type uniformly at the start of all equalities in `FlatTerm` and `CodeStack` to avoid unsoundness in the two-variable equality case. Some refactors were necessary to do this cleanly.

Debug tests were run, so far no assrtion violations. Correctness checks have been performed with the current SAT subsumption, showing that there is (i) no unsoundness, but (ii) code tree subsumption cannot detect set-based subsumption resolution, only multiset-based.

The final experimental data:

|  | master | regression | codetree |
|--------|--------|--------|--------|
| discount | 9981 (47) | 9981 (47) | 10150 (216) |
| otter | 9720 (42) | 9720 (43) | 9967 (289) | 

In parentheses the number of uniques are shown of master to codetree, regression to codetree and codetree to master, respectively.

### DIFF
--- a/Indexing/CodeTree.cpp
+++ b/Indexing/CodeTree.cpp
@@ -303,6 +303,8 @@ CodeTree::CodeOp CodeTree::CodeOp::getLitEnd(ILStruct* ils)
   CodeOp res;
   res.setAlternative(0);
   res._setData(ils);
+  // order matters
+  res._setInstruction(LIT_END);
   ASS(res.isLitEnd());
   return res;
 }
--- a/Indexing/CodeTree.cpp
+++ b/Indexing/CodeTree.cpp
@@ -683,6 +683,8 @@ void CodeTree::Compiler<forLits>::updateCodeTree(CodeTree* tree)
 template<bool forLits>
 void CodeTree::Compiler<forLits>::handleTerm(const Term* trm)
 {
+  ASS(!forLits || trm->isLiteral());
+
   static Stack<unsigned> globalCounterparts;
   globalCounterparts.reset();
 
--- a/Indexing/CodeTree.cpp
+++ b/Indexing/CodeTree.cpp
@@ -657,25 +657,20 @@ std::ostream& operator<<(std::ostream& out, const CodeTree& ct)
 
 //////////////// insertion ////////////////////
 
-void CodeTree::CompileContext::init()
-{
-  nextVarNum=0;
-  varMap.reset();
-  nextGlobalVarNum=0;
-  globalVarMap.reset();
-}
+template<bool forLits>
+CodeTree::Compiler<forLits>::Compiler(CodeStack& code) : code(code), nextVarNum(0), nextGlobalVarNum(0) {}
 
-void CodeTree::CompileContext::nextLit()
+template<bool forLits>
+void CodeTree::Compiler<forLits>::nextLit()
 {
-  nextVarNum=0;
+  ASS(forLits);
+  nextVarNum = 0;
   varMap.reset();
 }
 
-void CodeTree::CompileContext::deinit(CodeTree* tree, bool discarded)
+template<bool forLits>
+void CodeTree::Compiler<forLits>::updateCodeTree(CodeTree* tree)
 {
-  if(discarded) {
-    return;
-  }
   //update the max. number of variables, if necessary
   if(nextGlobalVarNum>tree->_maxVarCnt) {
     tree->_maxVarCnt=nextGlobalVarNum;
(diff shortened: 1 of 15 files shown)


easychair:
> Debug tests were run, so far no assrtion violations. Correctness checks have been performed with the current SAT subsumption, showing that there is (i) no unsoundness, but (ii) code tree subsumption cannot detect set-based subsumption resolution, only multiset-based.

Surely it can - we did it ages ago. The only difference is the
interpretation of the instruction in charge of setting the literal in
the query clause. For multisets, you iterate through previously
untried literals. For sets, you iterate through all literals.

A

Shell/ConditionalRedundancyHandler.cpp
```diff
@@ -106,28 +106,17 @@ class ConditionalRedundancyHandler::ConstraintIndex
...
-          code.push(CodeOp::getTermOp(ASSIGN_VAR, *varNumPtr));
-        }	else {
-          code.push(CodeOp::getTermOp(CHECK_VAR, *varNumPtr));
-        }
-      }
-      else {
-        ASS(t.isTerm());
-        compileTerm(t.term(), code, cctx, false);
+        compiler.handleVar(t.var());
+        continue;
```
quickbeam123, line 114: Did some of the now missing complexity get hidden elsewhere?

Indexing/CodeTree.hpp
```diff
@@ -124,11 +124,8 @@ class CodeTree
     unsigned depth;
     ILStruct* previous;
 
-    unsigned isVarEqLit:1;
     unsigned varCnt:31;
```
quickbeam123, line 127: Maybe we can drop the 31 now?

mezpusz:
> Surely it can - we did it ages ago. The only difference is the interpretation of the instruction in charge of setting the literal in the query clause. For multisets, you iterate through previously untried literals. For sets, you iterate through all literals. A

Thanks, it's good to know that it should work. Then perhaps a subsequent PR could fix this, I'm not sure at this point where this should happen in the current code.

mezpusz:
> This looks awesome. Thank you, @mezpusz .
> 
> Even before checking the code change in full, I suggest we make `-cts on` the default.

I agree, and big thanks to whoever implemented this in the first place. 🙂 Additionally, since the SAT subsumption cross-check on top is relatively cheap after we have found an instance, maybe we could add that as a permanent debug check to make sure no bugs creep in.

Shell/ConditionalRedundancyHandler.cpp
mezpusz: Yes, the "code compilation" is now delegated to a `CodeTree::Compiler` instance.

easychair:
Just to make it clear: set subsumption is super incomplete. Set subsumption
can be used for subsumption resolution because it always resolves away a
literal, so it creates a strictly smaller clause.

A

Shell/ConditionalRedundancyHandler.cpp
quickbeam123: Great!

MichaelRawson:
In #265 I removed a dead file called `CTFwSubsAndRes` which looks a bit like your `CodeTreeForwardSubsumptionAndResolution`. I think your new shiny code is probably better but might be worth a look to see if you forgot anything.

Indexing/CodeTree.hpp
```diff
@@ -209,11 +206,11 @@ class CodeTree
     inline ILStruct* getILS()
     {
       ASS(isLitEnd());
-      return _data<ILStruct>();
+      return reinterpret_cast<ILStruct*>(reinterpret_cast<uint64_t>(_data<ILStruct>()) & ~LIT_END);
```
MichaelRawson, line 209: Could this be `uintptr_t` please?

Indexing/CodeTree.cpp
```diff
@@ -666,25 +657,20 @@ std::ostream& operator<<(std::ostream& out, const CodeTree& ct)
...
-}
+template<bool forLits>
+CodeTree::Compiler<forLits>::Compiler(CodeStack& code) : code(code), nextVarNum(0), nextGlobalVarNum(0) {}
 
-void CodeTree::CompileContext::nextLit()
+template<bool forLits>
+void CodeTree::Compiler<forLits>::nextLit()
 {
-  nextVarNum=0;
+  ASS(forLits);
```
MichaelRawson, line 666: Is it worth templating just some of the methods in `Compiler` rather than the whole class to avoid this weirdness? I'm not sure, just wondering.

MichaelRawson (approved):

mezpusz:
> In #265 I removed a dead file called `CTFwSubsAndRes` which looks a bit like your `CodeTreeForwardSubsumptionAndResolution`. I think your new shiny code is probably better but might be worth a look to see if you forgot anything.

Thanks, that's a good suggestion! I have checked the code, fortunately it looks very similar, apart from (i) setting the age of the result for some reason and (ii) doing something with `Clause::_auxData`. I believe none of these are necessary anymore, but let me know what you think.

MichaelRawson:
Agreed. Whatever `auxData` is I'm happy it's dead, but maybe the age needs to be set: @quickbeam123 is the resident expert, I think.

Indexing/CodeTree.hpp
mezpusz: The `uint64_t` comes from `_content` as usual. But if you say it makes more sense, I obey.

Indexing/CodeTree.cpp
mezpusz: You mean the entire `CodeTree`? Well, there is the `bool _clauseCodeTree` member which I want to convert to a template flag in the future, and that would render templating `Compiler` unnecessary. I can try to do this now, not sure though how much would be changed.

Indexing/CodeTree.hpp
mezpusz: Do you think setting `DATA_BITS_START` [here](https://github.com/vprover/vampire/blob/master/Indexing/CodeTree.hpp#L236) to `INSTRUCTION_BITS_END` would work? Then, we would need to check somehow that whatever `Type` we're casting from and to satisfy `alignof(Type)>SEARCH_STRUCT` or something...

Indexing/CodeTree.cpp
MichaelRawson: Just the `Compiler` - but I tried and it sucked so forget I said anything!

Indexing/CodeTree.hpp
MichaelRawson: You could, but I'm not sure why you would: that would mean that you'd have two separate pieces of data that do not share the same area of memory. Maybe I misunderstood you?

MichaelRawson: By the way, the `getBits`/`setBits` calls [here](https://github.com/vprover/vampire/blob/master/Indexing/CodeTree.hpp#L246-L249) look useless to me, we can just read/assign `_content` directly in this case. We only need the bit-twiddling if you want to set some partial area of the bitfield.

mezpusz: Yes, we would use a separate 61 bits for our data pointers but we can only do this if the data is align in such a way (which we btw don't check for all types that are used as data here). This is unproblematic for `SUCCESS` which is 0, but not so for `LIT_END` which is 2, hence the above bit magic.

mezpusz: > By the way, the `getBits`/`setBits` calls [here](https://github.com/vprover/vampire/blob/master/Indexing/CodeTree.hpp#L246-L249) look useless to me [...]

I agree, but similarly [here](https://github.com/vprover/vampire/blob/master/Kernel/Term.hpp#L355) it is useless but only so because `TermTag::REF` is 0. Not saying that we should remove this though. Anyways, I can make some changes to make this more consistent for `CodeTree`.

Indexing/CodeTree.cpp
mezpusz: I'm sorry, what did you try? 😃

MichaelRawson: Tried removing the template parameter from the `Compiler` class and inserting it for the methods that need it, but if anything it's more error-prone.

PR #607, opened 2024-09-20
proof-extra with data rather than strings

MichaelRawson:
We have started to receive interest in more proof outputs (e.g. Dedukti @anjapetkovic, Metamath @digama0) and these require a greater degree of granularity in proofs as they are machine-checked.

We already have `--proof_extra`, but it's stringly-typed and hard to use for this purpose. Make explicit objects recording information about the proof for downstream use.

While we're at it:
1. Include resolution, demodulation, and equality resolution in the extra'd inferences. Not (equality) factoring as they're painful to adapt at the moment - more later.
2. Only on/off "proof extra" flag, I couldn't find much use for the halfway `FREE` option. Happy to be persuaded.
3. Change the format and data stored slightly. In my view this makes more sense.

### DIFF
--- a/Kernel/Clause.cpp
+++ b/Kernel/Clause.cpp
@@ -219,7 +219,7 @@ void Clause::destroy()
   static Stack<Clause*> toDestroy(32);
   Clause* cl = this;
   for(;;) {
-    if (env.options->proofExtra()) {
+    if (env.options->proofExtra() == Options::ProofExtra::FULL) {
       env.proofExtra.remove(cl);
     }
     Inference::Iterator it = cl->_inference.iterator();
@@ -383,7 +383,7 @@ std::string Clause::toString() const
   // print inference and ids of parent clauses
   result += " " + inferenceAsString();
 
-  if(env.options->proofExtra()){
+  if(env.options->proofExtra() != Options::ProofExtra::OFF){
     // print statistics: each entry should have the form key:value
     result += std::string(" {");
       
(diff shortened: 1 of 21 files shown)


digama0:
This will conflict with some of the changes I've made, but that's my fault for sitting on the s-expr PR for too long. I'll rebase it on top of this, I don't think the version of InferenceExtra I implemented is too different from this.

Kernel/Clause.cpp
```diff
@@ -383,7 +383,7 @@ std::string Clause::toString() const
   // print inference and ids of parent clauses
   result += " " + inferenceAsString();
 
-  if(env.options->proofExtra()!=Options::ProofExtra::OFF){
```
quickbeam123, line 386: This is a place I was using `-proof_extra free` for. To get more detailed clause output without having to pay for it.

MichaelRawson: Fair enough - if you use it I'll put it back.

MichaelRawson:
@digama0 - I was quite pleased with this design until today, when we tried to use it and were told we can't `static_cast` from a `InferenceExtra *` to a derived class for annoying technical reasons with virtual inheritance.

@anjapetkovic and I will likely fix it together ("composition over inheritance" is the way to go, I expect) in the next few days so we can make progress on Dedukti. Let us know if you need something.

digama0:
I'll try to make a PR with my version of this thing. Sorry for the delay, my time for this is more limited than it was a few weeks ago.

PR #608, opened 2024-09-20
SAT subsumption debug check for code trees

mezpusz:
Random strategy tested up to 10 seconds over the entire TPTP.


RobCoutel:
It seems ok for me. I made the documentation slightly more accurate for the new SR function and removed an unnecessary condition.

RobCoutel (approved):
All good to me

PR #610, opened 2024-09-23
Add `-DREV_COUNT=OFF` build option

kazarmy:
This PR adds a `-DREV_COUNT=OFF` build option to prevent the addition of the Git revision count to the Vampire binary name. That solves 2 problems with the current setup that attaches a revision count:

1. I cannot do a `vampire_latest` symbolic link because the Vampire binary name keeps changing with every commit.
2. The Vampire binaries start accumulating in the `bin/` directory. At 14M per binary, they are not exactly small.

This PR solves the above problems without the hassle of a post-build script. If I wanted to know what commit a binary is at, apparently I can just use `--version`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -913,6 +913,7 @@ endif() # COMPILE_TESTS
 # automated generation of Vampire revision information from git #
 #################################################################
 set(VAMPIRE_VERSION_NUMBER 4.9)
+option(REV_COUNT "Add Git revision count to binary name" ON)
 
 execute_process(
     COMMAND git rev-parse --is-inside-work-tree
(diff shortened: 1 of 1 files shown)


MichaelRawson:
I am also irritated by the accumulating Vampires! Thanks.

MichaelRawson:
BTW, this reminded me of a discussion we had internally ages ago - it turns out most of the CMake users agreed that they don't need this, so on master I've removed the feature altogether. :smile:

kazarmy:
Great! I agree. :smile:

TIL that a command of the following form `cp build/bin/vampire_* vampire` can be useful, if it can be guaranteed that there is only 1 binary.

kazarmy:
Also, the commit hash shown by `--version` is only guaranteed to update if one does `cmake ..` before `make`, but that probably should be SOP.

MichaelRawson:
Hm. That's a good point - we're probably Doing it Wrong™ there and should rebuild some meta-target thing every time the commit hash changes.

PR #614, opened 2024-10-01
add missing includes

anjapetkovic:
These were required to make it compile on my machine: mac, arm64, OS Sonoma, Apple clang version 13.1.6 (clang-1316.0.21.2).

Include these, pretty please with a cherry on top?


MichaelRawson (approved):
I can confirm that Anja's toolchain rejects these without the includes - probably my system gets them from system headers including each other. Unlikely to break anything IMO.

PR #615, opened 2024-10-04
Polymorphism + Arithmetic

joe-hauns:
Fixes compatibility issue of polymorphism and arithmetic operations.

The problem was that in polynomial normalization the type argument information was not carried around.
Also updates BottomUpEvaluation to allow for optionally ignore or not ignore type arguments in the bottom-up traversal.

### DIFF
--- a/Kernel/BottomUpEvaluation.hpp
+++ b/Kernel/BottomUpEvaluation.hpp
@@ -11,7 +11,6 @@
 #ifndef __LIB__BOTTOM_UP_EVALUATION_HPP__
 #define __LIB__BOTTOM_UP_EVALUATION_HPP__
 
-#define DEBUG(...) // DBG(__VA_ARGS__)
 
 /**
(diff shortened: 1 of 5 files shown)


Kernel/BottomUpEvaluation.hpp
```diff
@@ -368,17 +370,23 @@ class BottomUpEvaluation {
...
 
 }
+#undef DEBUG
 
 #include "Kernel/Term.hpp"
 
 namespace Lib {
+
+struct TermListContext {
+  bool ignoreTypeArgs = true;
```
MichaelRawson, line 387: I'm OK with this solution but I'm not sure I see where this is ever false - everything seems to use the default constructor.

MichaelRawson (approved):

Kernel/BottomUpEvaluation.hpp
MichaelRawson: Ahah, but you want to give people the option, right.

PR #616, opened 2024-10-21
Implement Geoff's idea to have an option stating the users "intent" for invoking Vampire (i.e., proving / disproving)

quickbeam123:
1) An easy update to use `intent` for creating the effect of `--mode casc_sat` (we can now use `--intent sat --mode casc` instead).
2) Update the samplers and, in particular, merge samplerFOF and samplerFNT (and delete samplerFNT).


MichaelRawson (approved):
Nice. This makes the options much more orthogonal: in future, does it make sense to have e.g. `--mode smtcomp --intent sat`?

quickbeam123:
Before we revert, let me just explain:
- as of now, the intent is only used in `casc mode`, to pick the right schedule. The fact that it could simplify the code (`casc_sat` was simply a copy-paste of `casc` with a different schedule choice) points in the direction that `intent` might be a useful concept (of course, the code could have been written differently, but still)
- in vampire mode (where it currently serves no role), we could make good use of `intent` to explicitly mark option-value combinations which lead to the loss of completeness: e.g. "You are trying to show SAT by finitely saturating the input, this might not be possible using sa=lrs, consider trying sa=otter instead."

> Why is (say) -- mode vampire --intent sat better --mode sat ?
We don't have mode sat at the moment. 

> When we are looking for a model, is it still the Vampire mode? If yes, then why?
This depends on the definition of Vampire mode. To be honest, I never heard one. For me it's just the default (universal?) mode of operation which does not use strategy schedules.

PR #618, opened 2024-10-26
Configure version.cpp at build time if git repo

kazarmy:
As per https://github.com/vprover/vampire/pull/610#issuecomment-2371012934, this PR makes the `version.cpp.in` transformation happen at build time rather than configure time if the Vampire source folder is from git (i.e. there is a proper `.git` subdirectory), thus ensuring that the git version number in `vampire ‑‑version` is always updated even if one forgets to do `cmake ..` before `make`.

I think I have tested this comprehensively, but there might be a sequence of:

1. moving the git HEAD commit; and/or
2. `cmake ..`; and/or
3. `make`

that results in unexpected behavior. Please inform me if that's the case!

References:
1. https://cmake.org/pipermail/cmake/2012-May/050227.html for the `cmake -P` solution.
2. https://stackoverflow.com/a/43206544 for the `add_custom_target` solution.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -916,20 +916,24 @@ execute_process(
     )
 
 if (GIT_IS_REPOSITORY STREQUAL true)
-  # GIT_COMMIT_DESCRIPTION used by version.cpp.in
-  execute_process(
-    COMMAND git log -1 --format=%h\ on\ %ci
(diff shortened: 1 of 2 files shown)


MichaelRawson (approved):
Nice. Thank you for figuring this out and documenting your findings, this is a solid solution.

PR #619, opened 2024-10-30
add (equality) factoring `proof_extra` information, simplify factoring

MichaelRawson:
Follow-up to #607, cherry-picked from the Dedukti efforts:
- add information for factoring and equality factoring
- while we're at it, somewhat simplify the factoring implementation
- also fix the proof_extra interfaces to use composition over inheritance so that they're usable (oops)

### DIFF
--- a/Inferences/Factoring.cpp
+++ b/Inferences/Factoring.cpp
@@ -14,15 +14,11 @@
 
 #include <utility>
 
-#include "Lib/Comparison.hpp"
 #include "Lib/Environment.hpp"
-#include "Lib/Int.hpp"
 #include "Lib/Metaiterators.hpp"
-#include "Lib/PairUtils.hpp"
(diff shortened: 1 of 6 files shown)


Inferences/Factoring.cpp
```diff
@@ -38,89 +34,55 @@
...
 
-    return pvi( pushPairIntoRightIterator(l2, unifs) );
-  }
-private:
-  Clause* _cl;
-  LiteralSelector& _sel;
-  RobSubstitutionSP _subst;
-};
+    RobSubstitution subst;
+    if(!subst.unify(TermList(l1), 0, TermList(l2), 0))
```
MichaelRawson, line 68: We know that `l1` and `l2` aren't equations, so there can only be one unifier and therefore no need to call `unifiers`.

PR #625, opened 2024-12-10
Added compiler-independed ALLOW_UNUSED

joe-hauns:


Debug/Assertion.hpp
```diff
@@ -343,7 +343,26 @@ void Debug::Assertion::violatedMethod(const char* file, int line, const T& obj,
 #endif // VDEBUG
 
 /** expression version of ASSERTION_VIOLATION */
-template<class T> T assertionViolation() 
+template<class T> T assertionViolation()
 { ASSERTION_VIOLATION }
 
+#if defined(__clang__)
+#  define __ALLOW_UNUSED(...)                                                             \
```
MichaelRawson, line 350: I think technically you're not supposed to use underscores for identifiers as they're reserved. But I guess that other code already uses this extensively so let's keep it for now, everyone does it.

MichaelRawson (approved):

PR #627, opened 2024-12-10
Lib Additions

joe-hauns:
Various helper functions, and some refactorings in various `Lib/` files. 
Every commit mentions the purpose of the respecitvely changed file.


Lib/Coproduct.hpp
```diff
@@ -33,12 +33,24 @@ namespace Lib {
 
 namespace TL = TypeList;
 
-template<unsigned v>
-struct Constant { static constexpr unsigned value = v; };
-
 template <class... As>
 class Coproduct;
 
+template<class Fs, class Ts>
```
MichaelRawson, line 39: Could these have a one-line comment explaining what they do?

Lib/Coproduct.hpp
```diff
@@ -637,6 +680,35 @@ template<class... Ts> struct std::hash<Lib::Coproduct<Ts...>>
...
+
+template<class Cond, class Cons, class... Rest>
+struct SelectOutput<Cond, Cons, Rest...> {
+  using type = TypeList::Into<Coproduct, 
+     TypeList::Concat< TypeList::List<std::result_of_t<Cons()>>
+                     , typename SelectOutput<Rest...>::type::Ts 
+                     >>;
+};
+
+// static_assert(std::is_same<sel)
```
MichaelRawson, line 695: forgotten static assertion?

Lib/DArray.hpp
```diff
@@ -38,9 +38,6 @@ namespace Lib {
 template<typename C>
 class DArray
 {
-private:
```
MichaelRawson, line 41: This looks suspicious - what's going on here? I'm OK with an explicitly-provided `operator=(const DArray &)` but none at all seems suspicious as the compiler-provided one will likely be wrong.

Lib/Metaiterators.hpp
```diff
@@ -152,6 +153,9 @@ class FilteredIterator
     return out;
   }
 
+  bool knowsSize() const { return false; }
+  bool size() const { ASSERTION_VIOLATION }
```
MichaelRawson, line 157: `size_t size()`? Not that it matters.

Lib/Metaiterators.hpp
```diff
@@ -476,11 +481,49 @@ class MappingIterator
    * the @b size() function, and if the @b knowsSize() function returns true.
    */
   inline size_t size() const { return _inner.size(); }
+
+  auto reverse() &&
```
MichaelRawson, line 485: Struggling to parse this. Equivalent to `auto &&reverse() {...}`?

Lib/Metaiterators.hpp
```diff
@@ -1656,7 +1775,15 @@ template<class Array> auto arrayIter(Array const& a) { return arrayIter(
...
 template<class Array> auto arrayIter(Array      & a) { return arrayIter(          a , a.size()); }
 
-}
+template<class Item, class... Items>
+auto iterItems(Item item, Items... items) 
+{ return arrayIter(Stack<Item>{std::move(item), std::move(items)...}); }
+
+template<class Item>
+auto iterItems()
+{ return arrayIter(Stack<Item>{}); }
```
MichaelRawson, line 1784: The empty iterator?

Lib/Recycled.hpp
```diff
@@ -211,8 +211,8 @@ class Recycled
   T* operator->() { return &self(); }
 
   auto size() const { return self().size(); }
-  template<class Idx> auto operator[](Idx idx) const { return self()[idx]; }
-  template<class Idx> auto operator[](Idx idx)       { return self()[idx]; }
+  template<class Idx> decltype(auto) operator[](Idx idx) const { return self()[idx]; }
+  template<class Idx> decltype(auto) operator[](Idx idx)       { return self()[idx]; }
```
MichaelRawson, line 215: `decltype(auto)` rather than `auto` here because you want to allow `T`, `T &`, `T &&` and cv-qualified friends?

Lib/STL.hpp
```diff
@@ -0,0 +1,32 @@
...
+namespace Lib {
+
+template<class T>
+T* move_to_heap(T t) { return new T(std::move(t)); }
+                
+template< typename T, 
+          std::enable_if_t<!std::is_pointer<T>::value, bool> = true  
+        >
+std::shared_ptr<T> make_shared(T t)
+{ return std::shared_ptr<T>(move_to_heap(std::move(t))); }
```
MichaelRawson, line 28: How does this differ from `std::make_shared`?

Lib/Stack.hpp
```diff
@@ -371,12 +371,12 @@ class Stack
...
-  { std::sort(begin(), end(), less); }
+  { std::sort(begin(), end(), std::move(less)); }
 
   /** like Stack::sort but moves the content out of `this` and returns the resulting Stack instead of changing the contents of this */
-  template<class Equal = std::equal_to<C>>
-  Stack sorted(Equal eq = std::equal_to<C>{})
-  { sort(); return std::move(*this); }
+  template<class Less = std::less<C>>
+  Stack sorted(Less less = std::less<C>{})
+  { sort(std::move(less)); return std::move(*this); }
```
MichaelRawson, line 379: Weird. Did this ever work?

Lib/Stack.hpp
```diff
@@ -848,6 +848,7 @@ class Stack
 
 
   friend std::ostream& operator<<(std::ostream& out, const Stack<C>& s) {
+    using namespace Kernel;
```
MichaelRawson, line 851: ?

Shell/UIHelper.cpp
```diff
@@ -115,7 +115,7 @@ bool szsOutputMode() {
   return (Lib::env.options && Lib::env.options->outputMode() == Shell::Options::Output::SZS);
 }
 
-ostream& addCommentSignForSZS(ostream& out)
+ostream& addCommentSignForSZS(std::ostream& out)
```
MichaelRawson, line 118: `std::ostream` in return type too please!

MichaelRawson (changes requested):
Really nice. I seem to have diff from #624 as well for some reason, I assume this is just GitHub being weird. Some comments but it's minor stuff or just for my understanding.

Lib/Coproduct.hpp
joe-hauns: it does "dead code" :)
Removed it!

Lib/Metaiterators.hpp
joe-hauns: No, it is a function that expects that this is a move reference 
Think of
```
auto reverse() && {}
auto reverse() const {}
auto reverse() {}
```
It's best practice to use this in a function that moves out of *this/invalidates it, which means that the user needs to explicitly `std::move` and there is no hidden moving data out of somewhere.

Lib/Metaiterators.hpp
joe-hauns: Yes. it's only there so we can use `iterItems<Type>(as...)` with an empty list `as` (as well as a non-empty one ofc).

Lib/Recycled.hpp
joe-hauns: Yes exactly. For example in a `Recycled<Stack<int>>` we want to be able to get an `int&` when we index into it, and not just an `int`.

Lib/STL.hpp
joe-hauns: Only in that we don't have to specify the type argument `T` as it's always equal to the function argument. That's quite helpful to not having to repeat long iterator types and similar things.

Lib/Stack.hpp
joe-hauns: No because I always put an explicit argument there :)

Lib/Stack.hpp
joe-hauns: This slipped though.. had been needed before refactoring `Lib/Output.hpp` as we needed to explicitly import `Kernel::operator<<`... basically compilcated and uninteresting namespacing issues. I just removed it agin :)

Shell/UIHelper.cpp
joe-hauns: hunted for some more of these.

Lib/STL.hpp
MichaelRawson: Ahah - it's some special case of the `Args...` constructor, got it.

Lib/DArray.hpp
joe-hauns: I was thinking one is implicitly deleted anyways as we have `DArray& operator=(DArray&& other)...` 
But as we have an implicity copy constructor we still get the copy assignment (before and after the PR).. I fixed this now :)

MichaelRawson: Ahah, I missed the copy constructor later. OK, sounds good - looks like you even managed to eliminate some copies in FMB, nice.

MichaelRawson (approved):

PR #628, opened 2024-12-10
Kernel Additions

joe-hauns:
Miscellaneous helper functions and tidies in `Kernel/`.
Biggest change is updating `NumTraits` and adding a compile time flag `NICE_THEORY_OUTPUT` that prints + as infix operator and so on. The former is needed for the latter (and will be merged with upcoming alsca stuff anyways).


Kernel/EqHelper.hpp
```diff
@@ -49,6 +49,11 @@ class EqHelper
 
   //WARNING, this function cannot be used when @param t is a sort.
   static Term* replace(Term* t, TermList what, TermList by);
+  static TermList replace(TermList t, TermList what, TermList by) {
+    return t == what ? by 
+         : t.isVar() ? t
+         : TermList(replace(t.term(), what, by));
+  }
```
MichaelRawson, line 56: Nice. I've needed the exact same function on `dedukti`, will be nice to have this in master.

Kernel/Term.cpp
```diff
@@ -122,6 +120,9 @@ bool TermList::isSafe() const
   return isVar() || term()->shared();
 }
 
+bool TermList::ground() const 
+{ return !isVar() && term()->ground(); }
```
MichaelRawson, line 124: I don't *think* it makes a difference, but why `!isVar()` rather than `isTerm()`?

Kernel/Term.cpp
```diff
@@ -678,14 +679,76 @@ std::string Term::toString(bool topLevel) const
...
+        return binary(" + ");
+      } else if (NumTraits::isMul(_functor)) {
+        return binary(" ");
+      } else if (NumTraits::isFloor(_functor)) {
+        return some(outputToString("⌊", termArg(0), "⌋"));
+      } else if (NumTraits::isMinus(_functor)) {
+        return uminus();
+      }
+    }
+    return Option<std::string>();
```
MichaelRawson, line 730: When does the "none" case occur? Could also consider using the empty string to signal that if the other cases are guaranteed non-empty.

Kernel/Term.cpp
```diff
@@ -678,14 +679,76 @@ std::string Term::toString(bool topLevel) const
     printArgs = isSort() || env.signature->getFunction(_functor)->combinator() == Signature::NOT_COMB;
   }
 
-  std::string s = headToString();
+#if NICE_THEORY_OUTPUT
+  auto theoryTerm = Kernel::tryNumTraits([&](auto numTraits) {
```
MichaelRawson, line 683: Would it make sense to move this to `NumTraits`?

Kernel/Term.hpp
```diff
@@ -1063,6 +1074,32 @@ class Literal
   /** set polarity to true or false */
   void setPolarity(bool positive)
   { _args[0]._setPolarity(positive); }
+
+  TermList eqArgSort() const;
+  
+  // prevent bugs through implicit bool <-> unsigned conversions
```
MichaelRawson, line 1080: This is a nice pattern, will be nicking this.

Kernel/TermIterators.hpp
```diff
@@ -44,6 +44,21 @@ class VariableIterator
   DECL_ELEMENT_TYPE(TermList);
   VariableIterator() : _stack(8), _used(false) {}
 
+  VariableIterator& operator=(VariableIterator&&) = default;
```
MichaelRawson, line 47: Does this do what you want? I don't think it calls the move-constructor.

In general, this looks like somewhere you might employ the copy-and-swap idiom.

Kernel/TermIterators.hpp
```diff
@@ -44,6 +44,21 @@ class VariableIterator
...
+    std::swap(_stack, other._stack);
+    std::swap(_used, other._used);
+    std::swap(_aux[0], other._aux[0]);
+    std::swap(_aux[1], other._aux[1]);
+    if (_stack.size() >= 1 && _stack[0] == &other._aux[1]) {
+      // fix pointer pointing into this object
+      ASS(_aux[0].isEmpty())
+      _stack[0] = &_aux[1];
+    }
+  }
```
MichaelRawson, line 60: Oh, boy - what does this bit do?! I know this is not your fault but if you understand what is going on here could we have a couple-line comment to explain?

Kernel/NumTraits.hpp
```diff
@@ -86,128 +94,314 @@ struct NumTraits;
...
-    static TermList name()                                                                                    \
-    { return TermList(name ## T()); }                                                                         \
-                                                                                                              \
-    static bool isName(const TermList& l)                                                                     \
-    { return l == name(); }                                                                                   \
-
-#define IMPL_NUM_TRAITS__QUOTIENT_REMAINDER(SHORT, X)                                                         \
-    IMPL_NUM_TRAITS__INTERPRETED_FUN( quotient ## X, SHORT,  _QUOTIENT_ ## X, 2)                              \
-    IMPL_NUM_TRAITS__INTERPRETED_FUN(remainder ## X, SHORT, _REMAINDER_ ## X, 2)                              \
+#define IMPL_NUM_TRAITS__INTERPRETED_SYMBOL(name, Name, SORT_SHORT, _INTERPRETATION)      \
```
MichaelRawson, line 97: I'm going to assume that you know what you're doing here. I think a lot of the diff is just introducing uppercase `Name`, right?

MichaelRawson (changes requested):
Some small changes if possible, looks good.

Kernel/NumTraits.cpp
```diff
@@ -9,8 +9,3 @@
  */
 #include "NumTraits.hpp"
 using namespace Kernel;
-
-// Sorry Joe, looks like Doxygen does not enjoy this solution.
-#ifndef IN_DOXYGEN
-__INSTANTIATE_NUM_TRAITS_ALL
-#endif
```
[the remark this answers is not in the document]
joe-hauns: it was removed since we don't need the `__INSTANTIATE_NUM_TRAITS_ALL` at all anymore :)

Kernel/OperatorType.cpp
```diff
@@ -137,7 +137,7 @@ std::string OperatorType::argsToString() const
 {
   unsigned ar = arity();
   ASS(ar);
-  std::string res =  ar > 1 ? "(" : "";  
+  std::string res = ar != 1 ? "(" : "";
```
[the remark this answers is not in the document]
joe-hauns: ah yeah my bad! Fixed it :)

Kernel/Term.cpp
joe-hauns: Added a clarifing comment. In this case we got an uninterprete term. Then we just format as usual.

Kernel/Term.cpp
joe-hauns: good question :) I was thinking of it as equivalent as well. Can change it.

Kernel/Term.cpp
joe-hauns: Hmmm i don't really think it would fit there. But I could factor it out and move it to a separate file if that's nicer you think.

Kernel/TermIterators.hpp
joe-hauns: You'r right. I thought for some reason it woud use the move constructor if we default. Fixed it. (There was no actual bug because the move assignment was never used anywhere)

Kernel/TermIterators.hpp
joe-hauns: added a long one. Does that work out?

Kernel/NumTraits.hpp
joe-hauns: Yes and also adding quite a bit of misc little helper functions to be able to query theory function symbols more nicel.

Kernel/TermIterators.hpp
MichaelRawson: I am horrified but enlightened. :-)

PR #632, opened 2024-12-11
Kbo move semantics

joe-hauns:
Made KBO & friends compatible with move semantics.

### DIFF
--- a/Kernel/KBO.hpp
+++ b/Kernel/KBO.hpp
@@ -22,6 +22,8 @@
 #include "Lib/DArray.hpp"
 
 #include "Ordering.hpp"
+#include "Debug/Tracer.hpp"
+#include <memory>
 
 #define SPECIAL_WEIGHT_IDENT_VAR            "$var"
 #define SPECIAL_WEIGHT_IDENT_INTRODUCED     "$introduced"
@@ -59,7 +61,7 @@ struct KboSpecialWeights<PredSigTraits>
   inline bool tryAssign(const std::string& name, unsigned weight) 
   { return false; }
 
-  inline static KboSpecialWeights dflt() 
+  inline static KboSpecialWeights dflt()
   { return { }; }
 
   bool tryGetWeight(unsigned functor, unsigned& weight) const;
@@ -74,6 +76,7 @@ struct KboSpecialWeights<FuncSigTraits>
   KboWeight _numInt;
   KboWeight _numRat;
   KboWeight _numReal;
+
   inline bool tryAssign(const std::string& name, unsigned weight) 
   {
     if (name == SPECIAL_WEIGHT_IDENT_VAR     ) { _variableWeight = weight; return true; } 
@@ -127,6 +130,8 @@ class KBO
 : public PrecedenceOrdering
 {
 public:
+  KBO(KBO&&) = default;
+  KBO& operator=(KBO&&) = default;
   KBO(Problem& prb, const Options& opt);
   KBO(
       // KBO params
@@ -145,9 +150,8 @@ public:
       // other
       bool reverseLCM);
 
-  static KBO testKBO();
+  static KBO testKBO(bool rand = false);
 
-  virtual ~KBO();
   void showConcrete(std::ostream&) const override;
   template<class HandleError>
   void checkAdmissibility(HandleError handle) const;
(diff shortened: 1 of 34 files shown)


Kernel/Ordering.hpp
```diff
@@ -24,13 +24,21 @@
...
 #include "Lib/Portability.hpp"
 #include "Kernel/SubstHelper.hpp"
 
 namespace Kernel {
 
+
+namespace PredLevels {
+  constexpr static int MIN_USER_DEF = 1; // equality has level 0, inequalities have level 1
+  constexpr static int EQ = 0;
+  constexpr static int INEQ = 0;
```
quickbeam123, line 39: The comment does not seem to represent what's written. Could you please check?

Kernel/KBO.hpp
```diff
@@ -249,24 +260,13 @@ class KBO
...
-    Result _lexResult;
-    /** The ordering used */
-    KBO& _kbo;
   }; // class KBO::State
 
   /**
    * State used for comparing terms and literals
    */
-  mutable State* _state;
+  mutable std::unique_ptr<State> _state;
```
MichaelRawson, line 268: Is there a reason this is a pointer at all? Looks like it could just be a field.

Kernel/KBOComparator.cpp
```diff
@@ -92,10 +92,10 @@ void KBOComparator::makeReady()
 
     // we only care about the non-zero weights and counts
     bool varInbalance = false;
-    auto state = kbo._state;
+    auto state = &*kbo._state;
 #if VDEBUG
     // we make sure kbo._state is not used while we're using it
-    kbo._state = nullptr;
+    auto __state = std::move(kbo._state);
```
MichaelRawson, line 98: I don't think this preserves the old behaviour - agree with the move, but shoudn't `kbo._state` still be assigned `nullptr`?

SAT/Z3Interfacing.cpp
```diff
@@ -1546,7 +1547,7 @@ Z3Interfacing::Representation Z3Interfacing::getRepresentation(Term* trm)
         //if constant treat specially
         if(trm->arity() == 0) {
           if(symb->integerConstant()){
-            IntegerConstantType value = symb->integerValue();
+            // IntegerConstantType value = symb->integerValue();
```
MichaelRawson, line 1550: Is this just dead?

MichaelRawson (approved):

SAT/Z3Interfacing.cpp
quickbeam123: Dead, feel free to delete. I was getting a warning and didn't want to kill the author's pattern completely.

Kernel/KBO.hpp
joe-hauns: Good question .. I don't really know whay it was a pointer in the first place .. I just kept it as it was I think to not mess up any magic :)

joe-hauns: I know there is some part of the code base that sets the pointer to `NULL` in order to be able to `ASSERT` that it's not being written to by multiple pocedures at the same time. But that's only debugging code. So I'm not sure..  maybe @mezpusz knows why it is a pointer at all?

Kernel/KBOComparator.cpp
joe-hauns: Isn't moving out of a `unique_ptr` the same thing as nulling it? It's then a pointer that cannot be accessed, right?

Kernel/KBO.hpp
mezpusz: I don't know where the code originates from or what the intention was. If thread safety was the issue it would be better to have the `State`s recycled in my opinion, otherwise setting them to null does not make too much sense. What do you think?

MichaelRawson: OK - let's keep it as-is for now to preserve the magic, but remember that it's weird and slate it for making it ...maybe just a `static` or plain members at some future point?

Kernel/KBOComparator.cpp
MichaelRawson: Yes, you're right - apparently the standard specifies the moved-from state of `unique_ptr`.

PR #633, opened 2024-12-11
Substitution Tree Variable Bank Refactor

joe-hauns:
Before this PR variable banks when querying a substitution tree were hard coded, to be the constants 
```
static constexpr int QUERY_BANK=0;
static constexpr int RESULT_BANK=1;
static constexpr int NORM_RESULT_BANK=3;
```

This has been replaced in this PR by adding fields `_queryBank`, `_normInternalBank`, `_internalBank` to the `SubstutionTree` retrieval algorithms `RobUnification` and `UnificationWithAbstraction`.

Why do we do this? The reason is that we prepare for the ALASCA merge, which involves a ternary inference. In order to implement that efficiently we need to match one clause with 2 substitution trees in sequence. The is only (soundly) possible if we use different `_normInternalBank` and `_internalBank` values for both of the queries. Thus we need to make this change.

As a further preparation for implementing the trinary inference rule we ad an overload `getUwa(AbstractingUnifier* state, ...)` to `LiteralSubstitutionTree` and `TermSubstitutionTree` which differs from the old overload that does not contain the `state` argument by starting the unification with a already initialized unifier, instead of a fresh one. Using this functoin we can use multiple substitutiontrees in sequence.

As this PR touches some bits of core routines of vampire (i.e. indexing) it would be great if @quickbeam123  could do some regression testing. The changes are not very drastic so I think there shouldn't be a difference in performance but we should be sure anyways. I already did some randomized runs and didn't run into any issues.


Indexing/SubstitutionTree.hpp
```diff
@@ -1364,10 +1385,19 @@ class SubstitutionTree final
 
       class RobUnification { 
         RobSubstitution _subs;
+        unsigned _queryBank;
```
MichaelRawson, line 1388: If you are trying to claw back some performance, you could 

```c++
template<unsigned queryBank, unsigned normInternalBank, unsigned internalBank>
class RobUnification {
};
```

propagate the template "up", etc, etc, and then eventually return a `VirtualIterator` wrapping it as necessary.

Indexing/SubstitutionTree.hpp
```diff
@@ -56,10 +62,16 @@
...
+inline unsigned subsTreeResultBank(unsigned n) 
+{ return 3 * n + 1; }
+
+inline unsigned subsTreeNormResultBank(unsigned n) 
+{ return 3 * n + 2; }
+
 
-static constexpr int QUERY_BANK=0;
-static constexpr int RESULT_BANK=1;
-static constexpr int NORM_RESULT_BANK=3;
```
MichaelRawson, line 62: So `NORM_RESULT_BANK` used to be 3, but now `subsTreeNormResultBank(0) == 2` - is this a problem?

Also, what happened to the "glue" bank we had at some point? Or did I misremember where that is?

Kernel/RobSubstitution.hpp
```diff
@@ -35,34 +35,35 @@
 #endif
 
 
-const int GLUE_INDEX=-3;
-const int SPECIAL_INDEX=-2;
+const int GLUE_INDEX=-2;
```
MichaelRawson, line 38: Ah, here is the glue. OK!

Kernel/RobSubstitution.hpp
```diff
@@ -35,34 +35,35 @@
 #endif
 
 
-const int GLUE_INDEX=-3;
-const int SPECIAL_INDEX=-2;
```
MichaelRawson, line 39: No more `SPECIAL_INDEX`? Happy news if so.

MichaelRawson (approved):

Kernel/RobSubstitution.hpp
```diff
@@ -35,35 +35,39 @@
...
 {
   VarSpec() {}
-  VarSpec(unsigned var, int index) : var(var), index(index) {}
+  VarSpec(TermList var, int index) : _code( 2 * var.var() + uint32_t(var.isSpecialVar()) ), index(index) {}
 
   friend std::ostream& operator<<(std::ostream& out, VarSpec const& self);
 
-  /** number of variable */
-  unsigned var;
+  /* we can store the variable + one bit to indicate whether it's a there is only 22 bits for variable number used in TermList */
```
quickbeam123, line 51: This comment is hard to parse. Could you please rephrase?

Indexing/SubstitutionTree.hpp
joe-hauns: Yeah this is a legacy thing because there used to be an unused `NORM_QUERY_BANK = 2`. It doesn't make a difference that the concrete numbers are changed, but only that they are distinct.

The glue bank is in `RobSubstitution`.

Kernel/RobSubstitution.hpp
joe-hauns: Yes indeed. How we got around it is a bit technical but the short answer is that one was never needed in the first place, and having it made things less general.

PR #634, opened 2024-12-12
Using GMP for Arithmetic

joe-hauns:
This PR replaces the datatype that is used to represent integers, reals and rationals from `int` to the gmp big-num type `mpz_t`. This means no more `MachineArithmeticExceptions` which helps for some verification problems reasoning about `MAX_INT` and friends, and makes coding things that involve arithmetic well nicer.

For that we ship with `mini-gmp` which can be compiled from source (see `mini-gmp-6.3.0/`). If a system-level installation of GMP is detected we link against that one instead of using `mini-gmp` as the full gmp library is much faster.

Besides changing the underlying datastructure the PR also tidies up some things around `IntegerConstantType` and friends, and adds some functionality to them.


Kernel/Theory.cpp
```diff
@@ -26,307 +26,211 @@
...
 
 #include "Theory.hpp"
-#define USES_2_COMPLEMENT (~0 == -1)
+
+std::string to_string(mpz_t const& self) {
+  auto s = mpz_sizeinbase(self, /* base */ 10);
+  auto str = new char[s + 2];
+  auto str_ = mpz_get_str(str, /* base */ 10, self);
+  ASS_EQ(str_, str)
+  return std::string(str);
```
MichaelRawson, line 45: I think this leaks, as `std::string(char *)` copies the input.

Kernel/Theory.hpp
```diff
@@ -54,77 +80,150 @@ class DivByZeroException         : public ArithmeticException
...
-  IntegerConstantType(IntegerConstantType&&) = default;
-  IntegerConstantType(const IntegerConstantType&) = default;
-  IntegerConstantType& operator=(const IntegerConstantType&) = default;
-  constexpr IntegerConstantType(InnerType v) : _val(v) {}
+  IntegerConstantType() { mpz_init(_val); }
+  explicit IntegerConstantType(int v) : IntegerConstantType() { mpz_set_si(_val, v); }
   explicit IntegerConstantType(const std::string& str);
 
+  IntegerConstantType(IntegerConstantType     && o) : IntegerConstantType() { mpz_swap(_val, o._val); }
+  IntegerConstantType(IntegerConstantType const& o) : IntegerConstantType() {  mpz_set(_val, o._val); }
```
MichaelRawson, line 153: Does this make a deep copy of `_val`? Otherwise when one destructor runs both will be invalidated.

MichaelRawson (approved):

easychair:
Has anybody checked what we should do to make sure that we can use GMP
under the LGPL license?

The last thing we want is to be bound by GPL.

Best, Andrei

MichaelRawson:
GMP is dual-licensed LGPL + GMP, so you can choose which you use. That said, I should look into the licensing before it gets merged.

Kernel/Theory.cpp
quickbeam123: Moreover, (as gcc pointed out), `str_` is unused (in release), which makes the code even more suspicious!

SAT/Z3Interfacing.cpp
```diff
@@ -76,18 +76,34 @@ struct BottomUpChildIter<z3::expr>
...
 z3::expr int_to_z3_expr(IntegerConstantType const& val, UInt64ToExpr toExpr) {
     auto sign = val.sign();
-    auto abs = val.abs().toInner();
+    auto abs = val.abs();
+
+    // TODO recycled stack
+    Stack<uint64_t> digits;
+    z3::expr base =  // <- == 2^64
+      toExpr(std::numeric_limits<uint64_t>::max()) + toExpr(1);
+    while(!abs.fits<unsigned long>()) {
```
quickbeam123, line 93: Is this how we communicate out arbitrary large numbers to z3's arbitrary large numbers?

Kernel/Theory.cpp
joe-hauns: > Moreover, (as gcc pointed out), `str_` is unused (in Release), which makes the code even more suspicious!

Well the assertion is only very defensive programming, we could actually ignore `str_`. 

> I think this leaks, as std::string(char *) copies the input.

Good poitn! I wasn't aware that that. Fixed now.

Kernel/Theory.hpp
joe-hauns: Do you mean `mpz_set`? That one makes a deep copy.

SAT/Z3Interfacing.cpp
joe-hauns: Yes exactly!

joe-hauns: Added a comment just now.

MichaelRawson:
LGPL appears OK for our purposes, merging.

PR #639, opened 2025-01-07
exponent noation real parsing

joe-hauns:
Fixed the parsing of exponent notation integers that was broken by introducing gmp as @quickbeam123  pointed out. 
Now we can parse problems like

```
tff(p_real_type,type,
    p_real: $real > $o ).

tff(reals,axiom,
    ( p_real(123.456e78)
    | p_real(123.456E78)
    | p_real(-123.456E78)
    | p_real(123.456E-78)
    | p_real(-123.456E-78) ) ).

```

again.


joe-hauns:
(this one depends on the signature PR)

PR #640, opened 2025-01-08
remove `TPTP::_isFof`, now redundant

MichaelRawson:
As numeral constants like `37` are no longer permitted in FOF, we could either:
1. Refuse to parse them altogether.
2. Do anything we like.

Let's say that (2) is better for us and our users. Interpret them as _typed, interpreted_ numeric constants, as we would for TFF, and simplify the parsing code.

### DIFF
--- a/Kernel/Signature.hpp
+++ b/Kernel/Signature.hpp
@@ -499,7 +499,7 @@ class Signature
   // Interpreted symbol declarations
 
   template<class Numeral>
-  unsigned addNumeralConstant(Numeral number_, bool defaultSort = false) {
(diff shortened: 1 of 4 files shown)


quickbeam123:
Looking good! I'll also run it across TPTP to see whether Geoff already did all of his part of the homework.

quickbeam123 (approved):

quickbeam123:
No obvious errors from non-HOL TPTP.

joe-hauns (approved):
Looks great! Thank you  :)

PR #641, opened 2025-01-08
add support for memory usage reporting

quickbeam123:
using the same sys/resource.h as we have for setMemoryLimit

### DIFF
--- a/Lib/Allocator.cpp
+++ b/Lib/Allocator.cpp
@@ -51,3 +51,14 @@ void Lib::setMemoryLimit(size_t limit) {
   // TODO should we warn here?
 #endif
 }
+
+long Lib::peakMemoryUsage() {
+#ifdef HAVE_RLIMIT
+  struct rusage usage;
+    if (getrusage(RUSAGE_SELF, &usage) == 0) {
+      return usage.ru_maxrss;
+    }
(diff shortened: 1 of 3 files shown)


Lib/Allocator.cpp
```diff
@@ -51,3 +51,14 @@ void Lib::setMemoryLimit(size_t limit) {
   // TODO should we warn here?
 #endif
 }
+
+long Lib::peakMemoryUsage() {
+#ifdef HAVE_RLIMIT
+  struct rusage usage;
```
MichaelRawson, line 57: Indentation weird but nbd.

Lib/Allocator.hpp
```diff
@@ -35,6 +35,7 @@
 namespace Lib {
 // attempt to set a memory limit for this process by system call
 void setMemoryLimit(size_t bytes);
+long peakMemoryUsage();
```
MichaelRawson, line 38: Maybe this should have a comment indicating that it is not clear what the unit is? Or move the Mac-specific thing inside this function?

MichaelRawson (approved):

quickbeam123:
Thanks for the feedback. Fixed!

PR #644, opened 2025-01-15
Fix OrderingComparator for non-specialized orderings

mezpusz:
As @joe-hauns pointed out, calling `OrderingComparator` for a term ordering that does not specialize this class or  `compareUnidirectional` results in some assertion violations. This PR attempts at rooting these out.


quickbeam123 (approved):

quickbeam123:
Do you want me, @mezpusz , to run any checks? 

(The change itself is below my resolution in this part of the codebase ;))

mezpusz:
> Do you want me, @mezpusz , to run any checks?

I will run some debug checks, I don't think performance tests are needed.

MichaelRawson (approved):

PR #647, opened 2025-01-20
Z3 intefacing fix

joe-hauns:
Bug fix in the translation of polymorphic vampire terms to monomorphic z3 terms.


MichaelRawson (approved):
Nice catch. What happens if we have something of still-polymorphic type, say `nil : list(X)`?

joe-hauns:
good question.. I think it shoud be alright, as we're just abstracting making `list(X)` and `list(int)` unrelated, not?
But the original code is from you right? haha 
I just updated what were doing for results to match what we're doing for arguments.

MichaelRawson:
Indeed. :innocent: IIRC that path wasn't handled in the code I replaced, so I didn't handle it either - it never seemed to actually happen, somehow it gets trapped earlier in the Vampire-to-Z3 pipeline.

PR #649, opened 2025-01-21
ALASCA

joe-hauns:
This PR contains ALASCA (Abstracting Linear Arithmetic Superposition CAlculus TACAS 2023).
It contains a full-fetched calculus: a couple of inference rules (`Inferences/ALASCA/*`), some new term orderings, a new preprocessor, and a couple of files with code shared among inferences ,orderings, and preprocessing.
Further it adds a git submodule for VIRAS (LPAR 2024) which is included as an inference rule inside of ALASCA.

In this PR we merge the finished real-arithmetic ALASCA, as well as unfinished (but sound & well-tested) extensions to mixed integer-real arithmetic. These extensions are not yet stable, but will be finished and stabilized very soon.

The options added in this PR:
- values for `-uwa` 
  - `alasca_one_interp`, `alasca_can_abstract`, `alasca_main`: strategies used for the real-arithmetic version of alasca (LPAR2023 paper  "Refining Unification with Abstraction")
  - `alasca_main_floor`: extension of `alasca_main` to mixed arithmetic. **unstable**
  - value `off` for `-ev`: evaluation can now be turned off.
- values for `-to`
  - `qkbo` ordering asd described in the ALASCA paper
  - `lakbo` new ordering that works for mixed arithmetic **unstable**
  - `incomp` makes any two terms incomparable. useful for debugging (e.g. checking if the ordering you are developing is incomplete)
- `-viras [ on | off ]` Turns on/off the VIRAS inference rule. this option is for mixed arithmetic and **unstable**
- `-alasca_demodulation [ on | off ]` specialized demodulation rule for alsca. this option is not yet well-tested and **unstable**
- `[ --alasca_strong_normalziation | -alasca_sn ] [ on | off ]` enables rewrites `s >= t ==> s > t | s == t` and `s != t ==> s > t | t > s`
- `[ --alasca_integer_conversion | -alascai ]  [ on | off ]` enables mixed integer-real reasoning in alasca. this option is **unstable**
- `[ --alasca_abstraction | -alascaa ] [ on | off ]` enables abstraction rules needed in corner cases when reasoning about mixed integer-real arithmetic. this option is **unstable**
- `-p smt2_proofcheck`, outputs the proof in a format where every function applicaiton is translated to an smt call. this is not very clean and stable but practically it helps a lot for proof-checking ALASCA proofs using other smt solvers like cvc5.

Further changes to the codebase are the following:
- added a `DBG_INDENT` macro. This macro can be in any block `{ DBG_INDENT; /* ... */ }` and all nested `DBG` outputs will be indented by 1 (+ 1 for every envlosing block with DBG_INDENT). this can be helpful for nice debug output in recursive-ish functions.
- `-p smt2_proofcheck`, outputs the proof in a format where every function applicaiton is translated to an smt call. this is not very clean and stable but practically it helps a lot for proof-checking ALASCA proofs using other smt solvers like cvc5.
- a parameter `qkbo` was added to a couple of functions in `Kernel/KBO*` and `Kernel/Ordering*`. This parameter is needed to adjust the precedence relation and the weight function of `KBO` and the underlying classes to work as expected by `QKbo`. The parameter should just be forwarded down to the places where it's needed (constructor of PresedenceOrdering and querying function weights in kbo)
- clause ids remain the same, whether we output theory axioms or not. Before the clause numbering used to change as a side effect of enabling the option `--show_theory_axioms on`.

The code of this PR has been extensively tested with random option combination runs over all of TPTP and SMTLIB.


Inferences/PolynomialEvaluation.hpp
```diff
@@ -21,33 +21,51 @@
...
+
+  mutable Memo::Hashed<PolyNf, PolyNf, StlHash> _memo;
 };
 
 
+class PolynomialEvaluationRule
+: public SimplifyingGeneratingLiteralSimplification
+{
+public:
+  USE_ALLOCATOR(PolynomialEvaluationRule);
```
MichaelRawson, line 57: No need for `USE_ALLOCATOR`, generally - but unlikely to do any harm.

Kernel/LiteralSelector.cpp
```diff
@@ -202,6 +202,7 @@ void LiteralSelector::select(Clause* c, unsigned eligibleInp)
   }
 
   unsigned eligible=1;
+  { /* we order the clause so that the literals with highest `getSelectionPriority` are in the front, and eligible is the number of literals with that selection priority */
```
MichaelRawson, line 205: Did you mean to introduce the new scope? No objection, just checking.

Kernel/Ordering.cpp
```diff
@@ -68,14 +70,23 @@ OrderingSP Ordering::s_globalOrdering;
...
 {
-
   if(s_globalOrdering) {
     return false;
   }
   s_globalOrdering = ordering;
   return true;
 }
 
+bool Ordering::unsetGlobalOrdering()
```
MichaelRawson, line 80: I have written some code that relies on there being a global ordering at some point. Should I not have?

Kernel/OrderingUtils.hpp
```diff
@@ -0,0 +1,501 @@
...
+    Stack<std::tuple<T, IntegerConstantType>> _elems;
+  public:
+    void integrity() const {
+#if VDEBUG
+      ASS(std::is_sorted(_elems.begin(), _elems.end(), [](auto l, auto r) { return std::get<0>(l) < std::get<0>(r); }))
+      for (auto e : _elems) {
+        ASS_G(std::get<1>(e), IntegerConstantType(0))
+      }
+#endif
+    }
```
MichaelRawson, line 36: Did you consider implementing this in terms of `std::multiset`?

Kernel/TypedTermList.hpp
```diff
@@ -20,16 +23,21 @@ class TypedTermList : public TermList
...
+  IMPL_COMPARISONS_FROM_TUPLE(TypedTermList);
+  IMPL_HASH_FROM_TUPLE(TypedTermList);
+
 
     // TODO get rid of default constructor
   TypedTermList() {}
   TypedTermList(TermList t, SortId sort) : TermList(t), _sort(sort) 
   { 
-    ASS_NEQ(sort, AtomicSort::superSort());
+    // ASS_NEQ(sort, AtomicSort::superSort());
```
MichaelRawson, line 37: Is this no longer true?

Lib/Sys/Multiprocessing.cpp
```diff
@@ -14,9 +14,11 @@
 
 #include <cerrno>
 #include <csignal>
+#include <errno.h>
 #include <unistd.h>
 #include <sys/types.h>
 #include <sys/wait.h>
+#include <cstring>
```
MichaelRawson, line 21: These don't seem like they want to be here?

Parse/TPTP.hpp
```diff
@@ -38,6 +39,7 @@ namespace Kernel {
 };
 
 namespace Parse {
+  using namespace Kernel;
```
MichaelRawson, line 42: I think this then means that you can use anything from `Kernel` as if it were in `Parse`, like `Parse::Clause` - was this the intention?

Shell/Options.hpp
```diff
@@ -1733,6 +1765,32 @@ bool _hard;
...
+
+
+    struct HasPolymorphism : OptionProblemConstraint{
+      USE_ALLOCATOR(HasHigherOrder);
+
+      bool check(Property*p){
+        ASS(p)
+        return (p->hasPolymorphicSym());
+      }
+      std::string msg(){ return " only useful with higher-order problems"; }
```
MichaelRawson, line 1790: "only useful with polymorphic problems"?

MichaelRawson (approved):
In the interest of time (with CAV and so on) I haven't read the completely-new code and will just trust you on that. There's a few things to fix here but nothing serious.

CMakeLists.txt
```diff
@@ -830,8 +890,9 @@ list(APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake-modules")
 # gmp stuff
 ################################################################
 
-find_package(GMP)
-if (GMP_FOUND)
+option(MINI_GMP "compile an own version of mini-gmp instead of linking with system level gmp" OFF)
```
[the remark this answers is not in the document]
joe-hauns: Yes. So the thing is that before this PR we would detect whether there is a systemlevel GMP and choose GMP or MINI_GMP depending on that. In order to make everything more stable for non-expert users we always compile with miniGMP now except user-enabled.
So we don't get potential version clashes if someone installed some old/new gmp version.

Indexing/ResultSubstitution.cpp
```diff
@@ -31,25 +31,26 @@ class RSProxy
   RSProxy(RobSubstitution* subst, int queryBank, int resultBank)
   : _subst(subst), _queryBank(queryBank), _resultBank(resultBank) {}
 
-  TermList applyToQuery(TermList t) override
+  TermList applyToQuery(TermList t) final override
```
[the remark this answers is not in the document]
joe-hauns: No not quite if I'm getting `final class` semantics right. You can still inherit from `RSProxy` but not override any of the methods. Not sure if you still could override constructor/destructor.

Inferences/ForwardSubsumptionDemodulation.cpp
```diff
@@ -490,6 +499,10 @@ bool ForwardSubsumptionDemodulation::perform(Clause* cl, Clause*& replacement, C
               // 3. L is lΘ = t, but t is larger that rΘ.
               // If all these checks fail, we try to find a literal M in D such that lΘ=rΘ < M.
               if (!_allowIncompleteness) {
+                bool dli_was_checked = false;
```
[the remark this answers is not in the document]
joe-hauns: Good spot. There was a line setting this to true, that was accidentally removed in the merging process. I pushed it again. The effect of the lack of this line was just a performance penalty but shouldn't have had any effects on the was proofsearch is being performed.

Kernel/OrderingUtils.hpp
joe-hauns: Good point. I wasn't aware this existed tbh. I'll give it a think for future optimizations! I think changing the current impl to that would be quite a bit of effort, but it might also be able to give some nice performance benefits, so I'll give it a think :)

Indexing/ResultSubstitution.cpp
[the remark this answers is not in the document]
MichaelRawson: Cool - as long as that's the intention!

CMakeLists.txt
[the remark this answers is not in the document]
MichaelRawson: Ah, OK - I like that! But: my reading of this code is that the mini-GMP is `OFF` by default.

Kernel/TypedTermList.hpp
joe-hauns: No because you can now call the bottom up evaluation of a `TypedTermList` with context arguments that let you either skip or not skip the type arguments of subterms. I added a test case `example_03__compute_size_with_context` in `tBottomUpEvaluation` showcasing that.

Parse/TPTP.hpp
joe-hauns: I just had a look again and before i added this we just had a `using namespace Kernel` in global scope. I thought this was a bit better because we don't leak into global scope for other files including `TPTP.hpp`. But I don't have any strong opiion on that. It was just accidentally broken when i merged things, so i fixed it again like this. I can also do the global import instead. as you like :)

CMakeLists.txt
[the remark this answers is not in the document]
joe-hauns: ah yea true! Just pushed a fix of that. Didn't notice bcause i always explicitly pass the flag in my scripts.

Parse/TPTP.hpp
MichaelRawson: Ugh, OK - this is better than that! I'll go around and do another `using namespace`-in-header file purge at some point. :-)

vampire.cpp
```diff
@@ -94,7 +94,11 @@ Problem* preprocessProblem(Problem* prb)
   // and that's the moment we want to set the random seed (no randomness in parsing, for the peace of mind)
   // the main reason being that we want to stay in sync with what profolio mode will do
   // cf ProvingHelper::runVampire
-  Lib::Random::setSeed(env.options->randomSeed());
+  if (env.options->randomSeed() != 0) {
+    Lib::Random::setSeed(env.options->randomSeed());
+  } else {
+    Lib::Random::resetSeed();
```
quickbeam123, line 100: I am not sure I want this. (I was happy that the default was reproducible also when not explicitly specifying any seed.)

Are you sure you need this, @joe-hauns ?

Inferences/ForwardSubsumptionDemodulation.cpp
```diff
@@ -83,7 +83,16 @@ bool ForwardSubsumptionDemodulation::perform(Clause* cl, Clause*& replacement, C
...
 
+
   Ordering const& ordering = _salg->getOrdering();
+  // some assertions won't hold if we use an alasca ordering i.e. in an alasca ordering we might have s subterm t but s not less than t, especially for non-atomic theory terms. 
+  // e.g. a subterm a - a
+  // but  a greater 0 equiv a - a
+  // we use this boolean flag to disable the assertions broken in this case. 
+  // everything is still sound but might become incomplete.
+  // TODO investigate casese like this indepth and adjust demodulation appropriately
+  static bool isAlascaOrdering = env.options->termOrdering () == Shell::Options::TermOrdering::QKBO
```
quickbeam123, line 94: Given this remark, should there be a separate line or two in `Options::complete` pointing to Alasca ordering as a potential completeness compromiser? Or are we fine on non-arithmetic inputs (and thus "covered" through the `// general properties causing incompleteness` there)?

vampire.cpp
joe-hauns: I see. Well i was thinking the default should be "actual random", but I'm fine with either. When i run i always set the seed explicitly.
I'm fine with either. We can simply change the default value of `randomSeed()` back to one so that the user will get reproducibility by defaults, but can choose for "actual randome" by setting it to `0`...?

Inferences/ForwardSubsumptionDemodulation.cpp
joe-hauns: For uninterpreted terms/literals the current alasca orderings will behave exactly like `kbo`. So that should be fine.

vampire.cpp
quickbeam123: OK, a good idea! Done :)

Shell/Statistics.hpp
```diff
@@ -260,6 +266,8 @@ class Statistics {
   unsigned finalActiveClauses;
   /** extensionality clauses at the end of the saturation algorithm run */
   unsigned finalExtensionalityClauses;
+  /** max number of literals of a generated clause  */
+  unsigned biggestGeneratedClause;
```
quickbeam123, line 270: `biggestGeneratedClause` to be "dead".

Should we make it work or rather properly get rid off?

Shell/Preprocess.cpp
```diff
@@ -136,10 +152,8 @@ void Preprocess::preprocess(Problem& prb)
     GoalGuessing().apply(prb);
   }
 
-  // If there are interpreted operations
+  // interpreted normalizations are not prepeared for "special" terms, thus it must happen after clausification
   if (prb.hasInterpretedOperations() || env.signature->hasTermAlgebras()){
-    // Normalizer is needed, because the TheoryAxioms code assumes Normalized problem
-    InterpretedNormalizer().apply(prb);
```
quickbeam123, line 142: Have you considered whether it's a problem for the non-alasca strategies that we no longer do `InterpretedNormalizer` here? 

I mean, in the light of 
"// Normalizer is needed, because the TheoryAxioms code assumes Normalized problem"

Shell/Preprocess.cpp
```diff
@@ -182,11 +204,6 @@ void Preprocess::preprocess(Problem& prb)
     LambdaElimination::addProxyAxioms(prb);
   }
 
-  if (prb.hasInterpretedOperations() || env.signature->hasTermAlgebras()){
-    // Some axioms needed to be normalized, so we call InterpretedNormalizer twice
-    InterpretedNormalizer().apply(prb);
```
quickbeam123, line 187: This was annoying, but I think we were lazy to supply some of the axioms in CNF. 

Not sure what's the situation now. 

Have you considered this?

Shell/Statistics.hpp
joe-hauns: removed it!

Shell/Preprocess.cpp
joe-hauns: Hmmm good question. I can't quite remember when/why i removed this. 
I did loads of randomized testing, so I think potential issues should have showed up. Maybe we should also do a regression test checking whether the performance of the old portfolios matches the new one just to be sure i didn't end up skipping adding some theory axioms that were necessary by changing that. 

In order to be safe we could also do a guarded normalization with `if(!env.options->alasca())`... But I think it's probably not necessary. What do you think?

Shell/Preprocess.cpp
joe-hauns: I made sure all axioms are in CNF now :)

quickbeam123: Perfect!

Shell/Preprocess.cpp
quickbeam123: Yes, that's the line of thinking I was hinting at! 

We had several issues in the past with the arithmetic code not doing what it should (performance-wise), because our pipeline was silently broken is such subtle ways like missing normalization steps at strange places during preprocessing. It's a good idea to do a performance regression. (And we should probably go through the pain of trying out both the old approach and alasca, to have them compared during spidering, before saying good buy to the old one completely.)

Kernel/Inference.cpp
```diff
@@ -398,7 +398,7 @@ Inference::Inference(const FormulaTransformation& ft) {
   init1(ft.rule,ft.premise);
 
   ASS_REP(isFormulaTransformation(ft.rule),ruleName(ft.rule));
-  ASS(!ft.premise->isClause());
+  // ASS(!ft.premise->isClause()); // TODO: should this assertion be here? it would keeps us from preprocessing clauses in InterpretedNormalizer
```
quickbeam123, line 401: The idea was: if you apply this to clauses, don't call it a FormulaTransformation.

Do you have a use case for calling FormulaTransformation on clauses? (I mean, it sounds sensible, but if that's the case, let's maybe think if we can single it out and assign it a different inference rule as well!)

joe-hauns: Ahhh i see. The reason is that I used InterpretdNormalization for Clauses as well as for formulas at some point (twice in the preprocessing chain). This is not the case anymore so adding the assertion back is fine i think. I just tried this and doing some debug runs now to make sure everything s still fine.

joe-hauns: So the ALASCA preprocessing is done after clausification. I didn't find a `ClauseTransformation` so I thought `FormulaTransformation` is the inference to go with. Or should I use `SimplifyingInference1` you reckon? Feels not quite right.

quickbeam123: I think we could add `ClauseTransformation`. I can do it if you agree that that's what feels like the right thing!

Shell/InterpretedNormalizer.cpp
```diff
@@ -404,8 +404,9 @@ Formula* InterpretedNormalizer::NLiteralTransformer::transform(Formula* f)
 // InterpretedNormalizer
 //
 
-InterpretedNormalizer::InterpretedNormalizer()
+InterpretedNormalizer::InterpretedNormalizer(InequalityNormalizer* norm)
 : _litTransf(new NLiteralTransformer())
```
quickbeam123, line 408: Not saying it's bad, but do you agree that a non-null `InequalityNormalizer` actually never passed to `InterpretedNormalizer`? (Maybe in tests? I only see a single use in `Preprocess.cpp` on line 89.)

Kernel/Signature.hpp
```diff
@@ -298,20 +300,43 @@ class Signature
...
+    bool numeralConstant(RationalConstantType*) const { return rationalConstant(); }
+    bool numeralConstant(IntegerConstantType*) const { return integerConstant(); }
+  public:
+
+    /** template version of {integer,rational,real}Constant */
+    template<class Number> bool numeralConstant() const
+    { return numeralConstant((Number*)nullptr); }
+
     /** return true if an interpreted number, note subtle but significant difference from numericConstant **/
     inline bool interpretedNumber() const
```
quickbeam123, line 314: Sorry, I think I asked before. But there is no `significant difference from numericConstant` anymore, right? (after all, there does not seem to be `numericConstant` anymore).

Shell/InterpretedNormalizer.cpp
joe-hauns: good spot. that was indeed dead code. removed it.

Kernel/Signature.hpp
joe-hauns: indeed. I updated the outdated comment.

quickbeam123 (approved):

PR #650, opened 2025-01-25
allow clauses to be iterators

MichaelRawson:
A minor irritation used to be that one cannot iterate over a clause with range-for. We now have `iterLits()`, which is a big improvement, but it's still nice to be able to write

```c++
// Clause *cl;
for(Literal *l : *cl)
  ;
```

Allow this.

### DIFF
--- a/Kernel/Clause.hpp
+++ b/Kernel/Clause.hpp
@@ -130,7 +130,9 @@ public:
   Literal** literals() { return _literals; }
   // support use of clauses as an iterator
   Literal **begin() { return _literals; }
+  Literal *const *begin() const { return _literals; }
   Literal **end() { return _literals + _length; }
+  Literal *const *end() const { return _literals + _length; }
 
   /** True if the clause is empty */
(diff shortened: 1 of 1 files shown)


quickbeam123 (approved):

quickbeam123:
Cool, a good idea!

(Would it also make sense to add const versions?)

MichaelRawson:
Yes! Although I had to think very hard about what the type signature should be. Apparently `Literal *const *begin() const` is the right one, according to my trusty [translator](https://cdecl.org/?q=struct+Literal+*const+*begin%28%29%3B). This seems to [do the right thing](https://godbolt.org/z/fM4c7Eqhn) (try making the parameter const and/or removing the assignment to `nullptr`).

quickbeam123:
We might even want 
```
const Literal *const *begin() const;
```
but let's wait what @joe-hauns thinks about this!

joe-hauns:
Looks all good to me :)

PR #653, opened 2025-02-04
identifiers beginning with an underscore followed by an uppercase let…

MichaelRawson:
C++ standard says

> - Each name that contains a double underscore (__) **or begins with an underscore followed by an uppercase letter** is reserved to the implementation for any use.
> - Each name that begins with an underscore is reserved to the implementation for use as a name in the global namespace.

Unfortunately this bites us on Windows (!) with Cygwin's compiler for some reason. Rename `_L` and `_M` in SAT subsumption (resolution) to have this compile.

@RobCoutel - if you prefer another name please go ahead and push to this branch.

@YvesMarcoux - this was the compiler issue I mentioned, just FYI.

### DIFF
--- a/SATSubsumption/SATSubsumptionAndResolution.cpp
+++ b/SATSubsumption/SATSubsumptionAndResolution.cpp
@@ -170,8 +170,8 @@ void SATSubsumptionAndResolution::loadProblem(Clause* L,
   cout << "----------------------------------------------" << endl;
   cout << "Setting up problem " << L->toString() << " " << M->toString() << endl;
 #endif
-  _L = L;
-  _M = M;
+  _sidePremise = L;
+  _mainPremise = M;
   _m = L->length();
   _n = M->length();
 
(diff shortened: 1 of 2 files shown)


RobCoutel:
I am not particularly opposed to this change. Maybe we could change the name to _mainPremise and _sidePremise instead. I would even be an improvement on the other notation. The L and M were just to be consistent with the paper. What do you think?

MichaelRawson:
> Maybe we could change the name to _mainPremise and _sidePremise instead. I would even be an improvement on the other notation. The L and M were just to be consistent with the paper. What do you think?

Good idea, done!

RobCoutel:
We might as well change the interface of the functions. When calling `checkSubsumption(clause L, clause M)`, maybe change to `checkSubsumption(clause mainPremise, clause sidePremise)`

hetzenmat:
If we follow this convention in the whole codebase (and I think we should), then we should also change all the include guards of the form `__FILE_NAME__` since the double-underscore rule [also applies to macros](https://clang.llvm.org/extra/clang-tidy/checks/bugprone/reserved-identifier.html).

MichaelRawson:
Good point, Matthias. I've put this on my list for a more aggressive purge.

PR #654, opened 2025-02-05
Fix precedences for interpreted functions

mezpusz:
This bug has already been fixed, but the fix must have got lost in some merge, so now I added a unit test, which is not much but better than nothing.

The bug could be reproduced by comparing the two terms in both directions which gives inconsistent results, so I suggest a debug check for this. For now I added this to forward demodulation since we are comparing pairs of terms there anyways, but I am open to putting it somewhere else.

I will run the entire TPTP with this check to see if there's any more such bugs.


quickbeam123 (approved):

mezpusz:
The tests did not give any further violation.

PR #655, opened 2025-02-06
Formula Tracing

joe-hauns:
Slight refactor to make the clause tracing options work with `Formula`s as well as with `Clause`s


quickbeam123:
Cool! But shouldn't we call the MACRO and the function `doUnitTracing` now? :)

quickbeam123 (approved):

joe-hauns:
Verily! I renamed the function :)

PR #659, opened 2025-03-13
Qa auto for Geoff - version 2

quickbeam123:
The major change (which could have consequences) is that in portfolio mode, we start modifying `env.options` in the child, while previously there was a copy (actually two). In my head, there is just one global Options object in vampire (for now), so I hope the change is only to the better (no copying needed).

Also, there is going to be a better place (in Options.cpp) for the
`// resolve away auto value before accessing it`
bit (now in Preprocess.cpp for `-qa`), once we merge the `alasca` PR.

### DIFF
--- a/CASC/PortfolioMode.cpp
+++ b/CASC/PortfolioMode.cpp
@@ -577,7 +577,7 @@ void PortfolioMode::runSlice(std::string sliceCode, int timeLimitInDeciseconds)
   ASS_GE(sliceTime,0);
   try
   {
-    Options opt = *env.options;
+    Options& opt = *env.options;
 
     // opt.randomSeed() would normally be inherited from the parent
     // addCommentSignForSZS(cout) << "runSlice - seed before setting: " << opt.randomSeed() << endl;    
@@ -608,17 +608,15 @@ void PortfolioMode::runSlice(std::string sliceCode, int timeLimitInDeciseconds)
 /**
  * Run a slice given by its options
  */
-void PortfolioMode::runSlice(Options& strategyOpt)
+void PortfolioMode::runSlice(Options& opt)
 {
   System::registerForSIGHUPOnParentDeath();
   UIHelper::portfolioParent=false;
 
-  Options opt = strategyOpt;
   //we have already performed the normalization (or don't care about it)
   opt.setNormalize(false);
   opt.setForcedOptionValues();
   opt.checkGlobalOptionConstraints();
-  *env.options = opt; //just temporarily until we get rid of dependencies on env.options in solving
 
   if (outputAllowed()) {
     addCommentSignForSZS(cout) << opt.testId() << " on " << opt.problemName() <<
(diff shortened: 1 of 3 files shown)


MichaelRawson:
Sorry @quickbeam123 , I could have caught with the first iteration had I thought about it enough. I am looking at it now, but it is a tangled web...

Shell/Options.cpp
```diff
@@ -2228,7 +2228,8 @@ void Options::init()
                                                                   {"auto","plain","synthesis","off"});
     _questionAnswering.description= "Determines whether (and how) we attempt to answer questions:"
        " plain - answer-literal-based, supports disjunctive answers; synthesis - designed for sythesising programs from proofs.";
-    _questionAnswering.addHardConstraint(If(notEqual(QuestionAnsweringMode::OFF)).then(ProperSaturationAlgorithm()));
+    _questionAnswering.addHardConstraint(If(equal(QuestionAnsweringMode::PLAIN)).then(ProperSaturationAlgorithm()));
+    _questionAnswering.addHardConstraint(If(equal(QuestionAnsweringMode::SYNTHESIS)).then(ProperSaturationAlgorithm()));
```
MichaelRawson, line 2232: I'm going to regret asking, but shouldn't `If(Or(equal(QuestionAnsweringMode::PLAIN), equal(QuestionAnsweringMode::SYNTHESIS))).then(ProperSaturationAlgorithm()` work?

CASC/PortfolioMode.cpp
```diff
@@ -608,17 +608,15 @@ void PortfolioMode::runSlice(std::string sliceCode, int timeLimitInDeciseconds)
...
 {
   System::registerForSIGHUPOnParentDeath();
   UIHelper::portfolioParent=false;
 
-  Options opt = strategyOpt;
   //we have already performed the normalization (or don't care about it)
   opt.setNormalize(false);
   opt.setForcedOptionValues();
   opt.checkGlobalOptionConstraints();
-  *env.options = opt; //just temporarily until we get rid of dependencies on env.options in solving
```
MichaelRawson, line 621: Good, I think we should admit defeat here, `env.options` isn't going away any time soon.

CASC/PortfolioMode.cpp
```diff
@@ -577,7 +577,7 @@ void PortfolioMode::runSlice(std::string sliceCode, int timeLimitInDeciseconds)
   ASS_GE(sliceTime,0);
   try
   {
-    Options opt = *env.options;
+    Options& opt = *env.options;
```
MichaelRawson, line 580: So this is now a reference to the post-fork `env.options`, right? But then why do we pass it around all the time? Couldn't it just be `env.options` everywhere?

MichaelRawson (approved):
I think this is OK - we now have one fewer copy of `Options` (very good thing!), and I don't believe there is a hazard as all paths lead to process termination quickly without relying on a distinction between `opt`, `env.options` and/or `strategyOpt`.

MichaelRawson:
I'm happy but I leave pulling the trigger to you - in case you have any last-minute thoughts!

PR #660, opened 2025-03-16
CaDiCaL

MichaelRawson:
@anjapetkovic @kfazekas @florianpollitt - just FYI. A long time coming, but we're getting there!

We'd like CaDiCaL because (i) it's a better SAT solver,  (ii) it produces proofs, and (iii) it has features like user propagation we might find useful. However, for very small instances solved repeatedly (like AVATAR) it's a bit slower than Minisat in some cases. Therefore, with regret, do *both*, exposed via an option, and let the strategy scheduling machinery figure it out.

@quickbeam123 - things to discuss before we merge:
1. How are we dealing with this as an external dependency? Do we want to vendor it like Minisat, `git submodule` it like Z3, or just straight-up dependency like GMP? I'd be inclined to do one of the first two, otherwise it's yet more things in the `README` that people don't read.
2. The "SAT solver" option logic is a bit wonky in places as we have either Minisat, CaDiCaL, or...Z3? Not everything makes sense with SMT and it's an increasingly leaky abstraction: should we have a separate "AVATAR modulo theories" flag or something?
3. There's a few things that aren't implemented in `CadicalInterfacing`, including the mysterious `interpolateViaAssumptions` - I'm not sure that we care (yet).


quickbeam123:
Hi @MichaelRawson, thanks for putting this together!

Some short answers; let's elaborate if needed:
1) I'd go the submodule way. (We are anyway getting another one with ALASCA, i.e., viras)
By the way, can we instruct `cmake` to visit and make a submodule (with some default settings) for users who "don't want to know"? (Viras is particularly undemanding in this respect, as it is header-only.)
2) I would definitely split the options (i.e., sat solver for avatar: mini,cadi,(z3); sat solver for fmb: mini,cadi; and maybe fix elsewhere, until we care: sat solver for global subsumption, for monotonicity, for interpolation -> minisat seems just fine; sat solver for proof reconstruction -> cadical) - I can do this
3) the way we approach 2) could help us resolve this.

quickbeam123 (approved):

PR #664, opened 2025-03-21
Capitalized CMAKE_BUILD_HOL value

Rockjack00:
Hello!

The release notes say to use `CMAKE_BUILD_HOL=ON` but `CMakeLists.txt` only supports `CMAKE_BUILD_HOL=On` (note the lowercase "n") with a string comparison ([here](https://github.com/vprover/vampire/blob/hol/CMakeLists.txt#L811)).  In order to keep the values consistent with other CMake arguments I propose changing it to `ON`.

Thanks!
~Tyler


MichaelRawson:
Thanks for this! I got bitten before. CI is just failing because we're not merging into master.

PR #666, opened 2025-03-24
an alternative fix

quickbeam123:
- this causes a much smaller difference to the code and to performance
- as far as I could check, it fixes the problematic 
```
./vampire Problems/SWV/SWV253-2.p --decode dis+1_1:1_sas=z3:si=on:sac=on:aac=none_0
```

### DIFF
--- a/Saturation/Splitter.cpp
+++ b/Saturation/Splitter.cpp
@@ -1393,6 +1393,14 @@ SplitLevel Splitter::tryGetComponentNameOrAddNew(unsigned size, Literal* const *
   else {
     RSTAT_CTR_INC("ssat_new_components");
 
+    // adding a component should mean "recompute model" (even if we actually don't end up adding a clause)
+    // this is connected to the subtle case in handleNonSplittable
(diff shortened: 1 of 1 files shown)


MichaelRawson (approved):
If it doesn't hurt performance, I'm very happy with this fix.

It seems we now have the invariant "if a component exists, we have a SAT assignment for it". Is this the case? If so can we simplify the code elsewhere in `Splitter` based on this assumption?

mezpusz (approved):

quickbeam123:
The point is that we ALMOST ALWAYS add clauses to the sat solver when we introduce a new component. (That's why the diff in behaviour and thus performance is minimal.)

"if a component exists, we have a SAT assignment for it"
- if you introduce a completely fresh component slot (not like here), you don't have assignment until you recompute the model (the SAT variable did not even exists until then
- another problem is that there is `MinimizingSolver` and `SATSolver::VarAssignment::DONT_CARE`

In sum, this change does not add any new obvious opportunities for simplifications.

PR #667, opened 2025-03-24
Structural induction one for one free variable

hzzv:
This PR adds a variant of structural induction one, where the induction literal contains a single existentially quantified variable:
```
C \/ ~L[t, x]
----------------------
premise => !y ?z. L[y, z]
```

In order for this to work, we track the skolems introduced in NewCNF. We then use this information in Induction to correctly resolve the clausified induction axiom (where `L[y, z]` is skolemized as `L[y, sK(...)]`) with the premise (`~L[t, x]`) -- i.e. to, map `x` to `sK(...)`.

To turn this feature on, use `--induction_ground_only off` (`-indgo` for short).
Example run on [nat-double.smt2](https://github.com/vprover/vampire_benchmarks/blob/master/synthesis/recursive/smt2/nat-double.smt2):
```
./vampire --forced_options ind=struct:indgo=off:tgt=off:erd=off:updr=off:indc=goal:indmd=1 nat-double.smt2
```

### DIFF
--- a/Shell/Statistics.cpp
+++ b/Shell/Statistics.cpp
@@ -45,6 +45,7 @@ Statistics::Statistics()
   : inputClauses(0),
     inputFormulas(0),
     formulaNames(0),
+    skolemFunctions(0),
     initialClauses(0),
     splitInequalities(0),
     purePredicates(0),
@@ -279,7 +280,7 @@ void Statistics::print(std::ostream& out)
   COND_OUT("Input formulas", inputFormulas);
   SEPARATOR;
 
-  HEADING("Preprocessing",formulaNames+purePredicates+trivialPredicates+
+  HEADING("Preprocessing",formulaNames+skolemFunctions+purePredicates+trivialPredicates+
     unusedPredicateDefinitions+functionDefinitions+selectedBySine+
     sineIterations+blockedClauses+splitInequalities);
   COND_OUT("Introduced names",formulaNames);


Shell/NewCNF.cpp
```diff
@@ -39,7 +39,7 @@ using namespace Kernel;
 
 namespace Shell {
 
-void NewCNF::clausify(FormulaUnit* unit,Stack<Clause*>& output)
+void NewCNF::clausify(FormulaUnit* unit,Stack<Clause*>& output, DHMap<unsigned, Term*>* bindings)
```
MichaelRawson, line 42: @quickbeam123 - I think you're the expert here. Any objections?

Shell/Options.cpp
```diff
@@ -1482,6 +1482,12 @@ void Options::init()
     _inductionOnComplexTerms.onlyUsefulWith(_induction.is(notEqual(Induction::NONE)));
     _lookup.insert(&_inductionOnComplexTerms);
 
+    _inductionOnlyGround = BoolOptionValue("induction_only_ground","indog",true);
+    _inductionOnlyGround.description = "Apply induction only on ground literals vs. literals with at most one free variable";
+    _inductionOnlyGround.tag(OptionTag::INDUCTION);
```
MichaelRawson, line 1487: What's *in*, dog?

It's a matter of taste, but I might prefer something like `induction_applies_to` with values `ground`, `one_var`. Then with the inevitable `two_var` extension it's just one more enumeration value. No strong preference either way.

MichaelRawson (approved):

Inferences/Induction.cpp
```diff
@@ -152,6 +153,15 @@ Formula* InductionContext::getFormula(const std::vector<TermList>& r, bool oppos
   return getFormula(tr, opposite);
 }
 
+Formula* InductionContext::getFormulaFreeVar(const std::vector<TermList>& r, bool opposite, unsigned freeVar, TermList& freeVarSub, Substitution* subst) const
+{
+  Formula* replaced = getFormula(r, opposite, subst);
+  static Substitution s;
```
mezpusz, line 159: Wouldn't simply a local `Substitution` suffice here?

Inferences/Induction.cpp
```diff
@@ -190,6 +200,20 @@ Formula* InductionContext::getFormulaWithSquashedSkolems(const std::vector<TermL
   return res;
 }
 
+bool InductionContext::getFreeVariable(unsigned& res) const {
+  // Note: we assume there is just one free variable in the literals of _cls
+  for (const auto& kv : _cls) {
+    for (const auto& lit : kv.second) {
+      FormulaVarIterator fvi(lit);
```
mezpusz, line 207: How about using the less heavy `VariableIterator`? Then we don't need the include above either.

Inferences/Induction.cpp
```diff
@@ -1567,6 +1618,76 @@ void InductionClauseIterator::performRecursionInduction(const InductionContext&
...
+/*
+Creates the structural induction axiom with an existentially quantified variable:
+?y,w. !u0,us,z. ?x. (L[0, u0] & (L[y, w] -> L[s(y), us]) -> L[z, x])
+*/
+void InductionClauseIterator::performStructInductionFreeVar(const InductionContext& context, InductionFormulaIndex::Entry* e)
+{
+  if (context._indTerms.size() > 1) return;
+  TermAlgebra* ta = env.signature->getTermAlgebraOfSort(SortHelper::getResultSort(context._indTerms[0]));
+  unsigned freeVar;
+  ALWAYS(context.getFreeVariable(freeVar)); // variable free in the induction literal
```
mezpusz, line 1630: Instead of asserting here, how about putting an assertion violation at the end of `getFreeVariable` (and make it return the value). In either case, if we fail in release we get undefined behaviour anyways.

Inferences/Induction.hpp
```diff
@@ -149,13 +148,19 @@ struct InductionContext {
     node->second.push(lit);
   }
 
-  // These two functions should be only called on objects where
+  // These functions should be only called on objects where
   // all induction term occurrences actually inducted upon are
   // replaced with placeholders (e.g. with ContextReplacement).
   Formula* getFormula(const std::vector<TermList>& r, bool opposite, Substitution* subst = nullptr) const;
+  Formula* getFormulaFreeVar(const std::vector<TermList>& r, bool opposite, unsigned freeVar, TermList& freeVarSub, Substitution* subst = nullptr) const;
```
mezpusz, line 155: This sounds like we are getting a free variable in a formula. How about `getFormulaWithExistentialVar` or something similar, to make it explicit?

Shell/Options.cpp
hzzv: I am wondering if there is much value in having the option so granular, allowing to choose between `one_var`, `two_vars`, `all_vars` etc. @mezpusz any opinions?

Anyway I don't have a strong preference (just thought of keeping it simple with a boolean).
I can also change it to `induction_ground_only` for the enthusiastic `indgo` :D (and to have it in line with `question_answering_ground_only`)

Inferences/Induction.cpp
hzzv: Yes, done.

Inferences/Induction.cpp
```diff
@@ -1036,6 +1085,7 @@ Clause* resolveClausesHelper(const InductionContext& context, const Stack<Clause
 {
   // first create the clause with the required size
   RobSubstitution renaming;
+  bool inductionWithFreeVar = subst.size() > 1;
```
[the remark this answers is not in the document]
hzzv: Good point. I re-plumbed it as you suggested.

Inferences/Induction.cpp
hzzv: Done.

Inferences/Induction.hpp
hzzv: I went with `getFormulaWithFreeVar`, since the returned formula is not quantified, and hence the "existential" sounds misleading to me. LMK if you object :)

Inferences/Induction.cpp
hzzv: Fair point, done.

Shell/Options.cpp
mezpusz: I prefer `indgo`! If we manage to generalize this to any number of free variables, it still wouldn't make sense to make it more granular, since we shouldn't get fundamentally bigger problems with more variables... I think.

mezpusz (approved):
I am fine with everything added, but I realized we don't have a standard method for performance/debug checking inductive reasoning. @quickbeam123 do you think we could use the same benchmarks/method as for the lemma discovery paper?

quickbeam123:
I could try running something, yes! 

What should be the main goal? To check that the old vampire is not getting worse, or that the new option solves more?

Shell/Options.cpp
MichaelRawson: Cool - you seem to have thought about this more than me. :-)

hzzv:
I renamed the option and added statistics on how much is this flavor of induction used.

While for now we only instantiate the "structural induction one" axiom with non-ground literals, the literal ground-ness is orthogonal to the axiom, and in the future this can be done also for other axioms. Therefore I added statistics independent on the axiom statistics, but rather as another alternative to `inductionApplication` and `generalizedInductionApplication`.

Shell/NewCNF.cpp
quickbeam123: Yes, look reasonable!

quickbeam123 (approved):

quickbeam123:
The code looks good! 

But let's do some randomizes assertion violation hunting and maybe a singe perf run to see how much this could help!

mezpusz:
> What should be the main goal? To check that the old vampire is not getting worse, or that the new option solves more?

I think mostly regression testing, but looking at the code this shouldn't give anything unexpected as the new code almost doesn't touch the old.

mezpusz (approved):

PR #668, opened 2025-03-24
More paranoid theory lemma checks (in DEBUG mode)

quickbeam123:
A question is, could this be slowing down even debug mode too much? Do we want to hide it behind compile flag (but then we would forget to use it). I hope the answer could be that with alasca, we will be adding much fewer theory axioms, so this will trigger less and less and won't slow us down.


MichaelRawson (approved):
Let's put it in by default and see if anyone notices/complains?

PR #671, opened 2025-03-25
save/restore logic for line numbers in TPTP parser

### LINKED ISSUE #670: Line number tracking is not reset when leaving an included file.
So that, e.g.,
```
include('Axioms/CAT003-0.ax').
cnf(assum)BROKEN(e_codomain_exists,hypothesis,
    there_exists(codomain(a)) ).
```

happily returns 
```
% Parsing Error on line 120: , expected at position 3867 (text: ))
```

MichaelRawson:
Fixes #670. There is also the ...byte offset? in e.g.

```
% Parsing Error on line 2: , expected at position 3867 (text: ))
```

but I'd be inclined to say that's useless and remove it.

### DIFF
--- a/Parse/TPTP.cpp
+++ b/Parse/TPTP.cpp
@@ -1229,6 +1229,7 @@ void TPTP::unitList()
     resetChars();
     delete _in;
     _in = _inputs.pop();
+    _lineNumber = _lineNumbers.pop();
     _includeDirectory = _includeDirectories.pop();
     delete _allowedNames;
     _allowedNames = _allowedNamesStack.pop();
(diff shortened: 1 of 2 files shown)


quickbeam123:
Cool, that was quick! 

Could maybe the byte offset be easily given the meaning "on that line"? (I agree that a global offset is useless.)

MichaelRawson:
>Could maybe the byte offset be easily given the meaning "on that line"?

Not ...easily, sorry. If you are curious, look at the `getChar(n)`, `shiftChars(n)` and `resetChars()` functions - one could probably do something ad-hoc, but the "proper fix" to get line/column tracking is to have only `getChar()` which would return "the next character". This would be far more conventional and perhaps more performant but is also a much bigger fix.

In the meantime, I removed the global offset information entirely - it was only used for error reporting ("ah yes, byte offset 32,489, I know it well") and should save us a few cycles.

PR #672, opened 2025-03-26
CMake refresh

MichaelRawson:
CMake had gotten very large and complex over time, but not for any particularly good reason. Split it up a bit, be ruthless with complexity, make it so we can actually use different CMake targets (Ninja at least now works).

Notably:
- Separate the "big ol' list of source files" into a different file.
- Ditch "source groups", I don't think anyone actually used them.
- Reduce the amount of (conceptual) "libraries" of object files that the build script conjured for some reason lost to time.
- Remove the IPO/LTO experiment, I don't think anyone actually used it as it took many hours and didn't make much difference.
- Factor out some duplicate logic.
- Reorganise the build such that similar logic happens in the same place.
- Always print `CMAKE_BUILD_TYPE`, it's difficult to work out what kind of build you have cached otherwise.
- Automatically checkout mandatory submodules (currently VIRAS, CaDiCaL) if not present.
- Do not build `vtest` in the `all` target. You can still build it easily with `make vtest`, but it's perhaps unfair on users to ask them to type `make vampire` if they don't want the unit tests.
- Always use mini-gmp.

The diff is not very useful here: better to read `CMakeLists.txt` directly as it's almost completely rewritten.


mezpusz (approved):
I am not very fluent in CMake-ese, but the changes are very welcome, I can now almost read `CMakeLists.txt` that it's so short!

I played around with it locally and it seems to work on Mac.

One thing, now that we're here, can we either add the source files that are in the repo but not in `sources.cmake` or delete them?

According to my calculations, there are quite many (`cpp/hpp`), excluding the ones from submodules and intermediate build sources, see file attached.
[missing-files.txt](https://github.com/user-attachments/files/19473938/missing-files.txt)

MichaelRawson:
Thanks for the list @mezpusz - I've done just that. It's a good idea to have every file listed explicitly in the CMake build, the automatic compiler dependency detection doesn't always work.

In some cases files weren't listed because if you list a file with a `.cpp` extension, CMake tries to compile it - and some of them are simply `#include`d into other files. I've renamed these to `.hpp` - any objections @joe-hauns?

There were also some forgotten things lying around. `MockedSaturationAlgorithm.cpp` seems to be unused. `UnitTests/dummyHash.hpp` isn't a unit test, so I've moved it to `Test/`. `tTimeTrace` was forgotten altogether, poor thing. It doesn't do anything unless `VTIME_PROFILING`, but :shrug:

CMakeLists.txt
```diff
@@ -35,1037 +28,139 @@ link_libraries(Threads::Threads)
...
-    SAT/SATLiteral.hpp
-    SAT/SATSolver.hpp
-    SAT/CadicalInterfacing.hpp
-    SAT/Z3Interfacing.hpp
-    )
-source_group(sat_source_files FILES ${VAMPIRE_SAT_SOURCES})
+# enable for (slightly) more deallocation
+add_compile_definitions(CHECK_LEAKS=0)
+# enable for time profiling
+add_compile_definitions(VTIME_PROFILING=0)
```
joe-hauns, line 63: What do you think about guarding these with 
```
if(DEFINED ENV{CHECK_LEAKS})
  add_compile_definitions(CHECK_LEAKS="$ENV{CHECK_LEAKS}")
else()
    add_compile_definitions(CHECK_LEAKS=0)
endif()
```
(and same for `VTIME_PROFILING`)

CMakeLists.txt
```diff
@@ -35,1037 +28,139 @@ link_libraries(Threads::Threads)
...
-    ${VAMPIRE_SMTCOMP_SOURCES}
-    ${VAMPIRE_MINISAT_SOURCES}
-    ${VAMPIRE_CADICAL_SOURCES}
-    ${VAMPIRE_CASC_SOURCES}
-    ${VAMPIRE_SAT_SUBSUMPTION_SOURCES}
-    Forwards.hpp
-    "${CMAKE_CURRENT_BINARY_DIR}/version.cpp"
-    )
+# TODO could this just be source files?
+include_directories(${CMAKE_CURRENT_SOURCE_DIR}/mini-gmp-6.3.0)
```
joe-hauns, line 129: I think yes they could be source files, but I think having them as an include dir makes more clear that we don't wanna touch them.

Makefile
```diff
@@ -23,7 +23,7 @@
 #   CHECK_LEAKS      - test for memory leaks (debugging mode only)
 #   VZ3              - compile with Z3
 
-COMMON_FLAGS = -DVTIME_PROFILING=0 -DVMINI_GMP
+COMMON_FLAGS = -DVTIME_PROFILING=0
 
 DBG_FLAGS = $(COMMON_FLAGS) -g  -DVDEBUG=1 -DCHECK_LEAKS=0 # debugging for spider
 REL_FLAGS = $(COMMON_FLAGS) -O3 -DVDEBUG=0 -DNDEBUG # no debugging
```
joe-hauns, line 29: There is the `NDebug` definition here, and it has been removed in the `CMakeLists.txt` I'm not sure what it does. Just wanted to check whether this was intentional?

joe-hauns (approved):
Looks great! Thank you for your work @MichaelRawson !

CMakeLists.txt
MichaelRawson: Good idea - I'll make them an option so we don't have to go around editing `CMakeLists` to activate them. Or did you specifically want an environment variable?

`CHECK_LEAKS` is kind of silly now, anyhow - I'll put it on my list to investigate better memory leak tracking.

Makefile
MichaelRawson: `NDEBUG` among other things means that `assert()` compiles to nothing - apparently it gets set by CMake on release builds.

PR #675, opened 2025-03-28
fix TPTP `include()` logic, report paths in errors

MichaelRawson:
- Group various save/restore TPTP logic into one struct.
- Finally move the resolve-include-file logic out from `Options` into `TPTP`.
- Fix it to conform with the TPTP standard, and only then with Vampire extensions.
- Report paths in TPTP parse errors.

### DIFF
--- a/Parse/TPTP.cpp
+++ b/Parse/TPTP.cpp
@@ -90,11 +90,9 @@ Clause* TPTP::parseClauseFromString(const std::string& str)
  * Initialise a lexer.
  * @since 27/07/2004 Torrevieja
  */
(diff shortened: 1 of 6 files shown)


Parse/TPTP.cpp
```diff
@@ -2061,7 +2061,7 @@ fs::path TPTP::resolveInclude(const fs::path included)
     return included;
 
   // 2
-  auto relativeToCurrentFileDirectory = currentFile.path.remove_filename() / included;
```
MichaelRawson: Oops. I'm not actually sure how this ever worked - thanks for the fix.

PR #677, opened 2025-03-31
CaDiCaL ensureVarCount needs to do something after all

MichaelRawson:
@mezpusz found a bug where AVATAR's polarity advice causes an assertion to fail in CaDiCaL where we didn't inform it about a new variable. I think the cleanest solution is to make `CaDiCaLInterfacing::ensureVarCount` actually do something - we probably should have been doing this all along but CaDiCaL was nice to us so far.


MichaelRawson:
There was another unpleasant interaction with `abs` where:
1. A new SAT variable is introduced
2. `suggestPolarity` is called, which invalidates CaDiCaL's model
3. The `BufferedSolver` suppresses the split clause the new variable was introduced for.
4. Something asks for the current assignment, which is now invalid.

Fix this by also buffering the polarity suggestions.

PR #680, opened 2025-04-01
KBO type term comparisons

mezpusz:
Normally, in KBO we wouldn't need comparisons between types and terms since no one calls the comparison with types and terms and subcomparisons are only done lexicographically. With term ordering diagrams, this changed as `TermPartialOrdering`, one of the backbones of efficiency of TODs, may unknowingly try to compare every element in its domain regardless of what ordering we use or whether it makes sense, resulting in an occasional assertion violation in KBO.

Comparing types and terms was already possible in LPO by making type constructors smaller than normal functors, so I just moved the common logic to `PrecedenceOrdering`. According to Ahmed this should work for KBO as well.

20 iterations of debug sampling with `-fdtod on` over the entire TPTPdid not show any related error, and the performance testing showed no difference.


MichaelRawson (approved):
From the description I feared the diff, but this is really nice! More general and less branching, bravo.

PR #683, opened 2025-04-08
Refactor code tree matchers

mezpusz:
`CodeTree` has two matcher implementations, even though they do *almost* the same thing. This PR merges them with some compile-time branching, so there should be no noticable overhead.

Usual sampled testing done for 20 iterations of the entire TPTP, no errors (so far).
Performance testing showed a miniscule amount of slowdown for some problems, I think it's just noise after testing them a couple of times locally.


mezpusz:
Maybe squashing the commits would make sense as this originates from some unrelated branch, the contents of which were merged or discarded already.

MichaelRawson:
@mezpusz - any reason for this to be on hold? Looks like something we want.

mezpusz:
Currently it's just one problem lost for Discount, none for Otter over FOL with `-i 20000`, but there is an increase in instructions spent/runtime apparently:

Discount:

run 0 unsat: 9208 (1) sat: 1030 (0) cputime: 56704.85 s instructions: 222910720 Mi memory: 1533982.30 MB
run 1 unsat: 9207 (0) sat: 1030 (0) cputime: 57026.97 s instructions: 222964103 Mi memory: 1529041.49 MB

Otter:

run 0 unsat: 9312 (0) sat: 1020 (0) cputime: 61537.83 s instructions: 222902233 Mi memory: 1739725.78 MB
run 1 unsat: 9312 (0) sat: 1020 (0) cputime: 61743.87 s instructions: 222978995 Mi memory: 1731210.00 MB

The problem lost is `LCL/LCL640+1.015.p`, and it is solved with 153 extra megainstructions.

So... 🤷

PR #685, opened 2025-04-14
modernization in Statistics

quickbeam123:
mainly relaxing here
- also, I wanted to make use of the modern cpp way of initializing fields in place
- I was also checking whether there are any leftover uninitialized fields in Statistics (of which I discovered a few in the recent past)
- then a modernization of the two enums there, which kind of cascaded everywhere, but I hope the diff is not so bad after all


MichaelRawson (approved):

MichaelRawson:
Nice, this was also bothering me.

PR #691, opened 2025-05-06
Added a preliminary schedule which knows about alasca

quickbeam123:
trained on ARI part of TPTP v9.0.0


CASC/Schedules.cpp
```diff
@@ -1953,8 +1953,8 @@ void Schedules::getStructInductionTipSchedule(const Shell::Property& property, S
   // Overall score 235.07056943321408 probs on average / budget 24401 Mi
 }
 
-void Schedules::getSnakeTptpUnsSchedule(const Shell::Property& property, Schedule& quick) {    
-  if (property.hasNumerals() || property.hasInterpretedOperations()) { 
+void Schedules::getSnakeTptpUnsSchedule(const Shell::Property& property, Schedule& quick) {
+  if (property.hasNumerals() || property.hasInterpretedOperations()) {
```
MichaelRawson, line 1957: @joe-hauns is this the right property invocation?

MichaelRawson (approved):

CASC/Schedules.cpp
quickbeam123: A good question. But I think so. It's the one that worker in the past. 

(But a different one, which perhaps boils down to the same thing on TPTP at the moment, could be more principled. Do you have a suggestion what could be better?)

MichaelRawson: No idea - it just looks a bit fishy, I could imagine:
1. problems which have numerals but only use them as constants
2. problems with interpreted non-arithmetic operations

So I guess really we want `property.hasArithmeticOperations()`?

quickbeam123: In fact, the whole "TPTP's SPC endswith ARI" is (on v9.0.0) equivalent to current Vampire's `hasNumerals=1`.

Even 
```
tff(a,type,c:$int).
tff(a,axiom,$less(c,c)).
```
triggers `hasNumerals`, I guess we could drop the whole `hasInterpreted` from the disjunction...

quickbeam123: Probably a job for a different PR, though!

.github/workflows/CI.yml
```diff
@@ -16,31 +16,45 @@ jobs:
...
-      working-directory: ${{runner.workspace}}/vampire/build
-      run: cmake .. -DCMAKE_BUILD_TYPE=Debug
+    - name: Configure and Build Vampire (Debug + Unit Tests)
+      working-directory: ${{ runner.workspace }}/vampire
+      run: |
+        mkdir -p build
+        cd build
+        cmake .. -DCMAKE_BUILD_TYPE=Debug
+        make -j8
+        make vtest -j8
```
MichaelRawson, line 46: I think just `vtest` would be OK here, right?

quickbeam123: Because it's mostly just a single linking job?

MichaelRawson: Because we don't need the `Debug` Vampire, only `vtest`.

quickbeam123: Aaaah! Cool!

PR #693, opened 2025-05-16
Precedence generation mini update

quickbeam123:
* don't shuffle twice for `si=on:sp=scramble`; reuse `shuffleArray` for `SCRAMBLE`
* don't tiebreak on `Occurrence` (it defeats the purpose of pre-Shuffling under `si=on`); but no: randomized quicksort is not stable, but that should not matter

### DIFF
--- a/Kernel/Ordering.cpp
+++ b/Kernel/Ordering.cpp
@@ -579,13 +579,13 @@ struct BoostWrapper : public SymbolComparator
   }
 };
 
-struct OccurenceTiebreak {
-  OccurenceTiebreak(SymbolType) {} // here the SymbolType is a dummy argument, required by the template recursion convention
+struct IncomaparableTiebreak {
+  IncomaparableTiebreak(SymbolType) {} // here the SymbolType is a dummy argument, required by the template recursion convention
 
-  Comparison compare(unsigned s1, unsigned s2) {  return Int::compare(s1,s2); }
+  Comparison compare(unsigned s1, unsigned s2) { return Comparison::EQUAL; }
 };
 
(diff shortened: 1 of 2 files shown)


Kernel/Ordering.cpp
```diff
@@ -579,13 +579,13 @@ struct BoostWrapper : public SymbolComparator
   }
 };
 
-struct OccurenceTiebreak {
-  OccurenceTiebreak(SymbolType) {} // here the SymbolType is a dummy argument, required by the template recursion convention
+struct IncomaparableTiebreak {
```
mezpusz, line 582: IncomaparableTiebreak -> IncomparableTiebreak

mezpusz (approved):
Do I get it right that under `-si on`, we first shuffle the input and do whatever we would do anyways, but in this case when the sorting would have to be tiebreaked, we don't want to destroy this shuffling, hence the new `noTieBreak` flag? Then, why not use `noTieBreak` also with other comparators that are used as `InnerComparator`s?

Otherwise I'm fine with these changes.

quickbeam123:
> Do I get it right that under `-si on`, we first shuffle the input and do whatever we would do anyways, but in this case when the sorting would have to be tiebreaked, we don't want to destroy this shuffling, hence the new `noTieBreak` flag? Then, why not use `noTieBreak` also with other comparators that are used as `InnerComparator`s?

That exactly the case and I had a previous version doing just that. However, the problem is that our implementation of quicksort in DArray uses randomization, so the quicksort is not only not stable, it's also `random_seed` dependent. I decided that's too much, especially having seen how easy it was to break some of test checks in `sanity` which I myself introduced last week...

MichaelRawson (approved):

PR #694, opened 2025-05-19
Fix UB in SATSubsumptionAndResolution::get(I|J)Matches

hetzenmat:
Before this commit, the code relied on UB because the indices in `_indexI` and `_indexJ` sometimes point to `_matchesByI.end()` and `_matchesByJ.end()`, respectively.
When running the vampire using a debug build, an assertion in the STL vector is violated because we try to access out of bounds.
But since we only need the pointer of the elements we can remove the dereferencing and just access the `vector::data()` pointer and add the desired offsets.

Moreover, `checks/sanity` the obsolete `egrep` is replaced by `grep -E` to remove all the warnings.

### DIFF
--- a/SATSubsumption/SATSubsumptionAndResolution.hpp
+++ b/SATSubsumption/SATSubsumptionAndResolution.hpp
@@ -166,9 +166,8 @@ private:
     Slice<Match> getIMatches(unsigned i)
     {
       ASS_L(i, _indexI.size())
-      return Slice<Match>(
-          &_matchesByI[_indexI[i]],
-          &_matchesByI[_indexI[i + 1]]);
+      return Slice(_matchesByI.data() + _indexI[i],
+                   _matchesByI.data() + _indexI[i + 1]);
     }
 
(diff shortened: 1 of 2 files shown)


SATSubsumption/SATSubsumptionAndResolution.hpp
```diff
@@ -166,9 +166,8 @@ class SATSubsumptionAndResolution {
     Slice<Match> getIMatches(unsigned i)
     {
       ASS_L(i, _indexI.size())
-      return Slice<Match>(
-          &_matchesByI[_indexI[i]],
-          &_matchesByI[_indexI[i + 1]]);
```
MichaelRawson, line 171: Nice. Yes, sorry for this bug - it works in C. ;-)

MichaelRawson:
@RobCoutel when you're happy and you know it, clap your hands and I will merge.

RobCoutel (approved):
Yes, it sounds good to me.

PR #695, opened 2025-05-21
SMT-based proof checking

MichaelRawson:
SMT-based proof checking as in the CADE submission *Ground Truth*. It's a bit rough around the edges but now

```
vampire <problem> --proof_extra full -p smtcheck | grep -v ^% | z3 /dev/stdin
```

does a form of proof checking. This was on `alasca-smtcheck` but the branch structure was a bit messy so I've moved everything to this branch in one mega-commit, history be damned.

@joe-hauns wanted this merged so he can check proofs. :-)
@anjapetkovic - just FYI. This isn't the Dedukti stuff yet but it moves `master` and `dedukti` closer together.

@joe-hauns since I assume you'll want to patch this at some point I'll ask you to review. Please feel free to fix anything and push directly to the branch if you feel enthusiastic - I'm on holiday this week so I'll try to do as little as possible. Otherwise I will eventually get to it. I'll add some comments inline to explain what's going on.

No big deal if not, but if we can get this merged by the CADE camera-ready we can drop the footnote with some random dead branch which inevitably somebody will ask us about in 15 years' time.

### DIFF
--- a/Kernel/SubstHelper.hpp
+++ b/Kernel/SubstHelper.hpp
@@ -24,6 +24,8 @@
 #include "TermIterators.hpp"
 #include "Kernel/BottomUpEvaluation.hpp"
 
+#include <unordered_map>
+
 namespace Kernel {
 
 using namespace Lib;
@@ -662,6 +664,37 @@ FormulaList* SubstHelper::applyImpl(FormulaList* fs, Applicator& applicator, boo
 inline TermList AppliedTerm::apply() const {
   return aboveVar ? SubstHelper::apply(term, *applicator) 
                   : term;
+}
+
+// a variable-term map that complies with the requirements for MatchingUtils and SubstHelper
+struct SimpleSubstitution
+{
+  std::unordered_map<unsigned, TermList> bindings;
+
+  TermList apply(unsigned var) {
+    if(auto bound = bindings.find(var); bound != bindings.end())
+      return bound->second;
+    return TermList(var, false);
+  }
+
+  void reset() { bindings.clear(); }
+
+  bool bind(unsigned var, TermList term)
+  {
+    auto [it, inserted] = bindings.insert({var, term});
+    return inserted || it->second == term;
+  }
+
+  void specVar(unsigned var, TermList term) { ASSERTION_VIOLATION; }
+};
+
+inline std::ostream &operator<<(std::ostream &out, const SimpleSubstitution &subst) {
+  out << "[";
+  for(auto [left, right] : subst.bindings) {
+    out << " " << left << " -> " << right;
+  }
+  out << "]";
+  return out;
 }
 
 #endif /* __SubstHelper__ */
--- a/Lib/Coproduct.hpp
+++ b/Lib/Coproduct.hpp
@@ -30,6 +30,7 @@
 #include <type_traits>
 #include <vector>
 #include <tuple>
+#include <vector>
 
 namespace Lib {
 
--- a/SATSubsumption/SATSubsumptionAndResolution.cpp
+++ b/SATSubsumption/SATSubsumptionAndResolution.cpp
@@ -69,6 +69,7 @@
    */
 
 #include "Kernel/Matcher.hpp"
+#include "Kernel/SubstHelper.hpp"
 #include "Lib/Environment.hpp"
 #include "Lib/Int.hpp"
 #include "Shell/Statistics.hpp"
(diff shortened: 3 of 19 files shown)


Kernel/SubstHelper.hpp
```diff
@@ -664,4 +666,35 @@ inline TermList AppliedTerm::apply() const {
                   : term;
 }
 
+// a variable-term map that complies with the requirements for MatchingUtils and SubstHelper
+struct SimpleSubstitution
```
MichaelRawson, line 670: I think this might be replaced by @mezpusz 's `MapBinderAndApplicator`.

Lib/Coproduct.hpp
```diff
@@ -30,6 +30,7 @@
 #include <type_traits>
 #include <vector>
 #include <tuple>
+#include <vector>
```
MichaelRawson, line 33: I think this un-broke Anja's build at some point.

SATSubsumption/SATSubsumptionAndResolution.hpp
```diff
@@ -510,6 +511,10 @@ class SATSubsumptionAndResolution {
                                              Kernel::Clause *mainPremise,
                                              unsigned resolutionLiteral);
 
+  // return the substitution required for the last subsumption resolution
+  // precondition: the last subsumption resolution succeeded
+  SimpleSubstitution getBindingsForSubsumptionResolutionWithLiteral();
```
MichaelRawson, line 516: This breaks @RobCoutel 's encapsulation somewhat. I don't think it's too horrible, but I welcome other suggestions.

cmake/sources.cmake
```diff
@@ -704,6 +704,8 @@ set(SOURCES
     Shell/Property.hpp
     Shell/Rectify.cpp
     Shell/Rectify.hpp
+    Shell/SMTCheck.hpp
+    Shell/SMTCheck.cpp
```
MichaelRawson, line 708: Should remember to add these to the `Makefile` too so that Martin is happy.

SATSubsumption/SATSubsumptionAndResolution.hpp
RobCoutel: In principle, this should be fine. But you do not need to ask the SAT solver for the model again. Line 936, the model is set in the variable `_model`, as used in `generateConclusion()`.

On the code aesthetics, it would be great if the documentation of the function matched the style of the file. Especially for a public function.

SATSubsumption/SATSubsumptionAndResolution.cpp
RobCoutel, line 1: I think the second guard
```cpp
// extra guard for below
if(lit.var().index() >= _bindingsManager.size())
  continue;
```
is redundant with the one above.

I am also a little uncertain about the second nested loop. What does it do?
If you have the match, you can just check its polarity to identify the resolution literal.
```cpp
if(!match.polarity) {
  // This is the literal you are interested in.
}
```

SATSubsumption/SATSubsumptionAndResolution.hpp
MichaelRawson: Ahah. This one's actually OK, I never call `checkSubsumptionResolution`, only `checkSubsumptionResolutionWithLiteral`.

Comments - sure, fixed.

SATSubsumption/SATSubsumptionAndResolution.cpp
MichaelRawson: This seems to happen only if there is a nullary predicate involved - so I think I'm OK with the check, if you are. BTW, we must have discussed this at the time, but does it make sense to also check for ground/ground matches in the same way and not insert a binding in that case?

MichaelRawson:
There are still a few questions about the thing I hacked into `SATSubsumptionAndResolution`, but after that we should be good to go. Thanks everyone!

SATSubsumption/SATSubsumptionAndResolution.cpp
MichaelRawson: Not really: we would need to know whether a match variable represents a normal or symmetric equality match. We could probably add a `bool` to `Match` without performance penalty as there is already `bool polarity` there and so it won't increase the size/alignment, but I leave this up to you as to which solution you prefer.

RobCoutel: The additional boolean is maybe not even necessary.
This is a little hacky, but we know in which order we match the literals e.g. (X = Y) -> (c = d) will create two variables b1 => (X->c, Y->d) and b2 => (X->d, Y->c). You also know that the variable id of b2 = b1 + 1. So you can check whether the match before has the same i,j, in which case you know the binding has to be the symmetry.
This approach prevents to add overhead on the subsumption and subsumption resolution part at the cost of a slightly more expensive reconstruction and reducing code maintainability (say if you change the order of binding)

RobCoutel: I like the new code (aside from the compiler error).
If there is a little bit of care for the efficiency, is there a way to merge the binding one you found earlier (lines 993 and 994), or do you necessarily have to run the matching procedure again?

PR #696, opened 2025-05-22
Remove combinatory superposition

hetzenmat:
In preparation for the HOL merge into main, we want to first remove the obsolete combinatory superposition implementation.


Kernel/Term.cpp
```diff
@@ -566,8 +566,6 @@ std::string Term::headToString() const
     if (!isSort() && Theory::tuples()->findProjection(functor(), isLiteral(), proj)) {
       return "$proj(" + Int::toString(proj) + ", ";
     }
-    bool print = (isLiteral() || isSort() ||
-                 (env.signature->getFunction(_functor)->combinator() == Signature::NOT_COMB)) && arity();
     std::string name = "";
```
quickbeam123, line 569: I am not sure I understand this is a harmless change (i.e. skipping code that can't anymore trigger).

I know it's just printing. But could you please explain it a bit? (Also the other changes in `Term.cpp` below).

hetzenmat: By default, every symbol in the signature gets assigned the `NOT_COMB` value for its `combinator()` field and only the real combinators have some different values.
As I removed all combinators, this means that we always want to print the arguments for symbols with positive arity.

So we can simplify the printing logic by removing the `print` and `printArgs` booleans and just check the arity if we need parentheses and print the arguments.

PR #697, opened 2025-05-26
Implement support for lambda terms and conversion to de Bruijn indices

hetzenmat:
This is a port from the `hol` branch with the difference that banks are not stored in variables anymore.

### DIFF
--- a/Shell/Options.cpp
+++ b/Shell/Options.cpp
@@ -308,7 +308,7 @@ void Options::init()
     _problemName.description="";
     //_lookup.insert(&_problemName);
 
-    _proof = ChoiceOptionValue<Proof>("proof","p",Proof::ON,{"off","on","proofcheck","tptp","property","smt2_proofcheck"});
+    _proof = ChoiceOptionValue<Proof>("proof","p",Proof::ON,{"off","on","proofcheck","tptp","property","smt2_proofcheck","smtcheck"});
     _proof.description=
       "Specifies whether proof (or similar e.g. model/saturation) will be output and in which format:\n"
       "- off gives no proof output\n"
@@ -316,9 +316,11 @@ void Options::init()
       "- proofcheck will output proof as a sequence of TPTP problems to allow for proof-checking by external solvers\n"
       "- tptp gives TPTP output\n"
       "- property is a developmental option. It allows developers to output statistics about the proof using a ProofPrinter "
-      "object (see Kernel/InferenceStore::ProofPropertyPrinter\n";
+      "object (see Kernel/InferenceStore::ProofPropertyPrinter\n"
+      "- smtcheck produces a ground SMT script for proof checking\n";
     _lookup.insert(&_proof);
     _proof.tag(OptionTag::OUTPUT);
+    _proof.addHardConstraint(If(equal(Proof::SMTCHECK)).then(_proofExtra.is(equal(ProofExtra::FULL))));
 
     _minimizeSatProofs = BoolOptionValue("minimize_sat_proofs","msp",true);
     _minimizeSatProofs.description="Perform unsat core minimization when a sat solver finds a clause set UNSAT\n"
--- a/FMB/FiniteModel.cpp
+++ b/FMB/FiniteModel.cpp
@@ -518,7 +518,7 @@ bool FiniteModel::evaluate(Formula* formula,unsigned depth)
 
      for(unsigned c=1;c<=_size;c++){
        Substitution s;
-       ALWAYS(s.bind(var,getDomainConstant(c)));
+       s.bindUnbound(var,getDomainConstant(c));
        Formula* next_sub = SubstHelper::apply(next,s);
        next_sub = SimplifyFalseTrue::simplify(next_sub);
        next_sub = Flattening::flatten(next_sub); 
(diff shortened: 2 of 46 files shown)


Indexing/TermSharing.cpp
```diff
@@ -88,21 +87,23 @@ void TermSharing::computeAndSetSharedTermData(Term* t)
...
+    bool hasLambda = t->isLambdaTerm();
     Color color = COLOR_TRANSPARENT;
     
     unsigned typeArity = t->numTypeArguments();
     for (unsigned i = 0; i < t->arity(); i++) {
       TermList* tt = t->nthArgument(i);
       if (tt->isVar()) {
         ASS(tt->isOrdinaryVar());
-        if(i >= typeArity){
+        if(i < typeArity) {
```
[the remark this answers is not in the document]
ibnyusuf: Yes, but on the `hol` branch we are recording `hasSortVar` rather than `hasTermVar`. Unfortunately, I no longer recall why I made the change from `hasTermVar` to `hasSortVar`!

Shell/Options.cpp
```diff
@@ -1935,8 +1948,16 @@ void Options::init()
...
 
-    _casesSimp = BoolOptionValue("cases_simp", "cs", false);
-    _casesSimp.description =
+    _equalityToEquivalence = BoolOptionValue("equality_to_equiv","e2e",false);
+    _equalityToEquivalence.description=
+      "Equality between boolean terms changed to equivalence \n"
+      "t1 : $o = t2 : $o is changed to t1 <=> t2";
+    _lookup.insert(&_equalityToEquivalence);
+    // potentially could be useful for FOOL, so am not adding the HOL constraint
+    _equalityToEquivalence.tag(OptionTag::HIGHER_ORDER);
```
quickbeam123, line 1957: I think we should either drop the `tag` or the comment. Do you agree? (I assume this is a preprocessing rule?)

Kernel/Term.hpp
```diff
@@ -754,7 +781,7 @@ class alignas(8) Term
 
   void setHasTermVar(bool b)
   {
-    ASS(shared() && !isSort());
```
quickbeam123, line 757: Could you please briefly explain the reason here?

quickbeam123:
This PR comes with a mild performance overhead for general FOL proving (probably due the changes in `Term` and `TermSharing`). Can we expect the followup HOL PRs to be all less invasive to the prover core?

Kernel/Signature.hpp
```diff
@@ -157,6 +142,7 @@ class Signature
     unsigned _letBound : 1;
     /** proxy type */
     Proxy _prox;
+    int _dbIndex;
```
quickbeam123, line 145: Maybe, could we spell out "DeBruijn" in full? 

For a second, I started thinking about databases ...

Indexing/TermSharing.cpp
```diff
@@ -70,7 +69,7 @@ TermSharing::~TermSharing()
 void TermSharing::setPoly()
 {
   //combinatory superposiiton can introduce polymorphism into a monomorphic problem
```
quickbeam123, line 71: We no longer have "combinatory superposiiton". We should fix the comment at least.

But is it also true for the new HOL calculus that it can bring poly, even if we did not have it?

Kernel/SortHelper.hpp
```diff
@@ -87,7 +87,7 @@ class SortHelper {
 
   static OperatorType* getType(Term const* t);
 
-  static void getTypeSub(const Term* t, Substitution& subst);
```
quickbeam123, line 90: Can you please add a comment here about what the new `bool` result means?

Kernel/Term.hpp
```diff
@@ -324,27 +354,29 @@ class TermList {
   // but it was *not* portable because the layout of the bitfield is not guaranteed
   //
   // now we use a manual bitfield, as follows
+#define MK_FIELD(typ, name, Name, NAME, start, size) \
```
quickbeam123, line 357: Can you help me understand why is the new stuff the same as the old? Or is it? 

@MichaelRawson , could you please take a quick look here too?

Indexing/TermSharing.cpp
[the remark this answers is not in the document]
hetzenmat: You are right, thanks.
If I am correct you added hasSortVar for `SortDeref`.

Kernel/Term.hpp
```diff
@@ -46,7 +45,7 @@
 #include "Lib/Recycled.hpp"
 
 // the number of bits used for "TermList::_info::distinctVars"
-#define TERM_DIST_VAR_BITS 22
+#define TERM_DIST_VAR_BITS 19
```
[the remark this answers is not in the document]
hetzenmat: The 3 bits that are added are for tracking if the term has a de Bruijn index, a redex, and a lambda expression.

Shell/Options.cpp
hetzenmat: You are right, this is used in `FOOLElimination` as a preprocessing rule. I dropped the `tag`.

Kernel/Signature.hpp
hetzenmat: I renamed all occurrences

Indexing/TermSharing.cpp
hetzenmat: Yes, this should be also true for the new HOL calculus. I updated the comment

PR #701, opened 2025-05-28
have Substitution conform to Binder, Applicator interfaces

MichaelRawson:
Recently @mezpusz mentioned that it would be nice if `Substitution` conformed to the usual interfaces. Do it.

A slight hitch is that `bind` was `void` in `Substitution` previously, and the implementation failed an assertion if a variable was already present in the substitution. I've changed it to return `bool` as you would expect, and with sneaky use of `[[nodiscard]]` found all the instances where it is called and changed them to `ALWAYS(subst.bind(...))`, so that we don't lose any assertions. Sometimes this is a little trivial, but we can always remove those later.

Some of the methods were not used, and all were trivial enough to be inlined IMO, so I have.

The upshot is you can now write e.g.

```c++
TermList foo(TermList base, TermList instance, TermList other) {
  Substitution subst;
  MatchingUtils::matchTerms(base, instance, subst);
  return SubstHelper::apply(other, subst);
}
```

without the compiler yelling at you. I'm agnostic about whether to replace several existing half-implementations with `Substitution`, but I hope this will happen organically.


Kernel/Substitution.hpp
```diff
@@ -12,61 +12,85 @@
...
-  void rebind(int v, TermList t);
-  bool findBinding(int var, TermList& res) const;
-  TermList apply(unsigned var) const;
-  void unbind(int var);
-  void reset();
-  bool isEmpty() const { return _map.isEmpty(); }
+  /**
+   * Bind `v` to `t`, regardless of what was there before (if anything).
+   */
+  void rebind(unsigned v, TermList t) { _map.set(v,t); }
```
mezpusz, line 51: Do these two still make sense? Can't we just use `bind` without checking the result (or possibly asserting negation)?

mezpusz:
Nice simplification. Do you know what classes remain in code that could be replaced by the newer `Substitution`? Would it make sense to hunt them down in this PR too?

Kernel/Substitution.hpp
mezpusz: Ah yes, `insert` does not override.

MichaelRawson:
Turns out that my `Substitution::bind(x, t)` is subtly wrong - to be useful for e.g. `MatchingUtils` it needs to return true if the binding is already present but is already exactly `x -> t`. @mezpusz got this right in `MapBinder`, of course.

I will first get #695 merged, then make a separate PR fixing this and removing the duplicate `Substitution`-like things. Sorry for the mess.

PR #702, opened 2025-05-30
bitfield-generating macro

MichaelRawson:
I nerd-sniped myself a bit yesterday afternoon. We currently have some mostly-manual bitfield stuff to do pointer tagging. In particular it requires a particular pattern with `_BITS_START` and `_BITS_END` which is quite annoying/fragile.

Implement this with a slightly odd-looking macro instead that works out the bit positions for you.

If you like it, great. If you think this is dark magic that should be banished, also fair enough.

### DIFF
--- a/Indexing/CodeTree.hpp
+++ b/Indexing/CodeTree.hpp
@@ -162,6 +162,8 @@ public:
     CHECK_VAR = 5,
     SEARCH_STRUCT = 6,
   };
+  static const unsigned INSTRUCTION_BITS = 3;
+  static_assert(SEARCH_STRUCT < 1 << INSTRUCTION_BITS, "Instruction should fit within INSTRUCTION_BITS");
 
   /** Structure containing a single instruction and its arguments */
   struct CodeOp
@@ -220,16 +222,15 @@ public:
     friend std::ostream& operator<<(std::ostream& out, const CodeOp& op);
 
     BITFIELD(64,
-      BITFIELD_MEMBER(unsigned, _arg, _setArg, 61,
-      BITFIELD_MEMBER(unsigned, _instruction, _setInstruction, 3,
+      BITFIELD_MEMBER(unsigned, _arg, _setArg, CHAR_BIT * sizeof(unsigned) - INSTRUCTION_BITS,
+      BITFIELD_MEMBER(unsigned, _instruction, _setInstruction, INSTRUCTION_BITS,
       END_BITFIELD
     )))
     static_assert(sizeof(void *) <= sizeof(uint64_t), "must be able to fit a pointer into a 64-bit integer");
-    static_assert(SEARCH_STRUCT < 8, "must be able to squash instructions into 3 bits");
     template<class T>
-    BITFIELD_PTR_GET(T, _data, 3)
+    BITFIELD_PTR_GET(T, _data, INSTRUCTION_BITS)
     template<class T>
-    BITFIELD_PTR_SET(T, _setData, 3)
+    BITFIELD_PTR_SET(T, _setData, INSTRUCTION_BITS)
 
   private:
     // bitfield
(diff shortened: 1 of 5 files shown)


Lib/BitUtils.hpp
```diff
@@ -117,16 +117,67 @@ class BitUtils
...
+ * For each BITFIELD_MEMBER 'foo' of type T (typically unsigned/bool), it generates two methods:
+ * T _foo() const;
+ * void _setFoo(T val);
+ */
+#define BITFIELD(bits, members)\
+  static_assert(bits >= CAR(members), "bitfield must have less than " #bits "bits");\
+  CDR(members)
+
+// generate code treating a whole bitfield as a pointer, masking out lower `mask` bits
+#define BITFIELD_PTR_GET(type, getter, mask) \
```
mezpusz, line 178: Why not use a single macro for these two as earlier?

mezpusz: I'm guessing because of the templated variants in `CodeTree`, right?

MichaelRawson: Right.

Indexing/CodeTree.hpp
```diff
@@ -219,31 +219,20 @@ class CodeTree
...
-      static_assert(alignof(T)>SEARCH_STRUCT);
-      return reinterpret_cast<T*>(BitUtils::getBits<DATA_BITS_START, DATA_BITS_END>(this->_content));
-    }
-    template<class T> void _setData(T* data) {
-      static_assert(alignof(T)>SEARCH_STRUCT);
-      BitUtils::setBits<DATA_BITS_START, DATA_BITS_END>(this->_content, reinterpret_cast<uint64_t>(data));
-    }
-    // end bitfield
+    template<class T>
+    BITFIELD_PTR_GET(T, _data, 3)
```
mezpusz, line 230: Same here, we should refer to the width of instructions.

Lib/BitUtils.hpp
mezpusz: Okay, I've just seen your first reply!

Kernel/Term.hpp
```diff
@@ -321,51 +320,28 @@ class TermList {
...
+  BITFIELD(64,
+    BITFIELD_MEMBER(uint32_t, _id, _setId, 32,
+    BITFIELD_MEMBER(uint32_t, _distinctVars, _setDistinctVars, TERM_DIST_VAR_BITS,
+    BITFIELD_MEMBER(unsigned, _order, _setOrder, 3,
+    BITFIELD_MEMBER(bool, _hasTermVar, _setHasTermVar, 1,
+    BITFIELD_MEMBER(bool, _sort, _setSort, 1,
+    BITFIELD_MEMBER(bool, _literal, _setLiteral, 1,
+    BITFIELD_MEMBER(bool, _shared, _setShared, 1,
+    BITFIELD_MEMBER(bool, _polarity, _setPolarity, 1,
+    BITFIELD_MEMBER(unsigned, _tag, _setTag, 2,
```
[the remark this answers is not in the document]
MichaelRawson: Good idea! I haven't done `bool`s on the basis that it's always 1 and shows that it should be just a number. Otherwise fixed.

mezpusz (approved):

quickbeam123:
Should we wait for [matthias-hol-terms](https://github.com/vprover/vampire/tree/matthias-hol-terms) who also changed the bitfields in `Term.hpp`. Will you know, @MichaelRawson , how to rebase/merge afterwards?

quickbeam123 (approved):
I think I like this! 

So the `_content` is hardwired via a convention, right?

quickbeam123:
I am not sure the flexibility is needed at the moment. I was just curious whether I understood where we stand in this regard.

PR #703, opened 2025-05-30
Some new option defaults

quickbeam123:
- based on extensive experiments, these lead to a more robust default on TPTP

obviously, this will affect any existing schedules and also sanity runs; let's not rush merging until we are sure we are happy with this

On TPTP v9.0.0, CNF+FOF, the performance of `-i 5000 -sa discount -awr 10` goes from `7777` up to `7829`. Goes further up to `8059` with `nwc=10`.

### DIFF
--- a/Shell/Options.cpp
+++ b/Shell/Options.cpp
@@ -1702,7 +1702,7 @@ void Options::init()
     _demodulationOnlyEquational.onlyUsefulWith(Or(_forwardDemodulation.is(notEqual(Demodulation::OFF)),_backwardDemodulation.is(notEqual(Demodulation::OFF))));
     _demodulationOnlyEquational.addProblemConstraint(hasEquality());
 
-    _extensionalityAllowPosEq = BoolOptionValue( "extensionality_allow_pos_eq","erape",true);
+    _extensionalityAllowPosEq = BoolOptionValue( "extensionality_allow_pos_eq","eape",true);
     _extensionalityAllowPosEq.description="If extensionality resolution equals filter, this dictates"
       " whether we allow other positive equalities when recognising extensionality clauses";
     _lookup.insert(&_extensionalityAllowPosEq);
--- a/Shell/Options.cpp
+++ b/Shell/Options.cpp
@@ -2273,6 +2273,13 @@ void Options::init()
     _lookup.insert(&_activationLimit);
     _activationLimit.tag(OptionTag::SATURATION);
 
+    // Even if AUTO_KBO resolves to "qkbo" or "lakbo", we still allow KBO suboptions (and possible ignore them)
+    // this is better than the default (to=auto_kbo) warning whenever we touch "kws" or "kmz" ...
+    auto KboLike = [this] {
+      return Or(_termOrdering.is(equal(TermOrdering::KBO)),
+                _termOrdering.is(equal(TermOrdering::AUTO_KBO)));
+    };
+
     _termOrdering = ChoiceOptionValue<TermOrdering>("term_ordering","to", TermOrdering::AUTO_KBO,
                                                     {"auto_kbo","kbo","qkbo","lakbo","lpo","incomp"});
     _termOrdering.description="The term ordering used by Vampire to orient equations and order literals.\n"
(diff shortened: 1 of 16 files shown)


Shell/Options.cpp
```diff
@@ -1064,7 +1064,7 @@ void Options::init()
     _positiveLiteralSplitQueueLayeredArrangement.onlyUsefulWith(_usePositiveLiteralSplitQueues.is(equal(true)));
     _positiveLiteralSplitQueueLayeredArrangement.tag(OptionTag::SATURATION);
 
-    _literalMaximalityAftercheck = BoolOptionValue("literal_maximality_aftercheck","lma",false);
+    _literalMaximalityAftercheck = BoolOptionValue("literal_maximality_aftercheck","lma",true);
     _literalMaximalityAftercheck.description =
                                    "For efficiency we perform maximality checks before applying substitutions. Sometimes this can "
                                    "lead to generating more clauses than needed for completeness. Set this on to add the checks "
```
MichaelRawson: This string probably needs changing.

Shell/Options.cpp
```diff
@@ -1698,7 +1698,7 @@ void Options::init()
     _demodulationOnlyEquational.onlyUsefulWith(Or(_forwardDemodulation.is(notEqual(Demodulation::OFF)),_backwardDemodulation.is(notEqual(Demodulation::OFF))));
     _demodulationOnlyEquational.addProblemConstraint(hasEquality());
 
-    _extensionalityAllowPosEq = BoolOptionValue( "extensionality_allow_pos_eq","erape",false);
+    _extensionalityAllowPosEq = BoolOptionValue( "extensionality_allow_pos_eq","erape",true);
```
MichaelRawson: While you're changing this, perhaps the unfortunate short name could be "eape"?

Shell/Options.cpp
```diff
@@ -2191,9 +2188,9 @@ void Options::init()
     _literalComparisonMode.addProblemConstraint(mayHaveNonUnits());
     _literalComparisonMode.addProblemConstraint(notJustEquality());
 
-    _nonGoalWeightCoefficient = NonGoalWeightOptionValue("nongoal_weight_coefficient","nwc",1.0);
+    _nonGoalWeightCoefficient = NonGoalWeightOptionValue("nongoal_weight_coefficient","nwc"); // default 10.0 is hard-wired to the constructor
```
MichaelRawson, line 2216: Does this still need to be a class that does something weird, rather than `FloatOptionValue`?

UnitTests/tEqualityResolution.cpp
```diff
@@ -55,7 +55,7 @@ TEST_GENERATION(test_02,
 TEST_GENERATION(test_03,
     Generation::AsymmetricTest()
       .input(     clause({  selected(x != f(a)), selected(p(x))  }))
-      .expected( exactly( clause({  p(f(a))                              })))
+      .expected( exactly())
```
MichaelRawson, line 58: Huh?

MichaelRawson (approved):

UnitTests/tEqualityResolution.cpp
quickbeam123: Marton's fix. I think it says that the inference will not trigger (because it does not pass the `lma` check at the end and `lma` is newly on by default.)

Shell/Options.cpp
quickbeam123: The code that uses the option actually expects two ints (which represent a ratio), but we pretend it's a float. 

I was just too lazy to touch it more. (And please appreciate that hiding the default and adding the comment was an improvement over the default there, whose change actually did not have any effect! ;))

Shell/Options.cpp
quickbeam123: A good idea! 

(Although, it also means changing all the samplers mentioning the option and my snake scripts... So I will sleep on it first.)

Shell/Options.cpp
MichaelRawson: >whose change actually did not have any effect!

Oh. *Oh*. Much appreciated!

PR #704, opened 2025-05-30
switch existing code to use Substitution

MichaelRawson:
Take 2 from #701. Now introduce `bindUnbound` for the old cases where we want to crash if there is already a binding, and use `bind` for the `MatchingUtils` interface. Replaces `MapBinderAndApplicator` and `SimpleSubstitution`.

### DIFF
--- a/SATSubsumption/SATSubsumptionAndResolution.cpp
+++ b/SATSubsumption/SATSubsumptionAndResolution.cpp
@@ -69,7 +69,6 @@
    */
 
 #include "Kernel/Matcher.hpp"
-#include "Kernel/SubstHelper.hpp"
 #include "Lib/Environment.hpp"
 #include "Lib/Int.hpp"
 #include "Shell/Statistics.hpp"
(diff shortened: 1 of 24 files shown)


SATSubsumption/SATSubsumptionAndResolution.cpp
```diff
@@ -987,6 +986,9 @@ SimpleSubstitution SATSubsumption::SATSubsumptionAndResolution::getBindingsForSu
     Match match = _matchSet.getMatchForVar(var);
     Literal *l = (*_sidePremise)[match.i];
     Literal *k = (*_mainPremise)[match.j];
+    // matchArgs doesn't like nullary for some reason
```
quickbeam123, line 989: Why did it work with `SimpleSubstitution`? Or did it?

MichaelRawson: I don't think it did, I should have tested my late changes to #695 more thoroughly.

quickbeam123: And what were the symptoms? Is master currently in danger of being mildly broken?

MichaelRawson: Crash with `-p smtcheck` in a weird case, i.e. a specific proof output mode crashes after the empty clause is found. Regular ol' Vampire stuff should be fine, hence my lack of caution.

quickbeam123: OK, thank you for the explanation!

quickbeam123 (approved):

PR #706, opened 2025-06-04
Add assert-synth keyword for synthesis

mezpusz:
Adding special support for synthesis with forall-exists conjectures, via a new SMTLIB2 command `assert-synth`. This assumes a forall and an exist quantifier block in the beginning implicitly, and makes it easier to recognize such "synthesis tasks" in the future.

### DIFF
--- a/Parse/SMTLIB2.cpp
+++ b/Parse/SMTLIB2.cpp
@@ -2824,31 +2824,24 @@ void SMTLIB2::readAssertSynth(LExpr* forall, LExpr* exist, LExpr* body)
     std::cout << "% WARNING: Found an assert-synth command but synthesis is not enabled. Consider running with '-qa synthesis'." << endl;
   }
 
-  auto fvars = VList::empty();
-  auto fsorts = SList::empty();
-  auto fRdr = READER(forall);
-  while (fRdr.hasNext()) {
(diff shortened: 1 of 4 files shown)


Parse/SMTLIB2.cpp
```diff
@@ -2804,6 +2816,55 @@ void SMTLIB2::readAssertNot(LExpr* body)
...
+    auto sort = parseSort(pRdr.readExpr());
+    tryInsertIntoCurrentLookup(name, var, sort);
+    VList::push(var.var(), fvars);
+    SList::push(sort, fsorts);
+  }
+
+  auto evars = VList::empty();
+  auto esorts = SList::empty();
+  auto eRdr = READER(exist);
+  while (eRdr.hasNext()) {
```
hzzv: Would it make sense to move this loop into a helper function to avoid code duplication?

quickbeam123 (approved):

Parse/SMTLIB2.cpp
mezpusz: Good idea! I couldn't find any other occurrences of exactly the same variable list parsing elsewhere, so I added a local lambda, please check.

hzzv: Looks good, thanks! :)

PR #712, opened 2025-06-09
chore: fix some typos

euprunin:


Saturation/SaturationAlgorithm.cpp
```diff
@@ -1280,7 +1280,7 @@ void SaturationAlgorithm::doOneAlgorithmStep()
 
   /*
    * Only after processing the whole input (with the first call to doUnprocessedLoop)
-   * it is time to recored for LRS the start time (and instrs) for the first iteration.
+   * it is time to recorded for LRS the start time (and instrs) for the first iteration.
```
MichaelRawson: "record"

SATSubsumption/SATSubsumptionAndResolution.cpp
```diff
@@ -47,7 +47,7 @@
    *
    * ----- Subsumption Resolution: -----
    * Let L and M be two clauses considered as sets. L and M are said to be the base and instance
-   * of a subsumption resolution inference, respectively iif
+   * of a subsumption resolution inference, respectively if
```
MichaelRawson: Probably "iff" was intended.

MichaelRawson (approved):
Thanks - I think a couple were disambiguated wrong, could you fix these? Then I'll merge.

SATSubsumption/SATSubsumptionAndResolution.cpp
euprunin: Makes sense. Fixed!

Saturation/SaturationAlgorithm.cpp
euprunin: Ah, of course. Thanks!

PR #713, opened 2025-06-10
SAT: only one assumption mechanism, remove some unused oddities

MichaelRawson:
First in a series of patches to clear up the `SATSolver` interface before I start hacking it about to do e.g. proofs.

@quickbeam123 left a nice comment here from ...2015?, reproduced below
```
 // The first three functions represent the original TWLSolver-style of interface ...

  /**
   * Add assumption to the solver. Perhaps process partially,
   * but mainly ensure the assumption is considered during the next calls to solve()
   */
  virtual void addAssumption(SATLiteral lit) = 0;

  /**
   * Retract all the assumptions added so far.
   * The solver becomes assumption-free.
   *
   * Note: this may destroy the model computed during a previous call to solve
   * (as it currently does in TWL).
   */
  virtual void retractAllAssumptions() = 0;

  /**
   * Test whether any assumptions are currently registered by the solver.
   */
  virtual bool hasAssumptions() const = 0;

  // ... a better alternative could be the interface below.

  // It is currently implemented in terms of the one above
  // (but may be overridden in SATSolver implementations).
  // Note the the below interface allows the solver
  // to identify a subset of the given assumptions that were only needed for
  // previous contradiction to be derived.

  // Note also, however, that solveUnderAssumptions cannot guarantee
  // access to the model after it returns SATISFIABLE, as it uses retractAllAssumptions
  // to clean in the end.
```

The "better alternative" is indeed better! And it conforms more easily with the API of the SAT solvers we use, so let's just have one.

Also:
- remove/rename a couple of confusing overloads, the way C++ resolves virtual calls is bonkers
- don't probe for failed literals during `solveUnderAssumptions`, we might not need it - do it all in `failedLiterals`

### DIFF
--- a/SAT/MinisatInterfacing.cpp
+++ b/SAT/MinisatInterfacing.cpp
@@ -40,30 +40,26 @@ unsigned MinisatInterfacing::newVar()
   return minisatVar2Vampire(_solver.newVar());
 }
 
-SATSolver::Status MinisatInterfacing::solveUnderAssumptions(const SATLiteralStack& assumps, unsigned conflictCountLimit)
+SATSolver::Status MinisatInterfacing::solveUnderAssumptionsLimited(const SATLiteralStack& assumps, unsigned conflictCountLimit)
 {
-  ASS(!hasAssumptions());
-
   // load assumptions:
+  _assumptions.clear();
   SATLiteralStack::ConstIterator it(assumps);
   while (it.hasNext()) {
     _assumptions.push(vampireLit2Minisat(it.next()));
   }
 
   solveModuloAssumptionsAndSetStatus(conflictCountLimit);
-
-  if (_status == Status::UNSATISFIABLE) {
-    // unload minisat's internal conflict clause to _failedAssumptionBuffer
-    _failedAssumptionBuffer.reset();
-    Minisat::LSet& conflict = _solver.conflict;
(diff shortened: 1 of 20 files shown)


MichaelRawson:
Clean shuffle with `i=1000`.

SAT/MinisatInterfacing.cpp
```diff
@@ -40,30 +40,26 @@ unsigned MinisatInterfacing::newVar()
...
-      _failedAssumptionBuffer.push(minisatLit2Vampire(conflict[i]).opposite());
-    }
-  }
-
-  _assumptions.clear();
+SATLiteralStack MinisatInterfacing::failedAssumptions() {
+  ASS_EQ(_status, Status::UNSATISFIABLE)
 
-  return _status;
+  SATLiteralStack result;
```
quickbeam123, line 59: Aren't we afraid this could be more expensive than return a references to the `_failedAssumptionBuffer`?

Betting on RWO and fast allocation instead?

MichaelRawson: It could be, but (i) this is safer and (ii) hopefully we are only calling `failedAssumptions` after at least one call to `solve`, so I'd argue it's not the biggest deal.

PR #716, opened 2025-06-10
chore: fix some typos

euprunin:


MichaelRawson (approved):
Thanks again. You could ignore `scripts/` for now if you like, it's very old stuff that I will get around to auditing and probably delete most of.

PR #718, opened 2025-06-11
chore: fix remaining known typos

euprunin:


MichaelRawson:
In short "yes", and we can break options a bit.

1. `_maintainanceStr` (sic) - this is fine as long as you replace all three occurrences in `Index.hpp` - you missed some above.
2. The FMB option `fmbsso=occurence` (sic) - this is also OK, as long as you replace all of those in the diff.

PR #724, opened 2025-06-27
prevent FMB crashes on FOOLish inputs

MichaelRawson:
The attached SMT-LIB input causes the finite model builder to crash while printing a model, mostly (completely?) down to FOOL constants. I don't think this is a complete fix, but it's a start.
[input.txt](https://github.com/user-attachments/files/20949028/input.txt)


quickbeam123:
Hmm, maybe I keep this PR as a reminder (for now), that I should properly understand debug model printing of unsorted / multisorted / fool-sorted inputs. Will use this as a starting point...

PR #727, opened 2025-07-07
Induction formula refactor

mezpusz:
This is the first step in trying to create induction formulas in a uniform way, in preparation for recursive synthesis, where we will need to take *all* our current induction features on universally quantified conjectures and apply them to forall-exists conjectures.

In particular, this PR abstracts away formula creation using `InductionTemplate`, `InductionCase` and `InductionUnit` structures. This covers all our current induction formulas, although I had to slightly modify `sik=three` to make it work, using a single binary subterm predicate for each term algebra instead of always introducing fresh unary ones that fix one of the arguments.

Additionally, some sanity tests are added to term algebra and inductive reasoning, and theory axiom addition is performed more carefully when ALASCA is enabled.


Inferences/Induction.cpp
```diff
@@ -1219,414 +1221,137 @@ void InductionClauseIterator::performInfIntInduction(const InductionContext& con
...
-  // make smallerThanY
-  unsigned sty = env.signature->addFreshPredicate(1,"smallerThan");
-  env.signature->getPredicate(sty)->setType(OperatorType::getPredicateType({sort}));
-
-  // make ( y = con_i(..dec(y)..) -> smaller(dec(y)))  for each constructor 
-  FormulaList* conjunction = FormulaList::singleton(Ly);
-  for(unsigned i=0;i<ta->nConstructors();i++){
-    TermAlgebraConstructor* con = ta->constructor(i);
-    unsigned arity = con->arity();
+  // cout << "using template" << endl;
```
hzzv: Remove this?
Similarly on line 1352.

Inferences/Induction.cpp
```diff
@@ -1219,414 +1221,137 @@ void InductionClauseIterator::performInfIntInduction(const InductionContext& con
...
-      j = 0;
-      auto rr = rn.apply(r->apply(typeSubst));
-      for (unsigned i = 0; i < rr->arity(); i++) {
-        if (indPos[i]) {
-          ts[j++] = *rr->nthArgument(i);
-        }
+    for (const auto& hyp : c.hypotheses) {
+      auto hypVars = VList::empty();
+      auto hypF = context.getFormula(hyp, typeBinder, var, &hypVars);
+      // quantify each hypotheses with variables replacing Skolems explicitly
```
hzzv: hypothesis

Inferences/Induction.cpp
```diff
@@ -1219,414 +1221,137 @@ void InductionClauseIterator::performInfIntInduction(const InductionContext& con
...
+        hypF = new QuantifiedFormula(Connective::FORALL, hypVars, SList::empty(), hypF);
       }
-      FormulaList::push(context.getFormula(ts,true),hyps);
+      FormulaList::push(hypF, hyps);
     }
-    FormulaList::push(hyps ?
-      new BinaryFormula(Connective::IMP,JunctionFormula::generalJunction(Connective::AND,hyps),right) : right, formulas);
-    var = rn.nextVar();
+    auto left = hyps ? JunctionFormula::generalJunction(Connective::AND, hyps) : nullptr;
+    auto hypVars = VList::fromIterator(c.hypUnivVars.iterFifo());
```
hzzv, line 1338: To be honest I got a bit confused at this point. Which variables are hypVars here compared to those on line 1329? Could you maybe clarify that in a comment?
I get that this method is working with _any_ induction template, but maybe it would be good to have some simple example similar to the comment on line 1289, e.g. for structural induction on a formula F[x] where x is a list (and possibly with some more variables??).

Kernel/InductionTemplate.cpp
```diff
@@ -0,0 +1,141 @@
...
+
+#include "Kernel/Term.hpp"
+#include "Kernel/TermIterators.hpp"
+
+using namespace std;
+
+namespace Kernel {
+
+template<typename S, typename T>
+ostream& outputImplication(ostream& out, const Stack<S>& antecedent, const T& succeedent, bool outerParentheses = true)
```
hzzv: I think it's "succedent"

Kernel/InductionTemplate.hpp
```diff
@@ -0,0 +1,86 @@
...
+  LiteralStack conditions;
+  VStack condUnivVars;
+};
+
+/**
+ * A formula template corresponding to a single case within an induction formula,
+ * of the form ∀(∀ x_1,...,x_k.(U_1 ⋀ ... ⋀ U_m) → U), where
+ * @b conclusion is the unit U,
+ * @b hypotheses is the list of units U_1,...,U_n, and
+ * @b hypUnivVars is the list of variables x_,...,x_k.
```
hzzv: "x_," -> "x_1,"

Kernel/InductionTemplate.hpp
```diff
@@ -0,0 +1,86 @@
...
+
+  InductionUnit conclusion;
+  Stack<InductionUnit> hypotheses;
+  VStack hypUnivVars;
+};
+
+/**
+ * An induction formula template corresponding to ∀F(C_1 ⋀ ... ⋀ C_n → C),
+ * where @b cases is the list C_1,...,C_n and @b conclusion is C.
+ * The sorts of @b F_terms members must equal to @b sorts and variables
```
hzzv: "must be equal"

hzzv (commented):
Overall looks good to me, I have just a couple of minor comments.

Inferences/Induction.cpp
mezpusz: I added some comments on the quantification of each part in `InductionTemplate.hpp` and some cross-references for examples and how parts connect.

I agree that the combination of all these features is a bit too complex at this point, but I think each individual feature is well documented and there are various unit tests with their combinations (although not showing the intermediate induction formulas, maybe we could do this somehow), so I wouldn't show an example in this particular place.

Let me know what you think about the current state!

hzzv: This works, thanks! :)

hzzv (approved):

Lib/DHMap.hpp
```diff
@@ -323,7 +335,7 @@ class DHMap
    * the value with @b initial, and return true. Otherwise,
    * return false.
    */
-  bool getValuePtr(Key key, Val*& pval, const Val& initial)
+  bool getValuePtr(Key key, Val*& pval, Val initial)
   {
```
quickbeam123, line 339: Did you somehow check this does not affect other users of this function?

(Shouldn't matter for small `Val`, but could for large...)

Shell/TheoryAxioms.cpp
```diff
@@ -957,154 +957,154 @@ void TheoryAxioms::addBooleanArrayWriteAxioms(TermList arraySort)
...
-                                 Theory::INT_MULTIPLY, Theory::INT_QUOTIENT_E, Theory::INT_DIVIDES,
-                                 Theory::INT_REMAINDER_E, Theory::INT_ABS, zero,one);
-      }
-      else if(haveIntDivides){ 
-        Stack<TermList>& ns = env.signature->getDividesNvalues(); 
-        Stack<TermList>::Iterator nsit(ns);
-        while(nsit.hasNext()){
-          TermList n = nsit.next();
-          addIntegerDividesAxioms(Theory::INT_DIVIDES,Theory::INT_MULTIPLY,zero,n); 
+  if (env.options->alasca()) {
```
quickbeam123, line 962: Can you recall what the change was here?

What does the ALASCA branch get which it did not before? I guess it want's to do arithmetic stuff its own way, but maybe other axioms should still be added as usual (FOOL, term algebras, arrays?)

Lib/DHMap.hpp
mezpusz: I see your point... How about adding one more variant with rvalue reference `Val&&`? (`DHMap` could also use some refactor as it contains lots of duplication, although maybe not the scope of this PR.)

Alternatively, I can just try to use a different structure, or try really hard to make it work with the current interface.

Shell/TheoryAxioms.cpp
mezpusz: This was the fix which reenabled all non-arithmetic theory axioms when `alasca=on`. In the end, we use default `alasca=off` but the theory axioms are still not used in the other case. We should be a bit careful though that I did not introduce any errors upon merging with master.

mezpusz:
A reminder that this needs debug sampling and regression testing over some induction benchmarks.

Lib/DHMap.hpp
quickbeam123: Checked already. It's OK. Sorry for the question!

Shell/TheoryAxioms.cpp
quickbeam123: Yes, we should be careful here. But don't have good sanity in for theories yet (that will come from SMTLIB). 

So unless you have some idea for a more thorough check we take this in now?

mezpusz: The term algebra functionality should be covered by some of the new sanity checks (see `checks/term-algebra`). Otherwise, if I ask GitHub to ignore whitespaces in the diff, it seems OK -- we branch on ALASCA or old arithmetic axioms, after which come array and term algebra axioms without branching.

mezpusz:
I put it on hold, but regression testing is underway.

PR #732, opened 2025-07-16
Casc2025

quickbeam123:
This is the version of Vampire Geoff got for CASC, I don't think it will change anymore, but maybe (unlikely) if he discovers a soundness bug. I mainly want to use the PR facility right now, so it's easier to see the diff. No need to merge quickly, but let's discuss how to proceed. 

Possible TODOs (that should for clarity go to different followup branches):
- FMB Models should support symbolic definitions of eliminated predicates and functions so that large-arity-redundant-symbols don't kill the model printing phase (this needs to be reflected on the side of the model_checking mode)
- I now have enough data to start the option elimination process we always dreamt of; we'll need new schedules when this is done, which I can also regenerate 
- it could be nice for debugging portfolio mode that branching schedules would drop a line to stdout about which branch got chosen and why (especially since the new and more faithful-to-reality way of printing strategies prevents an easy "CTRL+F" in Schedules.cpp)
- Marton has a proper fix for the `proc=on:drc=encompass` combination's incompleteness (which for Geoff we just explicitly marked as such); I tried to make the changes in such a way that there will be a merge conflict in Options where we will lift the incompleteness case - but maybe I should rerun all `proc=on` strategies we care about, so that Marton's change is reflected in the final planned schedules

So let's here just discuss what's looking too ugly to go to master immediately and I will try to polish it, before building the above branches on top.


CASC/PortfolioMode.cpp
```diff
@@ -237,65 +237,37 @@ bool PortfolioMode::prepareScheduleAndPerform(const Shell::Property& prop)
...
-    addScheduleExtra(main,schedule,"si=on:rtra=on");
-    addScheduleExtra(fallback,schedule,"si=on:rtra=on");
-
-  } else if (env.options->schedule() == Options::Schedule::SMTCOMP) {
-    // Normally we do main fallback main_extra fallback_extra
-    // However, in SMTCOMP mode the fallback is universal for all
-    // logics e.g. it's not very strong. Therefore, in SMTCOMP
-    // mode we do main main_extra fallback fallback_extra
-
+  if (env.options->schedule() == Options::Schedule::SMTCOMP) {
```
MichaelRawson, line 240: Just for backwards-compatibility, right? No inherent reason we should do this for SMT-COMP.

CASC/PortfolioMode.cpp
```diff
@@ -237,65 +237,37 @@ bool PortfolioMode::prepareScheduleAndPerform(const Shell::Property& prop)
...
 
   if (schedule.isEmpty()) {
     USER_ERROR("The schedule is empty.");
   }
 
-  return runScheduleAndRecoverProof(std::move(schedule));
+  // depending on _numWorkers, we use a certain number of champions to go first (and run for very long)
+  // - champions are selected to cover as much as possible by themselves
+  // - at the same time, "quick" is build so that it covers (again) even those problems covered by champions,
+  //   but does not go after them as eagerly as those that remained truly uncovered
```
MichaelRawson, line 254: This is a really nice solution to a thorny problem. Good stuff.

CASC/PortfolioMode.cpp
```diff
@@ -372,19 +344,27 @@ void PortfolioMode::getSchedules(const Property& prop, Schedule& quick, Schedule
     Schedules::getSnakeTptpSatSchedule(prop,quick);
     break;
 
-  case Options::Schedule::CASC_2024:
+  case Options::Schedule::CASC_2025:
```
MichaelRawson, line 347: Is now the time to drop `CASC_1998` or whatever the oldest schedule is?

CASC/Schedules.hpp
```diff
@@ -27,22 +27,25 @@ class Schedules
...
+
+  static void getSmtcomp2018Schedule(const Shell::Property& property, Schedule& quick);
 
   static void getLtb2017Hh4Schedule(const Shell::Property& property, Schedule& sched);
   static void getLtb2017IsaSchedule(const Shell::Property& property, Schedule& sched);
   static void getLtb2017HllSchedule(const Shell::Property& property, Schedule& sched);
   static void getLtb2017MzrSchedule(const Shell::Property& property, Schedule& sched);
   static void getLtb2017DefaultSchedule(const Shell::Property& property, Schedule& sched);
 
-  static void getInductionSchedule(const Shell::Property& property, Schedule& quick, Schedule& fallback);
```
MichaelRawson, line 41: Is there a pithy reason why we don't have fallback schedules any more? Or why we had them in the first place? Just curious.

FMB/FiniteModelMultiSorted.cpp
```diff
@@ -792,7 +862,327 @@ void FiniteModelMultiSorted::eliminateSortFunctionsAndPredicates(const Stack<uns
   }
 }
 
-bool FiniteModelMultiSorted::evaluate(Unit* unit)
+void FiniteModelMultiSorted::restoreEliminatedFunDef(Problem::FunDef* fd)
```
MichaelRawson, line 865: In the interest of time I'm going to assume you know what you're doing here. Do we have to do this in duplicate for mono- and multi-sorted logic?

Kernel/Problem.hpp
```diff
@@ -67,23 +70,70 @@ class Problem {
...
-  void addTrivialPredicate(unsigned pred, bool assignment);
+  enum class IntereferenceKind : unsigned char {
+    FUN_DEF = 0,
+    PRED_DEF = 1,
+    GLOB_FLIP = 2,  // backs up random_polarities
+    COND_FLIP = 3,  // backs up partially  predicates (Plaisted-Greenbaum-style p<=>q reduced to, e.g. p=>q) and eliminated blocked clauses
+  };
+
+  /* Intereference (satisfiability-preserving-only transformations) recording */
+  struct Intereference {
```
MichaelRawson: Typo - "intereference"/"interference" - normally I'd let it slide in the interest of authenticity, but I think this would drive me crazy, sorry.

Kernel/Problem.hpp
```diff
@@ -67,23 +70,70 @@ class Problem {
...
-   * Return map of trivial predicates into their assignments.
-   *
-   * Trivial predicates are the predicates whose all occurrences
-   * can be assigned either true or false.
-   */
-  const TrivialPredicateMap& trivialPredicates() const { return _trivialPredicates; }
+   * Pushing here during preprocessing,
+   * replaying backwards to get a model of the original signature.
+  */
+  Stack<Intereference*> intereferences;
```
MichaelRawson: Could this be `unique_ptr`'d or otherwise managed?

MichaelRawson: `std::vector` may help, I'm not sure how correct `Stack` is yet for move semantics, weird initialisers and so on. Also, death to homebrew containers.

vampire.cpp
```diff
@@ -560,8 +560,9 @@ void dispatchByMode(Problem* problem)
     env.options->setOutputMode(Options::Output::SZS);
     env.options->setProof(Options::Proof::TPTP);
     env.options->setOutputAxiomNames(true);
-    env.options->setNormalize(true);
-    env.options->setRandomizeSeedForPortfolioWorkers(false);
+
```
MichaelRawson, line 563: Did you mean to leave these in?

Shell/Options.cpp
```diff
@@ -2087,6 +2113,7 @@ void Options::init()
     _lookup.insert(&_splittingMinimizeModel);
     _splittingMinimizeModel.tag(OptionTag::AVATAR);
     _splittingMinimizeModel.onlyUsefulWith(_splitting.is(equal(true)));
+    _splittingMinimizeModel.addHardConstraint(If(equal(SplittingMinimizeModel::SCO)).then(_cleaveNonsplittables.is(notEqual(true))));
```
MichaelRawson, line 2128: Didn't we fix this with the stickiness?

Shell/Options.cpp
```diff
@@ -2046,6 +2066,12 @@ void Options::init()
     _splitAtActivation.onlyUsefulWith(_splitting.is(equal(true)));
     _splitAtActivation.tag(OptionTag::AVATAR);
 
+    _cleaveNonsplittables = BoolOptionValue("cleave_nonsplittables","cn",false);
+    _cleaveNonsplittables.description="Tentatively propose single-literal component strengthenings. Sometimes useful for bringing about finite saturations.";
```
MichaelRawson, line 2082: Weirdly seems to be in unsat schedules too. I guess these aren't just noise? I really don't know what I unleashed.

MichaelRawson (approved):
This is much less unpleasant than I thought it would be! Thanks for all the hard work fixing non-theorem output, that was a big one.

Shell/Options.cpp
quickbeam123: Are you sure? Note that EPR mixes SAT and UNS into one schedule.

Shell/Options.cpp
quickbeam123: No, stickiness just works despite `amm=all`. SCO brings a new nightmare. 

However, SCO is my first candidate for option (value) removal. Stay tuned!

Shell/Options.cpp
```diff
@@ -3577,6 +3616,10 @@ bool Options::complete(const Problem& prb) const
   if (_demodulationRedundancyCheck.actualValue == DemodulationRedundancyCheck::OFF) {
     return false;
   }
+  if (_demodulationRedundancyCheck.actualValue == DemodulationRedundancyCheck::ENCOMPASS &&
```
quickbeam123: This is the bit we want to remove when merging with Marton's proper fix (and thinking about schedules at the same time).

Shell/Options.cpp
MichaelRawson: Ahah - yes, fooled by EPR.

Kernel/Problem.hpp
quickbeam123: I trust `Stack` but am fine with `vector` too. `unique_ptr` is a good idea.

quickbeam123:
> Should any of the existing PRs come before/after this one?

The whole point of this branch was that master can move on independently. No strict ordering I am aware of (just being aware of the interactions with https://github.com/vprover/vampire/pull/726)

CASC/PortfolioMode.cpp
quickbeam123: Indeed. This can go as soon as we have new SMTCOMP schedules. Hopefully in 2026 the latest :)

CASC/PortfolioMode.cpp
quickbeam123: It was useful to have last year, to compare against during testing. I would keep it there a bit longer. (Nothing older for CASC proper there than 2024!)

There is a different question about `getLtb2017*` but I think Josef still used it relatively recently.

CASC/Schedules.hpp
quickbeam123: First I checked that `fallback` does not add anything useful with any of the schedules we had. Functionally, you can just push to `quick` with the semantics of "fallback" when you are done with your "quick". See e.g.
```
  Schedule& fallback = quick;
```
in `getSmtcomp2018Schedule`.

Then I decided to try this new game with `champions` which need to travel through the wider interface to `PortfolioMode.cpp` because their use depends on the number of workers to spawn.

quickbeam123: So the short answer is. I killed `fallback`, but newly use the second `Schedule&` for this new thing, semantically.

FMB/FiniteModelMultiSorted.cpp
quickbeam123: I don't maintain `FiniteModel`(non-multisorted) and I don't think we have to (currently, all model printing goes through `FiniteModelMultiSorted`). Feel free to remove it!

Kernel/Problem.hpp
quickbeam123: Fixed!

CASC/Schedules.hpp
MichaelRawson: Cool. Thanks for the explanation!

vampire.cpp
quickbeam123: I think Vampire is strong enough to live without these. (Normalizing is not brave and it's adding one layer of complexity to the strat building process.)

So vote for deleting, but let's wait for the results. Maybe Geoff really shuffles and we do want to normalize after all.

Shell/Options.cpp
```diff
@@ -3420,10 +3453,17 @@ std::string Options::generateEncodedOptions() const
     forbidden.insert(&_decode);
     forbidden.insert(&_sampleStrategy);
     forbidden.insert(&_normalize);
+    forbidden.insert(&_outputAxiomNames);
+    forbidden.insert(&_randomizSeedForPortfolioWorkers);
```
mezpusz: "_randomizSeedForPortfolioWorkers" -> "_randomizeSeedForPortfolioWorkers"

FMB/FiniteModelBuilder.cpp
```diff
@@ -413,7 +413,7 @@ void FiniteModelBuilder::init()
   if(env.options->fmbAdjustSorts() == Options::FMBAdjustSorts::PREDICATE){
     DArray<bool> deleted_functions(env.signature->functions());
     for(unsigned f=0;f<env.signature->functions();f++){
-      deleted_functions[f] =  (bool)_prb.getEliminatedFunctions().findPtr(f) || env.signature->getFunction(f)->usageCnt()==0;
+      deleted_functions[f] = env.signature->getFunction(f)->usageCnt()==0;
```
mezpusz, line 416: Just a passing thought, as I have recently seen that usage counts are not updated for type constructors for example [here](https://github.com/vprover/vampire/blob/513a213c3a39720520eef6d1c02b6c744288031f/Shell/Property.cpp#L111). Is there a reason this (or any similar) operation is skipped for them?

FMB/FiniteModelMultiSorted.cpp
```diff
@@ -214,31 +218,34 @@ std::string FiniteModelMultiSorted::toString()
   //Constants
   for(unsigned f=0;f<env.signature->functions();f++){
     Signature::Symbol* symb = env.signature->getFunction(f);
-    if(symb->usageCnt()==0) continue;
+    // if(symb->usageCnt()==0) continue;
```
mezpusz, line 221: Can you please add a comment why this is temporarily disabled?

FMB/FiniteModelMultiSorted.hpp
```diff
@@ -86,23 +86,46 @@ class FiniteModelMultiSorted {
...
+  void restoreEliminatedDefinitions(Kernel::Problem* prob);
 
   std::string toString();
 
 private:
+  unsigned evaluateTerm(TermList, const DHMap<unsigned,unsigned>& subst);
+  bool evaluateLiteral(Literal*, const DHMap<unsigned,unsigned>& subst);
+  bool evaluateFormula(Formula*, DHMap<unsigned,unsigned>& subst);
+
+  // if term evaluation encounters a missing record, it assumes the correspondig symbol has been implicitly eliminated
```
mezpusz: correspondig -> corresponding

FMB/FiniteModelMultiSorted.hpp
```diff
@@ -86,23 +86,46 @@ class FiniteModelMultiSorted {
...
+  // (e.g., eliminated unused function definition f(X) = g(X,c) might have eliminated c, if it did not occur anywhere else)
+  // such symbols are restored (just after restoreEliminatedDefinitions; although, formally it should happen before) in the simplest possible way:
+  // functions == 1 (the first domain element of the respective sort) everwhere
```
mezpusz: everwhere -> everywhere

Kernel/Problem.cpp
```diff
@@ -192,49 +192,171 @@ void Problem::copyInto(Problem& tgt, bool copyClauses)
...
- * This information may be used during model output
+ * This information may be used during model output.
  */
 void Problem::addTrivialPredicate(unsigned pred, bool assignment)
 {
-  ALWAYS(_trivialPredicates.insert(pred, assignment));
+  // create linear literal (~)pred(X0,X1,...X_arity)
+  TermStack args;
+  for (unsigned v = 0; v < env.signature->getPredicate(pred)->arity(); v++) {
+    args.push(TermList(v,false));
```
mezpusz: Maybe use the new sugary `TermList::var(v)`?

Kernel/Problem.cpp
```diff
@@ -192,49 +192,171 @@ void Problem::copyInto(Problem& tgt, bool copyClauses)
...
  *
  * This information may be used during model output
  */
-void Problem::addEliminatedFunction(unsigned func, Literal* definition)
-{
-  ASS(definition->isEquality());
+void Problem::addPartiallyEliminatedPredicate(unsigned pred, Formula* remainingImplication)
+{
+  Formula* f;
+  bool isQuantified=remainingImplication->connective()==FORALL;
```
mezpusz: If `remainingImplication` is not needed for anything else, is it possible to assign `remainingImplication->qarg()` into that, shortening the code a bit?

Kernel/Problem.cpp
```diff
@@ -192,49 +192,171 @@ void Problem::copyInto(Problem& tgt, bool copyClauses)
...
+  ASS_EQ(f->connective(),IMP)
+
+  Formula* cond;
+  Literal* new_atom_val;
+  bool neg = false;
+  if (f->left()->connective()==LITERAL && f->left()->literal()->functor()==pred) {
+    new_atom_val =f->left()->literal();
+    cond = f->right();
+  } else {
+    ASS(f->right()->connective()==LITERAL && f->right()->literal()->functor()==pred)
```
mezpusz, line 263: Could you please replace this with two `ASS_EQ`s?

Kernel/Problem.cpp
```diff
@@ -192,49 +192,171 @@ void Problem::copyInto(Problem& tgt, bool copyClauses)
...
+  }
+
+  interferences.push(res);
+}
+
+void Problem::addEliminatedBlockedClause(Clause* cl, unsigned blockedLiteralIndex) {
+  Literal* bll = (*cl)[blockedLiteralIndex];
+  Formula* cond= Formula::fromClause(cl,/* closed= */false);
+
+  bool needsFixpoint = false;
```
mezpusz: This could be replaced with some shorter iterator magic, if you like those.

Shell/Options.cpp
```diff
@@ -210,10 +212,16 @@ void Options::init()
...
     _slowness.onlyUsefulWith(UsingPortfolioTechnology());
 
-    _randomizSeedForPortfolioWorkers = BoolOptionValue("randomize_seed_for_portfolio_workers","",true);
-    _randomizSeedForPortfolioWorkers.description = "In portfolio mode, let each worker process start from its own independent random seed.";
-    _lookup.insert(&_randomizSeedForPortfolioWorkers);
-    _randomizSeedForPortfolioWorkers.onlyUsefulWith(UsingPortfolioTechnology());
+    _randomizeSeedForPortfolioWorkers = BoolOptionValue("randomize_seed_for_portfolio_workers","",true);
+    _randomizeSeedForPortfolioWorkers.description = "In portfolio mode, let each worker process start from its own independent random seed.";
+    _lookup.insert(&_randomizeSeedForPortfolioWorkers);
+    _randomizeSeedForPortfolioWorkers.onlyUsefulWith(UsingPortfolioTechnology());
```
mezpusz, line 222: Do these have a meaningful tag?

FMB/FiniteModelBuilder.cpp
quickbeam123: I think they come from an age when type constructors were not a thing. This does not matter much (at all) for FMB which (at least mentally) I keep compatible only with cnf/fof/tff. 

But it's a good point. And unused sort elimination is a valid (although trivial) model building technique.

FMB/FiniteModelMultiSorted.cpp
quickbeam123: It's not temporary. So I should probably delete the line. 

Explanation: eliminated symbols must be given a representation in the final model, otherwise it does not model check the original formula.

Why not done yet: I plan to keep refactoring FiniteModelMultiSorted.cpp either in this PR or on a new branch.

Kernel/Problem.cpp
quickbeam123: I don't think this can be done in a nice way. As purely propositional definitions are not quantified. And if you look at the call site, the information has already been lost there (so we would have to drag it along from `makeImplFromDef` in `PredicateDefinition.cpp`

Kernel/Problem.cpp
quickbeam123: A good suggestion!

Kernel/Problem.cpp
quickbeam123: It would take me longer to figure out how to write it than to do this. Maybe the same for reading. It's not that bad to read, is it? ;)

Shell/Options.cpp
quickbeam123: We should think the TAGs through a bit more. They use to have little semantics until recently, but now some uses for tagging start materializing and some, like PREPROCESSING must be set correctly for good option warnings. 

Here we could have a new tag PORTFOLIOS ?

Kernel/Problem.cpp
mezpusz: It's not bad, I also usually write something similar and then realize that with the iterator the variable is sometimes not needed, I cannot forget `break`ing out, etc. It's up to you though, the code is something like:
```c++
bool needsFixpoint = iterTraits(cl->iterLits()).any([&](Literal* other) {
  return other->functor() == bll->functor() && other->polarity() != bll->polarity();
});
```

Kernel/Problem.cpp
mezpusz: I just meant here locally, writing:
```c++
if (remainingImplication->connective()==FORALL) {
  remainingImplication = remainingImplication->qarg();
}
```
and then using `remainingImplication` instead of `f`. But I'm not opposed to this either, and I realize that this is much longer to write than `f`...

FMB/FiniteModelBuilder.cpp
mezpusz: Sure, maybe they have no meaning here. Thanks for checking!

FMB/FiniteModelMultiSorted.cpp
mezpusz: Okay, this makes sense!

Shell/Options.cpp
mezpusz: PORTFOLIO sounds good! Otherwise we can come back to it whenever tags are rethought.

Saturation/Splitter.cpp
```diff
@@ -396,39 +396,54 @@ void SplittingBranchSelector::updateSelection(unsigned satVar, SATSolver::VarAss
   SplitLevel posLvl = _parent.getNameFromLiteral(SATLiteral(satVar, true));
   SplitLevel negLvl = _parent.getNameFromLiteral(SATLiteral(satVar, false));
 
+  bool posUsed = _parent.isUsedName(posLvl);
+  bool negUsed = _parent.isUsedName(negLvl);
```
mezpusz, line 400: Does it make sense to add an assertion if these are mutually exclusive or is this a SAT solver invariant?

Saturation/Splitter.cpp
```diff
@@ -779,8 +795,36 @@ bool Splitter::shouldAddClauseForNonSplittable(Clause* cl, unsigned& compName, C
...
+    _clausesAdded = true;
+  }
+}
+
 bool Splitter::handleNonSplittable(Clause* cl)
 {
+  if (_cleaveNonsplittables && cl->length() > 1) {
+    auto it = cl->iterLits();
+    while (it.hasNext()) {
+      conjectureSingleton(it.next(),cl);
```
mezpusz, line 824: I'm a bit confused by this. If a non-unit clause is nonsplittable, isn't it always the case that in any literal there is some variable which is present also in some other literal? Then, how is this sound? Maybe I'm missing something, do you have an example problem where this is used?

Shell/Property.cpp
```diff
@@ -1059,6 +1063,139 @@ bool Property::hasXEqualsY(const Formula* f)
...
 
+
+/**
+ * True if f should clausify to only introduce constants as Skolems.
+ *
+ * @warning Works correctly only closed formulas!
+ */
+bool Property::onlyExistsForallPrefix(UnitList* units)
+{
+  Stack<const Formula*> forms;
```
mezpusz: If these always move in triplets, it would make the function a bit more robust in my opinion to use a stack of 3-tuples (or a local `struct`).

Shell/Property.hpp
```diff
@@ -149,6 +149,8 @@ class Property
   static const uint64_t PR_HAS_DT_CONSTRUCTORS = 1099511627776ul; // 2^40
   /* has co-algebraic data type constructors */
   static const uint64_t PR_HAS_CDT_CONSTRUCTORS = 2199023255552ul; // 2^41
+  /* signature has only constants and formulas have an exists-forall quantifier prefix */
+  static const uint64_t PR_ESSENTIALLY_BSR = 4398046511104ul; // 2^42
```
mezpusz, line 153: What does BSR stand for? (I'm guessing not backward subsumption resolution.:))

MichaelRawson: :) Bernays-Schönfinkel-Ramsey, I think.

Saturation/Splitter.cpp
MichaelRawson: This is my fault. There is an experimental option that "splits harder" than is sound (i.e. even through variables), then backtracks via AVATAR if it proves false conditionally on the split. It's sometimes useful for saturating.

mezpusz: That's actually an interesting thing, thanks for explaining!

Shell/Property.hpp
mezpusz: Okay, that I would never have guessed! Would you mind if I added this as an extra pointer in the comment?

MichaelRawson: Not at all, I think that would be sensible - if in fact that is what it is, just a guess.

Saturation/Splitter.cpp
quickbeam123: I think one of them implies the other, but they are not mutex. Assertion can be added, but I don't remember by heart which one :(

Shell/Property.cpp
quickbeam123: There are further opportunities for improvement here.

I think the function will not work correctly under FOOL. There is also a different function doing a similar formula traversal. Maybe they could be merged ...

Shell/Property.hpp
quickbeam123: Feel free to add!

Kernel/Problem.cpp
quickbeam123: Thank you, I will take it! 

(I will need to learn what `iterTraits` really does. Feels intimidating as a name...)

Kernel/Problem.cpp
quickbeam123: Right, this would be a good trick (up to the problem of changing the meaning of the input), but in the rest of the function, there would be many `remainingImplication`s ...

Saturation/Splitter.cpp
quickbeam123: Actually, this also depends on SplittingAddComplementary. I don't think the assertion would be very legible even if I managed to write it down.

Shell/Property.cpp
quickbeam123: OK, I gave it a shot!

Could you please check: https://github.com/vprover/vampire/commit/eae4c809d80f872a6df11bf659d7a4f385faf5d5
?

The compactification into the struct is probably fine, but I also collapsed the FORALL case with the EXISTS and the transformation is a bit subtle.

mezpusz: Looks nice, thank you!

mezpusz (approved):

PR #735, opened 2025-07-17
simplify ProofPrinter

MichaelRawson:
I have been working with Vampire's TSTP output recently. For some reason lost to time, it doesn't print units in the order of derivation (i.e. sorted by `Unit::number`) with `-p tptp`. This is annoying for transforming TSTP.

Also, `ProofPrinter` seems quite complicated for the job it does. Simplify it.

### DIFF
--- a/Kernel/InferenceStore.cpp
+++ b/Kernel/InferenceStore.cpp
@@ -46,6 +46,8 @@
 
 #include "InferenceStore.hpp"
 
+#include <set>
+
 //TODO: when we delete clause, we should also delete all its records from the inference store
 
 namespace Kernel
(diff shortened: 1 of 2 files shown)


quickbeam123:
Could this have performance impact on TPTP proof printing? Should we measure something before/after?

MichaelRawson:
Essentially you have to compute the transitive inference closure backwards from falsum, then order and deduplicate by unit number. Before, this was done with a `Stack`, a `DHSet`, and `sort`. Now, it uses `std::set` (i.e. an ordered set). I suspect this to be faster, if anything, but we could measure if you're concerned.

Trouble is, what on? Most proofs we find are teeny tiny (relatively) so we'll be looking at a lot of noise. Do you know any particularly pathological proofs?

PR #737, opened 2025-07-17
SMTLIB parser: separate functions/predicates and type constructors

MichaelRawson:
Internally, the SMT-LIB parser treats all symbols the same, and tags them as functions/predicates/type constructors as it goes. This causes some difficulties as SMT-LIB allows a type constructor and a function/predicate to have the same name.

The previous solution added `()` to type constructors, so you got the famous `![N : 'nat()']: ...` output. Also it wasted a few cycles in some places disambiguating what kind of symbol something is, when we should know what it is. Therefore: separate functions/predicates from type constructors internally. We can also remove the parens from sort names now.

This does introduce a "bug" where if you have the following highly-confusing SMT-LIB input:
```
(declare-datatype nat ((zero) (succ (pred nat))))

(declare-fun nat (nat) nat)
(assert (forall ((n nat)) (= (nat n) n)))

(assert (not (= (nat zero) zero)))
```
and ask for `-p tptp`, you'll get this proof (lightly edited for clarity):
```tff(type_def_5, type, nat: $tType).
tff(func_def_0, type, zero: nat).
tff(func_def_1, type, succ: nat > nat).
tff(func_def_2, type, pred: nat > nat).
tff(func_def_3, type, nat: nat > nat).
tff(f10,plain,(
  $false),
  inference(trivial_inequality_removal,[],[f9])).
tff(f9,plain,(
  zero != zero),
  inference(superposition,[],[f8,f7])).
tff(f7,plain,(
  ( ! [X0 : nat] : (nat(X0) = X0) )),
  inference(cnf_transformation,[],[f1])).
tff(f1,axiom,(
  ! [X0 : nat] : nat(X0) = X0),
  file('test.smt2',unknown)).
tff(f8,plain,(
  zero != nat(zero)),
  inference(cnf_transformation,[],[f6])).
tff(f6,plain,(
  zero != nat(zero)),
  inference(flattening,[],[f2])).
tff(f2,axiom,(
  ~zero = nat(zero)),
  file('test.smt2',unknown)).
```
...which is, I presume, not valid TSTP. Don't do that, `WONTFIX`.

### DIFF
--- a/Parse/SMTLIB2.cpp
+++ b/Parse/SMTLIB2.cpp
@@ -61,7 +61,6 @@ namespace Parse {
 using namespace std;
 
 static const char* PAR = "par";
-static const char* TYPECON_POSTFIX = "()";
 
 SMTLIB2::SMTLIB2(UnitList::FIFO formulaBuffer)
 : _logicSet(false),
@@ -145,7 +144,7 @@ void SMTLIB2::readBenchmark(LExpr* bench)
 
       auto name = ibRdr.readAtom();
       // // We strictly disallow shadowing, as opposed to the standard
-      if (isAlreadyKnownSymbol(name)) {
+      if (isAlreadyKnownSort(name)) {
         USER_ERROR_EXPR("Redeclaring built-in, declared or defined sort parameter: "+name);
       }
       ASS_EQ(_globalSortParamLookup.size(),_nextVar);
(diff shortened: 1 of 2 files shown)


mezpusz (approved):
It looks all fine to me, and I see the tests also pass. Thanks for sorting this out!

quickbeam123:
Did you also check it parses (at least to the same degree as before) some portion of SMTLIB?

quickbeam123:
Did realize we have parsing checks in `sanity`, but hey, let me run a parsing round on SMTLIB problems ...

Issue #722, opened 2025-06-14
LaTeX output option may change proving capabilities?


mbalfakeih:
Earlier, I defined the following functions on lists using the definitions from [here](https://tptp.org/cgi-bin/SeeTPTP?Category=Problems&Domain=DAT&File=DAT081_1.p):
```
tff(list,type,
    list: $tType ).

tff(nil,type,
    nil: list ).

tff(cons,type,
    cons: ( $int * list ) > list ).

tff(head,type,
    head: list > $int ).

tff(tail,type,
    tail: list > list ).

%----Selectors
tff(head_def,axiom,
    ! [I: $int,L: list] : ( head(cons(I,L)) = I ) ).

tff(tail_def,axiom,
    ! [I: $int,L: list] : ( tail(cons(I,L)) = L ) ).

%----Constructors
tff(list_constructors,axiom,
    ! [L: list] :
      ( ( L = nil )
      | ( L = cons(head(L),tail(L)) ) ) ).

tff(cons_requirement,axiom,
    ! [I: $int,L: list] : ( cons(I,L) != nil ) ).

tff(get_min,type,
    get_min: list > $int ).

tff(get_min_single,axiom,
    ! [I: $int] : ( get_min(cons(I,nil)) = I ) ).

tff(get_min_list,axiom,
    ! [I: $int,L: list] :
      ( get_min(cons(I,L)) = $ite($less(I,get_min(L)),I,get_min(L)) ) ).
```

Now, the `get_min` function is ill-defined (`get_min_list` also considers the situation where the list is a single element, which overlaps with `get_min_single`), which I did not realize when first writing this. I believe this caused a contradiction and therefore the principle of explosion came into effect, since:
```
tff(test_equal,conjecture,
    get_min(cons(1,nil)) = 1 ).
```
and
```
tff(test_unequal,conjecture,
    get_min(cons(1,nil)) != 1 ).
```
could both be proven with `vampire --mode casc lists.p` in around 0.5 seconds on my machine using the release version of Vampire 4.9.

However, when running `vampire --mode casc --latex_output  test lists.p`, only `test_equal` could be proven, while `test_unequal` times out after 60 seconds. While I understand that the problem was ill-formed, I find it strange that specifying for Vampire to output to LaTeX would drastically change whether or not it could prove a conjecture.

quickbeam123:
Yes, the input is contradictory even without the conjecture.

But adding 
```
tff(test_equal,conjecture,
[...]

PR #738, opened 2025-07-17
minimum viable fixes for LaTeX output (or maybe remove LaTeX altogether)

### LINKED ISSUE #721: LaTeX output confuses < and ≮
I was working on a bigger project and noticed that $<$ and $\not<$ seemed to be getting confused in the LaTeX output, so I tried recreating this with the example:
```
tff(t,conjecture,
    ! [X: $int] : ~ $less(X,X) ).
```
I then ran `vampire --latex_output test toreport.p`, giving the output:
```
% SZS status Theorem for toreport
% SZS output start Proof for toreport
1. ! [X0 : $int] : ~$less(X0,X0) [input]
2. ~! [X0 : $int] : ~$less(X0,X0) [negated conjecture 1]
8. ~$less(X0,X0) [theory axiom 150]
15. ? [X0 : $int] : $less(X0,X0) [ennf transformation 2]
16. ? [X0 : $int] : $less(X0,X0) => $less(sK0,sK0) [choice axiom]
17. $less(sK0,sK0) [skolemisation 15,16]
18. $less(sK0,sK0) [cnf transformation 17]
19. $false [subsumption resolution 18,8]
% SZS output end Proof for toreport
% Version: Vampire 4.9 (commit 5ad494e78 on 2024-06-14 14:05:27 +0100)
% Linked with Z3 4.12.3.0 79bbbf76d0c123481c8ca05cd3a98939270074d3 z3-4.8.4-7980-g79bbbf76d
% Termination reason: Refutation

% Memory used [KB]: 440
% Time elapsed: 0.005 s
% Instructions burned: 5 (million)
```
and then ran `pdflatex test` (I have `texlive-scheme-full` on fedora installed) to get the pdf output.

This gives me the following output:
[test.pdf](https://github.com/user-attachments/files/20739218/test.pdf)

Where for step 8, it states that $x_0 < x_0$, which is obviously untrue and is negated in the text output. Similarly step 15 uses the $\not<$ in the LaTeX output, but is not negated in the text output.

I have tested this on Overleaf as well and it gives the same results, so I do not believe this is due to my LaTeX installation.

MichaelRawson:
Closes #721, #722. Thanks for the report, @mbalfakeih! However, do not get too excited, this not what you might call a proper fix:
- In quite a few cases the LaTeX output flipped polarity, and somehow nobody noticed. Don't do that.
- Special terms (like `$ite`) are not handled properly: this used to crash, now it prints out the TPTP format instead when encountering a special term.

---

More generally, I think the LaTeX mode needs to die. It's clearly unloved (nobody noticed these quite glaring deficiencies until now), and people who do actually want this also tend to want to customise the output a bit. What they should do instead is read a TSTP proof, and write some Prolog to format it how they please. Maybe we can put an example on the wiki or something.

If there are not screams of outrage, I will upgrade this PR to remove LaTeX altogether. Let's focus on what we're actually good at.

### DIFF
--- a/Shell/LaTeX.cpp
+++ b/Shell/LaTeX.cpp
@@ -331,7 +331,7 @@ std::string LaTeX::toString (Literal* l) const
 
   //Check if this symbol has an interpreted LaTeX name
   // this should be true for all known interpreted symbols and any recorded symbols
-  std::string template_str = theory->tryGetInterpretedLaTeXName(l->functor(),true,l->isNegative());
+  std::string template_str = theory->tryGetInterpretedLaTeXName(l->functor(),true,l->isPositive());
 
   if(template_str.empty()){
     std::string res;
@@ -495,9 +495,13 @@ std::string LaTeX::toString (TermList* terms,bool single) const
      //Check if this symbol has an interpreted LaTeX name
      // this should be true for all known interpreted symbols and any recorded symbols
       std::string template_str = theory->tryGetInterpretedLaTeXName(trm->functor(),false);
-   
+
       if(template_str.empty()){
-        result += symbolToString(trm->functor(), false) + toString(trm->args());
+        if(trm->isSpecial())
+          // not ideal, but what can you do
+          result += trm->toString();
+        else
+          result += symbolToString(trm->functor(), false) + toString(trm->args());
       }
       else{
         // replace arguments in the template, arg0 replaces a0 etc.


kazarmy:
I actually want a node graph to go with the LaTeX but that graph probably will need to be simplified somehow and the code needed to support such customization will probably be a maintenance nightmare (I think).

quickbeam123:
Fine with the PR, fine with the removal (at some future point).

MichaelRawson:
OK, let's merge this and I'll put LaTeX on the kill list.

kazarmy:
> @kazarmy I believe [this repository](https://github.com/gleiss/saturation-visualization) aims to do what you are suggesting, however it appears to be unmaintained and I was unable to get it to work.

Interesting. The guy's thesis at https://repositum.tuwien.at/handle/20.500.12708/16444 provides further details. Sadly, AI is definitely trending nowadays so it is unlikely for me that the time and effort needed to study both the code and Vampire can be justified.

>TSTP vs SatVis

>(Pg. 102 of the aforementioned thesis. I assume the doctrine of "fair use" applies here.)

Heh.

kazarmy:
> From the proof, or from proof search? Bernhard Gleiss' tool is for proof search.
> 
> For proof, I think you could do a node graph fairly easily with a Prolog script, it's what I do for such things. Let me know if interested and I'll put a sample on the wiki.

Proof search has more information but yes please I would be interested in a sample since my Prolog skills are definitely rusty.

PR #739, opened 2025-07-21
Disentangled ImmediateSimplificationEngine::{simplify,simplifyMany}

joe-hauns:
This PR removes the function `simplifyMany` from `ImmediateSimplificationEngine` and moves it to its own class.
There was no reason for both methods to be present in the same class in the first place as subclasses would always only implement exactly `simplify` or `simplifyMany` and fail `NOT_IMPLEMENTED` in the other method.
By moving the method into a new class `ImmediateSimplificationEngineMany` we fix the broken class hierarchy.

Further the PR changes them interface to return an `Option` instead of an interator. This allows for implementors of the method to return `some(VirtualIterator::getEmpty)` in order to signal that the clause that is being simplified should be deleted, which was not possible in the old interface.

### DIFF
--- a/Inferences/CasesSimp.cpp
+++ b/Inferences/CasesSimp.cpp
@@ -100,7 +100,7 @@ struct CasesSimp::RewriteableSubtermsFn
 };
 
 
-ClauseIterator CasesSimp::simplifyMany(Clause* premise)
+Option<ClauseIterator> CasesSimp::simplifyMany(Clause* premise)
 {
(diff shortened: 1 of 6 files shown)


MichaelRawson:
Nice.

>in order to signal that the clause that is being simplified should be deleted, which was not possible in the old interface

Does this mean there are/could be some duplicating simplifying/deleting inferences somewhere that were created in order to achieve this?

PR #741, opened 2025-07-26
TPTP: Fix OR term order in proof

kazarmy:
After running Vampire 6d35e0bef on the following problem:
```tptp
% Law of excluded middle (Pel86 problem 6; tff variant of SYN387+1.p)

tff(p, type, p: $o).
%---p ∨ ¬p
tff(excluded_middle, conjecture, p | ~p).
```
the first line of the proof is:

```
1. ~p | p [input(conjecture)]
```

Somewhat disconcertingly, the two terms have switched places. This PR fixes the `OR` term concat order so that the first line of the proof becomes:

```
1. p | ~p [input(conjecture)]
```

### DIFF
--- a/Parse/TPTP.cpp
+++ b/Parse/TPTP.cpp
@@ -3389,7 +3389,7 @@ void TPTP::endFormula()
     case AND:
     case OR:
       f = _formulas.pop();
-      f = makeJunction((Connective)con,_formulas.pop(),f);
+      f = makeJunction((Connective)con,f,_formulas.pop());
       if (conReverse) {
 	f = new NegatedFormula(f);
       }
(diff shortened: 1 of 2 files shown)


MichaelRawson:
I think the CI failure might just be noise as a result of swapping the ORs, don't sweat it too much. Thanks for the PR, this is something I've been meaning to fix for ages (and the fact that equations are sometimes flipped), it makes proof checking annoying.

kazarmy:
Yeah the CI failure doesn't happen locally.

easychair:
There is a school of thought that says that a clause is a multiset of
literals. If you follow this school, your zen should not be spent for this
discussion.

The word "fix" is probably too strong and unfair to Vampire who should
spend CPU time for finding true enlightenment (proofs).

Cheers, Andrei

easychair:
Having the input preserved verbatim is, essentially, impossible without
considerable efforts. The problem is input variable names that are getting
normalized already during parsing.

Checking that two formulas are equivalent up to variable renaming is
polynomial time equivalent to the graph isomorphism, which is a hard
problem believed to be between P and NP.

I am not sure about checking equivalence up to reordering AND/OR but it
seems to me it has the same complexity (graph isomorphism), it is
definitely not harder.

Cheers, Andrei

MichaelRawson:
>Having the input preserved verbatim is, essentially, impossible without
considerable efforts. The problem is input variable names that are getting
normalized already during parsing.

This is true (although we do have some support for this already for question answering - essentially we maintain the map in some cases). However, for the ITP case e.g. Agda is quite happy to accept ɑ-equivalent terms but refuses to accept that logical OR is commutative, so this would make that export easier.

I'm inclined to merge as it makes little difference to Vampire but makes my life easier. However, @quickbeam123 is this likely to mess up your nicely prepared schedules? If so we could wait until after the 5.0 release.

PR #742, opened 2025-07-28
Fix code tree crash with gcc

mezpusz:
Debug sample tested on TPTP and inductive benchmarks.

Credit goes to Michael for finding this.

As a remark, maybe we should `-fsanitize` on a more regular basis.

### DIFF
--- a/Indexing/CodeTree.hpp
+++ b/Indexing/CodeTree.hpp
@@ -172,7 +172,6 @@ public:
     {
       ASS(ptr); //data has to be a non-zero pointer
       CodeOp res;
-      res.setAlternative(0);
       res._setData(ptr);
       res._setInstruction(SUCCESS_OR_FAIL);
       ASS(res.isSuccess());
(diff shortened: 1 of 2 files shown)


MichaelRawson:
I will put `-fsanitize` on my list, I fixed it at some point but things have clearly degraded in the meantime (or the sanitizer got better).

Indexing/CodeTree.hpp
```diff
@@ -234,7 +234,7 @@ class CodeTree
 
   private:
     // bitfield
-    uint64_t _content;
+    uint64_t _content = 0;
 
     /**
      * Pointer to an alternative operation
```
MichaelRawson, line 239: Should `_alternative` also be `nullptr`? No objection if you're sure it's safe.

MichaelRawson (approved):

Indexing/CodeTree.hpp
mezpusz: Probably yes. There is no such thing as safe here, I think, so let's do that too.

mezpusz: The new round of sampling looks okay, please merge.

PR #743, opened 2025-07-30
remove unused SATSolver methods, streamline SATLiteral

MichaelRawson:
Mostly obvious. `SATLiteral` is now a wrapper for a non-zero `int`, matching the convention almost everywhere else: I hope that the compiler is smart enough to optimise out the conversion methods.

### DIFF
--- a/SAT/SATClause.cpp
+++ b/SAT/SATClause.cpp
@@ -132,18 +132,12 @@ void SATClause::setInference(SATInference* val)
   }
 }
 
-
-static bool litComparator(SATLiteral l1, SATLiteral l2)
-{
-  return l2 < l1;
-}
-
 /**
  * Sort literals in descending order.
  */
 void SATClause::sort()
 {
-  std::sort(&_literals[0], &_literals[length()], litComparator);
+  std::sort(_literals, _literals + length());
 }
 
 SATClause* SATClause::removeDuplicateLiterals(SATClause* cl)


SAT/SATClause.cpp
```diff
@@ -135,7 +135,7 @@ void SATClause::setInference(SATInference* val)
 
 static bool litComparator(SATLiteral l1, SATLiteral l2)
 {
-  return l1.content()>l2.content();
+  return l2 < l1;
```
quickbeam123: Is the flip in polarity here important?

MichaelRawson: No flip as I've sneakily exchanged the arguments - it's just that I only defined `operator<` and not `operator>`.

SAT/SATLiteral.hpp
```diff
@@ -18,77 +18,49 @@
...
-  inline unsigned oppositePolarity() const { return 1-_polarity; }
-  inline unsigned content() const { return _content; }
+  SATLiteral(unsigned var, bool polarity)
+  : _content(polarity ? var : -int(var))
+  { ASS(var) }
 
-  unsigned defaultHash() const { return DefaultHash::hash(content()); }
-  unsigned defaultHash2() const { return content(); }
+  unsigned var() const { return abs(_content); }
+  bool polarity() const { return _content > 0; }
```
quickbeam123: Could we still call this `positive`? Polarity confuses me a bit ...

SAT/SATSolver.hpp
```diff
@@ -302,11 +276,6 @@ class PrimitiveProofRecordingSATSolver : public SATSolverWithAssumptions {
       _refutation->setInference(_refutationInference);
     }
 
-  virtual ~PrimitiveProofRecordingSATSolver() {
-    // cannot clear the list - some inferences may be keeping its suffices till proof printing phase ...
```
quickbeam123, line 306: OK, because `SATSolver` already has a virtual destructor?

MichaelRawson: Exactly - and this one doesn't do anything, so (I think) it's pointless.

SAT/SATLiteral.hpp
MichaelRawson: As in instead of `polarity`, or as well as?

quickbeam123: Instead. 

Or `isPositive`.

Since my work on Minisat with its
```
inline  Lit  mkLit     (Var var, bool sign  = false)
```
where `sign` means "negated". I don't trust vaguely named booleans ;)

MichaelRawson: Good! I don't care what it's called, just that there is only one, not three.

SAT/SATLiteral.hpp
```diff
@@ -18,77 +18,49 @@
...
-  }
-
-  inline bool isPositive() const { return _polarity; }
-  inline bool isNegative() const { return !_polarity; }
-  inline unsigned var() const { return _var; }
-  inline unsigned polarity() const { return _polarity; }
-  inline unsigned oppositePolarity() const { return 1-_polarity; }
-  inline unsigned content() const { return _content; }
+  SATLiteral(unsigned var, bool polarity)
+  : _content(polarity ? var : -int(var))
```
quickbeam123: Would it make sense to assert that `var < 2^31`?

MichaelRawson: Yes, I think so. Or whatever the fancy C++ way of checking `unsigned` fits in `int` is. I'll look it up.

quickbeam123:
Funnily, there is this ("single") performance regression for me:
```
./vampire_z3_rel_master_10130 -sa discount -awr 10 -p off Problems/GEO/GEO589+1.p 
% SZS status Theorem for GEO589+1
% Version: Vampire 5.0.0 (commit 695a78fff on 2025-07-29 14:28:19 +0200)
% Linked with Z3 4.14.0.0 3c47fd96cf5645d0c42b2c819d9e9a84380aa721 z3-4.8.4-9178-g3c47fd96c
% Termination reason: Refutation

% Time elapsed: 0.537 s
% Peak memory usage: 34 MB
% Instructions burned: 2417 (million)


./vampire_z3_rel_michael-sat-2_10131 -sa discount -awr 10 -p off Problems/GEO/GEO589+1.p 
% SZS status Theorem for GEO589+1
% Version: Vampire 5.0.0 (commit ad8dddbb4 on 2025-07-30 08:08:10 +0200)
% Linked with Z3 4.14.0.0 3c47fd96cf5645d0c42b2c819d9e9a84380aa721 z3-4.8.4-9178-g3c47fd96c
% Termination reason: Refutation

% Time elapsed: 1.678 s
% Peak memory usage: 69 MB
% Instructions burned: 8546 (million)
```

Can you reproduce it on your end?

MichaelRawson:
Thanks - reproduced. There are two things going on here:

1.  `SATLiteral::operator<` is not the same as it used to be, because the pattern of bits compared is not the same. If I correct this in a silly way (reconstructing the original bit pattern with `var() << 1 | positive()`) the divergence goes away, apart from...
2. For some reason we sort non-splittable AVATAR clauses but not splittable ones. I'm not sure it's worth the effort to sort it, but we should be consistent.

When these are both "corrected" the difference goes away. *However* I think they're both a bit silly. I propose:
- We define whatever order is most sensible on `SATLiteral` and ignore whatever it used to be.
- Don't sort either kind of clauses.

Issue #733, opened 2025-07-16
Wrong order of quantifiers with type variables


mezpusz:
Vampire sometimes produces output where type variables come after types that depend on them, such as the following line:
```
tff(f7336,plain,(
  ! [X0 : assn,X1 : list(X2),X2 : $tType, ...
```
Then, for example GDV complains about `X2` not being quantified.
This can be reproduced with:
```
-p tptp --decode dis+1010_32:1_sil=2000:bsd=on:fd=off:slsq=on:random_seed=161430699:cts=off:i=1282:sd=1:av=off:gtg=position:ss=axioms_2877 $TPTP/Problems/ITP/ITP218_2.p
```

A couple of other occurrences:
- `-p tptp --decode lrs+10_7_to=lpo:sil=64000:tgt=ground:drc=off:fde=none:sp=const_frequency:slsqc=4:slsq=on:random_seed=3561718340:i=10680:bd=all:gtg=exists_sym_2779 $TPTP/Problems/ITP/ITP220_4.p`
- `-p tptp --decode lrs+1002_1_to=lpo:sil=2000:sos=on:random_seed=1239504037:i=1674:sd=3:ss=axioms:sgt=16_2958 $TPTP/Problems/ITP/ITP222_4.p`
- `-p tptp --decode lrs+10_3_sil=8000:tgt=ground:fde=none:sp=reverse_arity:lcm=predicate:random_seed=3224555319:i=3317:sd=1:av=off:ss=axioms:sgt=50_2790 $TPTP/Problems/ITP/ITP223_4.p`
- `-p tptp --decode lrs+1002_1_to=lpo:sil=2000:sos=on:random_seed=2065495457:i=1674:sd=3:ss=axioms:sgt=16_2959 $TPTP/Problems/ITP/ITP225_4.p`

mezpusz:
Then it's worse than I thought.. we can delay it until HOL/polymorphism becomes a more popular use case.

quickbeam123:
For sure we could sort when printing. 

I just thought we should maybe be more explicit about type quantification for our own future's sanity's sake ;) 

(Not sure whether TPTP itselfs allows the mixing in a single block. I vaguely remember we do it to ourselves during Rectify, maybe.)

PR #745, opened 2025-07-31
report SMT-LIB quantifiers in the order given

MichaelRawson:
We parse and bind SMT-LIB quantifiers in hash-table iteration order. This is not ideal: report them in the left-to-right input order instead.

While I'm at it, also promote some type aliases to proper types (for more type safety) and remove a pointless heap allocation.

Not tested on SMT-LIB (or even inspected very closely) yet, so don't merge. I don't have a copy here so I'll wait until next week, unless one of you has a copy to hand and is willing to test.

### DIFF
--- a/Parse/SMTLIB2.hpp
+++ b/Parse/SMTLIB2.hpp
@@ -338,11 +338,44 @@ private:
   /** For generating fresh vampire variables */
   unsigned _nextVar;
 
-  /** < termlist, vampire sort id > */
-  typedef std::pair<TermList,TermList> SortedTerm;
-  /** mast an identifier to SortedTerm */
-  typedef DHMap<std::string,SortedTerm> TermLookup;
-  typedef Stack<TermLookup*> Lookups;
+  struct Binding {
+    TermList term;
+    TermList sort;
+  };
+
+  struct Lookup {
+    // `map` keys into `bindings`
+    std::unordered_map<std::string, size_t> map;
+    // the bindings in left-to-right order
+    std::vector<Binding> bindings;
+
+    size_t size() const { return bindings.size(); }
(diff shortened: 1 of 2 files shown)


Parse/SMTLIB2.hpp
```diff
@@ -338,11 +338,50 @@ class SMTLIB2 {
...
-  typedef std::pair<TermList,TermList> SortedTerm;
-  /** mast an identifier to SortedTerm */
-  typedef DHMap<std::string,SortedTerm> TermLookup;
-  typedef Stack<TermLookup*> Lookups;
+  struct Binding {
+    TermList term;
+    TermList sort;
+  };
+
+  struct Lookup {
```
mezpusz, line 350: Could you please document that this struct is needed to preserve the order?

Parse/SMTLIB2.hpp
```diff
@@ -338,11 +338,50 @@ class SMTLIB2 {
...
+    // `map` keys into `bindings`
+    std::unordered_map<std::string, size_t> map;
+    // the bindings in left-to-right order
+    std::vector<Binding> bindings;
+
+    // copy constructor is probably a bug
+    Lookup(const Lookup &) = delete;
```
mezpusz, line 357: Don't you have to list the assignment operators similarly to avoid copying? Or are you being careful not to use them?

mezpusz:
Maybe partially solves #733? Probably not fully though.

Parse/SMTLIB2.hpp
MichaelRawson: I *thought* the rule was that if you defined one, they all had to be provided explicitly. But my compiler fails to do this properly (or I misremembered the rule) so I'll fix this, thanks!

PR #746, opened 2025-07-31
align Coproduct bytes

MichaelRawson:
UBSan runs were complaining about unaligned memory accesses from `Coproduct`. I think this is the appropriate fix: the performance impact may be non-zero, but correctness trumps speed.

There again, performance could be better if we're not constantly doing misaligned loads. Hard to say.


joe-hauns (approved):
great catch! thank you!

PR #747, opened 2025-08-01
Option elimination

quickbeam123:
An ongoing effort to eliminate some options that are actually not needed to cover the problems we care about solving (in TPTP) and thus arguably only obscure the codebase and clutter the schedules.


Saturation/Splitter.cpp
```diff
@@ -424,7 +400,7 @@ void SplittingBranchSelector::updateSelection(unsigned satVar, SATSolver::VarAss
   {
     bool posSticky = posUsed && _parent.isSticky(posLvl);
     bool negSticky = negUsed && _parent.isSticky(negLvl);
-    if(_eagerRemoval) {
+    { // hardcoding (_eagerRemoval == true)
```
MichaelRawson, line 403: Did you mean to keep the block scope?

MichaelRawson: No big deal, just wondered. I can remove it afterwards if it's not needed.

PR #748, opened 2025-08-01
neaten SATSolver infrastructure, merge MinisatInterfacing{,NewSimp}

MichaelRawson:
...in a few descriptive commits. :innocent: In particular the two `MinisatInterfacing`s are now one class with a template parameter for the inner SAT solver. `MinisatInterfacingNewSimp` is now a type alias.


MichaelRawson:
Ready for review. I haven't tested much as _in principle_ this shouldn't change anything. If you're at all worried prod me and I'll run some.

Saturation/Splitter.hpp
```diff
@@ -91,10 +91,10 @@ _solver=0;
 private:
   friend class Splitter;
 
-  SATSolver::Status processDPConflicts();
```
quickbeam123, line 94: Status feels a bit too general here. How about saying `SAT::Status`` like I tried?

MichaelRawson: Fine by me - the nice thing about `enum class` is we should get a type error if it resolves ambiguously or surprisingly.

PR #752, opened 2025-08-06
IntNameTable mostly pointless

MichaelRawson:
Thanks to @mezpusz for pointing this out. Also change `int` to `unsigned` for variables in the TPTP parser (no idea why it was like this) and de-heap `_questionVariableNames`.

### DIFF
--- a/Parse/TPTP.hpp
+++ b/Parse/TPTP.hpp
@@ -351,6 +351,7 @@ public:
   unsigned lineNumber(){ return currentFile.lineNumber; }
   std::string currentPath(){ return currentFile.path; }
 
+  // careful: the returned pointer will be invalidated if _questionVariableNames is changed
   static Map<unsigned,std::string>* findQuestionVars(unsigned questionNumber) {
     return _questionVariableNames.findPtr(questionNumber);
   }


mezpusz (approved):

Parse/TPTP.hpp
```diff
@@ -352,9 +351,8 @@ class TPTP
   unsigned lineNumber(){ return currentFile.lineNumber; }
   std::string currentPath(){ return currentFile.path; }
 
-  static Map<int,std::string>* findQuestionVars(unsigned questionNumber) {
-    auto res = _questionVariableNames.findPtr(questionNumber);
-    return res ? *res : nullptr;
+  static Map<unsigned,std::string>* findQuestionVars(unsigned questionNumber) {
+    return _questionVariableNames.findPtr(questionNumber);
```
quickbeam123, line 356: So a danger would be if someone requests `findQuestionVars`, keeps the pointer stored with them, TPTP later inserts to `_questionVariableNames` some more, the `DHMap` grows, `Map`s inside it get `move`ed and the old pointer gets invalidated.

If you considered this option and have checked this cannot happen, it's fine (well, maybe except that we have a bomb ticking in the interface that should be marked as such - maybe add a comment? or maybe I am wrong?)

MichaelRawson: I agree with the hazard - I was happy with it given that I imagine that everything gets parsed ahead of time and only then (once the map is effectively frozen) is `findQuestionVars` called, but I appreciate it's hazardous. I can just put it back or `std::shared_ptr` it, it's no big deal.

quickbeam123: I think adding a comment should be fine! Thank you!

PR #754, opened 2025-08-08
isolate post-hoc unsat minimisation

MichaelRawson:
There are currently two SAT inferences which may need minimisation: global subsumption and AVATAR. AVATAR we can minimise _in situ_ as we know it will lead to a proof directly. Global subsumption we need to do lazily, as we're not sure what will end up in the final proof.

However, the two are somewhat entangled. Separate them out:
- Rename `FromSatRefutation` to `NeedsMinimization`.
- Use `NonSpecificInferenceMany` with pre-minimised premises for AVATAR.
- `NeedsMinimization` inferences *definitely require* minimisation, and it *will be* possible since it's only used by global subsumption, and global subsumption only produces SAT inferences that can be minimised by Minisat post-hoc.
- Oddity: `Metaiterator.hpp`'s `makeUnique` never did anything, so just remove it and calls to it.
- To support the AVATAR use-case, add a `minimizePremises` method to `SATSolver` that tries to minimise supplied unsatisfiable premises on a best-effort basis.
  + `CadicalInterfacing` forwards to `MinisatInterfacing`, which does something homebrew. In the near future I will extract DRAT proofs from CaDiCaL and use those instead, which should be much more efficient.
  + `Z3Interfacing` does nothing with this (as usual), but now it should be possible to minimise these as we have access to the meaning of theory literals.
- Temporarily remove `getRefutation` from `ProofProducingSATSolver`, as nothing uses this. In the near future this will be replaced with something that produces a (first-order) inference, either the usual "by sat" or a detailed proof via DRAT.

### DIFF
--- a/Kernel/Inference.cpp
+++ b/Kernel/Inference.cpp
@@ -59,10 +59,10 @@ UnitInputType Kernel::getInputType(UnitList* units)
 /**
  * To be kept around in _ptr2 of INFERENCE_FROM_SAT_REFUTATION
  **/
-struct FromSatRefutationInfo {
-  USE_ALLOCATOR(FromSatRefutationInfo);
+struct NeedsMinimizationInfo {
+  USE_ALLOCATOR(NeedsMinimizationInfo);
 
-  FromSatRefutationInfo(const FromSatRefutation& fsr) : _satPremises(fsr._satPremises), _usedAssumptions(fsr._usedAssumptions)
+  NeedsMinimizationInfo(const NeedsMinimization& fsr) : _satPremises(fsr._satPremises), _usedAssumptions(fsr._usedAssumptions)
   { ASS(_satPremises); }
 
   SAT::SATClauseList* _satPremises;
@@ -73,8 +73,8 @@ struct FromSatRefutationInfo {
 void Inference::destroyDirectlyOwned()
 {
   switch(_kind) {
-    case Kind::INFERENCE_FROM_SAT_REFUTATION:
-      delete static_cast<FromSatRefutationInfo*>(_ptr2);
+    case Kind::SAT_NEEDS_MINIMIZATION:
+      delete static_cast<NeedsMinimizationInfo*>(_ptr2);
       // intentionally fall further
     case Kind::INFERENCE_MANY:
       UnitList::destroy(static_cast<UnitList*>(_ptr1));
@@ -90,8 +90,8 @@ void Inference::destroy()
       if (_ptr1) static_cast<Unit*>(_ptr1)->decRefCnt();
       if (_ptr2) static_cast<Unit*>(_ptr2)->decRefCnt();
       break;
-    case Kind::INFERENCE_FROM_SAT_REFUTATION:
-      delete static_cast<FromSatRefutationInfo*>(_ptr2);
+    case Kind::SAT_NEEDS_MINIMIZATION:
+      delete static_cast<NeedsMinimizationInfo*>(_ptr2);
       // intentionally fall further
     case Kind::INFERENCE_MANY:
       UnitList* it=static_cast<UnitList*>(_ptr1);
@@ -105,7 +105,7 @@ void Inference::destroy()
   }
 }
 
-Inference::Inference(const FromSatRefutation& fsr) {
+Inference::Inference(const NeedsMinimization& fsr) {
   initMany(fsr._rule,fsr._premises);
 
   ASS_REP(isSatRefutationRule(fsr._rule),ruleName(fsr._rule));
@@ -114,8 +114,8 @@ Inference::Inference(const FromSatRefutation& fsr) {
     return; // SAT solver did not support minimization anyway
   }
 
-  _kind = Kind::INFERENCE_FROM_SAT_REFUTATION;
-  _ptr2 = new FromSatRefutationInfo(fsr);
+  _kind = Kind::SAT_NEEDS_MINIMIZATION;
+  _ptr2 = new NeedsMinimizationInfo(fsr);
 }
 
 /**
@@ -130,7 +130,7 @@ Inference::Iterator Inference::iterator() const
       it.integer=0;
       break;
     case Kind::INFERENCE_MANY:
-    case Kind::INFERENCE_FROM_SAT_REFUTATION:
+    case Kind::SAT_NEEDS_MINIMIZATION:
       it.pointer = _ptr1;
       break;
   }
@@ -159,7 +159,7 @@ bool Inference::hasNext(Iterator& it) const
       }
       break;
     case Kind::INFERENCE_MANY:
-    case Kind::INFERENCE_FROM_SAT_REFUTATION:
+    case Kind::SAT_NEEDS_MINIMIZATION:
       return (it.pointer != nullptr);
     default:
       ASSERTION_VIOLATION;
@@ -187,7 +187,7 @@ Unit* Inference::next(Iterator& it) const
       }
       break;
     case Kind::INFERENCE_MANY:
-    case Kind::INFERENCE_FROM_SAT_REFUTATION: {
+    case Kind::SAT_NEEDS_MINIMIZATION: {
       UnitList* lst = static_cast<UnitList*>(it.pointer);
       it.pointer = lst->tail();
       return lst->head();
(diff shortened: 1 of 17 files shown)


MichaelRawson:
>AVATAR we can minimise in situ as we know it will lead to a proof directly.

Although, thinking about it, perhaps I should care about the `-p off` case.

SAT/SATSolver.hpp
```diff
@@ -177,6 +177,13 @@ class SATSolver {
   }
 
   SATLiteralStack explicitlyMinimizedFailedAssumptions(unsigned conflictCountLimit, bool randomize);
+
+  /**
+   * Assuming that the last solving call was unsatisfiable,
+   * try to produce a minimal set of premises that it still unsatisfiable
+   * (with respect to assumptions)
+   */
+  virtual SATClauseList *minimizePremises(SATClauseList *premises) = 0;
```
quickbeam123, line 189: Q: It this is about "last solving call", what are `premises` here? Could we drop them? Or is it more convenient to receive them again here? If so, maybe add a comment they must be the ones from the last call?

Style: Do you really prefer `Type<SPACE>*method_name`? I find it strange, as splits the important, the overall type, with a space. (A tiny bit harder to read for me...)

SAT/ProofProducingSATSolver.hpp
```diff
@@ -73,61 +67,25 @@ class ProofProducingSATSolver final : public SATSolver {
...
-  }
+  // all premises ever added
+  SATClauseList *premiseList() const { return _addedClauses; }
+
+  /*
+   * only those premises required to get unsat from the last solve() call
+   * may be a superset of those actually necessary depending on
+   * how well the minimisation process goes
+   */
+  SATClauseList *minimizedPremises() { return _inner->minimizePremises(_addedClauses); }
```
quickbeam123, line 83: It's not clear to me here, why:
1) we need the overload (not sure I understand the comment. Is the first line a separate sentence or should I really read: "only those premises ... may be a superset"? Overall, it feels the most obvious first line of the description of the function is missing)
2) why is it separated from its namesake by `premiseList`

Kernel/Inference.cpp
```diff
@@ -484,14 +484,15 @@ std::string Inference::name() const {
 
 void Inference::minimizePremises()
 {
-  if (_kind != Kind::INFERENCE_FROM_SAT_REFUTATION)
+  if (_kind != Kind::SAT_NEEDS_MINIMIZATION)
+    return;
+  if(!env.options->minimizeSatProofs())
     return;
-  if (_ptr2 == nullptr)
-    return; // already minimized
```
quickbeam123, line 490: Was safe to drop?

Kernel/Inference.cpp
```diff
@@ -484,14 +484,15 @@ std::string Inference::name() const {
 
 void Inference::minimizePremises()
 {
-  if (_kind != Kind::INFERENCE_FROM_SAT_REFUTATION)
+  if (_kind != Kind::SAT_NEEDS_MINIMIZATION)
+    return;
+  if(!env.options->minimizeSatProofs())
```
quickbeam123, line 489: Interesting!

... but the option worked until now. Is the check now needed this deep inside Kernel?

SAT/SATInference.cpp
```diff
@@ -24,15 +24,6 @@ namespace SAT
 // SATInference
 //
 
-/**
- * Collect first-order premises of @c cl into @c res. Make sure that elements in @c res are unique.
- */
-void SATInference::collectFOPremises(SATClause* cl, Stack<Unit*>& acc)
```
quickbeam123, line 30: Didn't you find the hardwired trivial lambda a nice convenience?

If this is for efficiency, maybe `collectFOPremises` could still live in the header?

Saturation/Splitter.cpp
```diff
@@ -182,34 +176,34 @@ static Color colorFromPossiblyDeepFOConversion(SATClause* scl,Unit*& u)
...
+    ));
 
-    Clause* foRef = Clause::fromIterator(LiteralIterator::getEmpty(),
-        FromSatRefutation(_solverIsSMT ? InferenceRule::AVATAR_REFUTATION_SMT : InferenceRule::AVATAR_REFUTATION, prems, satPremises));
     // TODO: in principle, the user might be interested in this final clause's age (currently left 0)
     throw MainLoop::RefutationFoundException(foRef);
   } else { // we must produce a well colored proof
-
-    // collect actually used SAT premises
-    SATClauseStack actualSatPremises;
```
quickbeam123, line 199: Is this gone because newly `satPremises` are always non-null?

quickbeam123: (Not sure, if I would recall quickly how to test the well-colored avatar case, but maybe we should.)

SAT/SATSolver.hpp
MichaelRawson: The inner SAT solver doesn't "remember" what its `premises` are, that's `ProofProducingSATSolver`'s job. In general they don't have to be those from the last call, but it does respect assumptions. I'll add a comment.

Style: yes, because `int* foo(), bar();` declares `int* foo();` but `int bar();` - and the same with references. This is, I agree, a deficiency with the language grammar. But I'm not religious about it, so we can keep to your preference if you feel strongly.

SAT/ProofProducingSATSolver.hpp
MichaelRawson: `ProofProducingSATSolver` is now just a wrapper around a SAT solver that remembers what was asserted. There is `minimizePremises` to conform with `SATSolver`, but I can't imagine when you would call that. `premiseList` gives you the list of all assertions. Then there is `minimize*d*Premises`, which calls `minimizePremises` on the internally-kept list of premises. I'll fix the comment.

Kernel/Inference.cpp
MichaelRawson: Yep, `ptr2` is never `nullptr` now.

Kernel/Inference.cpp
MichaelRawson: This check replaces `ptr2 == nullptr`. We could set `ptr2` nullable if `!env.options.minimizeSatProofs` in global subsumption if you'd prefer that?

SAT/SATInference.cpp
MichaelRawson: Didn't happen often enough for me to keep it (at some point mid-hacking all lambdas were non-trivial) but I'll put it back.

Saturation/Splitter.cpp
MichaelRawson: Exactly - there used to be some convoluted logic like "if the SAT solver doesn't know what the premises are, ask it for an explicit proof and then try to get it from there" - but the SAT solver did know what the premises were, and if it didn't then the proof was effectively just "false [avatar sat refutation every,last,premise]" - i.e. a weird way of saying "use all the premises".

MichaelRawson: I also don't know how to test the AVATAR colouring thing. I convinced myself that my changes above that blob were compatible with its preconditions.

SAT/SATSolver.hpp
quickbeam123: Comment, yes, please.

Thank you for the explanation! (Would never have thought to use a comma when declaring functions/methods; but now I recall the same rule applies to variables. Not too strongly. Just a bit undecided about how much I care about consistency with these things. On the other hand, it's going to be easier for me to recognize what you wrote ;))

SAT/ProofProducingSATSolver.hpp
quickbeam123: Oh, I missed the "d". Thank you for explaining!

Kernel/Inference.cpp
quickbeam123: No, not really. Just curious!

SAT/SATInference.cpp
quickbeam123: If you don't mind. I just remembered how proud I was those 10+ years ago to think of it in the first place. (And how I don't like the long `Filtered` version, which we however still need.)

Saturation/Splitter.cpp
quickbeam123: Definitely, the logic was convoluted. (And from me.) Thanks for making things more streamlined!

MichaelRawson: Didn't mean to call your code names - I think it made sense in the "we have 4 SAT solvers" era!

PR #755, opened 2025-08-13
Refactoring answer literal producing logic into a separate simplification rule (for synthesis only)

hzzv:
This PR moves the if-then-else forming logic for answer literals (ALs) into a separate simplification rule instead of having it in BinaryResolution and Superposition. Since the new simplification rule is triggered before any other rule, it derives the clauses that BR/Sup derived before this PR.

Additionally, the new rule also derives the clauses with `r != r’ \/ ans(r)`, thus effectively finally implementing the right-hand column versions of rules from Fig. 3 of the CADE’23 synthesis paper: https://easychair.org/publications/preprint/KRmQ/open

E.g. for binary resolution, before we only had:
```
C \/ L \/ ans(r)      D \/ ~L' \/ ans(r')
-------------------------------BR with AL
(C \/ D \/ ans(if L then r’ else r))θ
```
where `θ := mgu(L, L')`.

Now we have both:
```
C \/ L \/ ans(r)      D \/ ~L' \/ ans(r')
-------------------------------BR
(C \/ D \/ ans(r) \/ ans(r’))θ
-------------------------------AL processing (if-then-else)
(C \/ D \/ ans(if L then r’ else r))θ
```
and:
```
C \/ L \/ ans(r)      D \/ ~L' \/ ans(r')
-------------------------------BR
(C \/ D \/ ans(r) \/ ans(r’))θ
-------------------------------AL processing (constraint)
(C \/ D \/ r != r' \/ ans(r))θ
```

Code-wise, this PR:
1. Renames `AnswerLiteralRemovals.{c,h}pp` to `AnswerLiteralProcessors.{c,h}pp`
2. Adds class `SynthesisAnswerLiteralProcessor` which triggers on clauses derived by BR or Sup, and if they contain 2 answer literals, it replaces the clause by both the if-then-else and the constraint versions with a joined answer literal.
3. To create the if-then-else joined AL, we have to have access to the literal upon which we resolved (for BR) or the equality literal (for Sup) after the substitution has been applied, and we also have to know which answer literal came from which premise. To this end, this PR introduces a new `proofExtra` field: `synthesisExtra`. This is populated in both BR and Sup.

The main advantage of this change is robustness, as it decouples the proving and synthesizing logic.
Further, following this PR it might be possible to refactor also the synthesis-recursive branch and then finally merge it into master, since the synthesis logic will be more-or-less decoupled from the proving logic also for recursive program synthesis.


Kernel/Clause.cpp
```diff
@@ -193,7 +193,8 @@ void Clause::destroy()
   static Stack<Clause*> toDestroy(32);
   Clause* cl = this;
   for(;;) {
-    if (env.options->proofExtra() == Options::ProofExtra::FULL) {
+    if ((env.options->proofExtra() == Options::ProofExtra::FULL) ||
+        ((env.options->questionAnswering() == Options::QuestionAnsweringMode::SYNTHESIS) && env.proofExtra.find(cl))) {
```
mezpusz: I think `remove` should work also if the entry is not in the map, so you can probably remove the `find` call.

Kernel/Inference.cpp
```diff
@@ -601,6 +601,10 @@ std::string Kernel::ruleName(InferenceRule rule)
     return "answer literal with input var skolemisation";
   case InferenceRule::ANSWER_LITERAL_REMOVAL:
     return "answer literal removal";
+  case InferenceRule::ANSWER_LITERAL_JOIN_WITH_CONSTRAINTS:
+    return "answer literal with constraints";
```
mezpusz: Why not add 'join' too? It would make it more descriptive.

Saturation/SaturationAlgorithm.cpp
```diff
@@ -746,16 +746,10 @@ Clause *SaturationAlgorithm::doImmediateSimplification(Clause* cl0)
...
-    }
-    onClauseReduction(cl, &simplCl, 1, 0);
-    return 0;
-  }
-
-  if (auto  cIt = _immediateSimplifierMany.simplifyMany(cl)) {
+  // Note: simplifyMany() has to go before simplify(), since the former
+  // postprocesses clauses with answer literals, while the latter deletes
+  // those which are invalid even after postprocessing.
+  if (auto cIt = _immediateSimplifierMany.simplifyMany(cl)) {
```
mezpusz, line 753: Note for future reference: it would probably be easier to remove `ImmediateSimplificationEngine` completely and use `ImmediateSimplificationEngineMany` to enforce a total order.

Shell/AnswerLiteralManager.cpp
```diff
@@ -528,10 +530,10 @@ bool SynthesisALManager::tryGetAnswer(Clause* refutation, Stack<Clause*>& answer
   while(it.hasNext()) {
     p = it.next();
     ASS(p.second.first != nullptr);
-    ASS(p.first == p.second.first->number());
     if (!proofNums.contains(p.first)) {
       continue;
     }
+    ASS(p.first == p.second.first->number());
```
mezpusz: Could this be an `ASS_EQ`?

Shell/AnswerLiteralManager.cpp
```diff
@@ -519,6 +519,8 @@ bool SynthesisALManager::tryGetAnswer(Clause* refutation, Stack<Clause*>& answer
   ALWAYS(it.hasNext());
   pair<unsigned, pair<Clause*, Literal*>> p = it.next();
   Literal* origLit = p.second.second;
+  while (p.first > 0 && !proofNums.contains(p.first) && it.hasNext()) p = it.next();
```
mezpusz, line 524: Can you explain what is happening in this block? Maybe also add it as a comment?

Shell/AnswerLiteralManager.hpp
```diff
@@ -162,6 +162,17 @@ class SynthesisALManager : public AnswerLiteralManager
     DHMap<unsigned, unsigned> _condFnToPred;
   };
 
+  class ComputableSelfTransformer : public BottomUpTermTransformer {
```
mezpusz: Is this used anywhere? I don't see it's implementation in `AnswerLiteralManager.cpp`.

Inferences/AnswerLiteralProcessors.cpp
```diff
@@ -0,0 +1,198 @@
...
+  static RobSubstitution helperSubst;
+
+  unsigned cLen = cl->length();
+  for (unsigned li = 0; li < cLen; li++) {
+    Literal* lit = (*cl)[li];
+    if (lit->isAnswerLiteral() && lit->isNegative()) {
+      for (Literal* toAvoid : _avoiders->iterLits()) {
+        if (lit->functor() == toAvoid->functor()) {
+          helperSubst.reset();
+          if (helperSubst.matchArgs(toAvoid,0,lit,1)) {
```
mezpusz, line 102: For this (including the functor check) I think you can use `MatchingUtils::matchTerms` from `Kernel/Matcher.hpp`.

Inferences/AnswerLiteralProcessors.cpp
```diff
@@ -0,0 +1,198 @@
...
+
+Option<ClauseIterator> SynthesisAnswerLiteralProcessor::simplifyMany(Clause* cl)
+{
+  if ((cl->inference().rule() == InferenceRule::ANSWER_LITERAL_JOIN_AS_ITE) ||
+      (cl->inference().rule() == InferenceRule::ANSWER_LITERAL_JOIN_WITH_CONSTRAINTS)) {
+    // The clause was produced by this very simplification, and it does not need further checking.
+    return none<ClauseIterator>();
+  }
+  // Count the answer literals in the clause.
+  int numAnsLits = 0;
```
mezpusz: Why not `unsigned`?

Inferences/AnswerLiteralProcessors.cpp
```diff
@@ -0,0 +1,198 @@
...
+  unsigned idx[2];
+  for (unsigned i = 0; i < cl->length(); ++i) {
+    if ((*cl)[i]->isAnswerLiteral()) {
+      ASS(numAnsLits < 2);
```
mezpusz: `ASS_L`?

Inferences/AnswerLiteralProcessors.cpp
```diff
@@ -0,0 +1,198 @@
...
+      idx[numAnsLits] = i;
+      numAnsLits++;
+    }
+  }
+  // This simplification rule deals with clauses with 2 answer literals produced by superposition
+  // or resolution (which fill in sythesisExtra).
+  // Clauses with a different number of answer literals or prodced by other rules are checked
+  // by other simplification rules.
+    if (numAnsLits != 2 ||
```
mezpusz: Probably wrong indentation here?

Inferences/AnswerLiteralProcessors.cpp
```diff
@@ -0,0 +1,198 @@
...
+      ((cl->inference().rule() != InferenceRule::SUPERPOSITION) &&
```
mezpusz, line 150: Shouldn't we assert instead that if `numAnsLits == 2` then the inference should be either of these three?

Inferences/BinaryResolution.cpp
```diff
@@ -241,8 +227,21 @@ Clause* BinaryResolution::generateClause(Clause* queryCl, Literal* queryLit, Cla
 
   inf_destroyer.disable(); // ownership passed to the the clause below
   Clause *cl = Clause::fromStack(*resLits, inf);
-  if(env.options->proofExtra() == Options::ProofExtra::FULL)
+  if (env.options->questionAnswering() == Options::QuestionAnsweringMode::SYNTHESIS) {
+    Literal *qAnsLit = queryCl->getAnswerLiteral(), *rAnsLit = resultCl->getAnswerLiteral();
+    Literal* sqAnsLit = qAnsLit ? subs->applyToQuery(qAnsLit) : nullptr;
```
mezpusz: Maybe I misunderstood but as far as I can tell here the answer literals can be null, but around `AnswerLiteralProcessors.cpp:167` you check that they shouldn't be null.

More generally, wouldn't it make more sense to completely skip creating the proof extra object if there is only one answer literal inherited?

Inferences/AnswerLiteralProcessors.cpp
```diff
@@ -0,0 +1,198 @@
...
+  //      C \/ ans(if cond then t else e)
+  // 2. we use just `ans(t)` and add a constraint on equality of `t` and `e`, producing:
+  //      C \/ t!=e \/ ans(t)
+  // TODO: if the answer literal has multiple arguments, maybe we should try all combinations
+  //       of options 1 and 2 for each argument?
+  ClauseStack res;
+  // Option 1, using if-then-else, only applies if the two answer literals are different.
+  if (thenLit != elseLit) {
+    Literal* newAnsLit = synthMan->makeITEAnswerLiteral(condition, thenLit, elseLit);
+    LiteralStack lits = collectLiteralsExceptForTwo(cl, cl->length()-1, idx[0], idx[1]);
```
mezpusz: This could be calculated only once, then `newAnsLit` can be popped after creating the first result. What do you think?

Inferences/AnswerLiteralProcessors.cpp
hzzv: No, also other inference rules might produce a clause with 2 ALs. If this happens though, the resulting clause should (and will) be removed be removed by another simplification rule.

Inferences/BinaryResolution.cpp
hzzv: Re the first point: that is okay, we only get to `AnswerLiteralProcessors.cpp:167` if the clause has two answer literals.

Re the second point: yes, thanks for the suggestion!

Shell/AnswerLiteralManager.hpp
hzzv: Good catch, it was a leftover from previous iteration. Removed, thanks!

Inferences/AnswerLiteralProcessors.cpp
hzzv: Actually the logic was wrong, it can be >= 2 (for clauses where this rule does not apply). Changed now.

Inferences/AnswerLiteralProcessors.cpp
hzzv: This actually isn't my code, it's just that I probably forgot to tell git that I'm just renaming the file.
Anyway, I can make the change if it really is equivalent -- @quickbeam123 could you please confirm?

quickbeam123: Happy to learn about this (for me) new function too :)

quickbeam123: (but also, feel free not touching this - as it ain't broken)

PR #757, opened 2025-08-20
Polymorphic NewCNF

mezpusz:
Regression over TPTP with default options (i.e. `-newcnf off`) shows no difference.

Regression over TPTP with `-newcnf on`:

Discount:
master unsat: 9122 (2) sat: 1074 (8)
poly-newcnf unsat: 9154 (34) sat: 1066 (0)

Otter:
master unsat: 9193 (5) sat: 1069 (8)
poly-newcnf unsat: 9219 (31) sat: 1061 (0)

I checked the lost problem instances manually and they can be attributed to adding FOOL axioms `($true = X0) | ($false = X0)` and `($true != $false)` to the input in `Shell/Preprocess.cpp` which (I believe) is needed to handle Boolean proper terms in some cases, but seems to be unnecessary when preprocessing can remove all FOOLish stuff completely. This should be somehow fixed before merging.

Debug sampling did not show any new bugs or assertions.


Shell/NewCNF.cpp
```diff
@@ -948,36 +949,70 @@ void NewCNF::ensureHavingVarSorts()
   }
 }
 
+TermList NewCNF::getVarSort(unsigned var) const
+{
+  ASS(_collectedVarSorts);
+  // TODO shouldn't we enforce that the sort is explicitly known?
+  auto sort = _varSorts.get(var, AtomicSort::defaultSort());
```
MichaelRawson, line 955: I assume this happens when we write e.g. `![X]: ...` rather than `![X : $i]: ...` as some TPTP dialects permit it. Would it be easy to insert `AtomicSort::defaultSort` when that happens and then we can make this lookup an error if it fails?

Also, what happens with unbound variables? Do we "implicitly universally quantify" them?

Shell/NewCNF.cpp
```diff
@@ -1136,12 +1174,23 @@ void NewCNF::process(QuantifiedFormula* g, Occurrences &occurrences)
...
+  LOG2("occurrences", occurrences.size());
+
   if (ts.isVar()) {
     processBoolVar(POSITIVE, ts.var(), occurrences);
     return;
   }
 
   Term* term = ts.term();
+  if (!term->isSpecial()) {
+    // TODO not sure this is the right way to do it, or whether it is even correct in all cases
```
MichaelRawson: Not sure anybody else is either. This seems reasonable for now.

Shell/NewCNF.cpp
```diff
@@ -948,36 +949,70 @@ void NewCNF::ensureHavingVarSorts()
...
-    unsigned uvar = vit.next();
-    domainSorts.push(_varSorts.get(uvar, AtomicSort::defaultSort()));
-    fnArgs.push(TermList(uvar, false));
-  }
+  TermList rangeSort=getVarSort(var);
+  Recycled<TermStack> termVarSorts;
+  Recycled<TermStack> typeVars;
+  Recycled<TermStack> termVars;
+
+  iterTraits(free->iter())
```
MichaelRawson: I think I'd prefer `for(unsigned uvar : iterTraits(free->iter())) { ... }` unless you have a strong preference for the closure thing.

Shell/NewCNF.cpp
```diff
@@ -948,36 +949,70 @@ void NewCNF::ensureHavingVarSorts()
...
+        termVarSorts->push(varSort);
+      }
+    });
+
+  SortHelper::normaliseArgSorts(*typeVars, *termVarSorts);
+  SortHelper::normaliseSort(*typeVars, rangeSort);
+
+  auto taArity = typeVars->size();
+
+  // TODO maybe this should be avoided for type cons
```
MichaelRawson: For efficiency reasons, you mean? Because there won't be any term arguments?

Shell/NewCNF.cpp
```diff
@@ -948,36 +949,70 @@ void NewCNF::ensureHavingVarSorts()
...
+    ASS(termVars->isEmpty() && termVarSorts->isEmpty());
+    ASS_EQ(taArity, arity);
+    unsigned typeCon = Skolem::addSkolemTypeCon(arity);
+    Signature::Symbol *sym = env.signature->getTypeCon(typeCon);
+    sym->markSkipCongruence();
+    if(_beingClausified->derivedFromGoal()){
+      sym->markInGoal();
+    }
+    if(_forInduction){
+      sym->markInductionSkolem();
```
MichaelRawson: Do you want to induct on a *Skolem type*? :exploding_head:

Shell/NewCNF.cpp
```diff
@@ -986,11 +1021,14 @@ Term* NewCNF::createSkolemTerm(unsigned var, VarSet* free)
...
-    res = Term::create(fun, arity, fnArgs.begin());
+    res = Term::create(fun, arity, args.begin());
   }
 
-  domainSorts.reset();
-  fnArgs.reset();
+  // Store type variables and their Skolemized form in a substitution
+  // which is then applied on variable sorts to get the right ones.
+  if (rangeSort == AtomicSort::superSort()) {
+    _skolemTypeVarSubst.bindUnbound(var, res);
```
MichaelRawson, line 1018: Shouldn't there be some logic to actually apply this substitution to sort variables? Maybe I missed it. e.g. the conjecture

`![X : $tType]: nil(X) = nil(X)`

becomes

`nil(sK) != nil(sK)`.

MichaelRawson (approved):

MichaelRawson:
Basically fine and much less painful than I expected! See comments. I agree about the FOOL axioms: perhaps we should only add them if some *property* `hasFOOL` is still the case after CNF transformation (but before preprocessing).

Shell/NewCNF.cpp
mezpusz: I think putting in `AtomicSort::defaultSort()` might be already done by `SortHelper::collectVariableSorts`, as it goes over all variables (even unbound ones). I will add an asserting lookup and see if it causes any trouble.

Shell/NewCNF.cpp
mezpusz: I didn't know about this idiom, thanks for the suggestion!

Shell/NewCNF.cpp
mezpusz: Yes, but iterating on an empty stack is still better than having different code for the 3 branches, right? 😟

Shell/NewCNF.cpp
mezpusz: I hope not. 😄 Thanks for catching this! Also, I refactored the symbol marking a bit, see new commit.

Shell/NewCNF.cpp
mezpusz: There is the new `getVarSort` which is used whenever a variable sort needs to be instantiated, otherwise terms get their proper sorts from type arguments and the clauses are substituted with the right Skolems at the end of clausification. Am I missing something?

MichaelRawson:
Maybe `hasFOOL`? Have a look in `Property.cpp` and see if it coincides with what you want - there is some refresh logic so it should represent the "current state" of the problem.

Shell/NewCNF.cpp
MichaelRawson: Right - and NewCNF is likely not the hot path.

Shell/NewCNF.cpp
MichaelRawson: >clauses are substituted with the right Skolems

Right - so I'd expect that `_skolemTypeVarSubst` would somehow be involved there, but I don't see it.

Shell/NewCNF.cpp
mezpusz: Okay, turns out there is the case when `SortHelper::collectVariableSorts` does not collect the sort of a variable, namely, when it has default sort and it's not present in the formula (meaning we wouldn't need Skolemization anyways). Should I put the default value back?

Shell/NewCNF.cpp
mezpusz: No, `_skolemTypeVarSubst` is only used for getting the Skolemized sorts quicker, otherwise each `GenClause` has a list of its bindings that are applied `NewCNF::toClause`.

mezpusz:
Yes, calling `hasFOOL` seems to work, most of lost problems disappeared.

Shell/NewCNF.cpp
MichaelRawson: I don't see why this function would ever be called in that case - but if it is somehow, go ahead! Thanks for the detective work.

mezpusz:
Updated regression test with TPTP:

DIscount:
master unsat: 9122 (2) sat: 1074 (0)
poly unsat: 9153 (33) sat: 1074 (0)

Otter:
master unsat: 9193 (6) sat: 1069 (0)
poly unsat: 9218 (31) sat: 1069 (0)

PR #758, opened 2025-09-01
Basic recursive induction (+ other minor synthesis stuff)

hzzv:
This PR includes:
- synthesis of recursive programs (*without* induction with a non-standard base case, this is expected to be supported by multi-literal induction with free variable)
- synthesis-related URR bugfixes
- TODO: marking of all symbols introduced after preprocessing as uncomputable

The overall overhead is expected to be small, since most of the new code is added only onto the synthesis path.


Inferences/Induction.cpp
```diff
@@ -23,21 +23,24 @@
 #include "Indexing/ResultSubstitution.hpp"
 #include "Lib/BitUtils.hpp"
 #include "Lib/DHMap.hpp"
+#include "Lib/DHSet.hpp"
```
mezpusz, line 26: `DHSet.hpp` includes `DHMap.hpp`, so this can be a replacement instead of adding one more include.

Inferences/Induction.cpp
```diff
@@ -519,7 +522,10 @@ void InductionClauseIterator::processClause(Clause* premise)
   // or it should be an integer constant comparison we use as a bound.
   if (InductionHelper::isInductionClause(premise)) {
     for (unsigned i=0;i<premise->length();i++) {
-      processLiteral(premise,(*premise)[i]);
+      Literal* lit = (*premise)[i];
+        if (!lit->isAnswerLiteral()) {
```
mezpusz: Misaligned tabulation?

mezpusz: Also, can you please use a range-based for loop here?

Inferences/Induction.cpp
```diff
@@ -817,7 +823,8 @@ void InductionClauseIterator::processLiteral(Clause* premise, Literal* lit)
         resolveClauses(kv.first, ctx, kv.second);
       }
     }
-  } else if (!env.options->inductionGroundOnly() && InductionHelper::isStructInductionOn() && InductionHelper::isNonGroundInductionLiteral(lit)) {
+  } else if (((env.options->questionAnswering() == Options::QuestionAnsweringMode::SYNTHESIS) || !env.options->inductionGroundOnly()) &&
```
mezpusz: Hm, shouldn't this be expressed with a constraint on the level of options? If we have QA and induction, then `indgo` should be `off`? Otherwise, this could be even unexpected for the user. What if someone wants to use induction to prove their synthesized program correct but do not want the program to use recursion?

Inferences/Induction.cpp
```diff
@@ -1150,17 +1158,36 @@ Clause* resolveClausesHelper(const InductionContext& context, const Stack<Clause
...
+        } else {
+          resLits->push(l);
         }
-        resLits->push(renaming.apply(l,1));
       }
     }
   }
 
+
+  if (indLitSubst) {
```
mezpusz, line 1171: Can you please explain why this is needed?

Inferences/Induction.cpp
```diff
@@ -1363,9 +1390,15 @@ void InductionClauseIterator::performStructInductionFreeVar(const InductionConte
   if (context._indTerms.size() > 1) return;
   TermList sort = SortHelper::getResultSort(context._indTerms[0]);
   TermAlgebra* ta = env.signature->getTermAlgebraOfSort(sort);
+
+  // Synthesis specific:
+  SynthesisALManager* synthMan = static_cast<SynthesisALManager*>((env.options->questionAnswering() == Options::QuestionAnsweringMode::SYNTHESIS) ? SynthesisALManager::getInstance() : nullptr);
+  List<Term*>* fnHeads = List<Term*>::empty();
```
mezpusz: Does this *really* have to be a `List`?

Inferences/Induction.cpp
```diff
@@ -1363,9 +1390,15 @@ void InductionClauseIterator::performStructInductionFreeVar(const InductionConte
...
+  vector<Shell::SkolemTracker> skolemTrackers;
+
   unsigned numTypeArgs = sort.term()->arity();
   unsigned freeVar = context.getFreeVariable(); // variable free in the induction literal
-  unsigned var = freeVar+1; // used in the following to construct new variables
+  unsigned var = freeVar + 1 + (synthMan ? synthMan->numInputSkolems() : 0); // used in the following to construct new variables
```
mezpusz, line 1399: Will these variable somehow collide with the input variables when giving the answer? Cannot they just be renamed at that point?

Inferences/Induction.cpp
hzzv: I'm not sure what is the best decision here -- it might be surprising for the user that `-qa synthesis -ind struct` is not sufficient to synthesize a recursive function. On the other hand, you make a good point with the use case of the proof using induction but the program not using recursion... (But then what if the user wants the proof to use induction with a free variable BUT does not want a recursive program as a result? :))

For now I am happy to make it such that to get recursive programs, all of `-qa synthesis -ind struct -indgo off` are needed. I guess the semantics of the synthesis-specific options can be also changed later.

Inferences/Induction.cpp
hzzv: What other structure do you propose? (I don't have a preference, it just needs to preserve the order.)

Inferences/Induction.cpp
hzzv: Yes, they would collide. In the function construction phase it would have to (1) construct the renaming substitution, and (2) rename both the function definition head and the body. I think it's simpler to just add the constant here.
Consider the program for computing the length of concatenation of lists `X0` and `X1`:
```
% Recursive function definitions:
rf20(nil) = len(X1)
rf20(cons(X4,X5)) = s(rf20(X5))
% SZS answers Tuple [[rf20(X0)]|_] for list-len_concat
```
The `X1`  is the input variable, the `X5` is coming from the constructor definition.

Inferences/Induction.cpp
hzzv: Done.

Inferences/Induction.cpp
hzzv: Done.

Inferences/Induction.cpp
hzzv: Done.

Inferences/Induction.cpp
```diff
@@ -518,8 +520,10 @@ void InductionClauseIterator::processClause(Clause* premise)
   // The premise should either contain a literal on which we want to apply induction,
   // or it should be an integer constant comparison we use as a bound.
   if (InductionHelper::isInductionClause(premise)) {
-    for (unsigned i=0;i<premise->length();i++) {
-      processLiteral(premise,(*premise)[i]);
+    for (Literal* lit : premise->iterLits()) {
+      if (!lit->isAnswerLiteral()) {
```
MichaelRawson, line 525: Could there be a more principled approach for this kind of thing in general? I don't mind for this PR but I could imagine doing `for(Literal *l : premise->reasonableThingsToDoInferenceWith()) ...`

Inferences/Induction.cpp
```diff
@@ -817,7 +821,7 @@ void InductionClauseIterator::processLiteral(Clause* premise, Literal* lit)
         resolveClauses(kv.first, ctx, kv.second);
       }
     }
-  } else if (!env.options->inductionGroundOnly() && InductionHelper::isStructInductionOn() && InductionHelper::isNonGroundInductionLiteral(lit)) {
+  } else if ((!env.options->inductionGroundOnly()) && InductionHelper::isStructInductionOn() && InductionHelper::isNonGroundInductionLiteral(lit)) {
```
MichaelRawson: This is the same, right?

Inferences/Induction.cpp
MichaelRawson: I'd use `std::vector` unless there's a good reason not to (and there could well be). Linked lists are far too easy not to free by mistake.

Inferences/URResolution.cpp
```diff
@@ -113,6 +113,7 @@ struct URResolution::Item
     _lits.reserve(litslen);
     unsigned nonGroundCnt = 0;
     for(unsigned i=0; i<clen; i++) {
+      if(!(*cl)[i]->ground()) nonGroundCnt++;
```
MichaelRawson, line 116: :grimacing: What happened before?

Kernel/Term.cpp
```diff
@@ -1920,33 +1912,59 @@ TermList Term::typeArg(unsigned n) const
...
-    TermList t = sit.next();
-    if (t.isTerm() && !env.signature->getFunction(t.term()->functor())->computable()) {
-      return false;
+  for (unsigned i = 0; i < _arity; i++) {
+    const TermList* t = nthArgument(i);
+    if (t->isTerm()) { // else we have a variable, which needs no check
+      if (idx) {
+        *idx = i;
+      }
+      if (!t->term()->computableOrVarHelper(recAncestors)) {
```
MichaelRawson: Ideally this wouldn't be recursive. If you're happy with an O(n) solution would `BottomUpEvaluation` be an option?

Kernel/TermTransformer.cpp
```diff
@@ -350,7 +350,8 @@ Formula* BottomUpTermTransformer::transform(Formula* f)
 TermList BottomUpTermTransformer::transform(TermList ts)
 {
   if (ts.isTerm()) {
-    return TermList(transform(ts.term()));
+    // try transform subterms and try transforming the whole term
+    return transformSubterm(TermList(transform(ts.term())));
```
MichaelRawson: Ooh, this looks scary. @quickbeam123 - do you immediately recall what goes on here?

Shell/AnswerLiteralManager.hpp
```diff
@@ -132,8 +134,42 @@ class PlainALManager : public AnswerLiteralManager
...
+typedef List<Binding> BindingList;
+
+struct SkolemTracker { // used for tracking skolem terms in the structural induction axiom (recursive program synthesis)
+  Binding binding;
+  unsigned constructorId; // A skolem constant will be considered computable only in the constructorId'th arg of rec(.)-term
+  bool recursiveCall; // Whether the constant corresponds to a recursive call result
+  unsigned indexInConstructor; // The index of the variable/Skolem in the constructor OR if 'recursiveCall' is true, to which argument of the constructor the recursive call coresponds to.
+  unsigned recFnId; // ID number of the associated rec-function
+
+  SkolemTracker() {}
```
MichaelRawson, line 148: In this constructor all the POD members (`constructorId`, `recursiveCall`, etc.) are left in unspecified state - do you really want this? If not then initialise e.g. `unsigned constructorId = 0;`. If so fair enough, but it's worth a check as the bugs are not fun.

Inferences/Induction.cpp
hzzv: I guess originally it was a given that all literals are reasonable to infer with :)
Anyway, maybe this could be a follow up, where what does `reasonableThingsToDoInferenceWith()` return depends on the value of `-qa`.

Inferences/Induction.cpp
hzzv: Yes, the parentheses were a leftover from going back and forth, removing them now!

Inferences/URResolution.cpp
hzzv: Bad merge, sorry 😅 
Fixed it now -- removing the other non-ground counter incrementation.

Kernel/Term.cpp
hzzv: No, when we are evaluating whether a sub-term is computable, we need to know in what context is it occuring.
E.g., it can be the case that `sK2` in `f(sK2)` is considered uncomputable, but the same `sK2` in `sk3(X, sK2, Y)` is computable (because `sK2` has a computable meaning within the second argument of `sK3`, but not anywhere else).
Any suggestions on how to do that better than with custom recursion?

Kernel/TermTransformer.cpp
hzzv: Example of what did not work before:
if we want to transform term by replacing the constant `a` by the constant `b`,
then transforming `a` itself would yield `a`, not `b`,
while transforming `f(a)` would yield `f(b)`.

hzzv (commented):
Changes and more comments to come.

Inferences/Induction.cpp
hzzv: Fair point. Turns out this also simplified processing down the line. Thanks for the suggestion!

Shell/AnswerLiteralManager.hpp
hzzv: Good point. I was thinking of deleting the default constructor, but unfortunately DHMap uses it when expanding its size, so that seems to be a no-go. (Unless I'm missing something?)
I added some default values.

Inferences/Induction.cpp
MichaelRawson: Follow-up sounds fine!

Shell/AnswerLiteralManager.hpp
MichaelRawson: Yeah, I think `DHMap` is broken in this sense - it should just allocate the memory and not construct any objects. Sometimes I use `std::unordered_map` instead. One day I will fix it...

Inferences/Induction.cpp
mezpusz: Thanks for the explanation!

Kernel/Term.cpp
```diff
@@ -15,15 +15,16 @@
...
 #include "SubstHelper.hpp"
 #include "TermIterators.hpp"
 #include "RobSubstitution.hpp"
+
+#include "Indexing/TermSharing.hpp"
 #include "Kernel/HOL/HOL.hpp"
+#include "Lib/Metaiterators.hpp"
+#include "Lib/Output.hpp"
+#include "Lib/StringUtils.hpp"
+#include "Shell/AnswerLiteralManager.hpp"
```
mezpusz: I am not too much in favor of having higher-level structures like `AnswerLiteralManager` in the implementation of something so low-level as `Term`. Can everything that requires `AnswerLiteralManager` be moved to e.g. `AnswerLiteralManager` itself?

Kernel/Term.cpp
```diff
@@ -1920,33 +1912,59 @@ TermList Term::typeArg(unsigned n) const
...
+  if (synthMan->isRecTerm(this)) {
+    ALWAYS(recAncestors->getValuePtr(functor(), idx, -1));
   }
-  SubtermIterator sit(this);
-  while (sit.hasNext()) {
-    TermList t = sit.next();
-    if (t.isTerm() && !env.signature->getFunction(t.term()->functor())->computable()) {
-      return false;
+  for (unsigned i = 0; i < _arity; i++) {
+    const TermList* t = nthArgument(i);
```
mezpusz: If you're not considering uncomputability of type arguments, I suggest looping up to `numTermArguments()` and using `termArg(i)`.

Kernel/Term.cpp
mezpusz: If I understand correctly, computability of a subterm depends on the position of the subterm and the function headers seen on the path leading there. So basically a sequence of pairs of natural numbers and function symbols. For example, this sequence for `sK2` in `sk3(X, sK2, Y)` is `((2, sK3))` and you can determine that `sK2` is computable by asking if `sK2` is computable in the 2nd position of `sK3` (and in general whether it is computable in the `i`th position of `f` for all `(i,f)` pairs in the sequence).

If this is really the case, you could (probably) implement this with a stack containing the current term path and a stack containing the indices, or probably also with `BottomUpEvaluation`. (If I understand correctly, in the worst case you would have to traverse the entire term anyways, right?)

Kernel/TermTransformer.cpp
mezpusz: I can testify to this, I tried to use it recently and it didn't work for the term itself. 😞 Could you maybe factor out the common `transformSubterm` call? Maybe
```C++
return transformSubterm(ts.isVar() ? ts : TermList(transform(ts.term())));
```
?

Inferences/Induction.cpp
mezpusz: Yes, we can change it later. This is a kind of vague part that will hopefully be unveiled in further papers. 🙂

Kernel/Term.cpp
quickbeam123: I second this. If at all possible...

Kernel/TermTransformer.cpp
quickbeam123: The main question is (since we never had spelled out what this should be doing really), whether some other uses of `BottomUpTermTransformer` aren't becoming incorrect due to the change ...

Kernel/Term.cpp
hzzv: @mezpusz yes, you describe the computability conditions well.
However, I'm not sure I get the implementation suggestions. With two stacks we would have to iterate through both of them to check the computability of each subterm, leading to worse-than-linear complexity, no? As for the `BottomUpEvaluation`, I'm not sure how to use that -- does the evaluation function for each subterm have access to the position in the superterm? (Although even if it does, it would again mean we have to go through the whole list of encompassing functions for each subterm, which again has a worse complexity.)

In my solution I am using a hash map indexed by the function id (in the example above, we know that `sK2` is only computable within the second argument of `sK3`, so we can easily check if the map contains `<sK3, 2>`), so the complexity should be linear.

And actually, what's the reason to avoid recursion?

Kernel/TermTransformer.cpp
quickbeam123: ... didn't @joe-hauns create a (more principled) replacement class for bottom up term transforming, which we eventually wanted to transition to everywhere?

joe-hauns: I also wouldn't dare changing this in `Kernel/TermTransformer.cpp` because i can imagine other usese of the transformer rely on this (rather unexpected) behaviour. I think when changing it one needs to carefully inspect all uses of the `BottomUpTermTransformer`. But I also agree that it actually should be changed.

Concerning my alternative to `BottomUpTermTransformer`:
There is  `Kernel/BottomUpEvaluation.hpp` (for sample usage see `UnitTests/tBottomUpEvaluation.cpp`).
It is faster as it does not require dynamic dispatch, can work with more different datatypes, allows for skipping of type arguments, allows different return types (e.g. bottomup-evaluating Term -> Int), optional use of memorization, and a few other things. If you wanna use it and have questions feel free to text me on zulip :)

Kernel/Term.cpp
MichaelRawson: > And actually, what's the reason to avoid recursion?

Good question - the rest of Vampire avoids it because you can hit stack overflow. But (i) this may not be true with modern stack sizes, and (ii) you would need a very deeply-nested term, so I can see arguments both ways. Especially as an explicit stack is usually slower than the implicit system call stack.

joe-hauns: I'm personally also not a big fan of our explicit recursion stack paradigm. It leads to a lot of very ugly code that could b very simple code if we'd use recursion. 

Andrei's been arguing for avoiding recursion though as there are some problems in SMTLIB/TPTP that have very deeply nested terms (terms of a few MB size or something like that ...?) and processing these does not work with the implicit stack.

Kernel/Term.cpp
hzzv: Done.

Kernel/Term.cpp
hzzv: Good point, I moved it to `SynthesisALManager` in `Shell/AnswerLiteralManager`.

Shell/AnswerLiteralManager.cpp
```diff
@@ -761,32 +780,294 @@ TermList SynthesisALManager::ConjectureSkolemReplacement::transformTermList(Term
...
+        return false;
+      }
+      return true;
+    }
+  }
+
+  // Top functor is computable, now recurse.
+  unsigned* idx = nullptr;
+  if (isRecTerm(t)) {
+    ALWAYS(recAncestors->getValuePtr(f, idx, -1));
```
mezpusz: Does this mean that there cannot be terms like `sK3(X0,sK3(X1,sK2,X2),X3)` with the same functor nested?

Kernel/Term.cpp
hzzv: I'm leaning towards keeping the recursion, since this function is not on the main proving path -- it is ever only called in the synthesis mode, and for that I don't expect very deeply nested terms to pop up.

@MichaelRawson if you strongly disagree, I can rewrite the same function to use explicit stack (though I think Joe is right, it will be ugly). Just let me know sooner rather than later, otherwise someone else will have to rewrite it :D

Inferences/InductionHelper.cpp
```diff
@@ -148,7 +148,8 @@ bool InductionHelper::isInductionClause(Clause* c) {
   static bool goal_plus = (kind == Options::InductionChoice::GOAL_PLUS);
   static unsigned maxD = env.options->maxInductionDepth();
   static bool unitOnly = env.options->inductionUnitOnly();
-  return ((!unitOnly || c->length()==1) && 
+  static bool ansLits = (env.options->questionAnswering() == Options::QuestionAnsweringMode::SYNTHESIS || env.options->questionAnswering() == Options::QuestionAnsweringMode::PLAIN);
```
mezpusz, line 151: Is there a case when `c->hasAnswerLiteral()` is true and we don't want to ignore the answer literal? If not, could computing and checking `ansLits` be skipped?

Kernel/Term.cpp
mezpusz: There is an example for a stack-based solution [here](https://github.com/vprover/vampire/blob/master/Kernel/KBO.cpp#L190), the trick is to use `TermList*` as a list and undo whatever you would do after the recursive call when you reach the [end of the list](https://github.com/vprover/vampire/blob/master/Kernel/KBO.cpp#L193). It's not pretty but not too ugly either in my opinion. :D

As for complexity, you're right, but I think it's only $$O(n\ \log\ n)$$, but using a hashmap makes it probably better as you do. I have another question for this code though (see other comment), so maybe I misunderstood the requirements.

MichaelRawson: > if you strongly disagree

I don't! That seems pragmatic to me - if it's not on the main path, no problem - if it becomes a problem later, somebody else can fix it without _too_ much trouble. :-)

Kernel/TermTransformer.cpp
hzzv: For now I chose the simplest solution -- reverted my change to `BottomUpTermTransformer` and added that one extra `transformSubterm` call from within the synthesis code. In this case the speed does not matter much, because we only call this once, upon successful derivation, to create the answer from the answer literal argument(s).

Shell/AnswerLiteralManager.cpp
hzzv: Hm, good point. I *think* there should be no such thing (and we did not encounter it in our examples so far). However, since my argument (below) is about why it doesn't make sense to derive that than about why it cannot ever happen, I'm changing the code to return "this is uncomputable" if that ever occurs.

Such content of an answer literal would correspond to a recursive function call, where one of the branches of the function is defined in two ways. In your example, the second branch should simultaneously be whatever `sK2` encodes, and the result of the call of the function itself on `X2` when the first branch is `X1`. This would however mean that the function is only a valid answer if those two things are equal, which should be dealt with within reasoning, not within function construction.
(And it probably also implies that `X0 = X1`?)

Whether this occurs in practice -- and if yes, whether we should deal with it by adding a condition like `sK2 != sK3(X1, sK2, X2)` -- might be something to investigate in the future.

Inferences/InductionHelper.cpp
hzzv: I added this so that we don't have to check `c->hasAnswerLiteral()` if there is no way that there will be an answer literal. The `c->hasAnswerLiteral()` is linear in the number of literals of the clause, while `ansLits` is just a single bit.

mezpusz: Yeah, that makes sense!

PR #759, opened 2025-09-16
Generalize induction to non-ground formulas

mezpusz:
This PR generalizes the current induction framework with all of its features (except synthesis) to forall-exist formulas (from just forall formulas). In other words, we may use non-ground literals in the induction formulas with the option `-indgo off` but we only induct on ground terms as before. This PR is mostly independent of the recently merged synthesis+induction PR.

Some of the main changes in detail:
- Since we now have to apply a substitution during the induction formula application on not only the induction clauses (the clausification of the induction formula) but also on the $$n$$ input clauses, the input clauses are in a fixed order inside `InductionContext` and the associated `InductionFormulaIndex` entry to make the application of this substitution easier. Because of this the `InductionContext` class was refactored and these objects are instantiated slightly differently.
- We need $$n+2$$ of the above-mentioned substitutions (one for all the induction clauses, one for the induction context, the rest for the input clauses which may contain variables that need to be taken distinct), so I just started using a `RobSubstitution` instead.
- We now quantify the exists part inside, this may need to be changed to some prenex quantification when we integrate this and synthesis together.
- I found some smaller bugs and refactored some things on the way.
- The synthesis code with the one structural induction formula is triggered when `-qa synthesis` is used, we have to revise this once everything is put together.

Core and synthesis regressions did not show any changes.

The induction regression showed a slight difference, namely:
```
master unsat: 952 (12) sat: 0 (0) cputime: 11631.22 s instructions: 45544131 Mi memory: 609500.76 MB
PR unsat: 959 (19) sat: 0 (0) cputime: 11807.60 s instructions: 45514255 Mi memory: 680203.14 MB
```
I checked the lost benchmarks manually and they are all solved locally by both versions, so it is probably just noise. However, we won some problems in the synthesis benchmark set that can be solved now for example with non-unit induction.

TODO: add sanity check benchmark for non-ground induction and synthesis.


Shell/NewCNF.cpp
```diff
@@ -39,7 +39,7 @@ using namespace Kernel;
 
 namespace Shell {
 
-void NewCNF::clausify(FormulaUnit* unit,Stack<Clause*>& output, DHMap<unsigned, Term*>* bindings)
+void NewCNF::clausify(FormulaUnit* unit,Stack<Clause*>& output, Substitution* subst)
```
quickbeam123, line 42: Could you please remind me what `bindings`/`subst` are for? Is this a non-compulsory "result" argument, perhaps only used by `Induction`?

mezpusz: It is the mapping of each variable in the formula to whatever they become in the resulting clauses (for example, renamed or Skolemized). In general this is not a function, because the formula may contain the same variable quantified multiple times, but for induction we make sure that the formula is rectified. Maybe we could give it a nicer name that is closer to what it actually does.

mezpusz: And yes, it is only used by induction currently, where we have to know how the (existential) variables are Skolemized to build the substitution for the induction application.

Indexing/InductionFormulaIndex.hpp
```diff
@@ -53,13 +58,13 @@ class InductionFormulaIndex
       for (const auto& cl : cls) {
         cl->incRefCnt();
       }
-      _st.push(std::make_pair(cls, subst));
-    }
-    const Stack<std::pair<ClauseStack,Substitution>>& get() const {
-      return _st;
+      _indInstances.push({ std::move(cls), std::move(subst) });
```
MichaelRawson, line 61: According to my understanding of `std::move` as cast-to-rvalue, these shouldn't be necessary.

Indexing/TermIndex.cpp
```diff
@@ -146,9 +143,12 @@ void InductionTermIndex::handleClause(Clause* c, bool adding)
   }
 
   // Iterate through literals & check if the literal is suitable for induction
-  for (unsigned i=0;i<c->length();i++) {
-    Literal* lit = (*c)[i];
-    if (!InductionHelper::isGroundInductionLiteral(lit)) {
+  for (const auto& lit : *c) {
```
MichaelRawson, line 146: Nitpicking, but why not just `auto lit : *c`?

Inferences/Induction.cpp
```diff
@@ -235,8 +236,42 @@ unsigned InductionContext::getFreeVariable() const {
...
+    }
+    if (resLits.isEmpty()) {
+      continue;
+    }
+    cls.emplace(cl, std::move(resLits));
+  }
+  return InductionContext(_indTerms, std::move(cls));
+}
+
+std::map<Term*,TermList> InductionContext::getReplacementMap(const std::vector<TermList>& r, RobSubstitution* subst) const
```
MichaelRawson: Do you need an *ordered* map here?

Indexing/InductionFormulaIndex.hpp
mezpusz: I understand what you're suggesting but the compiler complains then that the copy ctor of `subst` is deleted (so it apparently wants to copy it). Maybe I understand the semantics wrong (and as I looked the current value categories up, I got even more confused), but to me the function signature
```c++
void add(ClauseStack&& cls, RobSubstitution&& subst)
```
only tells that the function takes ownership of its arguments and the caller has to explicitly decide whether to pass a copy or not -- the objects are still "up for grabs" until the ends of the function scope. Otherwise I wouldn't be able to write something like
```c++
_indInstances.push({ cls, subst });
_indInstances.push({ std::move(cls), std::move(subst) });
```
which makes copies of the objects first and then passes ownership of the originals.

Indexing/TermIndex.cpp
mezpusz: I had some bad experiences with `auto` copying stuff, so I try to use this idiom wherever possible. This would never break even if we ever changed what we iterate over here. But please change my mind!

Inferences/Induction.cpp
mezpusz: Probably not, good catch, thanks!

Indexing/InductionFormulaIndex.hpp
mezpusz: Okay, [apparently](https://stackoverflow.com/questions/28483250/is-an-rvalue-reference-treated-as-an-lvalue-when-used-within-a-function) `cls` and `subst` are lvalues but they have *rvalue reference* types. Weird.

Indexing/TermIndex.cpp
MichaelRawson: Fine by me! Just wondered as it's (presumably?) a `Literal *`. I admire your discipline.

Indexing/InductionFormulaIndex.hpp
MichaelRawson: Crazy stuff. Thanks for looking it up, apparently I have much to learn.

PR #765, opened 2025-09-27
Refactor of stats

mezpusz:
The main motivation of this PR is to replace ad-hoc counters in `Statistics` for inferences with a counter array indexed by `InferenceRule`, and to try to count them semi-automatically. The automatic counting only works for clauses that are actually added to saturation, and for induction formulas. This covers most of the inferences but not, for example, preprocessing rules -- these need to be handled in the future.

I also added automatic counting for in-proof stats, which just counts how many of each `InferenceRule` occurs in the proof.

Some refactors and cleanups also happened on the way.


Inferences/TautologyDeletionISE.cpp
```diff
@@ -43,7 +43,7 @@ Clause* TautologyDeletionISE::simplify(Clause* c)
       continue;
     }
     // l is positive
-    if (_deleteEqTautologies && EqHelper::isEqTautology(l)) {
```
MichaelRawson, line 46: Why was this ever a thing?

MichaelRawson:
This is really nice.

> This covers most of the inferences but not, for example, preprocessing rules -- these need to be handled in the future.

Could this be done in a similar way when we construct e.g. a `FormulaUnit`? Not necessarily now, just thinking.

Inferences/TautologyDeletionISE.cpp
mezpusz: I have no idea, it just came across it, and it would have seemed absurd that it is ever false. It was indeed never false.

mezpusz:
> Could this be done in a similar way when we construct e.g. a `FormulaUnit`?

Sure, we just need to find the right place to call the stats object. We could even do it in the `Unit` ctor, but it seemed a bit riskier there. Is there only one stats object per thread?

MichaelRawson:
>Is there only one stats object per thread?

Per *thread* yes, but there are only two threads in Vampire: the main thread and the timeout thread. When we `fork` for the portfolio the stats object is duplicated along with the rest of Vampire's memory. Did I understand the question right?

Inferences/TautologyDeletionISE.cpp
MichaelRawson: Weird. Thanks!

mezpusz:
I realized my question was more about whether portfolio workers use the same statistics, but of course if they forked, this is not an issue.

PR #766, opened 2025-10-08
Beta and eta reduction machinery for lambda terms

hetzenmat:
This PR adds support for beta and eta reduction for lambda terms that uses the existing `TermTransformer` infrastructure.

### DIFF
--- a/Kernel/HOL/BetaNormaliser.cpp
+++ b/Kernel/HOL/BetaNormaliser.cpp
@@ -16,8 +16,6 @@
 #include "Kernel/HOL/HOL.hpp"
 
 TermList BetaNormaliser::normalise(TermList t) {
-  // term transformer does not work at the top level...
-  // t = transformSubterm(t);
   return transform(t);
 }
 
--- a/Kernel/HOL/RedexReducer.cpp
+++ b/Kernel/HOL/RedexReducer.cpp
@@ -23,10 +23,6 @@ TermList RedexReducer::reduce(TermList head, TermStack& args) {
   TermList t1Sort = *head.term()->nthArgument(1);
   _t2 = args.pop();
 
-  TermList transformed = transformSubterm(t1);
-
-  if(transformed != t1)
-    return HOL::create::app(t1Sort, transformed, args);
   return HOL::create::app(t1Sort, transform(t1), args);
 }
 
@@ -36,7 +32,7 @@ TermList RedexReducer::transformSubterm(TermList t) {
     if (index == _replace) {
       // any free indices in _t2 need to be lifted by the number of extra lambdas
       // that now surround them
-      return TermShifter::shift(_t2, _replace).first;
+      return _replace == 0 ? _t2 : TermShifter::shift(_t2, _replace).first;
     }
     if (index > _replace) {
       // free index. replace by index 1 less as now surrounded by one fewer lambdas
--- a/Kernel/HOL/TermShifter.cpp
+++ b/Kernel/HOL/TermShifter.cpp
@@ -16,10 +16,8 @@
 
 std::pair<TermList, Option<unsigned>> TermShifter::shift(TermList term, int shiftBy) {
   TermShifter ts = TermShifter(shiftBy);
+  TermList result = ts.transform(term);
 
-  const TermList transformed = ts.transformSubterm(term);
-  const TermList result = transformed == term ? ts.transform(term)
-                                              : transformed;
   return {result, ts._minFreeIndex};
 }
 
--- a/Kernel/TermTransformer.cpp
+++ b/Kernel/TermTransformer.cpp
@@ -200,12 +200,10 @@ Term* TermTransformer::transform(Term* term)
 
     TermList dest = transformSubterm(tl);
     if (tl != dest) {
-      args.push(dest);
       modified.setTop(true);
-      continue;
     }
-    if (tl.isVar()) {
-      args.push(tl);
+    if (dest.isVar() || !exploreSubterms(tl, dest)) {
+      args.push(dest);
       continue;
     }
 


Kernel/HOL/BetaNormaliser.hpp
```diff
@@ -0,0 +1,44 @@
...
+
+using namespace Kernel;
+
+// reduce a term to normal form
+// uses a applicative order reduction strategy
+// Currently use a leftmost outermost strategy
+// An innermost strategy is theoretically more efficient
+// but is difficult to write iteratively TODO
+class BetaNormaliser : public TermTransformer {
+  unsigned reductions;
```
MichaelRawson: If you prefer, you can write `unsigned reductions = 0;` which means you don't have to remember to initialise in the constructor.

Kernel/HOL/HOL.hpp
```diff
@@ -37,18 +37,32 @@ inline bool isFalse(TermList term) {
 std::string toString(const Term &term, bool topLevel);
 
 TermList matrix(TermList t);
-void getHeadAndArgs(TermList term, TermList &head, Kernel::TermStack &args);
+void getHeadAndArgs(TermList term, TermList &head, TermStack &args);
```
MichaelRawson: Is there a reason this and similar functions use out-parameters rather than returning e.g. `std::pair<TermList, TermStack>`?

Kernel/TermTransformer.hpp
```diff
@@ -54,11 +54,35 @@ class TermTransformerCommon {
 class TermTransformer : public TermTransformerCommon {
 public:
   virtual ~TermTransformer() {}
+  TermTransformer() : _sharedResult(true),
+                      _dontTransformSorts(false) {}
```
MichaelRawson: Could the logic be flipped so that the member is `_transformSorts`, initialised to `true`? Could also just be a public field, I don't see any benefit to encapsulation here.

Kernel/HOL/TermShifter.hpp
```diff
@@ -0,0 +1,61 @@
...
+
+#ifndef __TermShifter__
+#define __TermShifter__
+
+#include "Kernel/TermTransformer.hpp"
+#include "Kernel/Term.hpp"
+
+using namespace Kernel;
+
+class TermShifter : public TermTransformer
```
MichaelRawson, line 36: Could classes like this have a brief comment to explain what they do, perhaps with an example?

Kernel/HOL/SubtermReplacer.hpp
```diff
@@ -0,0 +1,50 @@
...
+
+#ifndef __SubtermReplacer__
+#define __SubtermReplacer__
+
+#include "Kernel/SortHelper.hpp"
+#include "Kernel/TermTransformer.hpp"
+
+using namespace Kernel;
+
+class SubtermReplacer : public TermTransformer {
```
MichaelRawson, line 29: How does this differ from e.g. `EqHelper::replace`?

Kernel/HOL/HOL.hpp
```diff
@@ -65,16 +79,30 @@ namespace HOL::create {
...
 
+inline TermList toNameless(Term* term) {
+  return toNameless(TermList(term));
+}
+
 } // namespace HOL::convert
 
+namespace HOL::reduce {
+TermList betaNF(TermList t, unsigned* reductions = nullptr);
+TermList etaNF(TermList t);
```
MichaelRawson, line 104: If we want beta-eta normal form, is there an order this should be applied in? First `betaNF`, then `etaNF`?

Kernel/HOL/EtaNormaliser.hpp
```diff
@@ -0,0 +1,32 @@
...
+
+#include "Kernel/Term.hpp"
+
+using namespace Kernel;
+
+// reduce to eta short form
+// normalises top down carrying out parallel eta reductions
+// for terms such as ^^^.f 2 1 0
+// WARNING Recursing lurks here (even during proof search!)
+// This is BAD! However, an  (efficient) iterative implementation is tricky, so
```
MichaelRawson, line 25: Better something than nothing! I am not convinced that recursion is so evil.

Kernel/HOL/Convert.cpp
```diff
@@ -118,9 +117,9 @@ static TermList toNameless(Kernel::Formula *formula, VarToIndexMap &map) {
       auto *lhs = formula->left();
       auto *rhs = formula->right();
 
-      auto constant = TermList(Term::createConstant(env.signature->getBinaryProxy(strMap.at(conn))));
-      const auto form1 = toNameless(lhs, map);
-      const auto form2 = toNameless(rhs, map);
+      auto constant = TermList(Term::createConstant(env.signature->getBinaryProxy(CONNECTIVE_STR_MAP.at(conn))));
```
MichaelRawson: Could the `CONNECTIVE_STR_MAP` lookup be a function `connectiveString` with a switch/case? I think the intent might be clearer.

Kernel/HOL/EtaNormaliser.hpp
hetzenmat: I also think that this is OK. This comment is originally from Ahmed and I just wanted to preserve it for the future, if we want to implement some improvements.

Kernel/HOL/HOL.hpp
hetzenmat: The reason is that sometimes (in future PRs) we want to pass an existing TermStack (i.e. declared static in the function) to collect the arguments. So we save some allocations.

But since this is not often the case, I also added a version that returns a pair, which resulted in much clearer code.

MichaelRawson:
Thanks for the explanation! That makes sense to me.

>Sure, I think this should also be documented somewhere for future references.

Wiki page could be a good idea if you find a spare minute. Something like "HOL notes".

hetzenmat:
Thanks for checking. I also find this strage.

If you explain how you exactly run these tests I can also try to have a closer look.

Kernel/TermTransformer.hpp
```diff
@@ -53,12 +53,26 @@ class TermTransformerCommon {
...
 protected:
   virtual TermList transformSubterm(TermList trm) = 0;
   Formula* transform(Formula* f) override;
   TermList transform(TermList ts) override;
+
+  virtual void onTermEntry(Term*) {}
+  virtual void onTermExit(Term*) {}
+
+  // default implementation, can override if required
+  virtual bool exploreSubterms(TermList orig, TermList newTerm) {
```
quickbeam123, line 69: This function seems to be inherited in several places, but never called anywhere. 

Am I missing something?

hetzenmat: Thanks for catching that. I missed a change that actually uses `exploreSubterms` in order to not having to recurse into every subterm.

PR #769, opened 2025-10-22
keep the original orientation of equations from the input

MichaelRawson:
Much less painful than I feared! Thanks to @quickbeam123 for (i) this solution and (ii) reminding me that he'd solved it for me when I forgot it the first time around.

For example, if you have the input file

```prolog
fof(decl, axiom, p(a)).
fof(decl, axiom, ~p(b)).
fof(reverses, axiom, b = a).
```

then mainline Vampire produces

```shell-session
$ vampire test.p
% SZS status Unsatisfiable for test
% SZS output start Proof for test
1. p(a) [input(axiom)]
2. ~p(b) [input(axiom)]
3. a = b [input(axiom)]
...
```

flipping the input equation `b = a` to the orientation it prefers, `a = b`, and not telling anyone about it. This is quite annoying for re-constructing Vampire proofs in other systems. With this patch Vampire doesn't do this, or at least does it less - it's quite hard to figure out *all* the places where equations are created, and whether or not they matter.

@ashinkarov - FYI, this should get rid of my disgusting `maybe_sym` hack. I will test this hypothesis shortly.

### DIFF
--- a/Kernel/Inference.cpp
+++ b/Kernel/Inference.cpp
@@ -640,6 +640,8 @@ std::string Kernel::ruleName(InferenceRule rule)
 //  case MINISCOPE:
   case InferenceRule::CLAUSIFY:
     return "cnf transformation";
+  case InferenceRule::REORIENT_EQUATIONS:
+    return "reorient equations";
   case InferenceRule::FORMULIFY:
     return "formulify";
   case InferenceRule::REMOVE_DUPLICATE_LITERALS:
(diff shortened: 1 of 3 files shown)


quickbeam123:
That CNF code looks terrible! :)

It's true that we already create part of the inference graph in the TPTP parser when negating the conjecture. 

But would what you propose mean that we would "as if" parse `fof` and only then jump to `cnf` with the reorienting step? If we could somehow pretend the original formula was a clause too, that would be the best...

PR #772, opened 2025-10-31
remove MultiCounter

MichaelRawson:
See #157. `MultiCounter` is essentially a `ZIArray` with a restricted interface, but it doesn't seem to be very useful in practice. Remove it.

### DIFF
--- a/Shell/Property.cpp
+++ b/Shell/Property.cpp
@@ -935,8 +935,6 @@ bool Property::hasXEqualsY(const Clause* c)
  * True if the subformula formula would have a literal X=Y
  * after clausification.
  *
- *
- * @warning Works correctly only with rectified formulas (closed or open)
  * @param f the formula
  * @since 11/12/2004 Manchester, true and false added
  * @since 27/05/2007 flight Frankfurt-Lisbon, changed to new datastructures


Kernel/FormulaVarIterator.hpp
```diff
@@ -74,9 +73,9 @@ class FormulaVarIterator
   unsigned _nextVar;
 
   /** Counter used to store bound variables, together with the number of times they are bound */
-  MultiCounter _bound;
+  ZIArray<unsigned> _bound;
   /** To store previously found free variables */
-  MultiCounter _free;
+  ZIArray<bool> _free;
```
MichaelRawson, line 78: `_free` is only ever used as a Boolean.

Shell/FunctionDefinition.cpp
```diff
@@ -122,21 +122,6 @@ struct FunctionDefinition::Def
   }
 }; // class FunctionDefintion::Def
 
-
```
MichaelRawson, line 125: Turns out the default constructor suffices.

Shell/FunctionDefinition.hpp
```diff
@@ -105,14 +103,7 @@ class FunctionDefinition
   Stack<Def*> _blockedDefs;
 
   Stack<Def*> _safeDefs;
-  /** Counters for occurrences of function symbols */
-  MultiCounter _counter;
-  /** The number of found definitions */
```
MichaelRawson, line 110: Unused fields.

Shell/Property.cpp
```diff
@@ -947,8 +947,7 @@ bool Property::hasXEqualsY(const Clause* c)
  */
 bool Property::hasXEqualsY(const Formula* f)
 {
-  MultiCounter posVars; // universally quantified variables in positive subformulas
-  MultiCounter negVars; // universally quantified variables in negative subformulas
```
MichaelRawson, line 951: `negVars` was never written to. This seems suspicious, but I'm not sure why you would want the negative vars (since they become Skolems, I guess).

PR #774, opened 2025-11-03
TPTP fof/tff formula should not contain an unquantified variable

quickbeam123:
Some users asked for this feature: to complain when we encounter an unquantified variable in fof, such as:
```
fof(a,axiom,![Entity] : human(Entity) -> mortal(Entyty) ).
```
as a good sanity check against typos. (Our current behavior is to silently treat the "free variables" as top-level universally quantified.)

There is a mild performance overhead connected with this. Will run the TPTP numbers soon.

### DIFF
--- a/Parse/TPTP.cpp
+++ b/Parse/TPTP.cpp
@@ -3577,7 +3577,7 @@ void TPTP::endFof()
   if (isFof) { // fof() or tff()
     env.statistics->inputFormulas++;
     if (freeVariables(f)) {
-      USER_ERROR("unquantified variable detected for an fof formula");
+      USER_ERROR("unquantified variable detected for a formula named '",nm,"'");
     }
     original = unit = new FormulaUnit(f,FromInput(_lastInputType));
     unit->setInheritedColor(_currentColor);


Parse/TPTP.cpp
```diff
@@ -3576,6 +3576,9 @@ void TPTP::endFof()
   Unit *unit, *original;
   if (isFof) { // fof() or tff()
     env.statistics->inputFormulas++;
+    if (freeVariables(f)) {
+      USER_ERROR("unquantified variable detected for an fof formula");
```
MichaelRawson: Could we say which formula and/or which variables? Otherwise great news.

quickbeam123: We already get the "(detected at or around line X)" with this.

The problem with the formula is that it may be large and it won't be formatted as in the file.

The problem with the variable is that we readily only hold the internal `unsigned`.

Please shout, if you have an idea for a reasonable compromise.

quickbeam123:
No performance difference over TPTP with `-i 5000`. Also, Geoff seems to be quantifying his variables well!

Parse/TPTP.cpp
MichaelRawson: We have the unit name `nm`? But otherwise I see the problem, fair enough. Later we could consider raising an error at the variable-lookup site if not bound and not parsing CNF.

PR #775, opened 2025-11-04
reset `_vars` after parsing a unit

MichaelRawson:
Clean shuffle at `i=1000`, probably saves some memory if not time.


MichaelRawson:
CI failure is the sanity-check script doing

```
-decode ott+11_1:1_to=lpo:thi=all:sas=z3:bd=all:fdtod=on:sp=const_max:foolp=on:gtgl=5:asg=cautious:gtg=all:rawr=on_1 -lma off -add on -nwc 1.0  Problems/SWC/SWC482_1.p
```

...any ideas, @quickbeam123? Is this one likely to be flaky because of butterflies?

PR #780, opened 2025-11-13
Modernize and clean

mezpusz:
- remove unused includes according to `clangd` (some were not removed because they induced too many other errors to fix)
- `Portiability.hpp` I tried not to remove, as this looks like something we shouldn't, even though it is not used
- apply `unused-override` from the `clang-tidy` rules

Now just testing whether this builds on the server...

### DIFF
--- a/Indexing/CodeTreeInterfaces.cpp
+++ b/Indexing/CodeTreeInterfaces.cpp
@@ -77,7 +77,7 @@ public:
 
   bool isIdentityOnQueryWhenResultBound() override {return true;}
 private:
-  void output(std::ostream& out) const final 
+  void output(std::ostream& out) const final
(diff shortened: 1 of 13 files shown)


mezpusz:
Yes, some places use `VAMPIRE_PERF_EXISTS`, but according to `clangd` only `PortfolioMode.cpp` uses it. If it suffices, I can remove the other occurrences.

mezpusz:
Good question, it turns out `Options.hpp` needs it and it is included in `Timer.cpp` too. But more generally, is there a way to ensure that whereever we check `#if VAMPIRE_PERF_EXISTS` we can make sure that it is actually defined (apart from adding `#ifndef VAMPIRE_PERF_EXISTS static_assert(false); #endif`)? I'm not a macro expert, but this would be good to have.

PR #783, opened 2025-11-19
Disallow unbound type variables

### LINKED ISSUE #744: the polymorphic `$let` binding thing
This from Martin on the new TPTP 9.1 TF1,

```
./vampire_z3_dbg_master_10130 --input_syntax tptp Problems/ITP/ITP259_4.p --show_preprocessing on
...
[PP] input: 6854. ! [X1 : $tType,X0 : $tType] : (linorder(X1) => ! [X29 : fun(X0,X1),X7 : X0,X8 : X0,X292 : list(X0)] : arg_min_list(X0,X1,X29,cons(X0,X7,cons(X0,X8,X292))) = $let(m2151: X0, m2151 := arg_min_list(X0,X1,X29,cons(X0,X8,X292)), ($ite(aa(X1,$o,aa(X1,fun(X1,$o),ord_less_eq(X1),aa(X0,X1,X29,X7)),aa(X0,X1,X29,m2151)), (X7,m2151))) [input(axiom)]
...
Condition in file Indexing/TermSharing.cpp, line 144 violated:
_wellSortednessCheckingDisabled || SortHelper::areImmediateSortsValidPoly(t)
Value of t->toString() is: aa(X1,X0,X2,m2151)
```

The problem here I think is that `m2151` should be a type-quantified function sort capturing the variables in its RHS, but isn't. I think @mezpusz had a proposed fix for this.

mezpusz:
The original intention of the PR was to fix #744, but I started with the assumption that we should disallow unbound type variables in all places. For example, we wouldn't allow an operator type of e.g. a function `f` to be `!>[X0: $tType]:((map(X0,X1) * X0) > X1)` but only `!>[X0: $tType, X1: $tType]:((map(X0,X1) * X0) > X1)` where all variables inside the type are prenex quantified.

I think this check would be reasonable to merge since unbound type variables of terms make the type variables fixed such that from then on those variables cannot be changed anywhere and `X1` cannot be used for anything else in any formula/clause where `f` is present.

Then, to make this work I got to refactoring `let` binders which resulted in a big amounts of code refactored and removed. Not sure if all paths are covered, but this is a relatively stable iteration.

There may be also other theories/operators that needs "polymorphization" like arrays, but I will also leave those for the future.

### DIFF
--- a/Kernel/SortHelper.cpp
+++ b/Kernel/SortHelper.cpp
@@ -110,21 +110,7 @@ TermList SortHelper::getResultSort(const Term* t)
   Signature::Symbol* sym = env.signature->getFunction(t->functor());
   TermList result = sym->fnType()->result();
 
-  /*
-  either
-  1. the substitution is non-empty
-  2. the result is ground (or $tType)
-  3. t is let-bound: consider the following TFF1
-
-      % polymorphic constant
-      tff(c_type, type, c: !>[A: $tType]: A).
-      % some polymorphic predicate, not important
-      tff(p_type, type, p: !>[A: $tType]: A > $o).
-
-      % let-bind a polymorphic identity function
-      tff(bug, axiom, ![A: $tType]: $let(id: A > A, id(X) := X, p(A, id(c(A))))).
-      % note that type of id is A > A, *not* !>[A: $tType]: A > A.
-  */
+  // If the substitution is empty, then the result sort must be necessarily ground.
   ASS(
     !subst.isEmpty() ||
     (result.isTerm() && (result.term()->isSuper() || result.term()->ground()))


Shell/SymbolDefinitionInlining.cpp
```diff
@@ -17,33 +19,45 @@ using namespace Lib;
...
   Substitution substitution;
 
-  VList::Iterator vit(_bindingVariables);
-  while (vit.hasNext()) {
-    unsigned var = vit.next();
+  for (const auto& baseArg : iterTraits(Term::Iterator(_lhs))) {
     ASS(tit.hasNext());
-    TermList arg = tit.next();
-    substitution.bindUnbound(var, arg);
+    // TODO find out if process needs to be called on tit.next() here
```
mezpusz, line 41: The answer to this is yes, but I'm not sure this is the right time to fix it, especially because I want to replace this with `BottomUpTermTransformer`.

Another problem is that if the inlined symbol is recursive, then we could diverge.

checks/fool/let-function.p
```diff
@@ -0,0 +1,20 @@
...
+        const(B, X, Y) := X,
+        p(A, const(C, c(A), d(C)))) ).
+
+tff(conj, conjecture,
+    ! [A: $tType, C: $tType, Z : C] :
+      (f(A, C, c(A), Z) = c(A) =>
+       $let(
+         const : !>[B: $tType] : (A * B) > A,
+         const(B, X, Y) := f(A, C, X, Z),
+         p(A, const(C, c(A), d(C))))) ).
```
MichaelRawson, line 20: This whole file is cursed, I love it. Reminds me of https://github.com/rust-lang/rust/blob/main/tests/ui/expr/weird-exprs.rs

Kernel/SortHelper.cpp
MichaelRawson, line 117: Perhaps this comment could be updated?

mezpusz: Hope the new version is not too simplistic, please check.

MichaelRawson:
Yeah, sorry - I wish it gave more informative e-mail when it's cancelled rather than failed. Gotta save those cycles.

PR #784, opened 2025-11-24
Option elimination

quickbeam123:
There was one more option I managed to eliminate in August. (I will continue with the effort, but only with new data next year. ) So let's have this last fruit of summer 2025.

### DIFF
--- a/Saturation/Splitter.cpp
+++ b/Saturation/Splitter.cpp
@@ -338,7 +338,7 @@ SAT::Status SplittingBranchSelector::processDPConflicts()
       // ... moreover, _dp->addLiterals will filter the set anyway
 
       _dp->reset();
-      _dp->addLiterals(pvi( LiteralStack::ConstIterator(gndAssignment) ));
+      _dp->addLiterals(pvi( LiteralStack::ConstIterator(gndAssignment) ), false);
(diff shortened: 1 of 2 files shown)


Saturation/Splitter.cpp
```diff
@@ -103,11 +103,8 @@ void SplittingBranchSelector::init()
   }
 
   if(_parent.getOptions().splittingCongruenceClosure()) {
-    _dp = new DP::SimpleCongruenceClosure(&_parent.getOrdering());
-    if (_parent.getOptions().ccUnsatCores() == Options::CCUnsatCores::SMALL_ONES) {
-      _dp = new ShortConflictMetaDP(_dp.release(), _parent.satNaming(), *inner);
-    }
-    _ccMultipleCores = (_parent.getOptions().ccUnsatCores() != Options::CCUnsatCores::FIRST);
+    _dp = new ShortConflictMetaDP(
```
MichaelRawson, line 106: Perhaps the declaration of `_dp` can now be `std::unique_ptr<ShortConflictMetaDP>` to avoid the vtable lookup?

MichaelRawson: Or `ScopedPtr` depending on your preference.

quickbeam123: Gave it a shot!

PR #786, opened 2025-12-02
Polymorphic arrays

mezpusz:
Changed arrays to be polymorphic w.r.t. their index and inner sorts.

This allowed me to also remove the infamous `MonomorphisedInterpretation` and some related dead code.

### DIFF
--- a/Kernel/Theory.cpp
+++ b/Kernel/Theory.cpp
@@ -1203,10 +1203,16 @@ std::string Theory::getInterpretationName(Interpretation interp) {
  */
 OperatorType* Theory::getOperatorType(Interpretation i)
 {
+  // Equality is of predicate type with two arguments.
+  // Note that equality is treated specially as monomorphic,
+  // as deducing the actual arguments types from the arguments
+  // (except for two variable equalities where the type argument
+  // is stored as an extra in the Literal).
   if (i == Interpretation::EQUAL) {
     return OperatorType::getPredicateType(2);
   }
 
+  // Array operators have two type arguments, index type and element type
   if (i == Interpretation::ARRAY_SELECT || i == Interpretation::ARRAY_STORE) {
     auto indexSort = TermList::var(0);
     auto innerSort = TermList::var(1);


Kernel/Theory.cpp
```diff
@@ -1249,33 +1197,24 @@ std::string Theory::getInterpretationName(Interpretation interp) {
...
-
 /**
  * Return type of the function representing interpreted function/predicate @c i.
  */
-OperatorType* Theory::getNonpolymorphicOperatorType(Interpretation i)
+OperatorType* Theory::getOperatorType(Interpretation i)
 {
-  ASS(!isPolymorphic(i));
+  if (i == Interpretation::EQUAL) {
+    return OperatorType::getPredicateType(2);
```
MichaelRawson, line 1212: I would expect either `![A: $tType]: A * A > $o` or a family of `A * A > $o` for each A. Why does this (which I think is `$i * $i > $o`) work?

Kernel/Theory.hpp
```diff
@@ -537,9 +498,7 @@ class Theory
   static unsigned getArity(Interpretation i);
   static bool isFunction(Interpretation i);
   static bool isInequality(Interpretation i);
-  static OperatorType* getNonpolymorphicOperatorType(Interpretation i);
-
-  static OperatorType* getArrayOperatorType(TermList arraySort, Interpretation i);
+  static OperatorType* getOperatorType(Interpretation i);
```
MichaelRawson, line 501: Could this have a comment saying roughly what it does in various cases? Especially equality, array functions, array functions on Booleans, ...

MichaelRawson (approved):
Nice, but a few questions:
1. What now happens with an array of Booleans and `ARRAY_BOOL_SELECT`? Are these now pseudo-Booleans?
2. How does the interpretation of equality work? See comment inline.
3. Don't we need to change the array theory code (if any exists) to expect 5 arguments, not 3?
4. Is it worth a run on some chunk of SMT-LIB? I think arrays were already (slightly) broken, but we might now be better or worse.

mezpusz:
> 1. What now happens with an array of Booleans and ARRAY_BOOL_SELECT? Are these now pseudo-Booleans?

Yes, I think it's not worth it adding them separately, given that no one seems to be interested in array reasoning in Vampire at the moment. We can change it back when needed!

> 2. How does the interpretation of equality work? See comment inline.

This comes from [here](https://github.com/vprover/vampire/blob/master/Kernel/Signature.cpp#L263) originally, I just moved it to the common place for all interpreted stuff.

> 3. Don't we need to change the array theory code (if any exists) to expect 5 arguments, not 3?

I think I did and there were only axioms regarding arrays, but let me know if you find any other place where they are used!

> 4. Is it worth a run on some chunk of SMT-LIB? I think arrays were already (slightly) broken, but we might now be better or worse.

Probably yes, I will look the related fragments up and give it a try!

Kernel/Theory.cpp
mezpusz: I agree, but I think the equality type argument comes from the two proper term arguments or `Term::twoVarEqSort()` otherwise. The original code is from `Signature::addEquality`, I decided to put it here.

Kernel/Theory.hpp
mezpusz: Sure, I will add a comment!

Kernel/Theory.cpp
MichaelRawson: Ahah - so hopefully this `OperatorType` is never used and equality remains a special case. Got it!

mezpusz:
It was a good call to run SMTLIB..

I ran with `-sa discount -i 20000` all 38217 benchmarks that claim to be related to arrays. In particular, ALIA, AUFDTLIA, AUFDTLIRA, AUFDTNIRA, AUFLIA, AUFLIRA, AUFNIA and AUFNIRA. There were almost 20k benchmarks where both master and this branch crashed due to an unrelated bug introduced with the let-bound PR a few weeks ago. So I would first fix that so we see clearly, and rerun this without the big diff.

MichaelRawson:
Thanks for looking into it! Good that you caught this, and the plan sounds good to me.

mezpusz:
With SMTLIB there is some trouble:

master unsat: 54782 (259) sat: 258 (0)
branch unsat: 54649 (126) sat: 258 (0)

I'll look into it when I have some time.

mezpusz:
> Unnecessary FOOL reasoning is more annoying, but perhaps this is the moment to do "Boolean promotion to literals" as a preprocessing step?

I'm not sure how this would be done, but sure, we can start thinking about it. It would make sense in the higher-order context too.

MichaelRawson:
Yuck, that looks uglier than I hoped. I was expecting to do something like

```
$$select_bool(Sort, Array, Index) <=> $select(Sort, Array, Index) = $true
```

and get back to something neater, but the more I think about it the less that makes sense to me. I'm certainly OK with it how it is, though.

mezpusz:
No, we don't treat such Booleans in any special way, as far as I know.

Shell/Property.cpp
```diff
@@ -773,16 +774,19 @@ void Property::scanForInterpreted(Term* t)
...
+  if (itp == Interpretation::EQUAL) {
     return;
   }
 
-  if (Theory::isPolymorphic(itp)) {
-    OperatorType* type = t->isLiteral() ?
-        env.signature->getPredicate(t->functor())->predType() : env.signature->getFunction(t->functor())->fnType();
+  if (itp == Interpretation::ARRAY_SELECT || itp == Interpretation::ARRAY_STORE) {
+    // TODO this could be separated into store and select
+    addProp(PR_HAS_ARRAYS);
```
quickbeam123, line 784: So we are adding this newly, on top of the other occurrence of `addProp(PR_HAS_ARRAYS);` elsewhere in Property.cpp, right?

Parse/TPTP.cpp
```diff
@@ -1964,7 +1964,7 @@ void TPTP::endTheoryFunction() {
    */
 
   Theory::Interpretation itp;
-  TermList args[3]; // all theory function use up to 3 arguments as for now
```
quickbeam123, line 1967: Marton, did you test this on some our own (TPTP-extension) files with arrays?

Kernel/Theory.hpp
quickbeam123: This is probably still a TODO?

Kernel/Signature.cpp
```diff
@@ -359,12 +357,10 @@ unsigned Signature::addInterpretedPredicate(Interpretation interpretation, Opera
  *
  * If no such symbol exists, it is created.
  */
-unsigned Signature::getInterpretingSymbol(Interpretation interp, OperatorType* type)
+unsigned Signature::getInterpretingSymbol(Interpretation interp)
 {
-  Theory::MonomorphisedInterpretation mi = std::make_pair(interp,type);
```
quickbeam123, line 364: @MichaelRawson , I think this is (one of the reasons) way `OperatorType`s are/were perfectly shared!

MichaelRawson: Right! As discussed I won't go messing with them, but later re-shuffling of Signature could profit from this.

Parse/TPTP.cpp
quickbeam123: All good here...

Shell/Property.cpp
mezpusz: This is a good question, something probably triggered me to do this, but I don't remember what... I'll try to find out.

mezpusz: Looks like the reason I added the `if` is because of inside the call `Theory::getOperationSort(itp);` on line 793 we assert `hasSingleSort(itp)` and `!isPolymorhic(itp)` which are both false for `ARRAY_SELECT` and `ARRAY_STORE`. (I think earlier equality and array operations were handled somewhere else, so this is why we need to handle it.)

And I added the `addProp(PR_HAS_ARRAYS)` line because, why not? 🙂  But I can remove it if you think it's not needed.

Kernel/Theory.hpp
mezpusz: I added comments for equality and the array operators, not sure if the others need explaining (and if yes, maybe I'm not the authority to add comments to those).

Shell/Property.cpp
quickbeam123: I think it's fine to have it here (too). 

I only raise warnings when I see changes that I fail to interpret as "keeping the original semantics" ;)

PR #787, opened 2025-12-04
enable UBSan and fix all the immediate flags

MichaelRawson:
Enable [UndefinedBehaviourSanitiser](https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html) for Debug builds with Clang and GCC. Then fix the warnings it gives us until I don't see any more, explanations inline.

Also found some unused gubbins in `Int`, remove that.

@quickbeam123 - if you want in on this, please patch the Makefile (probably just append `-fsanitize=undefined` to `CXXFLAGS` and `LDFLAGS`) as you see fit.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,9 +47,6 @@ if(CCACHE_PROGRAM)
   message(STATUS "CCACHE_PROGRAM: " ${CCACHE_PROGRAM})
 endif()
 
-# wrap up common object files so not built multiple times for each target
-add_library(common OBJECT ${SOURCES})
-
 ################################################################
 # compiler configuration
 ################################################################
@@ -104,6 +101,11 @@ if(GCC_OR_CLANG)
   add_compile_options(-fno-threadsafe-statics)
   # ...or RTTI
   add_compile_options(-fno-rtti)
+
+  if(CMAKE_BUILD_TYPE STREQUAL Debug)
+    add_compile_options(-fsanitize=undefined)
+    add_link_options(-fsanitize=undefined)
+  endif()
 endif()
 
 # switch on coverage options per-compiler if desired
@@ -204,6 +206,15 @@ else ()
 endif()
 
 ################################################################
+# Vampire
+################################################################
+# wrap up common object files so not built multiple times for each target
+add_library(common OBJECT ${SOURCES})
+
+# the actual Vampire executable
+add_executable(vampire vampire.cpp $<TARGET_OBJECTS:common>)
+
+################################################################
 # UNIT TESTING
 ################################################################
 
@@ -267,8 +278,3 @@ add_executable(subsat
     SATSubsumption/subsat/subsat_main.cpp
     $<TARGET_OBJECTS:common>
 )
-
-################################################################
-# Vampire
-################################################################
-add_executable(vampire vampire.cpp $<TARGET_OBJECTS:common>)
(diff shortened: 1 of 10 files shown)


Lib/Int.hpp
```diff
@@ -87,221 +74,8 @@ class Int
   static bool stringToFloat(const char*,float& result);
   static bool stringToUnsigned64(const std::string& str,long long unsigned& result);
   static bool stringToUnsigned64(const char* str,long long unsigned& result);
-  static int sign(int i);
-  static int sign(long i);
-  static int sign(double);
-  static int max(int i,int j);
-  static int min(int i,int j);
```
MichaelRawson, line 94: `max` and `min` were still used, but as far as I can tell were completely equivalent to `std::max` or `std::min`.

Lib/SharedSet.hpp
```diff
@@ -525,7 +525,7 @@ class SharedSet {
   }
 
   auto iter() const
-  { return arrayIter(_items, size()); }
+  { return arrayIter(static_cast<const T *>(_items), size()); }
```
MichaelRawson, line 528: This is declared as `T _items[1];`, but is used as a dynamic-length array with `operator new`. The generated iterator then captures the type `T[1] &`, and of course indexes greater than `0`, which is what UBSan catches. I'm not sure about what the standard says about this, but it's a bit dodgy so let's not. Instead cast to pointer (morally `T[]`) and do what was expected.

Saturation/PredicateSplitPassiveClauseContainers.cpp
```diff
@@ -27,10 +27,6 @@ using namespace std;
 using namespace Lib;
 using namespace Kernel;
 
-int computeLCM(int a, int b) {
-  return (a*b)/Int::gcd(a, b);
```
MichaelRawson, line 31: This expression overflowed in quite a few cases! Replace with `std::lcm` which seems more well-behaved.

CMakeLists.txt
```diff
@@ -47,9 +47,6 @@ if(CCACHE_PROGRAM)
   message(STATUS "CCACHE_PROGRAM: " ${CCACHE_PROGRAM})
 endif()
 
-# wrap up common object files so not built multiple times for each target
-add_library(common OBJECT ${SOURCES})
```
MichaelRawson, line 51: It's actually important for this to be below any configuration of compiler flags, otherwise CMake doesn't apply the flags to this target.

Indexing/CodeTree.cpp
```diff
@@ -230,6 +230,13 @@ bool CodeTree::ILStruct::equalsForOpMatching(const ILStruct& o) const
   if(varCnt!=o.varCnt) {
     return false;
   }
+  // avoids passing nullptr to memcmp below,
+  // which is technically UB
+  if(!varCnt)
+    return true;
+
+  ASS(globalVarNumbers)
```
mezpusz, line 238: It would have been strange if this check didn't find anything in the code tree implementation! 🙂

Lib/Int.hpp
mezpusz: They should be!! 😟

Indexing/CodeTree.cpp
MichaelRawson: But much less than I thought given the memory shenanigans, bravo!

Lib/SharedSet.hpp
mezpusz: Uff, this is a nice catch!

MichaelRawson:
Yes, exactly - as a result of the `mini-gmp` switch in another PR ages ago.

PR #788, opened 2025-12-04
Fix most paths for HOL until saturation

mezpusz:
This PR is meant to put some of the (already resurrected) HOL structures into place such that (first-order) saturation works without crashing on most HOL problems from TPTP.

In particular:
1. cherry picked a solution for a commonly occurring parsing error that @joe-hauns already fixed on `thf-parser-fix`,
2. enabled conversion of lambdas to ones with De Bruijn indices in `FOOLElimination` such that we don't get any problems with trying to substitute something wrong on the first-order level into lambda-bound variables,
3. changed all occurrences of `ApplicativeHelper` calls to the newly implemented `HOL` calls, and removed `ApplicativeHelper`,
4. removed blocker user errors from all places where they are not reasonable anymore.

I ran regression over TPTP with Discount and Otter and yielding that
a. problems with reported errors go down from 5480 to 743,
b. of the now newly attempted 4737 problems we solve miraculously by pure FO reasoning 985 and 937 problems with Discount and Otter, respectively,
c. no problems were lost.

I can run more regression tests if needed.

### DIFF
--- a/Inferences/Choice.cpp
+++ b/Inferences/Choice.cpp
@@ -49,17 +49,18 @@ Clause* Choice::createChoiceAxiom(TermList op, TermList set)
     if (var > max) {
       max = var;
     }
-  } 
+  }
   TermList freshVar = TermList(max+1, false);
 
   TermList t1 = HC::app(setSort, set, freshVar);
   TermList t2 = HC::app(op, set);
   t2 =          HC::app(setSort, set, t2);
 
-  return Clause::fromLiterals({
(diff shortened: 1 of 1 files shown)


mezpusz:
Thanks for the quick reviews!

Here's the list of the remaining TPTP problems that throw errors:

[tptp-error-throwing-problems.txt](https://github.com/user-attachments/files/23956400/tptp-error-throwing-problems.txt)

Note that some of the errors are not related to HOL, I will upload a filtered list too.

PR #789, opened 2025-12-05
remove SafeRecursion

MichaelRawson:
`SafeRecursion` can be modelled with `BottomUpEvaluation` and is only used in one place. Replace it and remove `SafeRecursion`.

There is an intermediate commit which I used to test that the two implementations agree. Also, I found a bug in `BottomUpEvaluation` where it doesn't respect `ignoreTypeArgs` for `TermList`, fix that.

### DIFF
--- a/DP/SimpleCongruenceClosure.cpp
+++ b/DP/SimpleCongruenceClosure.cpp
@@ -257,6 +257,10 @@ struct Memo {
  */
 unsigned SimpleCongruenceClosure::convertFO(TermList trm)
 {
+  unsigned cached;
+  if(_termNames.find(trm, cached))
+    return cached;
+
   return BottomUpEvaluation<TermList, unsigned>()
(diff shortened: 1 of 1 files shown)


DP/SimpleCongruenceClosure.cpp
```diff
@@ -230,53 +229,23 @@ unsigned SimpleCongruenceClosure::getPairName(CPair p)
...
-      }
-
-      Term::Iterator argIt(t.term());
-      while(argIt.hasNext()) {
-        TermList arg = argIt.next();
-        childCallbackFn(arg);
-      }
-    }
+// memoising structure used in convertFO below
+// TODO: should this be provided as part of the BUE machinery?
```
MichaelRawson, line 233: @joe-hauns - this is just `Memo::Hash` but with a `DHMap` instead of a `Map`. Perhaps you have some idea to abstract this?

At some point the two should just be the same container, but we're a way off from there.

PR #791, opened 2025-12-06
Fix bool lets

mezpusz:
This amends #783 from the SMTLIB side with the case of Boolean let binders.

I tested with:
1. TPTP Discount/Otter where 2 problems previously almost timing out with Otter flipped to timeout,
2. SMTLIB2 with Discount on ALIA, AUFDTLIA, AUFDTLIRA, AUFDTNIRA, AUFLIA, AUFLIRA, AUFNIA and AUFNIRA, where the 20k benchmarks crashing mentioned in #786 all disappeared, and 18951 flipped to being solved and we lose 6 benchmarks for whatever reason.

Not sure if I should test on more SMTLIB benchmarks or look into the differences, but I want to eventually run longer term regressions over SMTLIB too to see if we somehow decline.

### DIFF
--- a/Kernel/Theory.cpp
+++ b/Kernel/Theory.cpp
@@ -1053,7 +1053,8 @@ unsigned Theory::Tuples::getConstructor(unsigned arity)
 
 bool Theory::Tuples::isConstructor(Term* t)
 {
-  return !t->isSpecial() && !t->isSort() && getConstructor(t->numTypeArguments()) == t->functor();
+  return !t->isSpecial() && !t->isSort() && SortHelper::getResultSort(t).isTupleSort()
+    && getConstructor(t->numTypeArguments()) == t->functor();
 }
 
 unsigned Theory::Tuples::getProjectionFunctor(unsigned arity, unsigned proj)
--- a/Shell/FOOLElimination.cpp
+++ b/Shell/FOOLElimination.cpp
@@ -620,11 +620,17 @@ void FOOLElimination::process(Term* term, Context context, TermList& termResult,
         auto vars = VList::append(bodyFreeVars, argumentVars);
         collectSorts(vars, typeVars, termVars, allVars, termVarSorts);
 
+        // this is ugly, but otherwise := would have to be special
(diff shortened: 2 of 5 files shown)


Parse/SMTLIB2.cpp
```diff
@@ -1500,7 +1500,16 @@ void SMTLIB2::parseLetEnd(LExpr* exp)
 
     auto varList = vars.list();
     auto args = TermStack::fromIterator(iterTraits(varList->iter()).map(unsignedToVarFn));
-    auto binder = Formula::createDefinition(Term::create(exprT->functor(), args), SubstHelper::apply(boundExpr,subst), varList);
+    Term* lhs;
+    if (exprSort == AtomicSort::boolSort()) {
+      // This solution is ugly, but either := has to be special or we have
+      // to wrap this as a formula to preserve the term-formula boundary.
```
MichaelRawson, line 1509: At some point we can re-consider - I think you're of the opinion that more interpreted symbols and fewer special terms are a good thing, which I agree with. But for now this is a good fix.

MichaelRawson (approved):

mezpusz:
I ran a regression from f0fb5f90467414f86fea3f52689f090a27b0da45 (before the let-bound PR) to this branch and I found another bug introduced in the let-bound PR. It was because of my seemingly permanent misunderstanding of the functions `isBoolean` and `isLiteral`.

Now it still seems a bit suspicious, because we can somehow solve ~500 more benchmarks than before the let-bound PR. ~50 are lost. I will look into it for some time.

mezpusz:
Looking at a couple of benchmarks in the difference, it seems that up to clausification the difference is trivial (for example, using `$difference(...,1)` instead of `$sum(...,$uminus(1)))` or having different order of definitions/namings). 🤷

Parse/SMTLIB2.cpp
mezpusz: Yes, I would be happy if the special term wrapper disappeared at some point, but let's see if we can get rid of it.

PR #792, opened 2025-12-09
do not use reference counting for VirtualIterator

MichaelRawson:
`VirtualIterator` manages its internal "core" by reference counting. This is weird, as you never want two copies of the same iterator pointing to the same core.

Instead, insist on unique ownership by making `VirtualIterator` non-copyable. This required a few more `std::move`s, which the compiler enforced.

### DIFF
--- a/Indexing/ClauseVariantIndex.cpp
+++ b/Indexing/ClauseVariantIndex.cpp
@@ -14,7 +14,6 @@
 
 #include "Lib/List.hpp"
 #include "Lib/Metaiterators.hpp"
-#include "Lib/SmartPtr.hpp"
 #include "Debug/TimeProfiling.hpp"
 
 #include "Kernel/Clause.hpp"
@@ -39,10 +38,6 @@ public:
   : _lits(lits), _length(length), _queryIndex(new LiteralMiniIndex(lits, length))
   {
   }
-  ~ResultClauseToVariantClauseFn()
-  {
-  }
-
 
   Clause* operator()(Clause* mcl)
   {
@@ -98,7 +93,7 @@ public:
 private:
   Literal* const * _lits;
   unsigned _length;
-  SmartPtr<LiteralMiniIndex> _queryIndex;
+  std::unique_ptr<LiteralMiniIndex> _queryIndex;
 };
 
 //-------------------//-------------------//-------------------//-------------------
--- a/Lib/Metaiterators.hpp
+++ b/Lib/Metaiterators.hpp
@@ -1417,7 +1417,7 @@ static auto __ifElseIter(Args... args)
         }
       }
   });
-  return coproductIter(out ? *out : Out::template variant<total/2>(tupleGetApplied(Constant<total - 1>{})));
+  return coproductIter(out ? std::move(*out) : Out::template variant<total/2>(tupleGetApplied(Constant<total - 1>{})));
 }
 
 template<class... Args>
@@ -1778,7 +1778,7 @@ template<class Array, class Size>
 auto arrayIter(Array      & a, Size s) { return range(0, s).map([&](auto i) -> decltype(auto) { return a[i]; }); }
 
 template<class Array, class Size>
-auto arrayIter(Array     && a, Size s) { return range(0, s).map([a = std::move(a)](auto i) { return std::move(a[i]); }); }
+auto arrayIter(Array     && a, Size s) { return range(0, s).map([a = std::move(a)](auto i) mutable { return std::move(a[i]); }); }
 
 template<class Array> auto arrayIter(Array const& a) { return arrayIter(          a , a.size()); }
 template<class Array> auto arrayIter(Array     && a) { return arrayIter(std::move(a), a.size()); }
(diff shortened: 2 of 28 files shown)


Indexing/ClauseVariantIndex.cpp
```diff
@@ -98,7 +93,7 @@ class ClauseVariantIndex::ResultClauseToVariantClauseFn
 private:
   Literal* const * _lits;
   unsigned _length;
-  SmartPtr<LiteralMiniIndex> _queryIndex;
+  std::unique_ptr<LiteralMiniIndex> _queryIndex;
```
MichaelRawson, line 96: While I was nosing around I also noticed that this `SmartPtr` is only ever used as unique ownership.

Lib/Metaiterators.hpp
```diff
@@ -1778,7 +1778,7 @@ template<class Array, class Size>
 auto arrayIter(Array      & a, Size s) { return range(0, s).map([&](auto i) -> decltype(auto) { return a[i]; }); }
 
 template<class Array, class Size>
-auto arrayIter(Array     && a, Size s) { return range(0, s).map([a = std::move(a)](auto i) { return std::move(a[i]); }); }
+auto arrayIter(Array     && a, Size s) { return range(0, s).map([a = std::move(a)](auto i) mutable { return std::move(a[i]); }); }
```
mezpusz, line 1781: What does `mutable` here refer to? I don't know what this keyword does in this context.

Lib/VirtualIterator.hpp
```diff
@@ -122,104 +110,29 @@ class EmptyIterator
  * @see IteratorCore
  */
 template<typename T>
-class VirtualIterator {
+class VirtualIterator final {
```
mezpusz, line 113: At this point, wouldn't it be easier to just typedef the class as `std::unique_ptr`? 🙂 The only problem I see is losing the nice `getEmpty` static function, but that could be `getEmptyIter<T>`?

Indexing/ClauseVariantIndex.cpp
mezpusz: So can it be deleted too?

quickbeam123:
```
Sort by SAT
976 ['problemsSTD_master10506_otter_i5K.pkl']
976 ['problemsSTD_mvi10507_otter_i5K.pkl']
994 ['problemsSTD_master10506_dis10_i5K.pkl']
994 ['problemsSTD_mvi10507_dis10_i5K.pkl']

Sort by UNS
8057 ['problemsSTD_master10506_otter_i5K.pkl']
8057 ['problemsSTD_mvi10507_otter_i5K.pkl']
8110 ['problemsSTD_master10506_dis10_i5K.pkl']
8111 ['problemsSTD_mvi10507_dis10_i5K.pkl']
```

Indexing/ClauseVariantIndex.cpp
MichaelRawson: `SmartPtr` probably yes - it's used a few other places but it's next on my list of dodgy things to get rid of.

Lib/Metaiterators.hpp
MichaelRawson: Dark magic. ;-) This one line took me an hour to figure out:
1. We want to _move_ the array into the closure, so `[a = std::move(a)]`
2. Each element then needs to be moved *out* of the array incrementally.
3. But closure captures are `const` by default unless you specify `mutable`.

Lib/VirtualIterator.hpp
MichaelRawson: Demoting to `getBoxedIterator(getEmptyIter<T>())` would make sense! Certainly this could be a very lightweight wrapper around `unique_ptr`, but the trouble is that you need it to support e.g. `it.hasNext()`, not `it.hasNext()`.

joe-hauns (approved):
Looks great! Thanks a lot for that work! I've been wanting to get rid of the ref-counting here for a while. 
Thanks for the nice tidy :)

Lib/Metaiterators.hpp
mezpusz: Uh, it's good to learn something new! So the compiler forces you to add this?

Lib/VirtualIterator.hpp
mezpusz: You mean `it->hasNext()`? Yeah, I missed that. Nevermind, looks good like this too!

Lib/Metaiterators.hpp
MichaelRawson: Yes, although it's not very helpful. It mostly complains that with `Array = Stack<VirtualIterator>` that `VirtualIterator(const VirtualIterator &)` is deleted (good, it's meant to be). This is because without `mutable` `a[i]` has type `const VirtualIterator &`, and `std::move(const VirtualIterator &)` somehow compiles but gives you back a const reference, which then blows up when you try to return `VirtualIterator` from the closure.

Lib/VirtualIterator.hpp
MichaelRawson: I do, oops. But stripping out the rest is a good plan, I'll do that. :)

MichaelRawson: OK, there's other things I'd like to do with the iterator hierarchy at the same time. Let's have another PR. :)

PR #793, opened 2025-12-10
IndexManager clean

mezpusz:
This PR adds some type safety to `IndexManager`, such that the matching the index enum values and casted-to index type does not have to be done by the caller. There is still some risk, namely that we map index types to the same values, but I hope that this will be solved by some upcoming feature of C++20/23 where there are `constexpr` type IDs and other nice things.

The other improvement is that now `IndexManager` cannot be supplied from the outside to `SaturationAlgorithm` which allows for using the actual `IndexManager` instance during unit testing instead of the hacky `setTestIndices`.

Regression is TBD, but there shouldn't be anyting drastic.

Otter TPTP:
master unsat: 10299 (0) sat: 1020 (0)
branch unsat: 10300 (1) sat: 1020 (0)

### DIFF
--- a/DP/ShortConflictMetaDP.hpp
+++ b/DP/ShortConflictMetaDP.hpp
@@ -41,7 +41,7 @@ public:
   : _inner(inner), _sat2fo(sat2fo), _solver(solver) {}
 
   void addLiterals(LiteralIterator lits, bool onlyEqualites) override {
-    _inner->addLiterals(lits, onlyEqualites);
+    _inner->addLiterals(std::move(lits), onlyEqualites);
   }
 
   void reset() override {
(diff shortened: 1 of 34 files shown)


mezpusz:
Discount TPTP:
master unsat: 10247 (0) sat: 1030 (0)
branch unsat: 10247 (0) sat: 1030 (0)

Indexing/IndexManager.cpp
```diff
@@ -35,229 +32,54 @@
...
-
-/**
- * Provide index form the outside
- *
- * There must not be index of the same type from before.
- * The provided index is never deleted by the IndexManager.
+ * IMPORTANT: Keep these values distinct from each other!
  */
-void IndexManager::provideIndex(IndexType t, Index* index)
+INDEX_ID_IMPL(AlascaIndex<ALASCA::Demodulation::Lhs>,                                   0)
```
mezpusz: If someone has a better idea to implement this, please let me know. I looked at solution like [this](https://www.reddit.com/r/cpp/comments/fijq6n/ecs_back_and_forth_part_8_type_id/) and the one in the comments but they don't seem to work for all compilers/linkers.

PR #796, opened 2025-12-19
remove SmartPtr in favour of std::unique_ptr or std::shared_ptr

MichaelRawson:
We have this `SharedPtr` thing, but it's essentially `std::shared_ptr` with extra hazard. Remove it.

Some of the occurrences were essentially `std::unique_ptr`, which is good news, we can save some cycles there. Still to do:
- ~check that `ResultSubstitutionSP` *really* needs to be shared~ they probably don't really, but it would be a huge change
- ~use `std::make_shared` [where possible](https://devblogs.microsoft.com/oldnewthing/20230815-00/?p=108602)~
- make sure nothing's gone wrong with perf

Also: remove some extraneous includes from header files and then fix up the resulting missing includes elsewhere. And the sanitiser was complaining correctly about a null dereference in the `TPTP` parser.

### DIFF
--- a/Indexing/SubstitutionTree.hpp
+++ b/Indexing/SubstitutionTree.hpp
@@ -631,8 +631,7 @@ public:
     template<class Query>
     VirtualIterator<QueryRes<ResultSubstitutionSP, LeafData>> getVariants(Query query, bool retrieveSubstitutions)
     {
-      auto renaming = retrieveSubstitutions ? std::make_unique<RenamingSubstitution>() : std::unique_ptr<RenamingSubstitution>(nullptr);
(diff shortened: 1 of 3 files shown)


MichaelRawson:
With `-t 1 -sa discount` on CNF/FOF I get 33/45 uniques, i.e. 12 problems lost, but all are regained within the time limit on re-run. Comparing Vampire on individual problems there is some difference in solution time, but I don't really see why this is: different functions take more or less time, but not in the places I touched.

If we care I can run some more tests or we can forget this. Ultimately what should probably happen is that I work out how to make `ResultSubstitutionSP` a `std::unique_ptr`.

PR #797, opened 2025-12-19
795 nullptr used in naming

mezpusz:
+ 2 other small fixes for runtime errors (1 uninitialized bool, 1 nullptr dereferenced).

### DIFF
--- a/Shell/Naming.cpp
+++ b/Shell/Naming.cpp
@@ -668,7 +668,7 @@ Formula* Naming::apply_iter(Formula* top_f) {
         Task t1;
         t1.fncTag = APPLY_LIST_TOP;
 
-        ASS(tal.negResults || tal.where != UNDER_IFF);
+        ASS_EQ(tal.negResults != nullptr, tal.where == UNDER_IFF)
         t1.taskApplyList = {tal.fs->tail(),tal.where,tal.results+1,tal.negResults ? tal.negResults+1 : nullptr};
 
         // Formula* g = apply_sub(fs->head(), where, results[0], neg);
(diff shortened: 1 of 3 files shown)


Shell/NewCNF.cpp
```diff
@@ -402,10 +402,9 @@ TermList NewCNF::findITEs(TermList ts, Stack<unsigned> &variables, Stack<Formula
 bool NewCNF::shouldInlineITE(unsigned iteCounter) {
   /*
    * MS : TODO:
-   * Since "_iteInliningThreshold = (unsigned)ceil(log2(namingThreshold))",
-   * which evaluates to 0 for 0, and since "namingThreshold == 0" means
-   * "never name anything", it would make sense to add
-   * "|| _iteInliningThreshold == 0" here.
+   * Since _iteInliningThreshold is when namingThreshold is 0,
```
quickbeam123: Did something disappear from the intended new wording of this comment?

Shell/Naming.cpp
```diff
@@ -667,7 +667,9 @@ Formula* Naming::apply_iter(Formula* top_f) {
         // FormulaList* gs = apply_list(fs->tail(), where, results + 1, negResults + 1);
         Task t1;
         t1.fncTag = APPLY_LIST_TOP;
-        t1.taskApplyList = {tal.fs->tail(),tal.where,tal.results+1,tal.negResults+1};
+
+        ASS(tal.negResults || tal.where != UNDER_IFF);
```
quickbeam123: Maybe we could try stating this even as an equivalence?

MichaelRawson: Seems OK!

Shell/NewCNF.cpp
MichaelRawson: Right, I missed a '0'. Fixed!

PR #800, opened 2026-01-16
Update macos 13 builder

mezpusz:
The builder image `macos-13` is no longer available, so we had to change to `macos-15-intel`. I also lowered the number of cores used because [we only get 4 cores](https://docs.github.com/en/actions/reference/runners/github-hosted-runners) and otherwise a too high number can cause issues.


MichaelRawson (approved):
Nice, just squashed the commits.

PR #803, opened 2026-01-19
Setup code coverage

mezpusz:
This PR adds code coverage via GCOV, which can be run locally and as a GitHub Action too.

The latest code coverage can be found here (open `index.html` after unzipping):
https://github.com/vprover/vampire/actions/runs/21145813459/artifacts/5180374070

Current coverage is:
| | |
|--------|--------|
| Lines | 35.5% |
| Functions | 53.4% |
| Branches | 20.6% |

Of course, we shouldn't put too much emphasis on this, but I think it's a good way to see where we can improve the tests.

Also, the current setup might not be the best, but we can improve it iteratively.


mezpusz:
It's a bit misleading because we are not mocking anything, so the lower level structures are tested unintentionally. Sure, I'll push a cleanup!

PR #804, opened 2026-01-20
Core inference unit tests

mezpusz:
Added unit tests to most of the testable core inferences.

Coverage change:
| | master | branch |
|--------|--------|--------|
| Lines | 35.5% | 40.0% |
| Functions | 53.4% | 55.5% |
| Branches | 20.6% | 22.9% |

Regression over FOL showed no difference.

### DIFF
--- a/Saturation/SaturationAlgorithm.cpp
+++ b/Saturation/SaturationAlgorithm.cpp
@@ -1708,8 +1708,8 @@ std::pair<CompositeISE*, CompositeISEMany> SaturationAlgorithm::createISE(Proble
 
   bool mayHaveEquality = couldEqualityArise(prb,opt);
 
-  // This was earlier at the beginning of forward simplifications,
-  // so let's put it for now to the end of immediate simplifications
+  // InnerRewriting is relatively expensive, so let's insert it first,
+  // so that it gets applied as the last ImmediateSimplification
   if (mayHaveEquality && opt.innerRewriting()) {
     res->addFront(new InnerRewriting(ordering));
   }


Inferences/InnerRewriting.hpp
```diff
@@ -25,11 +25,17 @@ namespace Inferences
 using namespace Kernel;
 using namespace Saturation;
 
+// TODO this should be an immediate simplification
```
MichaelRawson: Seems it is now?

Inferences/URResolution.hpp
```diff
@@ -51,7 +49,7 @@ class URResolution
 
   bool _full;
   bool _emptyClauseOnly;
-  bool _selectedOnly;
+  bool _selectedOnly = false;
```
quickbeam123, line 52: This means nobody's ever requesting "selectedOnly", right?

Inferences/URResolution.cpp
```diff
@@ -53,6 +49,7 @@ void URResolution<synthesis>::attach(SaturationAlgorithm* salg)
 
   Options::URResolution optSetting = _salg->getOptions().unitResultingResolution();
```
quickbeam123, line 50: Did you try thinking of passing Options::URResolution to a constructor and setting `_full`/`_emptyClauseOnly` there?; maybe it would be more congruent to how we do things elsewhere ...

Inferences/URResolution.hpp
MichaelRawson: Looks like it. Do we remember what it did/should it be revived?

quickbeam123: I think it's about selected literals somehow; and the field is passed as arguments to the internal data structures that the main class uses as a client. So I subliminally decided not to go for elimination at the moment, but feel free to explore more.

Inferences/InnerRewriting.hpp
mezpusz: You're right, removing.

Inferences/URResolution.hpp
mezpusz: I did not want to eliminate any functionality unless it really got into the way, so I left this. Maybe we can expose this, alongside other stuff, like the usual `preordered` flag that is fixed now for subsumption demodulation but we actually have an option for it.

Inferences/URResolution.cpp
mezpusz: Yes, I'm actually working on a branch where we would remove the `attach`/`detach` madness and set everything in the constructor. So I would prefer doing it in all places all at once, if you agree. :)

quickbeam123: Yes, of course. This was just a suggestion!

PR #806, opened 2026-01-23
Add ToPlaceholders TermTransformer

hetzenmat:
The `ToPlaceholders` term transformer allows to replace higher-order subterms by a special polymorphic placeholder constant.
This is a preparation for the HOL unification machinery, as we can then use a `SubstitutionTree` as a filter for the first-order term skeletons before we even start with the expensive HOL unification.

Also, I renamed the HOL unit tests in order to be able to run them all with the command `ctest -R HOL_`.

### DIFF
--- a/UnitTests/HOL/tHOL_ToPlaceholders.cpp
+++ b/UnitTests/HOL/tHOL_ToPlaceholders.cpp
@@ -20,7 +20,7 @@ TypedTermList ph(TermList sort) {
 }
 
 HOL_TEST_FUN(to_placeholders_1) {
-  auto term = app({D.h, D.f, app(x(1, D.fSrt), x(2))});
+  auto term = app(D.h, D.f, app(x(1, D.fSrt), x(2)));
   ASS_EQ(term.sort(), D.srt)
 
   ASS_EQ(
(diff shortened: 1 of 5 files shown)


UnitTests/HOL/tHOL_ToPlaceholders.cpp
```diff
@@ -0,0 +1,54 @@
...
+#include "Test/UnitTesting.hpp"
+
+using namespace Test::HOL;
+
+TypedTermList ph(TermList sort) {
+  return {HOL::create::placeholder(sort), sort};
+}
+
+HOL_TEST_FUN(to_placeholders_1) {
+  auto term = app({D.h, D.f, app(x(1, D.fSrt), x(2))});
```
mezpusz: For future reference, maybe it would be beneficial to use variadic template arguments in `app` instead of an initializer list, to get rid of the braces.

Kernel/HOL/ToPlaceholders.cpp
```diff
@@ -0,0 +1,50 @@
...
+ */
+/**
+ * @file ToPlaceholders.cpp
+ */
+
+#include "Kernel/HOL/ToPlaceholders.hpp"
+#include "Kernel/HOL/HOL.hpp"
+
+ToPlaceholders::ToPlaceholders(std::optional<Options::FunctionExtensionality> funcExtMode)
+      : TermTransformer(false),
```
mezpusz: Could you please write `/* argname= */` to indicate what argument is set to false here?

Kernel/HOL/ToPlaceholders.cpp
```diff
@@ -0,0 +1,50 @@
...
+        _topLevel(true),
+        _mode(funcExtMode.value_or(env.options->functionExtensionality())) {}
+
+TermList ToPlaceholders::replace(TermList term) {
+  TermList transformed = transformSubterm(term);
+
+  if (transformed != term)
+    return transformed;
+
+  _topLevel = false;
```
mezpusz: Since `_topLevel` and `_nextIsPrefix` are not reset to their initial values, I'm assuming the object cannot be called twice with `replace`. Do you think it would make sense to either reset the values, so that the object can be reused, or only allow a static `replace` use and discard the object, to prevent this from happening?

mezpusz: Maybe `_nextIsPrefix` is reset in `onTermExit`?

Kernel/HOL/HOL.hpp
```diff
@@ -18,13 +18,15 @@
...
+#include <optional>
+
 /**
  * This namespace contains several helper functions to deal with higher-order terms.
  */
 namespace HOL {
 
 using Kernel::Term;
-
+
```
mezpusz: Unnecessary whitespace, please remove.

Kernel/HOL/ToPlaceholders.cpp
hetzenmat: Done, I also updated the other call sites.

Kernel/HOL/ToPlaceholders.cpp
hetzenmat: I made `replace` static, since all call sites in the `hol` branch just discard the `ToPlaceholders` object after one call.

PR #807, opened 2026-01-27
Shuffle minisat

quickbeam123:
Finally found way to inject noise into the sat solver, which does not cause too much of a slowdown.

### DIFF
--- a/Minisat/core/Solver.h
+++ b/Minisat/core/Solver.h
@@ -312,9 +312,7 @@ protected:
         int len = v.size();
         for(int i=from;i<len;i++){
             int j = irand(random_seed,len-i)+i;
-            T tmp = v[i];
-            v[i] = v[j];
-            v[j] = tmp;
+            std::swap(v[i],v[j]);
         }
     }
 };


Minisat/core/Solver.h
```diff
@@ -302,6 +305,18 @@ class Solver {
...
+    template <typename T>
+    void shuffle(vec<T>& v, int from = 0)
+    {
+        int len = v.size();
+        for(int i=from;i<len;i++){
+            int j = irand(random_seed,len-i)+i;
+            T tmp = v[i];
+            v[i] = v[j];
+            v[j] = tmp;
+        }
```
MichaelRawson, line 316: Is this a reasonable shuffle? I don't have strong views, just would be unfortunate if this turned out to auto-correlate somehow.

Also could use `std::swap` if you like.

quickbeam123: I think it is:
1) It asks for n*(n-1)*(n-2)*... random ints (and thus allows for the possibility of i==j, i.e., not shuffling)
2) It's the code from Shuffling::shuffleArray; I just didn't want to introduce the dependency by including it into minisat

I will take `std::swap`, thanks! For some reason a was trying to avoid an include here as well...

MichaelRawson: OK! I think we decided `shuffleArray` was reasonable and looks a bit like a Fisher-Yates shuffle.

PR #812, opened 2026-02-25
Sorts carried around (compulsorily) with quantifiers and lambdas

quickbeam123:
- no forgetting of the sort information during preprocesssing
- we should be able to avoid calling `SortHelper::tryGetVariableSort` or `SortHelper::getVariableSort` ever again
(not that these guys scan the whole formula, looking for your variable somewhere to guess its type from its use site, thus expensive to call and also don't work for non-rectified formulas)

### DIFF
--- a/Kernel/Formula.cpp
+++ b/Kernel/Formula.cpp
@@ -417,12 +417,7 @@ Formula* Formula::createDefinition(Term* lhs, TermList rhs, VList* uVars)
     VList::Iterator vit(uVars);
     while (vit.hasNext()) {
       unsigned v = vit.next();
-      TermList s;
-      ASS(varSortMap.find(v, s)) // MS: if this doesn't hold, the code below will cope, if not perfectly
-      if (!varSortMap.find(v, s)) {
-        s = AtomicSort::defaultSort();
-      }
-      vsfifo.pushBack({v, s});
+      vsfifo.pushBack({v, varSortMap.get(v)});
     }
     res = new QuantifiedFormula(Connective::FORALL, vsfifo.list(), res);
   }
--- a/Shell/Rectify.cpp
+++ b/Shell/Rectify.cpp
@@ -87,11 +87,7 @@ FormulaUnit* Rectify::rectify (FormulaUnit* unit0, bool removeUnusedVars)
     VList::Iterator vit(vars);
     while (vit.hasNext()) {
       unsigned v = vit.next();
-      TermList s;
-      if (!varSorts.find(v, s)) {
-        s = AtomicSort::defaultSort();
-      }
-      vsfifo.pushBack({v, s});
+      vsfifo.pushBack({v, varSorts.get(v)});
     }
     unit = new FormulaUnit(new QuantifiedFormula(FORALL, vsfifo.list(), g),FormulaClauseTransformation(InferenceRule::CLOSURE,unit));
   }


FMB/FunctionRelationshipInference.cpp
```diff
@@ -225,29 +225,45 @@ void FunctionRelationshipInference::addClaimForFunction(TermList x, TermList y,
                                                TermList arg_srt, TermList ret_srt, VList* existential,
                                                ClauseList*& newClauses)
 {
-    VList* xy = VList::cons(0,VList::cons(1,VList::empty()));
+    VSList* xy = VSList::cons(std::make_pair(0, arg_srt), VSList::cons(std::make_pair(1, arg_srt), VSList::empty()));
```
mezpusz: If you agree that `{a,b}` gives better readability than `std::make_pair(a,b)`, could you please change these to the first format?

FMB/FunctionRelationshipInference.cpp
```diff
@@ -225,29 +225,45 @@ void FunctionRelationshipInference::addClaimForFunction(TermList x, TermList y,
...
 
     if(existential){
-      injective  = new QuantifiedFormula(EXISTS, existential, 0, injective);
-      surjective = new QuantifiedFormula(EXISTS, existential, 0, surjective);
-      ing_and_nons = new QuantifiedFormula(EXISTS, existential, 0, ing_and_nons);
-      sur_and_noni = new QuantifiedFormula(EXISTS, existential, 0, sur_and_noni);
+      // Build VSList from existential VList - need to determine sorts from function type
+      // For now, collect from the formulas
+      DHMap<unsigned, TermList> varSorts;
+      SortHelper::collectVariableSorts(injective, varSorts);
```
mezpusz, line 250: Is there no way to collect the variable sorts while building `injective`? Otherwise this is probably not a big deal here, either.

MichaelRawson: I think there could be, it's probably induced by the type of `f`. But this would require changing the function signature - any of you know this code well?

Kernel/Term.hpp
```diff
@@ -433,9 +433,8 @@ class alignas(8) Term
       } _formulaData;
       struct {
         TermList lambdaExp;
-        VList* _vars;
-        SList* _sorts;  
-        TermList sort; 
+        VSList* _vars;  // variables with their sorts together
+        TermList sort;
```
MichaelRawson, line 437: So if a lambda-expression has sorted variables, what is `sort`? Or even `expSort`??

Parse/TPTP.cpp
```diff
@@ -45,6 +45,18 @@ using namespace Kernel;
...
+static VSList* zipVarsSorts(VList* vars, SList* sorts) {
+  VSList::FIFO fifo;
+  VList::Iterator vit(vars);
+  SList::Iterator sit(sorts);
+  while (vit.hasNext()) {
+    ASS(sit.hasNext());
+    fifo.pushBack({vit.next(), sit.next()});
+  }
+  return fifo.list();
+}
```
MichaelRawson, line 58: This seems like a useful function! Could it be promoted to zip any two lists into a list of pairs?

Shell/Rectify.cpp
```diff
@@ -78,9 +79,21 @@ FormulaUnit* Rectify::rectify (FormulaUnit* unit0, bool removeUnusedVars)
...
-    //TODO do we know the sorts of vars?
-    unit = new FormulaUnit(new QuantifiedFormula(FORALL,vars,0,g),FormulaClauseTransformation(InferenceRule::CLOSURE,unit));
+    DHMap<unsigned, TermList> varSorts;
+    SortHelper::collectVariableSorts(g, varSorts);
+    VSList::FIFO vsfifo;
+    VList::Iterator vit(vars);
+    while (vit.hasNext()) {
+      unsigned v = vit.next();
+      TermList s;
+      if (!varSorts.find(v, s)) {
```
MichaelRawson: This pattern seems kinda suspect. So we have a list of variables free in a formula `vars`, then if a variable is not in `varSorts` it's of type `$i` - but shouldn't that be in `varSorts`?

Kernel/Formula.cpp
```diff
@@ -420,7 +411,20 @@ Formula* Formula::createDefinition(Term* lhs, TermList rhs, VList* uVars)
...
   if (uVars) {
-    res = new QuantifiedFormula(Connective::FORALL, uVars, nullptr, res);
+    DHMap<unsigned,TermList> varSortMap;
+    SortHelper::collectVariableSorts(res, varSortMap);
+    VSList::FIFO vsfifo;
+    VList::Iterator vit(uVars);
+    while (vit.hasNext()) {
+      unsigned v = vit.next();
+      TermList s;
+      ASS(varSortMap.find(v, s)) // MS: if this doesn't hold, the code below will cope, if not perfectly
```
MichaelRawson: Ahah, so this should always be the case?

FMB/FunctionRelationshipInference.cpp
quickbeam123: It's a Giles' code :)

mezpusz: Okay, no worries then! I have no idea about this part.

Kernel/Formula.cpp
mezpusz: Have you seen this happen? Maybe we could also get the variable sorts from the outside. As far as I remember, the SMTLIB2 let binding should only contain type variables, for example. The TPTP should be a bit harder.

Kernel/Term.hpp
mezpusz: The field `sort` is the sort of the entire lambda. Not sure about the other, I also wondered recently if this could be removed.

Parse/TPTP.cpp
mezpusz: There is also `zip` in `MetaIterators.hpp`, although using it with `FIFO` here would be a bit convoluted.

Shell/Rectify.cpp
```diff
@@ -171,26 +193,16 @@ Term* Rectify::rectifySpecialTerm(Term* t)
...
-    while(slit.hasNext()){
-      TermList sort = slit.next();
-      TermList rectifiedSort = rectify(sort);    
-      if(sort != rectifiedSort){
-        modified = true;
-      }
-      rectifiedSorts = SList::addLast(rectifiedSorts, rectifiedSort); // careful: quadratic complexity
+    VSList* vs;
+    {
+      ScopedLet<bool> ruv(_removeUnusedVars,false); // set _removeUnusedVars to false, temporarily
```
mezpusz, line 194: I didn't know about this structure, neat!

Shell/Rectify.cpp
```diff
@@ -222,6 +234,11 @@ Term* Rectify::rectify (Term* t)
     return t;
   }
 
+  // cannot "recreate" super below, so let's special-case it
+  if (t->isSuper()) {
```
mezpusz, line 234: Wasn't this a problem before?

Shell/Rectify.cpp
```diff
@@ -478,78 +476,75 @@ unsigned Rectify::Renaming::bind (unsigned var)
...
 {
-  if (VList::isEmpty(vs)) {
+  if (VSList::isEmpty(vs)) {
     return vs;
   }
 
-  Stack<VList*> args;
-  while (VList::isNonEmpty(vs)) {
+  Stack<VSList*> args;
+  while (VSList::isNonEmpty(vs)) {
```
mezpusz, line 508: I'm trying to understand the logic here. The variables are in reverse order of their dependencies (due to type variables), so we have to rectify them in reverse-reverse order, right?

Kernel/Term.hpp
MichaelRawson: OK, that makes sense. But why do we have this? Seems like this would be deducible. I guess it might be expensive or fiddly in some cases.

Kernel/Formula.cpp
MichaelRawson: I've no idea one way or the other. Martin has a few occurrences of this pattern in this PR - see my immediately-preceding comment. But there I thought it looked wrong because we should always have `varSortMap.find(v, s)` - and here we have an assertion which suggests that it should always be the case.

mezpusz: I agree, we should enforce this invariant, if not yet enforced.

Kernel/Term.hpp
mezpusz: It's mostly for the case when we have just a variable inside. Maaaybe it is deducible, but I'm not sure about all corner cases. Nested special terms can be nasty.

PR #813, opened 2026-02-25
actually find SAT duplicate literals and tautologies

MichaelRawson:
Thanks to @RobCoutel for noticing this.

Somehow we have code for detecting propositional tautologies and duplicate literals in SAT clauses, but it doesn't work because the SAT clauses are not sorted correctly. Now, I did touch this recently so I could well have broken it, but it doesn't look like it ever worked. :shrug: 

Fix this.

### DIFF
--- a/UnitTests/tSATClause.cpp
+++ b/UnitTests/tSATClause.cpp
@@ -8,37 +8,42 @@
  * and in the source directory
  */
 #include "Debug/Assertion.hpp"
-#include "Lib/Stack.hpp"
 #include "SAT/SATClause.hpp"
 #include "SAT/SATLiteral.hpp"
 #include "Test/UnitTesting.hpp"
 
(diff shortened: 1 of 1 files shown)


mezpusz:
Thanks for the unit tests! I added some sugar on top, and cleaned them up a bit.

I think there is no need to add more, this is just enough to catch if something goes wrong again with this part.

mezpusz:
Btw, shouldn't `fromStack` already call `sort`? I don't see any place where this does not happen immediately.

mezpusz:
I don't necessarily mean non-tautological, just sorted. But if that can break things, never mind!

mezpusz:
@quickbeam123 is right that we can have situations where sorting is not (yet) needed, but we spend lots of time with it. So I'm completely okay if you ignore what I suggested.

PR #819, opened 2026-03-11
Remove attach detach

mezpusz:
Inferences should be tied to a single saturation algorithm during their existence, and to all properties of the saturation algorithm, such as options, ordering, etc. This simplifies reasoning, for example indices can be passed as references instead of shared pointers (not yet implemented), and hopefully opening up further simplifications down the line.

So far I have run FOL regression where only one problem was lost (MED011+1 or MED012+1, I have to check), due to somehow spending 1.5 seconds instead of 1.4 seconds of runtime.

Will run more tests, especially for ALASCA, which had to be changed the most.

### DIFF
--- a/Inferences/BackwardSubsumptionAndResolution.cpp
+++ b/Inferences/BackwardSubsumptionAndResolution.cpp
@@ -79,7 +79,7 @@ void BackwardSubsumptionAndResolution::perform(Clause *cl,
       auto it = _bwIndex->getInstances(lit, false, false);
       while (it.hasNext()) {
         Clause *icl = it.next().data->clause;
-        if (!_checked.insert(icl))
+        if (!_checked.insert(icl->number()))
           continue;
         env.statistics->backwardSubsumed++;
         List<BwSimplificationRecord>::push(BwSimplificationRecord(icl), simplificationBuffer);
@@ -93,7 +93,7 @@ void BackwardSubsumptionAndResolution::perform(Clause *cl,
       while (it.hasNext()) {
         auto res = it.next();
         Clause *icl = res.data->clause;
-        if (!_checked.insert(icl))
+        if (!_checked.insert(icl->number()))
           continue;
         Clause *conclusion = SATSubsumption::SATSubsumptionAndResolution::getSubsumptionResolutionConclusion(icl, res.data->literal, cl, /*forward=*/false);
         ASS(conclusion)
@@ -138,7 +138,7 @@ void BackwardSubsumptionAndResolution::perform(Clause *cl,
     auto it = _bwIndex->getInstances(lit, false, false);
     while (it.hasNext()) {
       Clause *icl = it.next().data->clause;
-      if (!_checked.insert(icl))
+      if (!_checked.insert(icl->number()))
         continue;
       // check subsumption and setup subsumption resolution at the same time
       bool checkS = _subsumption && !_subsumptionByUnitsOnly;
@@ -170,7 +170,7 @@ void BackwardSubsumptionAndResolution::perform(Clause *cl,
     auto it = _bwIndex->getInstances(lit, true, false);
     while (it.hasNext()) {
       Clause *icl = it.next().data->clause;
-      if (!_checked.insert(icl))
+      if (!_checked.insert(icl->number()))
         continue;
       // check subsumption resolution
       Clause *conclusion = _satSubs.checkSubsumptionResolution(cl, icl, /*forward=*/false, false);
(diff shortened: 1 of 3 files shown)


Inferences/GlobalSubsumption.hpp
```diff
@@ -32,13 +31,12 @@ using namespace Indexing;
 using namespace Saturation;
 using namespace SAT;
 
+// TODO consider making this an ImmediateSimplificationEngine if it really doesn't depend on any indices
```
MichaelRawson, line 34: I think this is another victim of performance considerations: global subsumption is pretty expensive (at least one SAT call) so I guess that's why it isn't one. Not to say it shouldn't be, with some more attention...

Inferences/InferenceEngine.hpp
```diff
@@ -110,9 +71,9 @@ class GeneratingInferenceEngine
 };
 
 class ImmediateSimplificationEngine
-: public InferenceEngine
 {
 public:
+  virtual ~ImmediateSimplificationEngine() = default;
```
MichaelRawson, line 76: I'd expect this to be `= 0` on the basis that `ImmediateSimplificationEngine` should be an abstract class.

Saturation/SaturationAlgorithm.cpp
```diff
@@ -288,32 +296,17 @@ SaturationAlgorithm::~SaturationAlgorithm()
...
   _passive->detach();
 
-  if (_generator) {
-    _generator->detach();
-  }
-  if (_immediateSimplifier) {
-    _immediateSimplifier->detach();
-  }
-
   while (_fwSimplifiers) {
```
MichaelRawson: These loops feel a lot like `List::destroyWithDeletion`. Even better, what about changing these to e.g. `std::vector<std::unique_ptr<SimplificationEngine>>`?

Inferences/GlobalSubsumption.hpp
mezpusz: Good point, so maybe we could eventually mark it an "expensive" immediate simplifier (if we get to implementing those). Would you mind if I left the TODO here regardless?

Inferences/InferenceEngine.hpp
mezpusz: Hm, a pure virtual destructor is an interesting concept that I have never considered before. Upon trying to implement this, I got some linker errors that the base class destructor symbol does not exist, even if I override it in the derived classes. Do you know what is expected here?

mezpusz: According to [this](https://stackoverflow.com/questions/630950/pure-virtual-destructor-in-c), we still have to define the destructor if the class is ever used, but then I don't see the point of making it pure virtual (if a class is not used, then delete it? 😄 ). Am I missing something?

MichaelRawson: Oh, yeah - the destructor will always be called from derived classes. Oops, ignore this!

Inferences/GlobalSubsumption.hpp
MichaelRawson: Not at all, just a comment in passing.

Saturation/SaturationAlgorithm.cpp
mezpusz: The deletion I replaced with the list static function, I'm not yet sure about vectors, as they don't support pushing to front. Do you have some other standard class in mind? Maybe `std::deque` or `Stack` with `iterFifo()`?

MichaelRawson: Another container could work. But I was thinking that the convention would simply switch - lists are O(1) to prepend, arrays are O(1) to append, so `addToFront` would actually append and then we'd iterate in reverse order. Up to you - anyway, let's not hold this up for this kind of detail, we can always come back to it.

mezpusz: I think we should do it, but in another PR. Maybe in the one where we remove lists altogether? 🙂

Saturation/SaturationAlgorithm.hpp
```diff
@@ -74,16 +80,11 @@ class SaturationAlgorithm : public MainLoop
...
-
   void setLabelFinder(LabelFinder* finder){ _labelFinder = finder; }
 
-  void addForwardSimplifierToFront(ForwardSimplificationEngine* fwSimplifier);
-  void addExpensiveForwardSimplifierToFront(ForwardSimplificationEngine* fwSimplifier);
-  void addSimplifierToFront(SimplificationEngine* simplifier);
-  void addBackwardSimplifierToFront(BackwardSimplificationEngine* bwSimplifier);
+  template<typename Inference> void addForwardSimplifierToFront();
+  template<typename Inference> void addExpensiveForwardSimplifierToFront();
+  template<typename Inference> void addBackwardSimplifierToFront();
```
quickbeam123: My guess is that these won't work if called from the outside (as the are defined in the cpp file). Shouldn't they be `private` or defined here in the header?

mezpusz: Changed them to private for now, because no one uses them outside.

PR #820, opened 2026-03-11
Update python scripts & small bugfixes

EpsilonPhoenix:


MichaelRawson:
Hi @EpsilonPhoenix - thanks for the contribution. Will merge when CI finishes. I wouldn't spend more time in `scripts/` though, I'm not sure that anyone uses them any more. Will go through them at some point and get rid of the unused stuff. Anyway, nice to have a new contributor, welcome!

MichaelRawson:
Ah, this is something else which broke. Waiting for #818...

PR #827, opened 2026-03-18
Optimize SAT assumption handling and Saturation queue

EpsilonPhoenix:


Indexing/ClauseCodeTree.cpp
```diff
@@ -97,7 +98,17 @@ void ClauseCodeTree::optimizeLiteralOrder(DArray<Literal*>& lits)
...
+  literalCodes.reserve(clen);
+  for (unsigned i = 0; i < clen; i++) {
+    CodeStack code;
+    LitCompiler compiler(code);
+    compiler.handleTerm(lits[i]);
+    literalCodes.push_back(std::move(code));
+  }
+
   CodeOp* entry=getEntryPoint();
+  bool stop=false;
```
mezpusz: What is the purpose of `stop`? As far as I see it's not used for anything.

Indexing/ClauseCodeTree.cpp
```diff
@@ -97,7 +98,17 @@ void ClauseCodeTree::optimizeLiteralOrder(DArray<Literal*>& lits)
 
   lits.sort(InitialLiteralOrderingComparator());
 
+  std::vector<CodeStack> literalCodes;
```
mezpusz: The original code could exit early without going over the clause and creating code blocks for all literals (see [here](https://github.com/vprover/vampire/pull/827/changes#diff-73d2665b27287b0a2740c26f74ea1ba7989372ddea5391faf0e824e9dda8bbf6L135)) but now all literals are converted to code blocks unconditionally, and an additional stack is created. Have you measured the performance impact of this change?

Indexing/ClauseCodeTree.cpp
EpsilonPhoenix: Leftover code from testing, I'll remove next commit

Indexing/ClauseCodeTree.cpp
```diff
@@ -97,6 +98,26 @@ void ClauseCodeTree::optimizeLiteralOrder(DArray<Literal*>& lits)
 
   lits.sort(InitialLiteralOrderingComparator());
 
+  std::vector<CodeStack> literalCodes(clen);
```
mezpusz: If the point is to reuse the `CodeStack` object in `evalSharing` as you described in your other comment, I would advise just changing `CodeStack code` to `Recycled<CodeStack> code` in the original code in `evalSharing`. The current changed version not only makes the code harder to read but it also uses a vector of `CodeStack` which might be even more detrimental for performance than not reusing the single `CodeStack` object.

EpsilonPhoenix: Ah yeah, I guess so.

I've rewritten it to remove the `vector<CodeStack>`. It now keeps a reusable scratch `CodeStack` in `optimizeLiteralOrder` and passes it to `evalSharing`, so it avoids repeated work without adding the extra caching.

Kernel/ClauseQueue.cpp
```diff
@@ -50,7 +50,7 @@ void ClauseQueue::insert(Clause* c)
 {
   // select a random height between 0 and top
   unsigned h = 0;
-  while (Random::getBit()) {
+  while (h < MAX_HEIGHT && Random::getBit()) {
```
quickbeam123: Could we even do `while (h <= _height && ...` here?

EpsilonPhoenix: I don’t think so. _height is the current height of the structure, while here we’re sampling the height of the new node. 

We need to allow that sample to go beyond the current _height; otherwise a new node can only grow by one extra level.

quickbeam123: What I was alluding to is that `h` seems to be being "clipped" to at most `_height` in the condition just below.

EpsilonPhoenix: Good point, you’re right that the later block clips sample h anyway. I updated it to use `while (h <= _height && h < MAX_HEIGHT && Random::getBit())` so it only samples up to the next possible level directly.

PR #836, opened 2026-04-09
HOL improvements (continued)

mezpusz:
This round of improving/merging HOL stuff includes:
1. Updating, cleaning up and adding unit tests to CNF-on-the-fly inferences, `Cases`, `CasesSimp`, `BoolSimp`.
2. Adding options to enable various HOL inferences, and adding back HOL options to the HOL sampler file (rest of the file is untouched, maybe @quickbeam123 can give some direction on whether we should try to merge the entire sampler, or start anew).
3. Adding `ImitateProject` which should perform some HOL unification. A note on this below.
4. Fixing definition introduction.

Some notes:
a. The diff seems to be really big, mostly because I moved *and* refactored some files, I can temporarily move them back to have a better diff if needed.
b. `ImitateProject` only works on flex-rigid negative equations, but for unification to work in general, we would need some kind of abstraction implemented for HOL too. Maybe it won't be needed at all, but I would add it for now, so it's easier to dig it up if we decide to remove it.
c. Some options require adding HOL axioms (functional extensionality, proxy, choice), but according to Ahmed's comments, not all of these axioms should be merged. For example, proxy axioms were not really helpful, so they shouldn't be merged in the first place (@hetzenmat could you comment on this?).

Performance:
FOL regression over TPTP mostly unaffected, one problem lost due to noise with Otter (`GRP/GRP682-11.p`).

HOL regression over TPTP:

Discount:

master unsat: 2102 (14) sat: 1 (0) cputime: 8779.49 s instructions: 43662656 Mi memory: 111225.78 MB
branch unsat: 2113 (25) sat: 1 (0) cputime: 8790.59 s instructions: 43898503 Mi memory: 114006.14 MB

Otter:

master unsat: 2080 (26) sat: 1 (0) cputime: 9241.68 s instructions: 44366574 Mi memory: 133447.29 MB
branch unsat: 2073 (19) sat: 1 (0) cputime: 9164.78 s instructions: 44823860 Mi memory: 136637.08 MB

Not sure if we should care about the above results too much, as some features have changed or are now behind flags.

Debug sampling showed some new bugs surfacing, mostly because new options have been added to the sampler, I will hunt these down eventually.


Inferences/DefinitionIntroduction.cpp
```diff
@@ -186,9 +190,12 @@ void DefinitionIntroduction::process(Clause *cl) {
 
   // process all the non-trivial terms in the clause
   for(Literal *l : cl->iterLits())
-    for(Term *t : iterTraits(NonVariableNonTypeIterator(l)))
+    for(Term *t : iterTraits(RewritableSubtermIterator<higherOrder>(l)))
```
mezpusz, line 193: I don't like the templated solution too much, but it was needed due to using two different iterator types. Alternatively, I can imagine something like the following to also work:
```
std::unique_ptr<IteratorCore> it(_higherOrder ? new FirstOrderSubtermIterator(l) : new NonVariableNonTypeIterator(l));
```

Kernel/Problem.hpp
```diff
@@ -151,7 +151,6 @@ class Problem {
   bool hasNumerals() const; // meaning the interpreted constants of arithmetic theories, e.g. 1,2, 3.1415,...
   /** Problem contains let terms or formulas, or term if-then-else */
   bool hasFOOL() const;
-  bool hasCombs() const;
```
MichaelRawson, line 154: Nice. Have you seen any other remaining combinatorial stuff lying around?

Kernel/HOL/HOL.hpp
```diff
@@ -93,7 +105,9 @@ namespace HOL::create {
   TermList bottom();
   TermList conj();
   TermList disj();
-  TermList imp();  
+  TermList imp();
+  TermList iff();
+  TermList xorP();
```
MichaelRawson, line 110: `iff` but `xorP`?

Kernel/HOL/HOL.hpp
```diff
@@ -65,6 +65,18 @@ inline bool canHeadReduce(const TermList& head, const TermStack& args) {
   return head.isLambdaTerm() && args.isNonEmpty();
 }
 
+// if flexTerm is of form X t1 t2 : i > i and t1 : int and t2 : tau
+// this function will fill stack with [i, tau, int]
```
MichaelRawson, line 69: This seems like odd behaviour - although I see the TODO below. Does `X t1 t2` need to be exactly `i > i`?

Inferences/DefinitionIntroduction.cpp
MichaelRawson: This seems OK! If it gets painful at some point we could consider using one class that does the runtime branch.

Inferences/HOL/CasesSimp.cpp
```diff
@@ -0,0 +1,80 @@
...
+      return lhs != arg.second && rhs != arg.second;
+    })
+    .flatMap([premise](pair<Literal*, TermList> arg) {
+      return pvi(iterItems(
+        performCaseSimp(premise, arg.first, arg.second, true),
+        performCaseSimp(premise, arg.first, arg.second, false)
+      ));
+    });
+
+  if (it.hasNext()) {
```
MichaelRawson, line 73: I realise you probably didn't write this, but this logic with the iterator/inference result looks a bit wonky - definitely what is intended?

Kernel/Problem.hpp
mezpusz: There is some in `TheoryFinder` but it seems orthogonal to combinatory superposition. Is that class used actively for anything?

Inferences/HOL/CNFOnTheFly.cpp
```diff
@@ -0,0 +1,391 @@
...
+
+#include "CNFOnTheFly.hpp"
+
+namespace Inferences {
+using namespace std;
+using namespace Indexing;
+
+static Clause* replaceLits(Clause* c, Literal* what, Proxy p, bool incAge, Literal* by1, Literal* by2 = nullptr);
+static TermList sigmaRemoval(TermList sigmaTerm, TermList expsrt);
+static TermList piRemoval(TermList piTerm, Clause* clause, TermList expsrt);
```
MichaelRawson, line 39: I've never quite understood what sigma and pi are. Could this have a little comment explaining if you now understand?

MichaelRawson (approved):
Cool! General question: some of these look very similar to FOOL inferences, by design. Any chance they could really be the same in future?

Kernel/Problem.hpp
MichaelRawson: I was told at some point that `TheoryFinder` looks for axiomatisations of e.g. arithmetic. Which I guess is cute/terrifying when it works. :-) No idea if it's used, I'll put it on my list of classes to look suspiciously at.

MichaelRawson: Ahah, but it only fills out some fields in `Property` - I thought it would then transform the problem to e.g. a problem with arithmetic. So yes, probably completely orthogonal.

quickbeam123: Indeed, it's used (and working for what it is), but basically there only to influence strategy dispatch...

Kernel/HOL/HOL.hpp
mezpusz: There is unfortunately such a [binary operator](https://en.cppreference.com/w/cpp/keyword/xor.html) in C++.

Inferences/HOL/CNFOnTheFly.cpp
mezpusz: They are coming from the dependently typed world, and I wouldn't attempt to explain them. How about just referring to [this](https://en.wikipedia.org/wiki/Dependent_type#%CE%A0_type) for pi and to [this](https://en.wikipedia.org/wiki/Dependent_type#%CE%A3_type) for sigma?

Inferences/HOL/CasesSimp.cpp
mezpusz: The `*SimplificationMany` logic is that *if the clause is redundant*, then it is replaced by this (possibly empty) set of clauses. So we have to distinguish between not redundant, and being replaced by the empty set...

Kernel/HOL/HOL.hpp
mezpusz: I have to summon @hetzenmat for this to check what the intention was, I just took these functions from his branch.

mezpusz:
> Cool! General question: some of these look very similar to FOOL inferences, by design. Any chance they could really be the same in future?

Good question, I was wondering the same myself. The new ones work under the HOL assumptions (only equalities, everything is applicative), but they should in principle be mergable with the FO(O)L ones if we find some common generalization. This does not seem to be doable currently, sadly.

On the upside, I think the inferences are really simple so they shouldn't be in the way, despite being duplicated.

Inferences/HOL/CasesSimp.cpp
MichaelRawson: Understood - but it seems weird that the iterator being empty signals that the clause is not redundant. But if you're happy, I'm happy.

Inferences/HOL/CNFOnTheFly.cpp
MichaelRawson: I was confused because I know _those_ sigma and pi, but we don't have dependent types. But then I found this in the [TPTP technical report](https://tptp.org/UserDocs/ProblemLibraryManual/TPTPTR.shtml#FormulaeSection):

> TH1 has five polymorphic constants: `!!` for Π (universal quantification), `??` for Σ (existential quantification), `@@+` for ε (indefinite description, aka choice), `@@-` for ι (definite description), and `@=` (equality). The type of the first four is `!>[A: $tType] : (A > $o) > $o` [...]

So I guess these are the logical analogue - interpreted constants with a particular type.

mezpusz: You're right, I realized now the confusion, I wasn't thinking about this aspect. As far as I see, we only use them for universal/existential quantification, not sure what happens in the TPTP benchmarks that use them. 🤷

Inferences/HOL/CasesSimp.cpp
mezpusz: Okay, yeah, this is indeed weird a bit, but `CasesSimp` can never delete a clause, so the empty case can only be the non-redundant one. Otherwise, I think this inference could be simply an `ImmediateSimplification`, as the two cases are equivalent (as far as I can tell). I will have to run some experiments to support this when we have fully functional HOL reasoning in master.

quickbeam123 (approved):

Kernel/HOL/HOL.hpp
hetzenmat: I admit that I also just took these functions from the hol branch.
The comment here just gives an example of the behaviour of the function.
It would be fine for me to add a TODO with my name to this comment (as well as all other non self-explanatory function from `HOl.hpp`) to add a proper explanation of what is going on.

mezpusz: Thanks for the explanation, I added now `TODO(HOL)` to mark this as a remaining issue from the HOL branch (I don't want to impose this task on you necessarily 🙂 ).

Inferences/HOL/CNFOnTheFly.cpp
mezpusz: Added comment with the two pointers.

PR #838, opened 2026-04-17
Add Leibniz equality elimination

mezpusz:
Relatively straightforward merge from the HOL branch.

FOL regression showed no change and HOL debug sampling showed no new bugs appearing.

HOL regression shows a bit of improvement:

Discount
run 0 unsat: 2113 (29) sat: 1 (0) cputime: 8818.80 s instructions: 43898569 Mi memory: 113981.23 MB
run 1 unsat: 2141 (57) sat: 1 (0) cputime: 8801.18 s instructions: 43776217 Mi memory: 111300.63 MB

Otter
run 0 unsat: 2073 (20) sat: 1 (0) cputime: 9264.58 s instructions: 44823890 Mi memory: 136645.76 MB
run 1 unsat: 2123 (70) sat: 1 (0) cputime: 9204.34 s instructions: 44402713 Mi memory: 133133.88 MB


MichaelRawson (approved):
Looks fine to me, keep them coming. :) Could this have a quick comment somewhere saying what this inference does? It took me a while, even knowing what Leibniz equality is.

quickbeam123 (approved):

mezpusz:
> Looks fine to me, keep them coming. :) Could this have a quick comment somewhere saying what this inference does? It took me a while, even knowing what Leibniz equality is.

Excellent point. I've added what the rule does syntactically, but I also have some problem understanding exactly why it does it. As far as I see we could "apply" the clausified version of Leibniz's rule without instantiating the variable in the clause. Probably @ibnyusuf knows?

MichaelRawson:
Yeah, me either - it's clearly sound but I have no idea why we do this in particular. Nevertheless, let's go.

ibnyusuf:
It has been sometime, but IIRC, I lifted this inference directly from Leo-III. I don't think there is anything deep about it. The inference is checking for some syntactic patterns that suggest instantiating a variable with equality or negative equality could be helpful. It is in no way a complete method of dealing with equality in higher-order logic.

image

PR #839, opened 2026-04-17
Improve code tree printing

mezpusz:
Mostly affecting debug printing for code trees:
- cleaned up what we print
- prints predicates correctly when printing a clause code tree
- type constructors are not printed properly (this is not really possible without adding `CHECK_TYPE_CON` as a new operation, which would slow our executation down), but at least it does not crash anymore

The main change that can affect performance is that code trees are not virtual, but the only virtual functions are called during deletion/debug printing, so effectively never in release mode.

FOL regression shows one problem lost (probably due to noise).


Indexing/CodeTree.hpp
```diff
@@ -41,21 +40,21 @@ class CodeTree
...
-  void (*_onCodeOpDestroying)(CodeOp* op);
-      
+  virtual void onCodeOpDestroying(CodeOp* op) {}
+  virtual void printSuccess(std::ostream& out, const CodeOp& op) const {}
+
 public:
-  CodeTree();
-  ~CodeTree();
-  
+  virtual ~CodeTree();
```
MichaelRawson, line 56: I think the question is whether there are likely to be a lot of `CodeTree` objects (and therefore whether the vtable makes a difference - experimentally it doesn't). I think not, right? There is only a lot of child objects.

MichaelRawson (approved):

Indexing/CodeTree.hpp
mezpusz: Right, I think each index instance should have at most one code tree instance, so if we don't instantiate indices like crazy, this shouldn't be a problem. Anyways, if we have too many indices, then maintenance becomes the bottleneck and not handling vtables, right?

quickbeam123 (approved):

PR #840, opened 2026-04-18
no synchronisation mechanism needed for portfolio workers

### LINKED ISSUE #548: don't synchronise proofs if we don't print them
Vampire uses some mechanism (currently semaphores, might be files in future - #540) to synchronise proof output. We don't need to bother with `-p off`, but we don't special-case this yet.

MichaelRawson:
At some point portfolio workers were synchronised with UNIX semaphores to prevent them trampling over a valid proof. 

In #540 I changed this mechanism to a more portable version based on files, but we can do better: just don't synchronise at all, have each child process write to their own temporary file, and have the parent process print the proof.

Closes #548.

### DIFF
--- a/CASC/PortfolioMode.cpp
+++ b/CASC/PortfolioMode.cpp
@@ -63,7 +63,7 @@ PortfolioMode::PortfolioMode(Problem* problem) : _prb(problem), _slowness(env.op
   unsigned cores = std::max(1u, std::thread::hardware_concurrency());
   _numWorkers = std::min(cores, env.options->multicore());
   if(!_numWorkers) {
-    // TODO this has been around since the ark, is it still worth the weirdness?
+    // only kicks in if the user explicitly sets "--cores 0" which cries "give me all" and it gives you fewer than "all" if you have plenty (i.e. on a server).
     _numWorkers = cores >= 8 ? cores - 2 : cores;
   }
 }
--- a/CASC/PortfolioMode.cpp
+++ b/CASC/PortfolioMode.cpp
@@ -37,7 +37,6 @@
 #include <unistd.h>
 #include <signal.h>
 #include <fstream>
-#include <cstdio>
 #include <random>
 #include <filesystem>
 //only for detecting number of cores, no threading here!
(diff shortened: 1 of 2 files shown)


CASC/PortfolioMode.cpp
```diff
@@ -57,38 +56,16 @@ using Lib::Sys::Multiprocessing;
...
 
 PortfolioMode::PortfolioMode(Problem* problem) : _prb(problem), _slowness(env.options->slowness()) {
-  unsigned cores = std::thread::hardware_concurrency();
-  cores = cores < 1 ? 1 : cores;
+  unsigned cores = std::max(1u, std::thread::hardware_concurrency());
   _numWorkers = std::min(cores, env.options->multicore());
-  if(!_numWorkers)
-  {
+  if(!_numWorkers) {
+    // TODO this has been around since the ark, is it still worth the weirdness?
```
mezpusz, line 66: 10/10 todo. 😄

CASC/PortfolioMode.cpp
```diff
@@ -401,19 +378,24 @@ void PortfolioMode::getSchedules(const Property& prop, Schedule& quick, Schedule
   }
 }
 
-bool PortfolioMode::runSchedule(Schedule schedule) {
+static fs::path proofPath(pid_t parent, pid_t child) {
+  std::string file = "vampire-proof-" +
+    Int::toString(parent) + "-" +
+    Int::toString(child);
+  return fs::temp_directory_path() / file;
```
mezpusz, line 385: Cool operator, my brain expected division at first.

mezpusz (approved):
Seems like a nice simplification.

CASC/PortfolioMode.cpp
MichaelRawson: Right? C++ folks getting with the times.

CASC/PortfolioMode.cpp
quickbeam123: This is not as weird as it sounds. 

It only kicks in if the user explicitly sets `--cores 0` which says "give me all" and it gives you less than "all" if you have plenty (i.e. on a server).

Still, open to discussion. At the very least, I can add a comment explaining the semantics...

MichaelRawson: Ah, OK - and this is desirable on a server, I guess. Comment would be great if you're already looking! Otherwise I can do it.

CASC/PortfolioMode.cpp
```diff
@@ -495,34 +469,43 @@ bool PortfolioMode::runScheduleAndRecoverProof(Schedule schedule)
...
-    //we don't delete the proof file
-    if(openSucceeded){
-      fs::remove(_path);
+      return false;
     }
   }
-
-  return result;
+  // done with the temporary file now
+  fs::remove(path);
```
MichaelRawson: @quickbeam123 this was what I had in mind for removing at least the proof that was printed. Works on my machine. ;-) Let me investigate a bit...

PR #841, opened 2026-04-20
Subsumption equality resolution

mezpusz:
Adds the new inference subsumption equality resolution that is similar in spirit to subsumption resolution and is based on equality resolution rather than resolution. The inference is behind a (default false) flag so it shouldn't cause much trouble now.

Some figures over FOL benchmarks when it's enabled:

Discount:
run 0 unsat: 9208 (5) sat: 1042 (0) cputime: 56575.07 s instructions: 222740733 Mi memory: 1461763.85 MB
run 1 unsat: 9215 (12) sat: 1042 (0) cputime: 57393.50 s instructions: 222608978 Mi memory: 1459148.81 MB

Otter:
run 0 unsat: 9322 (7) sat: 1025 (0) cputime: 61514.09 s instructions: 222752334 Mi memory: 1694391.73 MB
run 1 unsat: 9327 (12) sat: 1025 (0) cputime: 62363.23 s instructions: 222637658 Mi memory: 1691635.40 MB

### DIFF
--- a/Inferences/SubsumptionEqualityResolution.cpp
+++ b/Inferences/SubsumptionEqualityResolution.cpp
@@ -1,0 +1,68 @@
+/*
+ * This file is part of the source code of the software program
+ * Vampire. It is protected by applicable
+ * copyright laws.
+ *
+ * This source code is distributed under the licence found here
+ * https://vprover.github.io/license.html
+ * and in the source directory
+ */
+/**
+ * @file SubsumptionEqualityResolution.cpp
+ * Implements class SubsumptionEqualityResolution.
+ */
+
+#include "Inferences/DemodulationHelper.hpp"
+#include "Lib/Stack.hpp"
+
+#include "Kernel/Clause.hpp"
+#include "Kernel/Inference.hpp"
+#include "Kernel/RobSubstitution.hpp"
+#include "Kernel/SubstHelper.hpp"
+
+#include "SubsumptionEqualityResolution.hpp"
+
+namespace Inferences
+{
+
+Clause* SubsumptionEqualityResolution::simplify(Clause* cl)
+{
+  for (const auto& lit : *cl) {
+    if (!lit->isEquality() || lit->isPositive()) {
+      continue;
+    }
+
+    auto [lhs, rhs] = lit->eqArgs();
+
+    struct Unifier : SubstApplicator {
+      bool unify(TermList lhs, TermList rhs) {
+        return subst.unify(lhs, 0, rhs, 0);
+      }
+      TermList operator()(unsigned v) const override {
+        return subst.apply(TermList::var(v), 0);
+      }
+      RobSubstitution subst;
+    } unifier;
+
+    if (!unifier.unify(lhs, rhs)) {
+      continue;
+    }
+
+    RStack<Literal*> resLits;
+    for (const auto& curr : *cl) {
+      if (lit == curr) {
+        continue;
+      }
+      if (!DemodulationHelper::isRenamingOn(&unifier, curr)) {
+        goto fail;
+      }
+      resLits->push(curr);
+    }
+    return Clause::fromStack(*resLits, SimplifyingInference1(InferenceRule::SUBSUMPTION_EQUALITY_RESOLUTION, cl));
+fail:
+    continue;
+  }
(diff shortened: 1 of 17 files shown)


cmake/sources.cmake
```diff
@@ -345,6 +336,8 @@ set(SOURCES
     Inferences/PushUnaryMinus.hpp
     Inferences/SubsumptionDemodulationHelper.cpp
     Inferences/SubsumptionDemodulationHelper.hpp
+    Inferences/SubsumptionEqualityResolution.cpp
+    Inferences/SubsumptionEqualityResolution.hpp
```
MichaelRawson, line 340: Probably need something in the Makefile too?

Inferences/SubsumptionEqualityResolution.cpp
```diff
@@ -0,0 +1,70 @@
...
+      bool unify(TermList lhs, TermList rhs) {
+        return subst.unify(lhs, 0, rhs, 0);
+      }
+      TermList operator()(unsigned v) const override {
+        return subst.apply(TermList::var(v), 0);
+      }
+      RobSubstitution subst;
+    } unifier;
+
+    if (!unifier.unify(lhs, rhs)) {
```
MichaelRawson, line 47: Isn't this just `subst.unify(lhs, 0, rhs, 0)`?

MichaelRawson (approved):
Good idea! Could we also look at any/every other generating inference in a similar light?

One thought: if we do enable this by default, I think it subsumes `trivial literal removal`.

cmake/sources.cmake
mezpusz: Thanks, you're right!

Inferences/SubsumptionEqualityResolution.cpp
mezpusz: Yes, but we need a `SubstApplicator` to pass to `isRenaming`. So either we have a `RobSubstitution` and pass it to some structure or wrap it directly, I chose the second option. Does it make sense?

mezpusz: Btw, `isRenaming` could be templatized a bit better to accept anything with an `apply(unsigned)` function, like `SubstHelper` functions, and not just `SubstApplicator`s.

mezpusz:
> Good idea! Could we also look at any/every other generating inference in a similar light?

I'm afraid the standard inferences are done ((subsumption) demodulation for superpositions, subsumption resolution and this for resolutions), but we could go crazy and implement `URResolution` or something from ALASCA maybe?

> One thought: if we do enable this by default, I think it subsumes `trivial literal removal`.

Yes. But depending on the order of simplifications, we might get a slowdown (trivial literal removal is much faster and can be applied up front)?

quickbeam123:
@mezpusz , could you please also create an entry for this in the sampler files?

mezpusz:
> @mezpusz , could you please also create an entry for this in the sampler files?

Sure, I set the ratio now to 9 off / 1 on. Do you think it's okay now?

Inferences/SubsumptionEqualityResolution.cpp
MichaelRawson: Sorry, I got this mixed up with the `Matcher` stuff. You're right, the interface is annoying, something to iron out later...

MichaelRawson:
> the standard inferences are done

True, I was thinking about the less-standard. I don't want to duplicate every inference to have a subsuming version, so I had two things in mind:
1. When performing any generating inference and the conclusion subsumes (any of) the parents, delete the parent. But perhaps this is just backward subsumption with extra steps.
2. Do generating inferences as a final ultra-expensive simplification to see if any conclusion simplifies the parent. But I think this is (1) in disguise.

Not one of my better thoughts. Happy to go!

mezpusz:
I think (1) would be reasonable, and there is already some logic with the `generateSimplify` that returns also a `premiseRedundant` bool. But AFAIK only some ALASCA inferences implement this fully, and not sure how it would look like with the general "perform inferences as usual but at the end try to subsume" logic.

It could be also strengthened by stopping all further inferences when we find redundancy and only adding the premises that actually make the clause redundant.

PR #845, opened 2026-04-21
cleanup Multiprocessing API

MichaelRawson:
Found some dusty corners of `Lib::Sys::Multiprocessing` while working on #840. Simplify the interface and remove stuff that is no longer used - for example we never stop (rather than kill) child processes any more.

### DIFF
--- a/CASC/PortfolioMode.cpp
+++ b/CASC/PortfolioMode.cpp
@@ -32,9 +32,7 @@
 #include "Shell/Shuffling.hpp"
 #include "Shell/TheoryFinder.hpp"
 
-#include <sys/wait.h>
 #include <limits>
-#include <unistd.h>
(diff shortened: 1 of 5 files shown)


quickbeam123 (approved):

quickbeam123:
Looking cool!

Maybe we should test running a bit of portfolio mode across platforms?

MichaelRawson:
Maybe - I was hoping that the semantic behaviour should be "obviously the same" so we can save some cycles, but I admit that this might be a stretch.

I could run some randomly-chosen TPTP problems in portfolio on desktop Linux - could you do the same on a Mac?

mezpusz (approved):
Looks good to me! ~~How should we test it?~~ I'll run some cluster tests when I get access.

PR #846, opened 2026-04-22
HOL improvements (continued)

mezpusz:
Yet another HOL PR, this time going over many small things, mostly preprocessing-related:
- FOOLElimination
- InequalitySplitting
- The heuristics instantiations and its inference counterpart in CNFOnTheFly
- TweeGoalTransformation
- Higher-order split queues
- Some adjustments in how HO inferences are added: some are now hidden behind the `prb.isHigherOrder()` flag, and `ArgCong` was added (it was accidentally left out previously)
- Updated the HOL sampler

After running some iterations of debug sampling with `samplerHOL.smp`, there were only a few dozen errors left, mostly related to substitutions using loose DB indices, which needs further work on indices/unification/matching.

Here are some results:

FOL TPTP:

Discount
run 0 unsat: 9208 (0) sat: 1042 (0) cputime: 56725.35 s instructions: 222732776 Mi memory: 1462299.73 MB
run 1 unsat: 9208 (0) sat: 1042 (0) cputime: 56951.17 s instructions: 222734542 Mi memory: 1462058.74 MB

Otter
run 0 unsat: 9321 (1) sat: 1025 (0) cputime: 61791.53 s instructions: 222705868 Mi memory: 1694925.40 MB
run 1 unsat: 9320 (0) sat: 1025 (0) cputime: 62014.35 s instructions: 222707271 Mi memory: 1694757.50 MB

(one problem lost due to noise)

HOL TPTP:

Discount
run 0 unsat: 2147 (7) sat: 1 (0) cputime: 8842.38 s instructions: 44028487 Mi memory: 113310.06 MB
run 1 unsat: 2232 (92) sat: 1 (0) cputime: 8992.22 s instructions: 44261476 Mi memory: 114153.27 MB

Otter
run 0 unsat: 2107 (5) sat: 1 (0) cputime: 9257.15 s instructions: 44915239 Mi memory: 135458.60 MB
run 1 unsat: 2201 (99) sat: 1 (0) cputime: 9419.50 s instructions: 44903685 Mi memory: 136672.79 MB

Note that here the big increase in solved problems can be mostly attributed to `ArgCong` kicking in.

### DIFF
--- a/Inferences/HOL/CNFOnTheFly.cpp
+++ b/Inferences/HOL/CNFOnTheFly.cpp
@@ -199,7 +199,6 @@ ClauseIterator produceClauses(Clause* c, bool generating, SkolemisingFormulaInde
               auto tS = subst.apply(t, 0);
               auto argS = subst.apply(args[0],1); 
 
-              // can't use replaceLits, as we need to apply the type unifier
               RStack<Literal*> resLits;
               for (const auto& curr : *c) {
                 resLits->push(curr == lit
--- a/Inferences/HOL/CNFOnTheFly.cpp
+++ b/Inferences/HOL/CNFOnTheFly.cpp
@@ -64,10 +64,6 @@ ClauseIterator produceClauses(Clause* c, bool generating, SkolemisingFormulaInde
   if(generating && (eager || simp)){ return ClauseIterator::getEmpty(); }
   if(!generating && gen){ return ClauseIterator::getEmpty(); }
 
-  if (instantiations) {
-    INVALID_OPERATION("Options::CNFOnTheFly::CONJ_EAGER not yet supported");
-  }
-
   static TermStack args;
 
   for (const auto& lit : *c) {
@@ -180,13 +176,39 @@ ClauseIterator produceClauses(Clause* c, bool generating, SkolemisingFormulaInde
       case Proxy::PI:
       case Proxy::SIGMA: {
         ASS_EQ(args.size(), 1);
+        ClauseStack instCls;
         TermList srt = *SortHelper::getResultSort(head.term()).term()->nthArgument(0);
         TermList newTerm;
         Proxy proxy;
-        if((prox == Proxy::PI && positive) ||
-          (prox == Proxy::SIGMA && !positive)){
+        if ((prox == Proxy::PI && positive) || (prox == Proxy::SIGMA && !positive)) {
           proxy = Proxy::PI;
           newTerm = piRemoval(args[0], c, srt);
+          if (instantiations && !args[0].isVar()) {
+            auto insts = env.signature->getInstantiations();
+            for (const auto& t : iterTraits(insts->iter())) {
+              ASS(t.isTerm());
+              static RobSubstitution subst;
+              subst.reset();
+
+              auto tSort = SortHelper::getResultSort(t.term());
+              auto aSort = srt.domain();
+
+              if (!subst.unify(tSort,0,aSort,1)) {
+                continue;
+              }
+              auto tS = subst.apply(t, 0);
+              auto argS = subst.apply(args[0],1); 
+
+              // can't use replaceLits, as we need to apply the type unifier
+              RStack<Literal*> resLits;
+              for (const auto& curr : *c) {
+                resLits->push(curr == lit
+                  ? boolEq(HOL::create::app(argS, tS), rhs)
+                  : subst.apply(curr, 1));
+              }
+              instCls.push(Clause::fromStack(*resLits, GeneratingInference1(InferenceRule::HEURISTIC_INSTANTIATION, c)));
+            }
+          }
         } else {
           ASS(term.isTerm());
           bool newTermCreated = false;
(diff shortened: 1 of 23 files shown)


quickbeam123:
Before I look closely, a random remark: When I checked in February, the higher-order version of TweeGoalTransformation was terrible (compared to its FO counterpart). 

I made a hypothesis, that it's a bad idea to give names to partially applied terms, the corresponding fix is [here](https://github.com/vprover/vampire/commit/158095b53ef1b39fd747e7143ffb4e5808f2cd0f) (In general, there are a few fixes in the [hol](https://github.com/vprover/vampire/tree/hol) branch, on top of ahmed-new-hol.) But even after this change, `tgt` strategies did not become "way better than without" (as is the case in FO), so I wonder whether there is still something missing we could explore (or maybe the HO problems / search space simply behave differently).

mezpusz:
@quickbeam123 Thanks for mentioning this, I did not know about this branch! Your observations may be valid, and we probably have to adapt many things on top of the `ahmed-new-hol` branch. The changes so far (including the new ones) are not really changing the behaviour of Twee goal transformation, they just adapt it to the HO setting (no loose DB indices are bound and "appification" happens). But the good news is that, apart from HO unification and matching, we are getting very close to having all important changes in master from `ahmed-new-hol`, after which I can look at the `hol` branch too and import changes from there. Is this a reasonable plan for now?

Inferences/HOL/CNFOnTheFly.cpp
```diff
@@ -180,13 +176,39 @@ ClauseIterator produceClauses(Clause* c, bool generating, SkolemisingFormulaInde
...
+              auto tSort = SortHelper::getResultSort(t.term());
+              auto aSort = srt.domain();
+
+              if (!subst.unify(tSort,0,aSort,1)) {
+                continue;
+              }
+              auto tS = subst.apply(t, 0);
+              auto argS = subst.apply(args[0],1); 
+
+              // can't use replaceLits, as we need to apply the type unifier
```
MichaelRawson: Should `replaceLits` be patched to do this?

Kernel/HOL/HOL.hpp
```diff
@@ -79,6 +79,8 @@ Stack<std::pair<TermList, UnificationInference>> getProjAndImitBindings(TermList
 
 TermList createGeneralBinding(TermList head, const TermStack& sorts, unsigned& freshVar, bool surround = true);
 
+TermStack getAbstractionTerms(Literal* lit);
```
MichaelRawson, line 88: What does this do?

Saturation/PredicateSplitPassiveClauseContainers.cpp
```diff
@@ -481,6 +482,66 @@ float TheoryMultiSplitPassiveClauseContainer::evaluateFeatureEstimate(unsigned,
...
+
+  static DHMap<const Term*,unsigned> cache;
+  unsigned* cached;
+  if (!cache.getValuePtr(tt,cached)) {
+    return *cached;
+  }
+
+  // it's OK that the entry in cache has already been created, will only possibly ask for proper subterms
+
+  // TODO(mhajdu): I think this is not okay because the DHMap may reallocate during the recursive call,
```
MichaelRawson: This is indeed a hazard. But it can be fixed by doing the hashtable lookup again at the end of the function?

Shell/TweeGoalTransformation.cpp
```diff
@@ -69,6 +70,8 @@ class Definizator : public BottomUpTermTransformer {
     bool _groundOnly;
 
     unsigned _typeArity;
+    TermStack _typeVars;
+    TermStack _termVars;
```
MichaelRawson, line 74: What do these do, and how do they relate to `_allVars`?

MichaelRawson (approved):
Sorry for delay. Let's keep it moving!

Saturation/PredicateSplitPassiveClauseContainers.cpp
quickbeam123: Indeed. This is something I have also committed in the past and was removing recently from a similar function.

Keeping the pointer "too long" may backfire. One has to hash again at the end to store at a proper place.

Shell/TweeGoalTransformation.cpp
mezpusz: These are just for the polymorphization of `TweeGoalTransformation`, so one of them contains the type variables, based on which we create the `OperatorType` (this needs normalization), and so on. Unfortunately variants of this exist all across the codebase.

Saturation/PredicateSplitPassiveClauseContainers.cpp
mezpusz: Thanks for the suggestion, I'll add the code at the end.

Kernel/HOL/HOL.hpp
mezpusz: Good question, I can add a comment once I figure out.

Inferences/HOL/CNFOnTheFly.cpp
mezpusz: Good point. Actually, `replaceLits` is different in multiple aspects, so I think patching it would be more work than this "edge case" here. In any case, I remove this comment if you agree.

MichaelRawson: Agreed!

quickbeam123 (approved):

PR #847, opened 2026-04-29
Choice inference

mezpusz:
Cleaning up the choice inferences, merging the relevant changes from the HOL branch, and adding some unit tests.

HOL regression (with choice reasoning enabled, `-chr on`), shows a bit of improvement, mostly because loose DB indices are not bound and hence some errors are avoided:

Discount
run 0 unsat: 2240 (0) sat: 1 (0) cputime: 8449.42 s instructions: 41927093 Mi memory: 112563.93 MB
run 1 unsat: 2252 (12) sat: 1 (0) cputime: 8696.54 s instructions: 44143448 Mi memory: 117507.94 MB

Otter
run 0 unsat: 2212 (1) sat: 1 (0) cputime: 8943.08 s instructions: 42962767 Mi memory: 138522.92 MB
run 1 unsat: 2220 (9) sat: 1 (0) cputime: 9103.20 s instructions: 44816098 Mi memory: 141062.58 MB

Once review is done, I would like to move the `Choice` and `ChoiceDefinitionISE` to `Inferences/HOL`.

### DIFF
--- a/Inferences/Choice.cpp
+++ b/Inferences/Choice.cpp
@@ -159,8 +159,7 @@ ClauseIterator Choice::generateClauses(Clause* premise)
 {
   return pvi(premise->getSelectedLiteralIterator()
     .flatMap([](Literal* lit) {
-      NonVariableNonTypeIterator nvi(lit);
-      return pvi(getUniquePersistentIteratorFromPtr(&nvi));
+      return getUniquePersistentIterator(NonVariableNonTypeIterator(lit));
     })
     .filter(IsChoiceTerm())
     .flatMap(ResultFn()));
--- a/Inferences/Choice.cpp
+++ b/Inferences/Choice.cpp
@@ -65,13 +65,13 @@ Clause* Choice::createChoiceAxiom(TermList op, TermList set)
 
 struct Choice::AxiomsIterator
 {
-  AxiomsIterator(Term* term)
+  AxiomsIterator(TermList term)
   {
-    _set = *term->nthArgument(3);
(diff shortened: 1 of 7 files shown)


quickbeam123 (approved):

Inferences/Choice.cpp
```diff
@@ -149,55 +136,34 @@ struct Choice::IsChoiceTerm
...
-  //apply rewrite rules to literals
-  auto it4 = getMapAndFlattenIterator(std::move(it3), ResultFn());
-
-
-  return pvi( std::move(it4) );
-
+  return pvi(premise->getSelectedLiteralIterator()
+    .flatMap([](Literal* lit) {
+      NonVariableNonTypeIterator nvi(lit);
+      return pvi(getUniquePersistentIteratorFromPtr(&nvi));
```
MichaelRawson: I think this is a hazard, as `nvi` will be destroyed at the end of the function when the sub-iterator is returned, and then when the sub-iterator is consumed it will use `nvi`. What did I miss? Also - probably you don't need `pvi` here because of Joe's templating magic, but I could be wrong about that too.

MichaelRawson (approved):
Basically good - just one question.

Inferences/Choice.cpp
mezpusz: As far as I know `UniquePersistentIterator` consumes its inner iterator in the ctor already, making it persistent, so the inner iterator is not needed anymore. I will try to simplify this part though.

MichaelRawson: Ahah. Smart - yeah, maybe you don't need `pvi` but either way I'm fine with this.

PR #848, opened 2026-04-29
Separate generalizing term index and use code trees

mezpusz:
This PR is a first step towards making code trees the default for matching (generalizing) indices. This will help us focus on code trees for matching in the future, e.g. extending it to HOL.
Most of the PR contains just removing interface code so that code trees can be more directly used as generalizing term indices.

One missing feature from term code trees was the handling of sorts of indexed variables. An ad-hoc fix was used in forward demodulation but this was not even used in the critical case where the variable has a non-ground sort, as those equations were not added to the demodulation LHS index due to another small bug. In any case, handling inside the code tree is more natural (IMO), but it could be later updated to share also the variable sorts in a separate code tree, such that sort matching would be also efficient.

I also switched all places where we used substitution trees for getting generalizations to use code trees. In particular, in induction, ALASCA demodulation and obtaining Skolems in CNF on the fly.

Regression results are shown below:

ALASCA SMTLIB+TPTP:
run 0 unsat: 44428 (0) sat: 135 (0) cputime: 48261.48 s instructions: 185056392 Mi memory: 996109.45 MB
run 1 unsat: 44428 (0) sat: 135 (0) cputime: 48617.18 s instructions: 185033849 Mi memory: 995655.16 MB

(Two problems flipped from timeout to error, I checked these locally and the error happens with a bigger timeout on master.)

FOL Discount TPTP:
run 0 unsat: 9208 (0) sat: 1042 (0) cputime: 56855.80 s instructions: 222732554 Mi memory: 1462215.63 MB
run 1 unsat: 9208 (0) sat: 1042 (0) cputime: 57448.59 s instructions: 222738803 Mi memory: 1462395.81 MB

FOL Otter TPTP:
run 0 unsat: 9321 (2) sat: 1025 (1) cputime: 61948.36 s instructions: 222706049 Mi memory: 1694946.39 MB
run 1 unsat: 9322 (3) sat: 1024 (0) cputime: 62511.80 s instructions: 222690853 Mi memory: 1695844.64 MB

(I will check the diff here more closely, most likely noise.)

HOL Discount TPTP:
run 0 unsat: 2112 (0) sat: 1 (0) cputime: 8822.22 s instructions: 43898419 Mi memory: 113978.75 MB
run 1 unsat: 2112 (0) sat: 1 (0) cputime: 8812.26 s instructions: 43899663 Mi memory: 113938.23 MB

HOL Otter TPTP:
run 0 unsat: 2073 (0) sat: 1 (0) cputime: 9278.63 s instructions: 44823797 Mi memory: 136652.94 MB
run 1 unsat: 2073 (0) sat: 1 (0) cputime: 9291.50 s instructions: 44824137 Mi memory: 136576.49 MB

Induction:
run 0 unsat: 973 (3) sat: 0 (0) cputime: 11091.90 s instructions: 45481819 Mi memory: 673205.01 MB
run 1 unsat: 973 (3) sat: 0 (0) cputime: 11088.78 s instructions: 45482307 Mi memory: 672135.89 MB

(Diff to be checked here too.)

Synthesis:
run 0 unsat: 438 (0) sat: 0 (0) cputime: 1226.62 s instructions: 5234739 Mi memory: 52817.96 MB
run 1 unsat: 438 (0) sat: 0 (0) cputime: 1227.77 s instructions: 5234630 Mi memory: 52778.30 MB


Indexing/TermCodeTree.cpp
```diff
@@ -51,9 +54,14 @@ void TermCodeTree<higherOrder, Data>::insert(Data* data)
   static CodeStack code;
   code.reset();
 
-  TermList t=data->term;
+  auto t = data->key();
   if (t.isVar()) {
     code.push(CodeOp::getTermOp(ASSIGN_VAR,0));
+    // we match the variable sort separately, but the binding array has to be prepared
```
mezpusz, line 60: Variable matching change 1.

Indexing/TermCodeTree.cpp
```diff
@@ -154,19 +163,31 @@ Data* TermCodeTree<higherOrder, Data>::TermMatcher::next()
...
   }
-  
-  Base::_matched=Base::execute();
-  if (!Base::_matched) {
-    return 0;
-  }
 
-  ASS(Base::op->isSuccess());
-  return Base::op->template getSuccessResult<Data>();
+  while ((Base::_matched=Base::execute())) {
```
mezpusz, line 167: Variable matching change 2.

Kernel/TypedTermList.hpp
```diff
@@ -40,6 +41,22 @@ class TypedTermList : public TermList
...
   TypedTermList(Term* t) : TypedTermList(TermList(t), SortHelper::getResultSort(t)) {}
 
+  VirtualIterator<TermList> varIter() const {
+    if (isVar()) {
+      return pvi(concatIters(getSingletonIterator(untyped()), Term::getVariableIterator(_sort)));
+    }
+    return Term::getVariableIterator(untyped());
+  }
+
+  bool containsAllVariablesOf(TypedTermList other) const {
```
mezpusz, line 51: Variable matching change 3.

Indexing/CodeTreeInterfaces.hpp
```diff
@@ -21,25 +21,57 @@
...
 
+template<class Data>
+class GenSubstitution
+  : public SubstApplicator
+{
+public:
+  GenSubstitution(CodeTree::BindingArray* bindings, Renaming* resultNormalizer)
+  : _bindings(bindings), _resultNormalizer(resultNormalizer) {}
+
+  USE_ALLOCATOR(GenSubstitution);
```
MichaelRawson: Does this have a performance impact? Otherwise we can drop it, I'm hoping to gradually remove these.

Indexing/TermCodeTree.hpp
```diff
@@ -60,12 +62,14 @@ class TermCodeTree : public CodeTree
     using Base = Matcher</*removing*/false,false,higherOrder>;
     using Base::ft;
 
-    void init(CodeTree* tree, TermList t);
+    void init(CodeTree const* tree, TypedTermList t);
```
MichaelRawson: Nitpick, but is there any time when `tree` is `nullptr`? If not this could be a const reference, converting to pointer internally if need be.

Inferences/HOL/CNFOnTheFly.cpp
```diff
@@ -214,7 +214,7 @@ ClauseIterator produceClauses(Clause* c, bool generating, SkolemisingFormulaInde
             if(results.hasNext()){
               auto tqr = results.next();
               TermList skolemTerm = tqr.data->value;
-              skolemTerm = tqr.unifier->applyToBoundResult(skolemTerm);
+              skolemTerm = tqr.unifier->apply(skolemTerm);
```
MichaelRawson, line 217: Just making sure this is intentional?

MichaelRawson (approved):
Cool! And if code trees are the preferred indexing strategy, then they should be on the hot path. 

One thing I don't really understand and would like to: one of the strengths of Vampire's TFF1 representation is that sorts are just terms and can be treated as such almost all of the time. So I wouldn't really expect you to need any sort information (except perhaps in the dread two-variable-equality case), `TypedTermList`, etc. It's possible to construct e.g. a retrieved generalisation that is ill-typed, but then it should not have been put in the index to start with. What am I missing?

Indexing/CodeTreeInterfaces.hpp
mezpusz: I don't think so, and we should probably have a non-heap `GenSubstitution` anyways, as it's basically a class wrapper for applying the substitution and does nothing.

Indexing/TermCodeTree.hpp
mezpusz: Good point, I will try this!

Inferences/HOL/CNFOnTheFly.cpp
mezpusz: Yes, this is the default for code trees (having fixed variables in the query, and only substituting in the result).

easychair:
> Cool! And if code trees are the preferred indexing strategy, then they should be on the hot path.

One has to be careful. Code trees are goof for FORWARD matching,
subsumption and subsumption resolution. They probably suck for the
corresponding backward operations.

> One thing I don't really understand and would like to: one of the strengths of Vampire's TFF1 representation is that sorts are just terms and can be treated as such almost all of the time. So I wouldn't really expect you to need any sort information (except perhaps in the dread two-variable-equality case), TypedTermList, etc. It's possible to construct e.g. a retrieved generalisation that is ill-typed, but then it should not have been put in the index to start with. What am I missing?

I come from logic and have no idea what weird abbreviations like TFF1
are, so I cannot comment on it. It would be good if we speak logic
when discussing Vampire.

If we speak FOL with sorts, then, as I mentioned, you can only violate
well-sortedness when you have equalities x = y.

Andrei

mezpusz:
Just to be clear, code trees are only used for forward matching in Vampire, and this PR just adds new places where code trees are used instead of substitution trees (which should be deprecated for forward matching).

mezpusz:
Regarding sorts matching, the only place where we have to match sorts explicitly when we match terms is for variables. What I meant by the comment that @MichaelRawson referred to was the following. Suppose we index many variables:

`X0 : nat`
`X0 : list(X1)`
`X0 : list(list(X1))`
`X0 : list(nat)`
...
Currently when we match a term `t`, it will match with all the above variables, and then we go through all sorts one by one to get the ones that actually match the sort of `t`. Instead, we could use a second code tree, to index just `nat`, `list(X1)`, `list(list(X1))`, `list(nat)`, ... This is not supposed to be too hard, I'm just not sure it's worth it (we have to check if this situation comes up a lot in practice).

MichaelRawson:
> I come from logic and have no idea what weird abbreviations like TFF1
are, so I cannot comment on it. It would be good if we speak logic
when discussing Vampire.

If you prefer functional programming terminology, it's "rank-1 polymorphism" (like Standard ML), but see also the (quite readable!) [TFF1 spec](https://www21.in.tum.de/~blanchet/tff1spec.pdf).

The point is that sorts are just terms of a special sort `$tType`, so you can write terms like `cons($real, X, Xs)` and know implicitly that `X : $real` without having to modify the `TermList` structure. Generally stuff like resolution "just works" and you don't have to make well-sortedness checks.

@mezpusz 's comment explained why we need to handle sorts specially in code trees in the variable case, thanks!

quickbeam123 (approved):

Issue #843, opened 2026-04-20
new TSTP Skolemization standard


MichaelRawson:
The [TPTP Format for Derivations](https://tptp.org/UserDocs/QuickGuide/Derivations.html) changed what it requires for Skolemisation steps slightly. I don't think we were ever compliant, so now might be a good time to do it. Shouldn't be so hard, I'll get to this (hopefully before the competition) if nobody feels enthused before.

mezpusz:
Maybe @i-am-a-teapot also wants to contribute to this.

MichaelRawson:
Fixed due to #849. Thanks @i-am-a-teapot!

PR #849, opened 2026-04-29
Skolemization format update

i-am-a-teapot:
Update for TPTP proof output for Skolemization for #843

### DIFF
--- a/Kernel/InferenceStore.cpp
+++ b/Kernel/InferenceStore.cpp
@@ -13,9 +13,11 @@
  */
 
 #include "Kernel/Theory.hpp"
+#include "Kernel/Unit.hpp"
 #include "Lib/Allocator.hpp"
 #include "Lib/Environment.hpp"
 #include "Lib/Int.hpp"
+#include "Lib/Metaiterators.hpp"
 #include "Lib/ScopedPtr.hpp"
 #include "Lib/SharedSet.hpp"
 #include "Lib/Stack.hpp"
@@ -42,7 +44,9 @@
 
 #include "InferenceStore.hpp"
 
+#include <memory>
 #include <set>
+#include <string>
 
 //TODO: when we delete clause, we should also delete all its records from the inference store
 
@@ -79,6 +83,15 @@ void InferenceStore::recordIntroducedSymbol(Unit* u, SymbolType st, unsigned num
 {
   SymbolStack* pStack;
   _introducedSymbols.getValuePtr(u->number(),pStack);
+  pStack->push(SymbolId(st,number));
+}
+
+void InferenceStore::recordIntroducedSkolemSymbol(Unit* u, SymbolType st, unsigned number, unsigned replacedVar, std::unique_ptr<DHSet<unsigned>> inScopeVars)
+{
+  SymbolStack* pStack;
+  _introducedSymbols.getValuePtr(u->number(),pStack);
+  _introducedSymbolReplacedVars.insert(number, replacedVar);
+  _introducedSymbolInScopeVars.insert(number, std::move(inScopeVars));
   pStack->push(SymbolId(st,number));
 }
 
@@ -555,19 +568,24 @@ protected:
   std::string getNewSymbols(std::string origin, std::string symStr) {
     return "new_symbols(" + origin + ",[" +symStr + "])";
   }
+
+  std::string getSymbolName(SymbolId sym) {
+    if (sym.first == SymbolType::FUNC ) {
+      return env.signature->functionName(sym.second);
+    } else if (sym.first == SymbolType::PRED){
+      return env.signature->predicateName(sym.second);
+    } else {
+      return env.signature->typeConName(sym.second);
+    }
+  }
+
   /** It is an iterator over SymbolId */
   template<class It>
   std::string getNewSymbols(std::string origin, It symIt) {
     std::ostringstream symsStr;
     while(symIt.hasNext()) {
       SymbolId sym = symIt.next();
-      if (sym.first == SymbolType::FUNC ) {
-        symsStr << env.signature->functionName(sym.second);
-      } else if (sym.first == SymbolType::PRED){
-        symsStr << env.signature->predicateName(sym.second);
-      } else {
-        symsStr << env.signature->typeConName(sym.second);
-      }
+      symsStr << getSymbolName(sym);
       if (symIt.hasNext()) {
         symsStr << ',';
       }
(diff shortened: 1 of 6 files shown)


i-am-a-teapot:
The TSTP standard has changed to allow multiple skolemizations now, which this new format *should* be compatible with.
I ran some test files which looked fine, however there may be edge cases which I am not aware of (seems like skolemization works slighlty differently for higher order things)

Shell/Skolem.cpp
```diff
@@ -487,11 +493,11 @@ Formula* Skolem::skolemise (Formula* f)
         if (arity > 0) {
           def = new QuantifiedFormula(FORALL, vArgs.list(), def);
         }
-
-        Unit* defUnit = new FormulaUnit(def,NonspecificInference0(UnitInputType::AXIOM,InferenceRule::SKOLEM_SYMBOL_INTRODUCTION));
-        UnitList::push(defUnit,_skolimizingDefinitions);
+        
+        //Unit* defUnit = new FormulaUnit(def,NonspecificInference0(UnitInputType::AXIOM,InferenceRule::SKOLEM_SYMBOL_INTRODUCTION));
+        //UnitList::push(defUnit,_skolimizingDefinitions);
```
MichaelRawson: I think this was the last use of `SKOLEM_SYMBOL_INTRODUCTION`. Maybe it can be removed now?

Kernel/InferenceStore.cpp
```diff
@@ -82,6 +88,15 @@ void InferenceStore::recordIntroducedSymbol(Unit* u, SymbolType st, unsigned num
   pStack->push(SymbolId(st,number));
 }
 
+void InferenceStore::recordIntroducedSkolemSymbol(Unit* u, SymbolType st, unsigned number, unsigned replacedVar, std::unique_ptr<std::vector<unsigned>> inScopeVars)
```
MichaelRawson: Why `unique_ptr<vector<unsigned>>` rather than just `vector<unsigned>`?

Kernel/InferenceStore.cpp
```diff
@@ -82,6 +88,15 @@ void InferenceStore::recordIntroducedSymbol(Unit* u, SymbolType st, unsigned num
...
+{
+  SymbolStack* pStack;
+  _introducedSymbols.getValuePtr(u->number(),pStack);
+  _introducedSymbolReplacedVars.insert(number, replacedVar);
```
MichaelRawson: All of these fields are keyed by an `unsigned` and global to Vampire, right? What prevents us from Skolemnising `X0` in different formulas and clashing here?

Kernel/InferenceStore.cpp
```diff
@@ -555,19 +570,24 @@ struct InferenceStore::TPTPProofPrinter
...
+
+  std::string getSymbolName(SymbolId sym) {
+    if (sym.first == SymbolType::FUNC ) {
+      return env.signature->functionName(sym.second);
+    } else if (sym.first == SymbolType::PRED){
+      return env.signature->predicateName(sym.second);
+    } else {
+      return env.signature->typeConName(sym.second);
+    }
+  }
```
MichaelRawson, line 579: Good idea. I feel like this could be promoted to `Signature.hpp` if it doesn't exist already?

Shell/Skolem.hpp
```diff
@@ -103,7 +103,9 @@ class Skolem
   DHMap<unsigned,TermList> _varSorts;
 
   // for some heuristic evaluations after we are done
-  Stack<std::pair<bool, unsigned>> _introducedSkolemSyms;
+  
+  // Type, numberOfSymbol, Variable that was replaced by the skolem symbol, Variables that are in scope for a skolem symbol (for proof recording purposes)
+  Stack<std::tuple<bool, unsigned, unsigned, std::unique_ptr<std::vector<unsigned>>>> _introducedSkolemSyms;
```
MichaelRawson: The first `bool` is whether the Skolem is for sorts or terms, I think. So: could this be a pair `<unsigned, Term *>` where the lhs is the variable being eliminated and the rhs is the Skolem term?

MichaelRawson (commented):
Hi @i-am-a-teapot - thanks for the patch! Got some questions about it but it's looking good, we should have this in for CASC. :)

Shell/Skolem.cpp
i-am-a-teapot: I agree, I just commented it out since i was not sure whether it may be good to have it in the future. I think we could remove it!

Kernel/InferenceStore.cpp
i-am-a-teapot: You are completely right, I also need to key it by the unit number I am afraid!

easychair:
Did I get it right that we drop a step from the proof (output)?

If we have Skolemization, we should have a rule for it. It is essential for
proof-checking.

Or is it something else?

Andrei

i-am-a-teapot:
Skolemization previously consisted of two steps in our proof output: Introducing the symbol and another step which is the eliminating the existential.
The TSTP format is now slightly changed so we need to be more detailed for skolemization (so we keep recording skolemization!). Pretty much the simplest way of adhering to the required format, is to just do everything in one single step (printing all relevant detail in this single step) instead of splitting it in two.

Kernel/InferenceStore.cpp
i-am-a-teapot: Using a term now, this is not necessary anymore.

Kernel/InferenceStore.cpp
i-am-a-teapot: I now key by the SymbolId (pair of symbolType and identifier) , so it should be unique.

Shell/Skolem.hpp
i-am-a-teapot: I modified it to use what you suggested and I *think*, it should be equivalent to check the term kind (if it is sort or not). (I also checked what is happening in Term and that seems to be ok, but maybe I am missing something)

In some tests I ran this seems to work, let me know if you see a problem with it!

Kernel/SubstHelper.hpp
```diff
@@ -529,10 +529,14 @@ Formula* SubstHelper::applyImpl(Formula* f, Applicator& applicator, bool noShari
...
+      TermList newSort = sort;
+      if(sort.isTerm()){
+        newSort = TermList(applyImpl<ProcessSpecVars>(sort.term(), applicator, noSharing));
+      }
       ASS(binding.isVar());
       unsigned newVar = binding.var();
-      VSList::push({newVar, sort}, newVars);
-      if(newVar!=v) {
+      VSList::push({newVar, newSort}, newVars);
+      if(newVar!=v || newSort!=sort) {
```
i-am-a-teapot, line 536: In case you are wondering what is happening here:
I tried out a problem (see below), and (default) vampire crashed in tptp proof output mode due to not replacing the skolemized type in the polymorphic container(A), leaving an unbound variable. This should fix it I hope.

```
tff(container_type, type, (
    container: $tType > $tType
)).
tff(contains_pred, type, (
    contains: !> [A: $tType] : ( A * container(A) ) > $o
)).
tff(trigger_axiom, axiom, (
    ! [A: $tType] : 
      ! [X: A] : 
        ? [C: container(A)] : contains(A, X, C)
)).

% 4. Conjecture to complete valid TPTP problem state
tff(dummy_conjecture, conjecture, (
    ! [A: $tType] : 
      ! [X: A] : 
        ? [C: container(A)] : contains(A, X, C)
)).
```

MichaelRawson (approved):

MichaelRawson:
@quickbeam123 - this is now screwing with `SubstHelper` because it wasn't substituting sort variables in quantifier blocks. Could you take a quick look to double check?

Shell/Skolem.hpp
```diff
@@ -103,8 +103,9 @@ class Skolem
   DHMap<unsigned,TermList> _varSorts;
 
   // for some heuristic evaluations after we are done
-  Stack<std::pair<bool, unsigned>> _introducedSkolemSyms;
-  
+
```
quickbeam123: I know this was undocumented before, but since we are here, could we have a comment about what the `bool` and the `Term*` stand for?

quickbeam123:
> @quickbeam123 - this is now screwing with `SubstHelper` because it wasn't substituting sort variables in quantifier blocks. Could you take a quick look to double check?

The change definitely looks like an improvement! Thank you for proposing it, @i-am-a-teapot !

Kernel/InferenceStore.cpp
quickbeam123: I agree, although `SymbolId` seems to be a local concept of `InferenceStore` at the moment ...

quickbeam123 (approved):

Kernel/InferenceStore.cpp
i-am-a-teapot: For now I figured that I would leave it local here (as I believe typically the code knows with what symbol it deals and it is somewhat unique to InferenceStore to need this).

PR #851, opened 2026-05-01
also allow imports relative to the _supplied_ input file

### LINKED ISSUE #850: TPTP includes relative to input problem not working
In the `TPTP::resolveInclude` logic, the second case (`// 2`) does not fire for the most important combination "supplied input file" -> "its include", because `currentFile.path` is empty for the "supplied input file"

This is because `UIHelper::parseFile(const std::string& inputFile` opens a stream for the TPTP parser, which then never learns about `inputFile`.  A fix would be to let the TPTP parser know about `inputFile` even though it was already opened on its behalf. 

I think case (`// 2`) could still work for the situation "correctly resolved include file" -> "its nested include".

MichaelRawson:
Hopefully closes #850. This is probably my bug, oops.

Weave an appropriate `std::filesystem::path` through the code. The whole logic could use a re-work but this works for now.

### DIFF
--- a/Shell/UIHelper.cpp
+++ b/Shell/UIHelper.cpp
@@ -189,10 +189,10 @@ static bool hasEnding (std::string const &fullString, std::string const &ending)
   }
 }
 
-void UIHelper::tryParseTPTP(istream& input)
+void UIHelper::tryParseTPTP(istream& input, std::filesystem::path path)
 {
   LoadedPiece& curPiece = _loadedPieces.top();
-  Parse::TPTP parser(input,curPiece._units);
+  Parse::TPTP parser(input,path,curPiece._units);
   try {
     parser.parse();
     curPiece._units = parser.unitBuffer();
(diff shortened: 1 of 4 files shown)


Shell/UIHelper.cpp
```diff
@@ -238,9 +238,11 @@ void UIHelper::parseSingleLine(const std::string& lineToParse, Options::InputSyn
   std::istringstream stream(lineToParse);
   try {
     switch (inputSyntax) {
-      case Options::InputSyntax::TPTP:
-        tryParseTPTP(stream);
+      case Options::InputSyntax::TPTP: {
+        std::filesystem::path path(lineToParse);
+        tryParseTPTP(stream, path);
```
quickbeam123: Shouldn't this better be "\<string\>" again?

quickbeam123 (approved):

Shell/UIHelper.cpp
MichaelRawson: Right, I confused myself.

PR #856, opened 2026-05-18
Hol improvements (continued)

mezpusz:
This merges another batch from the HOL branch, and also all current changes from branch `hol` on top of it.

Now enabling checking for loose DB indices by default in `Substitution` to avoid huge changes in code. It should be a trivial check in most cases, using the shared term's "contains DB index" flag.

Regression over TPTP showed no difference, which is somewhat interesting because of the above check. I think either `Substitution` is not used in that many places, or it is mostly not enabled with the default options.

Note that I changed a bunch of seemingly other stuff on the way, I will try to mark these inline.


Kernel/MLMatcher.hpp
```diff
@@ -41,6 +42,29 @@ inline std::ostream& operator<<(std::ostream& os, MLMatchStats const& stats)
   return os;
 }
 
+typedef DHMap<unsigned,unsigned, IdentityHash, DefaultHash> UUMap;
+
+/**
+ * Binder that stores bindings into a specified array. To be used
+ * with MatchingUtils template methods.
+ */
+struct ArrayStoringBinder final
```
mezpusz, line 51: This was duplicated, adding the common implementation here.

Kernel/Ordering.cpp
```diff
@@ -569,18 +569,18 @@ struct BoostWrapper : public SymbolComparator
   }
 };
 
-struct OccurenceTiebreak {
-  OccurenceTiebreak(SymbolType, // here the SymbolType is a dummy argument, required by the template recursion convention
+struct OccurrenceTieBreak {
```
mezpusz, line 572: Typo fixed.

Saturation/SaturationAlgorithm.cpp
```diff
@@ -1430,7 +1430,7 @@ SaturationAlgorithm *SaturationAlgorithm::createFromOptions(Problem& prb, const
     gie->addFront(new Induction(*res));
   }
 
-  if (opt.instantiation() != Options::Instantiation::OFF) {
+  if (opt.instantiation()) {
```
mezpusz, line 1433: This option had only two values, so changed it `BoolOptionValue`. Btw, instantiation does not seem to be very useful (only used a few times in the schedules). Should we consider deleting it?

Kernel/MLMatcher.hpp
MichaelRawson: I feel like we have some version of this elsewhere too, it feels familiar. But this is definitely an improvement!

Kernel/Substitution.hpp
```diff
@@ -33,29 +33,45 @@ namespace Kernel {
  * This is a simple map from variables to terms.
  * If you want something with 'banks' to implement renaming variables apart, you probably want RobSubstitution.
  */
-class Substitution
+template<bool higherOrder>
+class Substitution_
```
MichaelRawson, line 37: Why is this a `Substitution_` now?

Shell/Options.cpp
```diff
@@ -525,6 +526,18 @@ void Options::init()
...
+    // TODO(HOL): investigate perhaps less invasive options of restraining
+    // general tgt in HOL, that would still be performant
+    _tweeSkipArrows = BoolOptionValue("twee_skip_arrows","tsa",true);
+    _tweeSkipArrows.description =
+      "During twee_goal_transformation, when in HOL, "
+      "don't introduce definitions for other subterms than those "
+      "whose type is not an arrow type.";
+    _tweeSkipArrows.tag(OptionTag::PREPROCESSING);
+    _tweeSkipArrows.setExperimental();
+    _lookup.insert(&_tweeSkipArrows);
```
MichaelRawson, line 537: Cool! But I don't quite understand the description. "don't introduce definitions for other subterms than those whose type is not an arrow type"? Don't introduce definitions for arrow-typed subterms?

MichaelRawson (approved):

Kernel/MLMatcher.hpp
mezpusz: There is also one with the same name but using an array of `TermList` instead of `unsigned`, so that it's hard to generalize.

Kernel/Substitution.hpp
mezpusz: This is a historic reason in the making! 😄  By just templatizing `Substitution`, I would have to rename all instances everywhere. So I thought it's easier to rename the templated class and name the special case `Substitution`.

Shell/Options.cpp
mezpusz: Yeah, there seems to be an unnecessary double negation. Changing to what you wrote.

Kernel/Substitution.hpp
MichaelRawson: Ah, I see - we like such history here. :) No objection then - if `template<bool higherOrder=true> class Substitution` works then consider that, but nbd either way.

mezpusz: Sadly this doesn't work without `<>` (or without `typename` in front in some cases).

Saturation/SaturationAlgorithm.cpp
quickbeam123: If instantiation still is what I remember it used to be, we should definitely consider deleting it. 

It was a giant heuristic hack used to solve a few easy but then-for-us elusive problems in TPTP ARI; in times even before we had theory instantiation, let alone ALASCA.

quickbeam123 (approved):

PR #857, opened 2026-05-19
Moving to C++20

mezpusz:
Compiled on my Mac and our Linux cluster, please try on your own machines too. There was some warning only on Linux and release referring to "array out of bounds" in one of the uses of `Allocator`.

I can regression test it also if all else looks good.


quickbeam123:
My Makefile paths are happy, both on Mac and Linux.

(With MacOs 26, it seems gcc is becoming less and less supported, so I am explicitly using clang there.)

quickbeam123 (approved):

MichaelRawson (approved):
All OK here (Debian trixie, Clang 19) - thanks for doing the work upstream in VIRAS as well.

I didn't get any warnings about `Allocator` - what do you get? I am perpetually suspicious of that thing, even after cutting it back.

mezpusz:
[c++20-allocator-warnings.txt](https://github.com/user-attachments/files/28144719/c%2B%2B20-allocator-warnings.txt)

These are all warnings (maybe overlapping), rooted in `ExtensionalityResolution` as far as I can see. The strange thing is that they only appear in release mode. My clang version is `Debian clang version 19.1.7 (3+b1)`. Let me know if you make something out of this.

MichaelRawson:
Weird, I have the exact same version of clang, but no warning in release mode. The error also seems weird, it's like the compiler hallucinates some type information somewhere - `unsigned char [4]` is the wrong type for `SmartPtr::RefCounter` and the error would be wrong even if it were. Let's merge and see if it persists somehow?

PR #861, opened 2026-06-04
Sort out FOOL paramodulation and Cases[Simp]

mezpusz:
There is an axiomatization and multiple inferences for handling Booleans, and they are kind of a mess. The goal of the present PR is uniformize the inferences, so that we can decide later which ones are still useful. My suggestion would be to eventually remove `FOOLParamodulation`, but we can also probably very easily implement it in `Cases` (basically taking the first result instead of all). In particular, the following changes are included:
1. FOOL paramodulation was fixed for polymorphism, and now handles the higher-order case.
2. Cases[Simp] handles now the first-order case too and the two inferences were merged as much as possible.
3. Cases also inferred a lot of confluent but unnecessary intermediate clauses, this was ruled out.
4. All inferences and the exhaustiveness axiom are turned on properly when needed. This had an unforseeable side effect, namely that with the exhaustiveness axiom we had `hasBoolVar()==true` in the `Property` object, which in turn toggles some additional inferences like primitive instantiation. Due to this, now with the missing (but unnecessary axiom) we seemingly lose some problems when Cases is turned on.

Detailed comparisons with master are to follow.

Cases FOL:
Discount master 10254 (0) vs branch 10260 (6)
Otter master 10354 (0) vs branch 10361 (7)

Cases HOL Discount:
master unsat: 2235 (13) sat: 1 (0) cputime: 8817.91 s instructions: 44196749 Mi memory: 118099.22 MB
branch unsat: 2231 (9) sat: 14 (13) cputime: 8820.62 s instructions: 43948531 Mi memory: 118149.41 MB

Cases HOL Otter:
master unsat: 2202 (20) sat: 1 (0) cputime: 9240.33 s instructions: 44841767 Mi memory: 141257.77 MB
branch unsat: 2200 (18) sat: 14 (13) cputime: 9239.07 s instructions: 44603251 Mi memory: 143950.84 MB

FOOLP HOL Discount:
master unsat: 2200 (37) sat: 14 (0) cputime: 8513.03 s instructions: 42983868 Mi memory: 118078.84 MB
branch unsat: 2183 (20) sat: 14 (0) cputime: 8735.23 s instructions: 43785102 Mi memory: 118122.17 MB

FOOLP HOL Otter:
master unsat: 2163 (28) sat: 14 (0) cputime: 8878.40 s instructions: 43640601 Mi memory: 144348.84 MB
branch unsat: 2151 (16) sat: 14 (0) cputime: 9175.95 s instructions: 44560333 Mi memory: 143307.99 MB


Shell/Options.cpp
```diff
@@ -2091,23 +2091,14 @@ void Options::init()
...
-    _cases.onlyUsefulWith(_casesSimp.is(equal(false)));
-    _lookup.insert(&_cases);
-    // potentially could be useful for FOOL, so am not adding the HOL constraint
-    _cases.tag(OptionTag::HIGHER_ORDER);
+    _boolCases = ChoiceOptionValue<BoolCases>("bool_cases","bc",BoolCases::AXIOM,{"axiom","generating","simplifying"});
+    _boolCases.description=
+      "Sets between three ways of dealing with the FOOL exhaustiveness axiom X0 = $true | X0 = $false:\n"
+      "(1) axiom: adds the axiom,\n"
+      "(2) generating: uses FOOL paramodulation as a generating inference rule,\n"
+      "(3) simplifying: a simplifying variant of the FOOL paramodulation inference rule.";
```
MichaelRawson, line 2099: Would "off" be a reasonable value here?

Inferences/HOL/Cases.cpp
```diff
@@ -26,47 +27,94 @@ namespace Inferences {
...
+  } else {
+    return Clause::fromStack(*resLits, GeneratingInference1(InferenceRule::BOOL_CASES, premise));
+  }
 }
 
-ClauseIterator Cases::generateClauses(Clause* premise)
+template<bool higherOrder>
+auto casesFilterFn = [](Term* t) {
+  if constexpr (!higherOrder) {
+    // TODO consider using iterators that only return booleans
```
MichaelRawson, line 62: Wouldn't it indeed be easier to use e.g. `BooleanSubtermIt` as you do for the HOL case below?

Inferences/HOL/Cases.cpp
```diff
@@ -26,47 +27,94 @@ namespace Inferences {
...
 
-  // Copy the literals from the premise except for the one at `literalPosition`,
-  // that has the occurrence of `booleanTerm` replaced with false
+  // Copy the literals from the premise except for `lit`,
+  // that has the occurrence of `t` replaced with `troo` or `fols`
   for (Literal* curr : iterTraits(premise->iterLits())) {
-    resLits->push( curr != lit 
-        ? curr
-        : EqHelper::replace(curr, t, troo));
+    resLits->push(EqHelper::replace(curr, TermList(t), replaceWithTroo ? troo : fols));
```
MichaelRawson, line 46: Would it be cleaner to pass the value of `replaceWithTroo ? troo : fols` directly to this function? Fairly minor though so nbd.

MichaelRawson (approved):

Shell/Options.cpp
mezpusz: We are hiding all this behind the `hasFOOL()==true` condition, but of course, we could also consider disabling it completely. I honestly cannot tell which one makes more sense, what's you opinion?

Inferences/HOL/Cases.cpp
mezpusz: Good idea, but as far as I see `BooleanSubtermIt` does not handle the non-applicative case, so there might be problems in the FO case.

Shell/Options.cpp
MichaelRawson: That seems fine - what I was thinking about is the case where we do have FOOL, but we don't need to do any reasoning about terms being Boolean. This may not actually happen in practice, of course.

Inferences/HOL/Cases.cpp
MichaelRawson: Got it. OK, we could consider patching `BooleanSubtermIt` to work properly in the future, but that seems like another PR.

mezpusz:
Turns out to be much harder to refactor them into one inference due to small differences in side conditions. Closing for now.

PR #864, opened 2026-06-10
Ignore Unrecognized Logic in SMTCOMP porfolio mode

joe-hauns:
Currently when using `--portfolio smtcomp` Vampire won't attempt solving problems where the smtlib logic is not set, or set to something vampire doesn't know (e.g. `(set-logic HORN)`).
This PR fixes this when vampire is supplied with the flag `--ignore_unrecognized_logic on`.

(Note that the flag already had an effect before in non-portfolio modes, but now it's also working in the smtcomp portfolios)

(The PR also contains the changes from #858 and #859, so it should only be merged after the other two.)


MichaelRawson:
Tag me here when ready!

mezpusz (approved):

MichaelRawson (approved):

PR #862, opened 2026-06-09
Preparation for CASC HOL 2026 #1

mezpusz:
FIrst batch of fixes and improvements in preparation for CASC HOL 2026.

1. Function definition introduction was adapted for HOL.
2. Backward simplifying indices (i.e. essentially `SubstitutionTree::FastInstancesIterator`) were adapted for HOL in somewhat inefficient way to avoid variable capture/loose DB indices.
3. A bug related to proxification (e.g. with `--cnf_on_the_fly lazy_gen`) was fixed, because proxification cannot handle type quantifiers. We simply don't proxify, but clausify such clauses instead of crashing.
4. A safeguard was added in term sharing to ensure that everything is appified when doing HOL
5. HOL sampler was adapted (for example, removing FMB options for now).

Multiple rounds of debug sampling TPTP did not show any related new errors (1 related to z3, 2 related to function definition inlining showed up, but these are in FOL).

Regression tests (run 0 is master, run 1 is this branch):

**Testing 1 (`-fdi 5`):**

Discount FOL
run 0 unsat: 5099 (0) sat: 878 (1) cputime: 67963.24 s instructions: 300768768 Mi memory: 2402548.99 MB
run 1 unsat: 5099 (0) sat: 877 (0) cputime: 67951.24 s instructions: 300764373 Mi memory: 2403903.48 MB

Otter FOL
run 0 unsat: 5448 (0) sat: 873 (0) cputime: 60802.40 s instructions: 293972229 Mi memory: 1592099.27 MB
run 1 unsat: 5448 (0) sat: 873 (0) cputime: 60791.79 s instructions: 293967069 Mi memory: 1592198.68 MB

Discount HOL
run 0 unsat: 1252 (15) sat: 1 (0) cputime: 14236.33 s instructions: 68655709 Mi memory: 301840.90 MB
run 1 unsat: 1690 (453) sat: 1 (0) cputime: 14082.85 s instructions: 59927145 Mi memory: 212328.85 MB

Otter HOL
run 0 unsat: 1373 (25) sat: 1 (0) cputime: 14229.10 s instructions: 66266612 Mi memory: 319064.22 MB
run 1 unsat: 1832 (484) sat: 1 (0) cputime: 13653.15 s instructions: 57393162 Mi memory: 246531.99 MB

**Testing 2 (`-bd all -bs on -bsr on -bsd on`):**

Discount FOL
[...]

PR #865, opened 2026-06-11
Basic HOL unification for Superposition and Equality Resolution

mezpusz:
The main improvement in this PR is functionality to iterate HOL unifiers (in a basic way, using Huet's preunification). There is some wrapper code that takes an abstracting unifier, takes the potentially HOL unification constraints from it, and does this iteration to a certain depth, before returning abstracted unification constraints.

To achieve this, we also have to abstract these parts already in UWA, hence a new HOL UWA variant was added (this is used by default with `uwa=auto` if the problem is HO).

The main change in "core" Vampire is the adaption of Equality Resolution, where things had to be shuffled around to deal with the potentially multiple unifiers. Here also the UWA functionality was slightly changed.

Regression tests results will follow as soon as #862 is merged into this branch.


mezpusz:
Regression tests for FOL showed no difference.

Regreesion for HOL over TPTP:

Discount

master unsat: 2237 (175) sat: 1 (0) cputime: 8818.44 s instructions: 44166631 Mi memory: 117445.72 MB
branch unsat: 2147 (85) sat: 1 (0) cputime: 9568.07 s instructions: 49203799 Mi memory: 136360.63 MB

Otter

master unsat: 2203 (161) sat: 1 (0) cputime: 9281.81 s instructions: 44845447 Mi memory: 140804.28 MB
branch unsat: 2126 (84) sat: 1 (0) cputime: 9881.58 s instructions: 49617218 Mi memory: 180818.93 MB

This is of course not very informative as we are comparing two different unification approaches, but it shows that the new method is not super slow and gives a good baseline for future improvement.

Kernel/HOL/HOL.hpp
```diff
@@ -65,6 +65,8 @@ inline bool canHeadReduce(const TermList& head, const TermStack& args) {
   return head.isLambdaTerm() && args.isNonEmpty();
 }
 
+void normaliseLambdaPrefixes(TermList& t1, TermList& t2);
```
hetzenmat, line 70: Could you add a short comment what this function does?

I think we should do (and I should have done) this for most of the functions in this file for future reference.

Kernel/HOL/Unifier.cpp
```diff
@@ -0,0 +1,333 @@
...
+      }
+
+      Stack<UnificationNode*> res;
+      for (const auto& [binding,inf] : bindings) {
+        DEBUG("binding ", flexTerm.head(), " ", binding);
+        res.push(new UnificationNode(*this, flexTerm.head().var(), binding));
+      }
+      return some<Stack<UnificationNode*>>(res);
+    }
+    // else flex-flex, which we ignore
```
hetzenmat, line 152: I think a debug assertion would be nice to enforce this when testing.

Saturation/SaturationAlgorithm.cpp
```diff
@@ -1460,7 +1464,7 @@ SaturationAlgorithm *SaturationAlgorithm::createFromOptions(Problem& prb, const
     if(prb.hasFOOL() && opt.booleanEqTrick()){
       gie->addFront(new BoolEqToDiseq(*res));
     }
-    if(true/* !opt.higherOrderUnifDepth() && !opt.applicativeUnify() */){
+    if(!opt.higherOrderUnifDepth()){
```
hetzenmat: I think writing it as `opt.higherOrderUnifDepth() == 0` makes the intention more readable.

hetzenmat (changes requested):
Looks very reasonable to me

Kernel/RobSubstitution.cpp
```diff
@@ -521,6 +523,28 @@ TermList RobSubstitution::apply(TermList trm, int index) const
     .apply(AutoDerefTermSpec(TermSpec(trm, index), this));
 }
 
+TermList RobSubstitution::applyGlue(TermList trm, int index)
+{
+  return BottomUpEvaluation<AutoDerefTermSpec, TermList>()
+    .function([&](auto const& orig, TermList* args) -> TermList {
+        if (orig.term.isVar()) {
+          // TODO: find out if this is reasonable to assert, probably not
+          // ASS(!orig.term.varSpec().special());
```
mezpusz, line 532: @joe-hauns this assertion was meant to prevent this method being called in non-leaves of the substitution tree, but the assertion was hit a couple of times, so I'm not sure this is true, and even in leaves special vars can be the final "deref" result of a variable. Can you confirm this?

Kernel/HOL/Unifier.hpp
```diff
@@ -0,0 +1,104 @@
...
+  {
+    TermList _lhs;
+    TermList _rhs;
+    TermList _sort;
+    TermList _lhead;
+    TermList _rhead;
+
+    Constraint(TermList lhs, TermList rhs, TermList sort);
+
+    inline bool flexFlex() const { return _lhead.isVar() && _rhead.isVar(); }
```
hetzenmat: The `inline` specifiers are redundant for functions defined entirely inside a class or function definition ([see](https://clang.llvm.org/extra/clang-tidy/checks/readability/redundant-inline-specifier.html)).

Kernel/HOL/Unifier.hpp
```diff
@@ -0,0 +1,104 @@
...
+  unsigned _freshVar;
+};
+
+/**
+ * Wrapper around an abstracting unifier that may contain HO unification constraints.
+ * It takes an abstracting unifier and tries to solves its constraints, or instantiates
+ * variables until a certain limit (this is set by the depth parameter, which is the
+ * number of variables instantiated on a given branch in the HOL unification tree).
+ */
+class AbstractingWrapper
```
hetzenmat, line 80: It is good practice to add `override` to `~AbstractingWrapper`, `hasNext`, and `next` in case the interface of the base class changes.

Kernel/HOL/Unifier.hpp
```diff
@@ -0,0 +1,104 @@
...
+    TermList _lhead;
+    TermList _rhead;
+
+    Constraint(TermList lhs, TermList rhs, TermList sort);
+
+    inline bool flexFlex() const { return _lhead.isVar() && _rhead.isVar(); }
+    inline bool rigidRigid() const { return _lhead.isTerm() && _rhead.isTerm(); }
+    inline bool flexRigid() const { return !flexFlex() && !rigidRigid(); }
+
+    bool derefHead(TermList& head, TermList& side, const Substitution& subs);
```
hetzenmat: `derefHead` can be `static`

hetzenmat (approved):
Thanks, looks really good

Kernel/UnificationWithAbstraction.hpp
```diff
@@ -150,14 +147,13 @@ class AbstractingUnifier
...
   Recycled<UnificationConstraintStack> _constr;
   Option<BacktrackData&> _bd;
-  AbstractionOracle _uwa;
 
   friend class RobSubstitution;
-  AbstractingUnifier(AbstractionOracle uwa) : _subs(), _constr(), _bd(), _uwa(uwa) { }
 public:
+  AbstractionOracle _uwa;
+
+  AbstractingUnifier(AbstractionOracle uwa) : _subs(), _constr(), _bd(), _uwa(uwa) { }
```
MichaelRawson: Since you're changing this anyway, `_subs()`, `_const()`, and `_bd()` are all pointless IIRC.

Inferences/EqualityResolution.cpp
```diff
@@ -37,112 +38,79 @@ using namespace Kernel;
...
-      _salg.getOptions().literalMaximalityAftercheck() && _salg.getLiteralSelector().isBGComplete(),
-      &_salg.getOrdering()));
+      auto [lhs, rhs] = lit->eqArgs();
 
-  auto it4 = getFilteredIterator(it3,NonzeroFn());
+      // We only care about non-trivial constraints where the top-symbol of the two literals are the same
+      // and therefore a constraint can be created between arguments
+      if(lhs.isTerm() && rhs.isTerm() &&
+        lhs.term()->functor() != rhs.term()->functor() && _salg.getOptions().unificationWithAbstraction() != Shell::Options::UnificationWithAbstraction::HOL){
+        abstractionOracle = AbstractionOracle(Shell::Options::UnificationWithAbstraction::OFF);
```
MichaelRawson, line 100: Seems like the `AbstractionOracle` (and maybe `AbstractingUnifier`) is created/destroyed very often. Should it be a member, maybe? This only-if-top-symbol-matches logic could stay, of course.

Indexing/TermSubstitutionTree.hpp
```diff
@@ -85,6 +89,13 @@ class TermSubstitutionTree
   VirtualIterator<QueryRes<AbstractingUnifier*, LeafData>> getUwa(TypedTermList t, Options::UnificationWithAbstraction uwa, bool fixedPointIteration) final
   { return pvi(getResultIterator<typename SubstitutionTree::template Iterator<RetrievalAlgorithms::UnificationWithAbstraction<AbstractingUnifier, RetrievalAlgorithms::DefaultVarBanks>>>(t, /* retrieveSubstitutions */ true, AbstractingUnifier::empty(AbstractionOracle(uwa)), AbstractionOracle(uwa), fixedPointIteration)); }
 
+  VirtualIterator<QueryRes<AbstractingUnifier*, LeafData>> getUwaHOL(TypedTermList t, Options::UnificationWithAbstraction uwa, bool fixedPointIteration, unsigned hoUnifDepth) override
+  {
+    return pvi(iterTraits(getUwa(t, uwa, fixedPointIteration))
+      .flatMap([hoUnifDepth](QueryRes<AbstractingUnifier*, LeafData> qr) { return pvi(pushPairIntoRightIterator(qr, vi(new HOL::AbstractingWrapper(qr.unifier, hoUnifDepth)))); })
```
MichaelRawson, line 98: In my head `getUwa` and `getUwaHOL` should be the same, with a different UWA oracle. What am I forgetting? And/or would it make sense to unify all the UWAing so that we can mix arithmetic and HOL and get something sensible anyway?

Not hugely important for now, but maybe something to clear up with a comment somewhere.

MichaelRawson (approved):
Cool! That was less painful than I thought. Some confusion from me but I think this is mostly my ignorance rather than a problem.

Inferences/EqualityResolution.cpp
mezpusz: Isn't `AbstractionOracle` a POD class? It only has a single member of enum type, so I'm guessing initializing locally is trivial (and anyways we have to reinitialize it for every literal). But you're the expert, maybe there is some overhead with it.

Indexing/TermSubstitutionTree.hpp
mezpusz: It's a bit confusing, you're right. The idea is that `getUwaHOL` should be only used on higher-order problems as it has some overhead that we should eventually iron out; but the UWA oracle is not necessarily HOL, as we might still want to use the FO unification.

Mixing it with other UWA oracles might be even bad, as the HOL unifier wrapper can (I think) hijack some of the abstracted equations but mixing arithmetic with HOL is not possible currently. In the future, maybe separating the HOL parts completely in abstraction could be beneficial, to allow for some more layered abstraction..

So definitely worth a comment, should I make any other changes in your opinion?

Inferences/EqualityResolution.cpp
MichaelRawson: Ahah - no, it just seemed like it should be heavier, I didn't check. Fine by me then.

Indexing/TermSubstitutionTree.hpp
MichaelRawson: No, this seems like a "research problem" for the future. Definitely fine for now, thanks for the comment!

Saturation/SaturationAlgorithm.cpp
```diff
@@ -1460,7 +1464,7 @@ SaturationAlgorithm *SaturationAlgorithm::createFromOptions(Problem& prb, const
     if(prb.hasFOOL() && opt.booleanEqTrick()){
       gie->addFront(new BoolEqToDiseq(*res));
     }
-    if(true/* !opt.higherOrderUnifDepth() && !opt.applicativeUnify() */){
+    if(opt.higherOrderUnifDepth() == 0){
```
quickbeam123: Maybe let's discuss this place before we merge. 

We have `"hol_unif_depth","hud",2` by default, yet it does nothing (I assume) unless we set `uwa=hol`. 

So shouldn't we have a different condition here to guard `ImitateProject`?

mezpusz: Good point! How about `opt.uwa() != HOL || opt.higherOrderUnifDepth() == 0`?

quickbeam123: Yes, please!

PR #868, opened 2026-06-15
removed unstable alasca options from samplers

joe-hauns:
I realized that the arithmetic sampler files contained two options that were never meant to be there, as they are not well-enough tested yet, and also might be removed in the future. Thus i commented them out, so no portfolios rely on unstable code.


quickbeam123 (approved):

quickbeam123:
Thank you, Joe, I remember I was trying to add as much as possible, but we converged to a smaller set in the end. It's good to have this also documented in the samplers...

PR #869, opened 2026-06-17
Fix divisibility

joe-hauns:
As divisibility appears to never have been benchmarked really, i ran into some more bugs with it when trying out some benchmarks that use it. 
This PR fixes SMT-LIB format proof-checker output for divisibility, as well as an issue in theory instantiation that lead to an unsoundness in some cases.

### DIFF
--- a/Inferences/TheoryInstAndSimp.cpp
+++ b/Inferences/TheoryInstAndSimp.cpp
@@ -444,7 +444,7 @@ void TheoryInstAndSimp::ConstantCache::reset()
 
 Term* TheoryInstAndSimp::ConstantCache::freshConstant(SortId sort) 
 { 
-  auto& cache = _inner.getOrInit(sort, [](){ 
+  auto& cache = _inner.getOrInit(sort, [&](){
       DEBUG("new constant cache for sort ", _inner.size());
       return SortedConstantCache(); 
     });
@@ -967,9 +967,9 @@ bool TheoryInstAndSimp::isTheoryLemma(Clause* cl, bool& couldNotCheck) {
   return !solutions.next().sat;
 }
 
-} // namespace Inferences
-
 std::ostream& operator<<(std::ostream& out, Inferences::Solution const& self)
 { return out << "Solution(" << (self.sat ? "sat" : "unsat") << ", " << self.subst << ")"; }
+
+} // namespace Inferences
 
 #endif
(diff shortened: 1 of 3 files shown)


Inferences/TheoryInstAndSimp.cpp
```diff
@@ -444,7 +444,7 @@ void TheoryInstAndSimp::ConstantCache::reset()
 
 Term* TheoryInstAndSimp::ConstantCache::freshConstant(SortId sort) 
 { 
-  auto& cache = _inner.getOrInit(sort, [](){ 
+  auto& cache = _inner.getOrInit(sort, [&](){
```
quickbeam123, line 447: Is this change (the "&") just cosmetics or part of the actual fix?

I am simply curious what's going on here...

SAT/Z3Interfacing.cpp
```diff
@@ -1641,11 +1641,8 @@ z3::expr Z3Interfacing::getRepresentation(Term* trm)
             switch(interp){
             // Numerical operations
             case Theory::INT_DIVIDES:
-              {
-              auto k = getNamingConstantFor(toEval, _context->int_sort());
```
quickbeam123, line 1645: What was this trying to do? Sorry, again just very curious!

quickbeam123 (approved):

SAT/Z3Interfacing.cpp
joe-hauns: So it was trying to say `a` divides `b` iff there is some `k` such that `a * k = b`, b introducing a skolemish constant `k`. But actually that's not correct in every context because the skolem `k` should actually be a function of the outer universal quantifiers or so.
Was a bit of not too well thought-out code that I wrote a while and apparently we never had any problems using divisibility which triggered the issue.

Inferences/TheoryInstAndSimp.cpp
joe-hauns: So when `DEBUG(...)` is enabled 
i.e.
```#define DEBUG(...) // DBG(__VA_ARGS__)```
is changed to
```#define DEBUG(...) DBG(__VA_ARGS__)```

we're capturing `this->_inner` in the lambda, and the code won't compile. That's why I put the `&` here.

quickbeam123: Right! I missed that `_inner` is part of the lambda

quickbeam123 (approved):

mezpusz (approved):

PR #870, opened 2026-06-18
fix the cnf-claim path

quickbeam123:
The claims feature was silently unsound for inputs in cnf, e.g.:
```
cnf(cl1,claim,p(X)).
```
would become
```
![X]: (claim <=> p(X)
```
in Vampire, which is wrong.

This PR fixes the problem by requesting a formula closure before "claimifying".

(The needed functionality was factored our from a piece of, likely dead, code in `Recrify.cpp` to prevent code duplication.)


MichaelRawson (approved):
Good catch. E apparently treats input clauses as weirdly-spelled formulae now, we could consider this to avoid future bugs of this form.

PR #873, opened 2026-07-01
Fix subst tree generalizations for hol

mezpusz:
Fixes bug in substitution tree fast generalizations binder:
- Bindings were erroneously inserted when bound term contains a loose DB index.
- This could result in incorrect results later, including unbound variables bound to random terms with loose DB indices.
- This also fixes SAT-subsumption+resolution for HOL.

Extra changes: one warning was silenced, one error in HOL sampler fixed.

### DIFF
--- a/Indexing/SubstitutionTree_FastGen.hpp
+++ b/Indexing/SubstitutionTree_FastGen.hpp
@@ -49,14 +49,11 @@ struct SubstitutionTree<LeafData_>::GenMatcher::Binder
    */
   bool bind(unsigned var, TermList term)
   {
-    if(var > _maxVar) {
+    if (var > _maxVar || term.containsLooseDBIndex()) {
       return false;
     }
     TermList* aux;
     if(_parent->_bindings.getValuePtr(var,aux,term)) {
-      if (term.containsLooseDBIndex()) {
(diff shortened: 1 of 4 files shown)


Indexing/SubstitutionTree_FastGen.hpp
```diff
@@ -49,14 +49,11 @@ struct SubstitutionTree<LeafData_>::GenMatcher::Binder
    */
   bool bind(unsigned var, TermList term)
   {
-    if(var > _maxVar) {
+    if (var > _maxVar || term.containsLooseDBIndex()) {
```
MichaelRawson, line 52: Was the bug here that `getValuePtr` was called regardless of `containsLooseDBIndex`?

MichaelRawson (approved):

Indexing/SubstitutionTree_FastGen.hpp
mezpusz: Exactly! We could remove the element also when `term.containsLooseDBIndex()==true` but I think experiments so far indicate that this check is not so expensive to unnecessarily complicate the code.

quickbeam123 (approved):

PR #874, opened 2026-07-01
Hol functional extensionality

mezpusz:
Adding functional extensionality to HOL, that can be enabled by `func_ext=abstraction`. The option was already there but didn't do anything until now. I switched the default option to `off` for now, as abstraction gives slightly fewer proofs on THX problems. First-order benchmarks are unaffected by the change.

We should later probably also do this for variable sorts, as well as Booleans (= Boolean extensionality), but I wasn't sure if those cases can be safely covered currently.

THX with `func_ext=abstraction`:

Discount:
master unsat: 2147 (77) sat: 1 (0) cputime: 9545.21 s instructions: 49204782 Mi memory: 136339.48 MB
branch unsat: 2136 (66) sat: 1 (0) cputime: 9729.12 s instructions: 49982726 Mi memory: 139925.18 MB

Otter:
master unsat: 2126 (70) sat: 1 (0) cputime: 9863.07 s instructions: 49617558 Mi memory: 180796.81 MB
branch unsat: 2121 (65) sat: 1 (0) cputime: 10025.58 s instructions: 50147647 Mi memory: 182001.22 MB

Note that ~30 problems switched from `INCOMPLETE` to being solved.

### DIFF
--- a/Kernel/UnificationWithAbstraction.cpp
+++ b/Kernel/UnificationWithAbstraction.cpp
@@ -237,11 +237,6 @@ bool AbstractionOracle::canAbstract(AbstractingUnifier* au, TermSpec const& t1, 
 
 Option<TermSpec> appHead(AbstractingUnifier* au, TermSpec t)
 {
-  // contains the @ arguments, innermost on top
-  Stack<TermSpec> args;
-  // contains the substituted arguments inside lambdas, the innermost on top
-  Stack<TermSpec> subst;
-
   for (;;) {
     t = au->subs().derefBound(t);
     // we don't deal with lambdas yet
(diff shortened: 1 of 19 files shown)


quickbeam123 (approved):

Kernel/UnificationWithAbstraction.cpp
```diff
@@ -295,8 +298,34 @@ Option<AbstractionOracle::AbstractionResult> hol(AbstractingUnifier* au, TermSpe
...
+    }
+  }
+
+  // otherwise decompose application normally
+  Stack<UnificationConstraint> unifs;
+  DEBUG_UNIFY(0, "decomposing ", t1, " (", t1.sort(), ") ", t2, " (", t2.sort(), ")");
+  while (t1.term.isApplication()) {
+    // It can happen that the substitution tree unifies terms before their sorts
+    // due to reordering. In this case, we may get terms with the same head but
+    // different number of applications, hence the check below and after the loop.
```
MichaelRawson, line 317: This sounds chaotic. Is there any hope of this being different in future?

MichaelRawson (approved):
Cool!

Kernel/UnificationWithAbstraction.cpp
mezpusz: I agree. I think adapting the reordering wouldn't be so difficult, it's just not very well documented so I have to understand it first.

MichaelRawson: Oh, great! I read this as an inevitability, phew.

PR #876, opened 2026-07-06
report correct file paths for axioms

MichaelRawson:
With `--output_axiom_names on -p tptp` we currently just report the file passed on the command line, not necessarily where the axiom was found (!). This can be confusing when multiple axioms called the same thing in different files end up in Vampire's output.

The code is not beautiful, but the whole thing needs a re-work at some point anyway, strings-that-should-be-ostreams everywhere.

### DIFF
--- a/Parse/TPTP.cpp
+++ b/Parse/TPTP.cpp
@@ -60,7 +60,7 @@ static VSList* zipVarsSorts(VList* vars, SList* sorts) {
 
 #define DEBUG_SHOW_UNITS 0
 #define DEBUG_SOURCE 0
-DHMap<unsigned, std::string> TPTP::_axiomNames;
+DHMap<unsigned, std::pair<std::string, std::filesystem::path>> TPTP::_axiomNames;
 DHMap<unsigned, Map<unsigned,std::string>> TPTP::_questionVariableNames;
 
 //Numbers chosen to avoid clashing with connectives.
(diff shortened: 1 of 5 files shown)


Parse/TPTP.cpp
```diff
@@ -4848,22 +4848,18 @@ unsigned TPTP::addUninterpretedConstant(const std::string& name, bool& added)
...
  */
-bool TPTP::findAxiomName(const Unit* unit, std::string& result)
+bool TPTP::findAxiomName(const Unit* unit, std::string& name, std::filesystem::path &path)
 {
-  return _axiomNames.find(unit->number(), result);
+  std::pair<std::string, std::filesystem::path> found;
+  if(!_axiomNames.find(unit->number(), found))
+    return false;
+  name = std::move(found.first);
+  path = std::move(found.second);
```
quickbeam123, line 4861: Move is fine, because you already "copied" into `found`? 

Could the compiler figure this out on its own?

quickbeam123 (approved):

Parse/TPTP.cpp
MichaelRawson: Right - maybe in principle the compiler could figure this out, but I'm not sure it would. I'd be fine with a `move`-less (stationary?) version if you prefer, this is not a hot part of the code.

quickbeam123: No, don't change anything. I was just curious how you think about these...

MichaelRawson: Thinking: OK, I should use `_axiomNames.find` because that's what was there before. That achieves the necessary copy. But I don't need to copy it again into the out-parameters, so let's just move it out of the locals `found.first` and `found.second`, these will definitely be dead on return.

As for whether the compiler will optimise it...[argh, it can](https://en.cppreference.com/cpp/language/copy_elision), but it's not a mandatory optimisation.

easychair:
Sorry guys, I am playing with SPF records, ignore this email!

Andrei

PR #877, opened 2026-07-14
A working branch for the casc submission

quickbeam123:


MichaelRawson (approved):
Nice. I don't think Geoff likes UTF-8 (as a good Anglo-Saxon) but we can try it.

starexec/bin/starexec_run_casc_sat
```diff
@@ -1,3 +1,3 @@
 #!/bin/sh
 
-exec ./vampire-ho --mode casc --intent sat -m 16384 --cores 7 -t $STAREXEC_WALLCLOCK_LIMIT $1
+exec ./vampire-ho --input_syntax tptp --output_axiom_names on --mode casc --intent sat -m 16384 --cores 7 -t $STAREXEC_WALLCLOCK_LIMIT $1
```
mezpusz, line 3: Is it okay that we use the HOL variant here?

mezpusz (approved):

starexec/bin/starexec_run_casc_sat
MichaelRawson: Yep - I agree it's weird but it makes sense this time around as Martin hasn't been doing neural things for FNT.

quickbeam123: Yes, we avoid using the giant 0.5GB vampire for this. Although it could do it too.

PR #880, opened 2026-07-16
Optimization: Share opposite-polarity literals in clause code trees

Synrom:
This PR optimizes clause code tree subsumption resolution by sharing literal information across polarities. Instead of storing separate positive and negative `LitInfo` entries for each literal, matching records whether a result came from the opposite polarity and handles that during matcher execution.

The implementation preserves subsumption priority over subsumption resolution by keeping non-opposite matches first in `ILStruct` and making `ClauseMatcher::matchGlobalVars` try non-opposite matches before falling back to opposite-polarity matches.

This optimization is currently not compatible with `CHECK_GROUND_TERM=1`.


Indexing/ClauseCodeTree.cpp
```diff
@@ -459,11 +475,13 @@ void ClauseCodeTree<higherOrder>::LiteralMatcher::recordMatch()
     //no need to record matches which we already know will not lead to anything
     return;
   }
-  if(!ils->matchCnt && Base::linfos[Base::curLInfo].opposite) {
+  if(!ils->matchCnt && opposite) {
     //if we're matching opposite matches, we have already tried all non-opposite ones
     ils->noNonOppositeMatches=true;
+  } else if (ils->noNonOppositeMatches && !opposite) {
```
mezpusz, line 480: Can you explain why this is needed?

Indexing/ClauseCodeTree.cpp
```diff
@@ -606,7 +619,8 @@ Clause* ClauseCodeTree<higherOrder>::ClauseMatcher::next(int& resolvedQueryLit)
       }
 
       bool seekOnlySuccess=lms.size()==query->length();
-      enterLiteral(newLitEntry, seekOnlySuccess);
+      bool canEnterOpposites=sres && sresLiteral == sresNoLiteral;
```
mezpusz: Could you please parenthesize this to make it clearer?

Kernel/FlatTerm.hpp
```diff
@@ -61,6 +61,9 @@ class FlatTerm
     inline bool isVar(unsigned num) const { return isVar() && _number()==num; }
     inline bool isFun() const { return _tag()==FUN || _tag()==FUN_UNEXPANDED; }
     inline bool isFun(unsigned num) const { return isFun() && _number()==num; }
+    /* Litearl headers encode polarity in their lowest bit, so xor-ing with 1
```
mezpusz: "Litearl" -> "Literal"

Kernel/FlatTerm.hpp
```diff
@@ -61,6 +61,9 @@ class FlatTerm
...
+     * gives the opposite (negated) predicate. Only meaningful for literal predicates. */
+    inline bool isOppositeFun(unsigned num) const { return isFun() && (_number()^1)==num; }
```
mezpusz, line 66: Would it make sense to assert that this actually gives the opposite literal, so if we ever change it, it automatically triggers? Maybe not here, since this is called very often.

Indexing/CodeTree.hpp
```diff
@@ -332,26 +345,48 @@ class CodeTree
...
+     */
+    class MarkedOp
+    {
+      static_assert(alignof(CodeOp) >= 2, "CodeOp must be at least 2-byte aligned so its lowest bit is free for the mark");
+      CodeOp* raw;
+    public:
+      MarkedOp(CodeOp* op, bool mark) :
+        raw(reinterpret_cast<CodeOp*>(reinterpret_cast<std::uintptr_t>(op) | mark)) {}
+      
+      inline bool getMark() const {
```
mezpusz: Could you use the `Lib/BitUtils.hpp` functionality for this? See `CodeOp` and `Term`.

Indexing/CodeTree.hpp
```diff
@@ -74,13 +74,21 @@ class CodeTree
 
   struct MatchInfo
   {
-    /** Index of the matched LitInfo in the EContext */
-    unsigned liIndex;
+    inline unsigned getLiIndex() const { return markedLiIndex & ~leftmost_bit; }
+    inline TermList* getBindings() { return &bindings[0]; }
```
mezpusz, line 79: Why add this function?

Indexing/CodeTree.hpp
```diff
@@ -74,13 +74,21 @@ class CodeTree
...
+    inline bool opposite() const { return markedLiIndex & leftmost_bit; }
+
+  private:
+    /** Index of the matched LitInfo in the EContext, with the opposite flag packed
+     * into the leftmost bit. Use getLiIndex()/opposite() above rather than reading
+     * this field directly. */
+    unsigned markedLiIndex;
```
mezpusz: Please make members `private` that are unsafe to use directly.

Indexing/CodeTree.cpp
```diff
@@ -277,7 +279,20 @@ void CodeTree::ILStruct::addMatch(unsigned liIndex, DArray<TermList>& bindingArr
   if(!matches[matchCnt]) {
     matches[matchCnt]=MatchInfo::alloc(varCnt);
   }
-  matches[matchCnt]->init(this, liIndex, bindingArray);
+  
+  /* Maintain the invariant that non-opposite matches occupy [0, nonOppositeMatchCnt):
+   * an opposite match is simply appended, while a non-opposite match is swapped into
+   * position nonOppositeMatchCnt before that boundary is advanced. */
+  if(opposite) {
+    matches[matchCnt]->init(this, liIndex, bindingArray, true);
```
mezpusz: Could you please add the name of the last argument commented? Similarly below.

mezpusz: Nice trick with the swapping btw!

Indexing/CodeTree.cpp
```diff
@@ -811,13 +851,33 @@ inline bool CodeTree::Matcher<removing, checkRange, higherOrder>::doCheckGroundT
   return true;
 }
 
+template<bool removing, bool checkRange, bool higherOrder>
+inline typename CodeTree::Matcher<removing, checkRange, higherOrder>::MarkedOp CodeTree::Matcher<removing, checkRange, higherOrder>::markOp(CodeOp *op)
+{
+  return MarkedOp(op, opposite);
```
mezpusz, line 866: Do we really need this bloated wrapping for this function?

Indexing/CodeTree.cpp
```diff
@@ -811,13 +851,33 @@ inline bool CodeTree::Matcher<removing, checkRange, higherOrder>::doCheckGroundT
...
+}
+
 template<bool removing, bool checkRange, bool higherOrder>
 inline bool CodeTree::Matcher<removing, checkRange, higherOrder>::doSearchStruct()
 {
   ASS_EQ(op->_instruction(), SEARCH_STRUCT);
 
   const FlatTerm::Entry* fte=&(*ft)[tp];
-  CodeOp* target=op->getSearchStruct()->getTargetOp(fte);
+  CodeOp* target=op->getSearchStruct()->getTargetOp(fte, false);
```
mezpusz: Please add commented name of argument for 2nd arg here too.

Indexing/CodeTree.cpp
```diff
@@ -811,13 +851,33 @@ inline bool CodeTree::Matcher<removing, checkRange, higherOrder>::doCheckGroundT
...
+  /* look up the branch for the negated predicate, so it
+   * can be tried on backtracking for subsumption resolution */
+  if (canEnterOpposites && tp == 0 && op->getSearchStruct()->kind == SearchStruct::FN_STRUCT) {
+    CodeOp* alt = op->getSearchStruct()->getTargetOp(fte, true);
+    if (alt && target != alt) {
+      /* 'opposite' will be determined by doCheckFun
+       * when this backtracking point is resumed */
+      if constexpr (removing) {
+        btStack.push(BTPointRemoving(tp, MarkedOp(alt, false), RemovingBase::firstsInBlocks->size()));
```
mezpusz: These could probably be `emplace` instead of `push`.

Indexing/CodeTree.cpp
```diff
@@ -811,13 +851,33 @@ inline bool CodeTree::Matcher<removing, checkRange, higherOrder>::doCheckGroundT
...
 inline bool CodeTree::Matcher<removing, checkRange, higherOrder>::doSearchStruct()
 {
   ASS_EQ(op->_instruction(), SEARCH_STRUCT);
 
   const FlatTerm::Entry* fte=&(*ft)[tp];
-  CodeOp* target=op->getSearchStruct()->getTargetOp(fte);
+  CodeOp* target=op->getSearchStruct()->getTargetOp(fte, false);
+  /* look up the branch for the negated predicate, so it
+   * can be tried on backtracking for subsumption resolution */
+  if (canEnterOpposites && tp == 0 && op->getSearchStruct()->kind == SearchStruct::FN_STRUCT) {
```
mezpusz, line 878: Could you please add a note in `execute` where `btStack` is pushed upon that we also push in this function?

mezpusz:
Nice work, once the conflict is resolved I will run some more tests to be sure. Do I get it right that this shouldn't affect anything if `fsr=off`? Maybe that could be an extra check to see that no bugs are introduced.

Indexing/ClauseCodeTree.cpp
Synrom: Because it is longer guaranteed that non-opposite literals are matched before opposite ones.

Before, placing opposite `LitInfo`s behind non-opposite ones guaranteed that.
If we recorded a match that was the first one (`!ils->matchCnt`) and it was opposite, we knew no non-opposite matches would be added later.

That no longer holds, so the additional checks here are needed. I also noticed the old comment is no longer true, so I removed it.

Kernel/FlatTerm.hpp
Synrom: In the method itself, it would be hard. `FlatTerm::changeLiteralPolarity` uses the same trick, so you could add an assert there.
But as far as I can tell, `FlatTerm::changeLiteralPolarity` will not be executed anymore.
Maybe, I can just add a unit test that checks the FlatTerm behavior against the Literal implementation?

Indexing/CodeTree.hpp
Synrom: Done.

Indexing/CodeTree.hpp
Synrom: After adding the opposite mark, you can only access `liIndex`, using the `getLiIndex` method.
I thought it was more canonical if all values of `MatchInfo` would be accessed via methods instead of attributes.
But I can also change it back :)

Indexing/CodeTree.hpp
Synrom: I've integrated `Lib/BitUtils.hpp` here as well now and the fields are private :)

Indexing/CodeTree.cpp
Synrom: I removed it now :)

Indexing/CodeTree.cpp
Synrom: I've added a comment :)

Synrom:
Thank you :) Yes with `fsr=off`, it should be identical. I ran:
```
./vampire  -sa discount -fsr off -al 1000 --random_seed 1 -t 0 ../../problems/condensed_detachment_4.p
```
On master and on `opposite-optimization` and the runtime statistics (clauses created, clauses deleted, ssat_new_components, ...) were exactly identical.

Indexing/CodeTree.hpp
mezpusz: No, I don't mind it, thanks!

Kernel/FlatTerm.hpp
mezpusz: Okay, the unit test would be awesome! Please feel free to remove any unused stuff from `FlatTerm` too.

mezpusz:
> This optimization is currently not compatible with CHECK_GROUND_TERM=1.

Could you please add an `ASSERTION_VIOLATION` and a TODO referring to this PR somewhere so that whenever we set `CHECK_GROUND_TERM=1` we notice this? Otherwise I think removing the feature as well as fixing it for this optimization is unnecessary in this PR.

Kernel/FlatTerm.hpp
Synrom: I've added a unit test now :) should I also remove the now-unused opposite field and related methods from `LitInfo`?
The methods are the only remaining callers of `FlatTerm::changeLiteralPolarity`, so that could go to.

Synrom:
Alright, I've added an assertion plus TODO now.

Indexing/CodeTree.cpp
mezpusz: So this function is now unused, isn't it? 🤔

Indexing/CodeTree.cpp
```diff
@@ -638,10 +671,18 @@ bool CodeTree::Matcher<removing, checkRange, higherOrder>::execute()
 }
 
 template<bool removing, bool checkRange, bool higherOrder>
-void CodeTree::Matcher<removing, checkRange, higherOrder>::init(const CodeTree& tree_, CodeOp* entry_, LitInfo* linfos_, size_t linfoCnt_, Stack<CodeOp*>* firstsInBlocks_)
+void CodeTree::Matcher<removing, checkRange, higherOrder>::init(const CodeTree& tree_, CodeOp* entry_, bool canEnterOpposites_, LitInfo* linfos_, size_t linfoCnt_, Stack<CodeOp*>* firstsInBlocks_)
 {
   tree=&tree_;
+  canEnterOpposites=canEnterOpposites_;
+#if GROUND_TERM_CHECK
+  if (canEnterOpposites) {
```
mezpusz: How about `ASS(!canEnterOpposites)`?

Synrom: Yes, that makes more sense!

Kernel/FlatTerm.hpp
mezpusz: Sorry I forgot about this. Yes, feel free to clean out everything that is not called anymore.

PR #881, opened 2026-07-17
More fixes to align with CASC rules

mezpusz:
We output ~~`definition`~~ `axiom` type for theory axioms and SAT conflicts, and since we need at least one introduced symbol if we use `introduced(definition,...)`, we instead have to use the God-given `introduced(theory,...)` construct in those cases.

Fixed parenthesizing formulas inside terms, this was never done on the term-formula boundary, hence we had some syntax errors in the output.

Fixed printing for Skolems too, in some cases we output `=` or `vAPP` instead of an introduced symbol, and we "deappify" higher-order non-nullary Skolems terms now to align with Geoff's current parsing.

I also added some sanity checks that a non-specific inference shouldn't be used for inference rules that we have other classes for, otherwise we could have for example theory axioms with premises, which would cause unforeseeable errors. Most of the diff comes from this.


Kernel/InferenceStore.cpp
```diff
@@ -475,6 +478,7 @@ struct InferenceStore::TPTPProofPrinter
     case InferenceRule::GENERAL_SPLITTING_COMPONENT:
     case InferenceRule::INEQUALITY_SPLITTING_NAME_INTRODUCTION:
     case InferenceRule::EQUALITY_PROXY_DEFINITION:
+    case InferenceRule::THEORY_TAUTOLOGY_SAT_CONFLICT:
       return "definition";
```
quickbeam123, line 483: Let me try to understand. Is this in line with the below comment: 
"// without introduced symbols we have to claim that the axiom comes from a theory"
?

Also the `isTheoryAxiomRule(rule)` case above. What if it does not introduce a new symbol?

And in general, since this is likely not for this year anymore, is it really a good idea to sell various theory axioms as definitions? (Sorry maybe I don't understand Geoff's definition of a definition.)

mezpusz: I don't like this solution either, but the point is that something has to either come from the problem, has to be a proper definition e.g. `p(X) <=> ...` or `f(X) = ...`, or it comes from somewhere else, like the axiom of choice, theory axioms, or whatnot (and in this case not accepted as "theorem", just ignored). All of the 2. and 3. fall into the `definition` type now, with various subcategories. I think this can be a discussion point at TP^3.

MichaelRawson:
Now that the Tea Party has finished for another year, does

> All of the 2. and 3. fall into the definition type now, with various subcategories.

still hold? Lots of things were discussed there, sorry. I'd be against this, on account of nothing sensible being defined, but if it's been firmly decided then :shrug: one more oddity in the TPTP world.

Inferences/HOL/CNFOnTheFly.cpp
```diff
@@ -246,14 +246,39 @@ Clause* replaceLits(Clause* c, Literal* what, Proxy p, bool incAge, Literal* by1
...
+        return InferenceRule::AND_PROXY_CLAUSIFICATION_SIMPLIFYING;
+      case Proxy::OR:
+        return InferenceRule::OR_PROXY_CLAUSIFICATION_SIMPLIFYING;
+      case Proxy::IMP:
+        return InferenceRule::IMP_PROXY_CLAUSIFICATION_SIMPLIFYING;
+      case Proxy::IFF:
+        return InferenceRule::IFF_PROXY_CLAUSIFICATION_SIMPLIFYING;
+      case Proxy::XOR:
+        return InferenceRule::XOR_PROXY_CLAUSIFICATION_SIMPLIFYING;
+      default:
```
MichaelRawson: I never remember; do we want to `default:` this kind of switch/case? I think I'd like to get a warning if I added another member to `Proxy`.

Kernel/Inference.hpp
```diff
@@ -363,6 +371,11 @@ enum class InferenceRule : unsigned char {
         (see also isGeneratingInferenceRule) */
   GENERIC_GENERATING_INFERENCE_LAST,
 
+  /** these inferences are hard to categorize, so we just have them here at the end */
+  GENERIC_NONSPECIFIC_INFERENCE,
+  /** not considered a theory axiom at the moment (it's a HOL creature) */
+  HILBERTS_CHOICE_INSTANCE,
```
MichaelRawson, line 377: Wouldn't mind this being a theory axiom or tautology, but no objection to the status quo either.

Kernel/InferenceStore.hpp
```diff
@@ -91,16 +90,13 @@ class InferenceStore
...
-  typedef Stack<SymbolId> SymbolStack;
+  typedef Stack<Signature::Symbol*> SymbolStack;
   // unit id -> stack of introduced symbols (in order of introduction)
   DHMap<unsigned,SymbolStack> _introducedSymbols;
   // symbol id -> existential variable name (number) that was replaced by the symbol
-  DHMap<SymbolId, unsigned> _introducedSymbolReplacedVars;
+  DHMap<Signature::Symbol*, unsigned> _introducedSymbolReplacedVars;
   // symbol id -> the term that is introduced when introducing the skolem symbol
-  DHMap<SymbolId, Term*> _introducedSkolemSymTerms;
+  DHMap<Signature::Symbol*, Term*> _introducedSkolemSymTerms;
```
MichaelRawson, line 99: Nice, as long as we don't iterate over these at any point?

Parse/TPTP.cpp
```diff
@@ -3727,7 +3728,9 @@ void TPTP::endFof()
     _unitSources->insert(original->number(),source);
   }
 
-  ALWAYS(_axiomNames.insert(original->number(), {nm, currentFile.path}));
+  // GDV only reads the first output file, but can dereference included axiom files,
+  // so to make all axioms findable, it's better to output the main file for now.
+  ALWAYS(_axiomNames.insert(original->number(), {nm, _inputFile}));
```
MichaelRawson: This is crazy talk, but I do understand we have some symbiosis with GDV now.

MichaelRawson (approved):

Inferences/HOL/CNFOnTheFly.cpp
mezpusz: I think for `enum class`es it makes sense. I can add the remaining `NOT_PROXY` with `ASSERTION_VIOLATION` (I realized after squinting at it for a while that it means "something is not a proxy" as opposed to "something is a `not` proxy").

Kernel/InferenceStore.hpp
mezpusz: Nice catch, thanks! Currently we don't iterate over them, but in the long-run we could add `id()` to `Symbol`s, if that's okay for you.

Parse/TPTP.cpp
mezpusz: I haven't given much thought to this until now, but we should simply look up what the specification says. Do you maybe know where this part is specified? If it says exact filenames can be (or must be) specified, then we should simply force Geoff to do it in GDV rather than having this here.

mezpusz: One more nuance is that Geoff might be using the `-p <problem_file>` flag of GDV, but I can't use it locally currently due to my experimental setup. In any case, I can remove this hack and try to find a better solution.

MichaelRawson: I only know of [this](https://tptp.org/UserDocs/QuickGuide/Derivations.html) - which says

>  File record: A file record contains the problem file name and the corresponding annotated formulae name in the problem file.

but not what the 'problem file' is. In my view there's little point in the the file record if it doesn't say where the axiom was from.

Do you have easy access to Geoff to check? Either way - no point holding this up, leave a TODO and we can fix it later.

mezpusz: I think I'll revert this, even Geoff says this is unreliable -- in which case having the proper path and not changing the code seems like the better choice.

Kernel/InferenceStore.cpp
mezpusz: The axioms in question have now role `axiom`.

quickbeam123 (approved):

PR #882, opened 2026-07-17
Fable tptp parser fixes

### LINKED ISSUE #553: A different operator precedence when parsing HOL?
Geoff writes:
```
Part of the cause of this confusion is that Vampire (and E) still parses ...
    ! [X:$o] : X = X
... incorrectly, as ...
    (! [X:$o] : X) = X
...
<and explains that uses a workaround for now, to always create     ! [X:$o] : (X = X) for us>
...
deep in your heart, don't you want to do the right thing and fix the parsing?
```
I guess the fix requires making sure that = binds stronger than a quantifier?

quickbeam123:
I went on a parser fixing spree with Claude and this is the result.

Every commit is hopefully self-contained, so we don't need to accept all of them. And we could try to understand what's going on in each in isolation. (Rather than just reading the whole diff at once as we usually do.)

The commit messages usually contain little tptp inputs showcasing what's getting fixed. I could still ask to create some true tests, likely in sanity.

The code base is likely getting more complex thanks to this. But I couldn't say I understood the parser as a whole before, so the situation is not changing too much for me.

Obviously, we need to test this thoroughly. It might be that new strange things will start popping up due to us being more correct! In the last commit, I asked it to start issuing warnings for some parses we allow but actually should not according to the BNF.

Most importantly, this should fix https://github.com/vprover/vampire/issues/553 ;)


Parse/TPTP.cpp
```diff
@@ -4394,16 +4394,15 @@ TermList TPTP::readSort()
...
-        int c = getChar(0);
-        //Polymorphic sorts of are of the form 
+        //Polymorphic sorts of are of the form
         //type_con(sort_1, ..., sort_n)
         //the same as standard first-order terms.
         //Code below works, but does not fit the philosophy of
         //this parser. However, recursive calls to readSort are
         //used in for array sorts and tuple sorts, so polymorphism
         //isn't uniquely evil on this front!
-        if(c == '('){
```
MichaelRawson, line 4413: This is horrifying, I can't believe this worked so long!

Parse/TPTP.hpp
```diff
@@ -637,6 +637,25 @@ class TPTP
...
+    /** 'r = ~ s' read as 'r = (~ s)'; 'g = ^[X]: t' as 'g = (^[X]: t)' */
+    NC_NON_UNITARY_EQUALITY_ARGUMENT,
+    /** 'r = s = t' read as 'r = (s = t)' */
+    NC_CHAINED_EQUALITY,
+    /** the number of kinds above */
+    NC_KINDS
+  };
+  /** which non-conformity kinds have already been warned about in this run */
+  bool _nonConformityWarned[NC_KINDS] = {};
+  void nonConformityWarning(NonConformity kind, const std::string& explanation);
```
MichaelRawson: This could be absorbed into `TPTP.cpp`, but nbd.

Parse/TPTP.cpp
```diff
@@ -3413,7 +3413,16 @@ TermList TPTP::createFunctionApplication(std::string name, unsigned arity)
     insertImplicitLetTypeArguments(ref, arity);
   } else {
     bool dummy;
-    if (arity > 0) {
+    if (_isThf && (name == "vAND" || name == "vOR" || name == "vIMP" ||
+                   name == "vIFF" || name == "vXOR")) {
```
MichaelRawson, line 3542: This seems iffy, shouldn't these simply be `$$internal_system_words` (NB `$$` prefix)?

MichaelRawson (approved):
I'm quite impressed; if you'd written these I'd just say 'merge' and not think about it any more. As it is - is it clean on TPTP with `--mode output`? If it's easy to do it might be sensible to run against master and check it parses the same (ish). And if it all looks rosy let's live dangerously and merge everything?

Parse/TPTP.hpp
quickbeam123: Incoming...

Parse/TPTP.cpp
quickbeam123: Decided not to do for now...

PR #883, opened 2026-07-24
remove output_axiom_names option

MichaelRawson:
I claim that the overhead of `--output_axiom_names on` is negligible and it's a nice thing to have. Remove the option and fix the 'on' behaviour.

If you disagree about the overhead, then I can run some performance tests.


quickbeam123:
I agree we should pay the price. And that it's a low one.

quickbeam123 (approved):

PR #884, opened 2026-07-25
output proofs in the TPTP fragment of the input problem

### LINKED ISSUE #730: Wrong TPTP output for arithmetic problems
The output should contain `tff` instead of `fof` commands when using `-p tptp` for these problems.

Known problem instances where it reproduces with the default strategy:
```
ARI496_1.p
ARI500_1.p
ARI501_1.p
ARI503_1.p
ARI504_1.p
ARI520_1.p
ARI523_1.p
ARI524_1.p
```

shalashaska117:
Fixes #730.

`InferenceStore` chose between `fof` and `tff` using `Problem::hasNonDefaultSorts()`. That property is recomputed from the current unit list (`refreshProperty` calls `Property::scan(_units)`), so it answers "do the units I am holding right now mention a non-`$i` sort" rather than "which fragment was the input problem in". Preprocessing can delete the last unit mentioning a sort, and the property then reports an untyped problem while the proof still contains typed formulas.

The eight problems listed in #730 hit this. The conjecture normalizes to `~$true`, the arithmetic disappears, and the proof comes out as:

```tptp
fof(f1,conjecture,(
  $is_int(6)),
  file('ARI496_1.p',mixed_types_problem_1)).
```

## What changed

The first version of this PR printed `tff` unconditionally, as @MichaelRawson suggested in #730. It is now the union of that and @mezpusz's counter-proposal on `fix-output-tptp-fragment`:

- `Problem` records `initiallyHasNonDefaultSorts` and `initiallyHigherOrder` when the input problem is set, and the proof printers pick `fof`, `tff` or `thf` from those flags. A pure FOF problem keeps its `fof` proof.
- Variable annotations use a per-formula test (`t != defaultSort() || initiallyHasNonDefaultSorts()`) in `getQuantifiedStr` and `Formula::toString`, so a `$int` variable surviving in a proof step never prints bare and reads back as `$i`.
- The `getQuantifiedStr` overload without a sort map is gone. It passed a static `DHMap` that nothing ever filled, and `printGeneralSplittingComponent` went through it: `ALWAYS(t_map.find(var,t))` on an empty map is undefined behaviour in release, and the general splitting definitions printed their variables without sorts either way. The definition printer now collects the sorts from the clause.
- `FunctionRelationshipInference` swaps the main problem twice during FMB sort bound detection, which runs after preprocessing, so the flags got recomputed from the preprocessed units and the bug came back for fmb strategies. Those two calls now skip the recording (`isInputProblem=false`).
- `outputSymbolTypeDeclarationIfNeeded` follows the saved flags, and the proof printers skip type declarations entirely when the recorded fragment is fof. The signature can hold a typed symbol that no unit ever used, and its declaration would put a `tff` line in front of an `fof` proof.

`-p on`, `-p smtcheck` and `-p smt2_proofcheck` are untouched, and SAT steps from AVATAR still print as `cnf`.

## Testing

- `checks/sanity` passes. The new checks test both directions: no `fof` in the proofs of ARI496_1 and `mixed-language.p`, only `fof` for PUZ001+1 and for a problem whose signature holds an unused typed symbol, and sort annotations on the gsp definitions (`checks/proof/gsp-typed.p` fails on master).
- 97/97 unit tests pass in debug, plus assertion-enabled runs including `-sa fmb -fmbdsb on`.
- `-p tptp` over the checks problems against master: ARI496_1, `let-bool.p` and `mixed-language.p` change `fof(` to `tff(`, `unused-typed-symbol.p` drops the unused declaration, every SZS status is unchanged, and the remaining outputs are identical up to the clause renumbering master already shows between runs.

I still have not checked the output against a strict TPTP parser like tptp4X; Vampire's own parser accepted the broken proofs, so round-tripping through Vampire does not catch the old behaviour.

### DIFF
--- a/Kernel/Formula.cpp
+++ b/Kernel/Formula.cpp
@@ -198,11 +198,9 @@ std::string Formula::toString () const
           res += Term::variableToString(var);
           // a variable of a sort other than $i must always be annotated, and
           // preprocessing is not expected to introduce such a sort into an
-          // initially untyped problem; if that ever happens, debug builds fail
-          // here and release builds still print the sort rather than drop the
-          // annotation
+          // initially untyped problem
           ASS(sort == AtomicSort::defaultSort() || env.initiallyHasNonDefaultSorts());
-          if (sort != AtomicSort::defaultSort() || env.initiallyHasNonDefaultSorts()) {
+          if (env.initiallyHasNonDefaultSorts()) {
             res += " : " + sort.toString();
           }
           first = false;
--- a/Kernel/Formula.cpp
+++ b/Kernel/Formula.cpp
@@ -196,7 +196,10 @@ std::string Formula::toString () const
             res += ",";
           }
           res += Term::variableToString(var);
-          if (sort != AtomicSort::defaultSort() || env.getMainProblem()->hasNonDefaultSorts()) {
+          // a variable of a sort other than $i must always be annotated: the
+          // problem's own hasNonDefaultSorts() is recomputed from the current
+          // unit list and forgets sorts that preprocessing removed
+          if (sort != AtomicSort::defaultSort() || env.initiallyHasNonDefaultSorts()) {
             res += " : " + sort.toString();
           }
           first = false;
--- a/Kernel/Formula.cpp
+++ b/Kernel/Formula.cpp
@@ -196,9 +196,12 @@ std::string Formula::toString () const
             res += ",";
           }
           res += Term::variableToString(var);
-          // a variable of a sort other than $i must always be annotated: the
-          // problem's own hasNonDefaultSorts() is recomputed from the current
-          // unit list and forgets sorts that preprocessing removed
+          // a variable of a sort other than $i must always be annotated, and
+          // preprocessing is not expected to introduce such a sort into an
+          // initially untyped problem; if that ever happens, debug builds fail
+          // here and release builds still print the sort rather than drop the
+          // annotation
+          ASS(sort == AtomicSort::defaultSort() || env.initiallyHasNonDefaultSorts());
           if (sort != AtomicSort::defaultSort() || env.initiallyHasNonDefaultSorts()) {
             res += " : " + sort.toString();
           }
--- a/Kernel/InferenceStore.cpp
+++ b/Kernel/InferenceStore.cpp
@@ -149,18 +149,6 @@ std::string getQuantifiedStr(const VarContainer& vars, std::string inner, DHMap<
 }
 
 /**
- * Return @b inner quentified over variables in @b vars
- *
- * It is caller's responsibility to ensure that variables in @b vars are unique.
- */
-template<typename VarContainer>
-std::string getQuantifiedStr(const VarContainer& vars, std::string inner, bool innerParentheses=true)
-{
-  static DHMap<unsigned,TermList> d;
-  return getQuantifiedStr(vars,inner,d,innerParentheses);
-}
-
-/**
  * Return std::string containing quantified unit @b u.
  */
 std::string getQuantifiedStr(Unit* u, List<unsigned>* nonQuantified=0)
@@ -442,9 +430,14 @@ struct InferenceStore::TPTPProofPrinter
 
   void print() override
   {
-    //outputSymbolDeclarations also deals with sorts for now
-    //UIHelper::outputSortDeclarations(out);
-    UIHelper::outputSymbolDeclarations(out);
+    //an fof proof needs no type declarations, and the signature can hold a
+    //typed symbol no unit ever used, whose declaration would put a tff line
+    //into an otherwise fof proof
+    if(env.initiallyHasNonDefaultSorts() || env.initiallyHigherOrder()){
+      //outputSymbolDeclarations also deals with sorts for now
+      //UIHelper::outputSortDeclarations(out);
+      UIHelper::outputSymbolDeclarations(out);
+    }
     ProofPrinter::print();
   }
 
@@ -780,6 +773,11 @@ std::string getSkolemizeMap(unsigned unitNumber, It symIt){
 
     Literal* nameLit=_is->_splittingNameLiterals.get(us->number()); //the name literal must always be stored
 
+    //sorts of the clause's variables, so the quantifiers of the definition
+    //below are annotated like every other formula in the proof
+    DHMap<unsigned,TermList> t_map;
+    SortHelper::collectVariableSorts(us, t_map);
+
     std::string defId=tptpDefId(us);
 
     out<<getFofString(tptpUnitId(us), getFormulaString(us),
@@ -821,11 +819,11 @@ std::string getSkolemizeMap(unsigned unitNumber, It symIt){
     }
     ASS(!first);
 
-    compStr=getQuantifiedStr(compOnlyVars, compStr, multiple);
+    compStr=getQuantifiedStr(compOnlyVars, compStr, t_map, multiple);
     List<unsigned>::destroy(compOnlyVars);
 
     std::string defStr=compStr+" <=> "+Literal::complementaryLiteral(nameLit)->toString();
-    defStr=getQuantifiedStr(nameVars, defStr);
+    defStr=getQuantifiedStr(nameVars, defStr, t_map);
     List<unsigned>::destroy(nameVars);
 
     SymbolId nameSymbol = SymbolId(SymbolType::PRED,nameLit->functor());
@@ -850,9 +848,12 @@ protected:
     InferenceRule rule = cs->inference().rule();
     UnitIterator parents= cs->getParents();
 
-    //outputSymbolDeclarations also deals with sorts for now
-    //UIHelper::outputSortDeclarations(out);
-    UIHelper::outputSymbolDeclarations(out);
+    //an fof proof needs no type declarations, see TPTPProofPrinter::print
+    if(env.initiallyHasNonDefaultSorts() || env.initiallyHigherOrder()){
+      //outputSymbolDeclarations also deals with sorts for now
+      //UIHelper::outputSortDeclarations(out);
+      UIHelper::outputSymbolDeclarations(out);
+    }
 
     // fragment of the unpreprocessed input, see the comment in getFofString
     std::string kind = "fof";
--- a/Kernel/InferenceStore.cpp
+++ b/Kernel/InferenceStore.cpp
@@ -117,9 +117,11 @@ std::string getQuantifiedStr(const VarContainer& vars, std::string inner, DHMap<
     std::string ty="";
     TermList t;
 
-    if(t_map.find(var,t) && env.getMainProblem()->hasNonDefaultSorts()){
-      //hasNonDefaultSorts is true if the problem contains a sort
-      //that is not $i and not a variable
+    if(t_map.find(var,t) &&
+       (t != AtomicSort::defaultSort() || env.getMainProblem()->hasNonDefaultSorts())){
+      //a variable of a sort other than $i must always be annotated: relying on
+      //hasNonDefaultSorts alone loses the annotation when preprocessing removes
+      //the last unit mentioning that sort. Same predicate as Formula::toString.
       ty=" : " + t.toString();
     }
     if(ty == " : $tType"){
@@ -530,9 +532,10 @@ protected:
 
   std::string getFofString(std::string id, std::string formula, std::string inference, InferenceRule rule, UnitInputType origin=UnitInputType::AXIOM)
   {
-    std::string kind = "fof";
-    if(env.getMainProblem()->hasNonDefaultSorts()){ kind="tff"; }
-    if(env.getMainProblem()->isHigherOrder()){ kind="thf"; }
+    // always TFF: it subsumes FOF, and choosing between them from
+    // hasNonDefaultSorts() is wrong, because that property is recomputed from
+    // the current unit list and so forgets sorts that preprocessing removed.
+    std::string kind = env.getMainProblem()->isHigherOrder() ? "thf" : "tff";
 
     return kind+"("+id+","+getRole(rule,origin)+",("+"\n"
 	+"  "+formula+"),\n"
@@ -858,9 +861,8 @@ protected:
     //UIHelper::outputSortDeclarations(out);
     UIHelper::outputSymbolDeclarations(out);
 
-    std::string kind = "fof";
-    if(env.getMainProblem()->hasNonDefaultSorts()){ kind="tff"; }
-    if(env.getMainProblem()->isHigherOrder()){ kind="thf"; }
+    // always TFF, see the comment in getFofString
+    std::string kind = env.getMainProblem()->isHigherOrder() ? "thf" : "tff";
 
     out << kind
         << "(r"<< cs->number()
--- a/Kernel/InferenceStore.cpp
+++ b/Kernel/InferenceStore.cpp
@@ -120,11 +120,9 @@ std::string getQuantifiedStr(const VarContainer& vars, std::string inner, DHMap<
     if(t_map.find(var,t)){
       //a variable of a sort other than $i must always be annotated, and
       //preprocessing is not expected to introduce such a sort into an
-      //initially untyped problem; if that ever happens, debug builds fail
-      //here and release builds still print the sort rather than drop the
-      //annotation. Same predicate as Formula::toString.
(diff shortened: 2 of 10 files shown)


Kernel/InferenceStore.cpp
```diff
@@ -117,9 +117,11 @@ std::string getQuantifiedStr(const VarContainer& vars, std::string inner, DHMap<
...
     TermList t;
 
-    if(t_map.find(var,t) && env.getMainProblem()->hasNonDefaultSorts()){
-      //hasNonDefaultSorts is true if the problem contains a sort
-      //that is not $i and not a variable
+    if(t_map.find(var,t) &&
+       (t != AtomicSort::defaultSort() || env.getMainProblem()->hasNonDefaultSorts())){
+      //a variable of a sort other than $i must always be annotated: relying on
+      //hasNonDefaultSorts alone loses the annotation when preprocessing removes
+      //the last unit mentioning that sort. Same predicate as Formula::toString.
```
MichaelRawson: In future we could consider just printing `$i`, it's not illegal any more because we never use `fof`.

MichaelRawson (approved):
Cool! I intend to merge after asking a few people about it.

MichaelRawson:
> Unit tests: 96 of 97 pass. Inferences_HOL_Injectivity fails identically with and without this change on https://github.com/vprover/vampire/commit/3677326861181f990ce3ef461e90471ba9749225, so it is not from here.

This doesn't seem good: CI passes this test. What system are you using?

shalashaska117:
> > Unit tests: 96 of 97 pass. Inferences_HOL_Injectivity fails identically with and without this change on https://github.com/vprover/vampire/commit/3677326861181f990ce3ef461e90471ba9749225, so it is not from here.
> 
> This doesn't seem good: CI passes this test. What system are you using?

I don't understand it either, I get the same error on main, I used Ubuntu latest version, now I'm not at the computer so can't see the exact version

MichaelRawson:
Uh-oh. Experience suggests that a compiler upgrade finally found a bug that was lurking - if you can, please try to debug the unit test. Otherwise I guess we'll let it play out.

MichaelRawson:
(but it's independent of this change of course, don't worry!)

shalashaska117:
> Uh-oh. Experience suggests that a compiler upgrade finally found a bug that was lurking - if you can, please try to debug the unit test. Otherwise I guess we'll let it play out.

I will work on it today.

shalashaska117:
@MichaelRawson success_3's expected clause was wrong (copy-pasted from success_2), and clang only hides it because it evaluates the two sides of the != in the opposite order to gcc, which makes the produced clause share a subterm with the expected one, and the test comparator returns true on identical subterms without checking that the variable renaming is consistent.

mezpusz:
I also like the fix, but I think the issue could also happen for `thf` problems, and we probably don't want to go full `thf` if we don't have to. I'm proposing to instead save the initial value of the flags that are used to output `fof`/`tff`/`thf`. More generally, we could do a per formula scanning before printing to use the weakest possible fragment we need, but this is probably expensive. Anyways, going weaker is currently allowed by Geoff Sutcliffe, while going stronger is not, so we should be careful here.

I will open a separate PR with my proposed solution, so that we can discuss.

shalashaska117:
> I also like the fix, but I think the issue could also happen for `thf` problems, and we probably don't want to go full `thf` if we don't have to. I'm proposing to instead save the initial value of the flags that are used to output `fof`/`tff`/`thf`. More generally, we could do a per formula scanning before printing to use the weakest possible fragment we need, but this is probably expensive. Anyways, going weaker is currently allowed by Geoff Sutcliffe, while going stronger is not, so we should be careful here.
> 
> I will open a separate PR with my proposed solution, so that we can discuss.

I am very curious about your solution, I will look into it

shalashaska117:
@mezpusz do you have any test idea that I could try to verify your claim about thf problems? trying something but I don't know how to check for all cases to be honest, so if you have better ideas to try to point to errors that I could not find

mezpusz:
I've checked it now with an "unsound" preprocessing step that just replaces all units with `$false`. The point is that we have to remove all higher-order constructs already in preprocessing. As expected the output contained `fof` in all output lines (except for type declarations, where we get `tff`).

I pushed the changes for the alternative solution to `fix-output-tptp-fragment`, I will run some tests before opening a PR.

shalashaska117:
@mezpusz any news about how to solve this problem?

mezpusz:
@shalashaska117 [This is my proposed solution](https://github.com/vprover/vampire/compare/fix-output-tptp-fragment) but I need to clear out a few other TPTP issues before I can verify that it works.

shalashaska117:
@mezpusz I went through `fix-output-tptp-fragment` and I think the two fixes combine well: your saved flags decide the fof/tff/thf header, and the per-formula test from this PR keeps the variable annotations right. I've pushed the union to this branch. Three things I ran into while reading your diff:

- `getQuantifiedStr` still reads `env.getMainProblem()->hasNonDefaultSorts()` while its comment says `initiallyHasNonDefaultSorts`, so clause printing keeps the stale-property bug: a proof step that still contains a `$int` variable prints it bare, and it reads back as `$i`. I used the per-formula test (`t != defaultSort() || initiallyHasNonDefaultSorts()`) there and in `Formula::toString`, so a typed variable never loses its annotation.
- `ALWAYS(t_map.find(var,t))` fails on `-gsp on` proofs: the `getQuantifiedStr` overload without a map passed a static `DHMap` that nothing ever filled, and `printGeneralSplittingComponent` went through it. In release `ALWAYS(Cond)` is `if (!(Cond)) __UNREACHABLE`, so a failed find is undefined behaviour. The empty map was a bug of its own: general splitting definitions printed their variables without sorts, `checks/proof/gsp-typed.p` shows it. The definition printer now collects the sorts from the clause, and the map-less overload is gone.
- `FunctionRelationshipInference` swaps the main problem twice during FMB sort bound detection, which runs after preprocessing, so the saved flags got recomputed from the preprocessed unit list and the bug came back for fmb strategies. Those two calls now skip the recording (`isInputProblem=false`).

Two more changes in the same area: `outputSymbolTypeDeclarationIfNeeded` now follows the saved flag, otherwise an originally higher-order problem whose constructs are all eliminated in preprocessing prints thf steps with tff type declarations. And the proof printers skip type declarations entirely when the recorded fragment is fof, since the signature can hold a typed symbol that no unit ever used, whose declaration would put a tff line in front of an fof proof.

`checks/sanity` now tests both directions: no fof in the proofs of ARI496_1 and mixed-language.p, only fof for PUZ001+1 and for the unused typed declaration, and sort annotations on the gsp definitions. Against master, `-p tptp` over the checks problems changes ARI496_1, let-bool.p and mixed-language.p (fof to tff) and unused-typed-symbol.p (drops the declaration of the unused symbol); every SZS status is unchanged and the remaining outputs are identical up to the clause renumbering master already shows between runs. 97/97 unit tests pass.

mezpusz:
> * `getQuantifiedStr` still reads `env.getMainProblem()->hasNonDefaultSorts()` while its comment says `initiallyHasNonDefaultSorts`

Good catch!

> 

> * `ALWAYS(t_map.find(var,t))` fails on `-gsp on` proofs

The point of this assertion was to enforce callers to fill the map, otherwise @MichaelRawson might know why we want UB in case the assertion fails in release.

> * I used the per-formula test (`t != defaultSort() || initiallyHasNonDefaultSorts()`) there and in `Formula::toString`, so a typed variable never loses its annotation.

I think it should never happen that `initiallyHasNonDefaultSorts()==false` and we introduce non-default sorts during preprocessing, so this should rather be an assertion too.

shalashaska117:
> > * `ALWAYS(t_map.find(var,t))` fails on `-gsp on` proofs
> 
> The point of this assertion was to enforce callers to fill the map, otherwise @MichaelRawson might know why we want UB in case the assertion fails in release.
> 

Wouldn't this create more problems?

> > * I used the per-formula test (`t != defaultSort() || initiallyHasNonDefaultSorts()`) there and in `Formula::toString`, so a typed variable never loses its annotation.
> 
> I think it should never happen that `initiallyHasNonDefaultSorts()==false` and we introduce non-default sorts during preprocessing, so this should rather be an assertion too. 

I can check this easily, but in my test it never occurred. Easy fix to the code.

MichaelRawson:
> The point of this assertion was to enforce callers to fill the map, otherwise @MichaelRawson might know why we want UB in case the assertion fails in release.

I defined `ASS` and friends to hint unreachability in a release build because it allows certain optimisations (or did something at the time, I'm not sure now). There was a small performance improvement, but I'm not attached to it - so if you prefer it can go.

Consider e.g.

```c++
void f(int x) {
  if(x < 0) throw Exception("expected positive integer");
  // ....
}

void g(int y) {
  ASS(y > 0)
  f(y);
  // ...
}
```

When inlining `f` into `g`, the optimiser can reason that `y > 0` (otherwise UB), so the conditional branch and the exception setup can disappear altogether.

shalashaska117:
@mezpusz agreed on the assertion. I'll add

```c++
ASS(t == AtomicSort::defaultSort() || env.initiallyHasNonDefaultSorts());
```

at the two sites of the per-formula test (`InferenceStore.cpp` and `Formula::toString`). I'd keep the disjunction as the print condition though: in release `ASS` compiles to `{}` (`Debug/Assertion.hpp`), so it costs nothing, and if some future preprocessing rule does introduce a non-default sort into an initially untyped problem, a release build prints the variable with its sort instead of dropping the annotation, which is the failure mode this PR is about. Debug builds fail the assertion and the rule gets found.

I could not come up with a counterexample either: theory axioms, induction and FOOL elimination all need the sorts to be in the problem already. For some confidence beyond that I'll run a debug build with the assertion over `checks/` and a batch of TPTP FOF problems with `-gsp on`, `-newcnf on` and `-sa fmb -fmbdsb on` before pushing.

On `ALWAYS(t_map.find(var,t))`: dropping the map-less overload keeps the contract, it just moves it to compile time. Every caller now has to build a filled map, and the `ALWAYS` in the remaining overload holds because `printGeneralSplittingComponent` collects the sorts from the clause.

@MichaelRawson thanks for the explanation. One detail: in release only `ALWAYS`/`NEVER`/`ASSERTION_VIOLATION` expand to `__UNREACHABLE`, plain `ASS` is `{}`, so the optimisation only concerns those three. Either way it looks independent of this PR, so I left the macros alone.

Kernel/InferenceStore.cpp
shalashaska117: @MichaelRawson this predates the union: proofs of initially untyped problems come out as `fof` again, where `:$i` is illegal. In `tff`/`thf` proofs every quantified variable is annotated now, including the `$i` ones.

shalashaska117:
Pushed the assertion in 5bf5aa600: `ASS(t == AtomicSort::defaultSort() || env.initiallyHasNonDefaultSorts())` now guards both per-formula sites, and the disjunction stays as the release print condition as discussed above.

Validation: 97/97 unit tests in debug, `checks/sanity` with both the debug and the release build, and a sweep of the debug build over 158 problems in four configs (default, `-gsp on`, `-newcnf on`, `-sa fmb -fmbdsb on`, all with `-p tptp -t 10`): 42 bushy MPTP2078 FOF problems, 53 typed and mixed TPTP problems, and the 63 files under `checks/`. Of the 632 runs, 220 printed refutations and 5 printed finite models; the new assertion never fired.

The sweep did surface two debug assertion failures that already exist on master (reproduced at 367732686), so they are not from this PR:

- `UIHelper.cpp:546`, `ASS(!s_expecting_unsat)`: some bushy MPTP2078 problems whose header says `Status : Theorem` come out CounterSatisfiable (MPT0651+1 by saturation even in the default config, MPT1151+1 / MPT1451+1 / MPT1501+1 / MPT1751+1 as fmb models). If those problems are really countersatisfiable the headers are at fault rather than the solver, but either way a debug build dies on the mismatch instead of reporting it.
- `LispParser.hpp:109`, `ASS(e->isList())`: when auto input_syntax fails to parse a file as TPTP (for example tuples without `-newcnf on`, or thf) and falls back to SMTLIB2, the lisp parser asserts on the non-SMTLIB text instead of raising a user error.

I can open issues for these if they are not already known.

MichaelRawson:
> @MichaelRawson thanks for the explanation. One detail: in release only ALWAYS/NEVER/ASSERTION_VIOLATION expand to __UNREACHABLE, plain ASS is {}, so the optimisation only concerns those three. Either way it looks independent of this PR, so I left the macros alone.

Oops, yes - evaluating the `ASS` condition in release is more expensive than it's worth. Whereas the semantics of `ALWAYS`  and so on requires evaluating the condition anyhow.

checks/proof/gsp-typed.p
```diff
@@ -0,0 +1,14 @@
+% ax1 chains four literals through single shared variables, so -gsp on names
```
mezpusz: Could you please shorten these comments to just the expected outcome, from the lengthy detailed explanation it's really hard to grasp what should happen and why.

checks/sanity
```diff
@@ -167,3 +220,16 @@ check_szs_status Unsatisfiable -qa synthesis synthesis/ex2-inverse_of_ixopiy.smt
 
 # Higher-order logic
 check_szs_status Theorem hol/hol1.p
+
+# Proof output language: the arithmetic goes away during preprocessing, so a
```
mezpusz: These tests potentially have value, but I wouldn't put them into our sanity check suite, as from time to time we will have to test our outputs with GDV over the entire TPTP anyways.

Kernel/Formula.cpp
```diff
@@ -196,7 +196,13 @@ std::string Formula::toString () const
...
           }
           res += Term::variableToString(var);
-          if (sort != AtomicSort::defaultSort() || env.getMainProblem()->hasNonDefaultSorts()) {
+          // a variable of a sort other than $i must always be annotated, and
+          // preprocessing is not expected to introduce such a sort into an
+          // initially untyped problem; if that ever happens, debug builds fail
+          // here and release builds still print the sort rather than drop the
+          // annotation
+          ASS(sort == AtomicSort::defaultSort() || env.initiallyHasNonDefaultSorts());
+          if (sort != AtomicSort::defaultSort() || env.initiallyHasNonDefaultSorts()) {
```
mezpusz: Having the `sort != AtomicSort::defaultSort()` check in release just hides potential errors that we should catch in testing, so I would remove it. Same for `InferenceStore.cpp`.

checks/proof/gsp-typed.p
shalashaska117: Ok, tried to be as specific as I could, I tend to be verbose in all my comments in multi-team projects

checks/sanity
shalashaska117: I will cancel the hook but keeping the tests

Kernel/Formula.cpp
shalashaska117: Ok wrongly assumed that a defensive-disjunct idea was the better one, but yes I can see why for release it should be avoided

shalashaska117:
@mezpusz all three done in 0e5a7e260:

* the checks/proof headers now just state the expected outcome
* checks/sanity is back to its state on master; I kept the three problem files under checks/proof so they can still be run by hand, but nothing in the suite uses them
* both print sites now test only `env.initiallyHasNonDefaultSorts()`, with the assertion unchanged, so a broken invariant shows up in release as a missing annotation instead of being papered over

Re-ran ctest (97/97, debug) and checks/sanity with the release build, plus the dropped checks by hand against debug: the -gsp definition still annotates its variables, PUZ001+1 stays fof, mixed-language.p comes out all tff, unused-typed-symbol.p all fof, and ARI496_1 has no fof lines with either -p tptp or -p proofcheck.

mezpusz (approved):
Thanks for the changes! We will have to wait until #881 is merged, then I will run another round of GDV tests, hopefully the rest of the errors are fixed by this.

Kernel/Formula.cpp
mezpusz: No worries, my reasoning is that this comes from the developer side and we usually test whether a new typed feature is turned off for FOF.

checks/proof/gsp-typed.p
mezpusz: No worries, I'm just trying to decrease maintenance burden, and longer comments are usually not read by the developers.

MichaelRawson:
#881 merged now, feel free to run tests.

PR #885, opened 2026-07-25
build $let definitions of Boolean symbols as formulas

### LINKED ISSUE #863: Bug with let definition shadowing
The following crashes with current master (`a3f660c0e3ce335808a1717ec0b4996bc38ea503`):
```
./vampire -newcnf on $TPTP/Problems/SWX/SWX006_1.p
```
Adding `--show_everything on` crashes even earlier, because some `$let` binding cannot be printed, due to having a special term `$ite(...)` in the LHS. This happens due to not shadowing properly while parsing (the input contains several instances of let definitions, defining `max`, `bad` and `i` symbols, shadowing each other), and some occurrence of `max` (namely `max1`) in the LHS gets replaced by an outer occurrence. The parsed input:
```
[PP] input: 7. ~$formula{$let(n: $int, n := $ite(~$less(i,n), ($term{$true},$term{bad}), ($let(max1: $int, max1 := $ite($less(max,select_int_int(a,i)), (select_int_int(a,i),max), ($let(i2: $int, i2 := $sum(i,1), ($let(0: $int, 0 := $ite(~$less(i2,n), ($term{$true},$term{bad0}), ($let(max4: $int, max4 := $ite($less(max1,select_int_int(a,i2)), (select_int_int(a,i2),max1), ($let(i5: $int, i5 := $sum(i2,1), ($let(max1: $int, max1 := $ite(~$less(i5,n), ($term{$true},$term{bad3}), ($let(max7: $int, max7 := $ite($less(max4,select_int_int(a,i5)), (select_int_int(a,i5),max4), ($let(i8: $int, i8 := $sum(i5,1), ($let(i2: $int, i2 := $ite(~$less(i8,n), ($term{$true},$term{bad6}), ($let(max10: $int, max10 := $ite($less(max7,select_int_int(a,i8)), (select_int_int(a,i8),max7), ($let(i11: $int, i11 := $sum(i8,1), ($let(1: $int, 1 := $ite(~$less(i11,n), ($term{$true},$term{bad9}), ($let(max13: $int, max13 := $ite($less(max10,select_int_int(a,i11)), (select_int_int(a,i11),max10), ($let(i14: $int, i14 := $sum(i11,1), ($term{bad12 | ! [X0 : $int] : (($greatereq(X0,0) & $less(X0,i14)) => $greatereq(max13,select_int_int(a,X0)))})))))))))))))))} [negated conjecture 6]
```
~~But the TPTP problem suggests that recursive definitions are allowed, where we would either inlining either won't terminate or will be incomplete according to the current implementation.~~ What I read as recursive definitions actually refer to the previous values (also for the first let).

shalashaska117:
Fixes #863.

`TPTP::endLet` built the definition with `Term::create(symbol, args)`. When the `$let` binds a Boolean symbol, `symbol` is a predicate number, so `Term::create` looked it up in the function half of the signature: `Signature::functionArity` then indexed past the end of the function stack. In a debug build it trips the bounds assertion in `Lib::Stack::operator[]`, in a release build it segfaults.

`SMTLIB2::parseLet` already handles this, with a comment about preserving the term-formula boundary, so this makes the TPTP parser do the same thing.

## Shadowing is not the cause

The issue puts it down to shadowing, and `SWX006_1.p` does shadow `bad`, `max` and `i` repeatedly, but shadowing turns out to be irrelevant. This is enough on its own:

```tptp
tff(c,conjecture, $let(p: $o, p := $true, p ) ).
```

Any `$let` over a Boolean symbol crashes on master, shadowing or not, fresh name or not. `SWX006_1.p` is just the first problem anyone ran into that contains one.

With the fix, `SWX006_1.p` gives `SZS status Theorem`, which matches its TPTP header, and `--show_everything on` no longer dies while printing.

## A separate problem this uncovers

Once the crash is gone, `-newcnf on` gives a wrong answer for a conjecture holding two or more Boolean `$let`s. It reports `CounterSatisfiable` for this theorem:

```tptp
tff(q,type, q: $o ).
tff(c,conjecture,
    ( $let(b: $o, b := ( q | ~ q ), b )
    & $let(d: $o, d := ( q | ~ q ), d ) ) ).
```

That one predates this change and I have left it alone. NewCNF gets the same shape from the SMT-LIB parser, which has always built Boolean let definitions this way, so it reproduces on master with no parser change at all:

```smt2
(set-logic ALL)
(declare-fun q () Bool)
(assert (not (and (let ((b (or q (not q)))) b) (let ((d (or q (not q)))) d))))
(check-sat)
```

`vampire --input_syntax smtlib2 -newcnf on` answers `Satisfiable`, on master and with this branch alike, where the default clausifier answers `Unsatisfiable`. A single Boolean `$let` is fine in both. Worth its own issue, and I am happy to file one with the reduction if that helps.

The regression tests below therefore use one Boolean `$let` per file, so they pass in all four clausifier configurations instead of encoding the NewCNF bug as expected behaviour.

## Testing

- `checks/theory/let-bool.p` (nullary) and `checks/theory/let-bool-pred.p` (arity 1), each shadowing a global of the same name, both under the default clausifier and `-newcnf on`, with and without `-ile off`. All eight combinations give `Theorem` with the fix and segfault without it. `checks/theory/let-bool-simultaneous.p` covers two symbols bound in one `$let`, so `endLet` builds a definition per symbol; on master that one is a sort error rather than a crash.
- Other shapes I ran by hand, each `Theorem` in default, `-newcnf on` and `-ile off`, each a segfault on master: predicate of arity 2, a Boolean `$let` nested in another one's definition, a Boolean `$let` under an outer quantifier whose definition uses the bound variable, a Boolean `$let` whose body rebinds the same name again, and two Boolean symbols bound simultaneously.
- Two shapes I could not get to the changed code at all: `$let(p: !>[A: $tType]: A > $o, ...)` and its implicit form `$let(p: A > $o, ...)`. Both are rejected earlier with a `User error`, identically on master and on this branch. So the `iTypeArgs` part of the new `args.size()` is not covered by anything I could construct, and I have left that path as it was.
- `checks/sanity` passes, 74 checks, built against Z3 4.14.0.
- Unit tests: 96 of 97 pass. `Inferences_HOL_Injectivity` fails the same way with `Parse/TPTP.cpp` reverted, so it is not from here.
- Debug build, assertions on, over the new tests and the existing `theory/let-*.p` ones. Nothing fires on the new tests. `theory/let-tuple-bool.p` in default mode does trip `Shell/LispParser.hpp:109`, but it does so on master too: the TPTP parser raises its own "set --newcnf on if using tuples" error, auto mode then falls back to SMT-LIB, and the Lisp parser asserts on the way out. That is presumably why `checks/sanity` only runs the tuple tests with `-newcnf on`.

### DIFF
--- a/checks/theory/let-bool-simultaneous.p
+++ b/checks/theory/let-bool-simultaneous.p
@@ -1,0 +1,9 @@
+% Two Boolean symbols bound in one $let, so endLet builds a definition for each.
+% One of them is Boolean and one is not, which used to give a sort error rather
+% than a crash.
+tff(let_bool_simultaneous,conjecture,
+    $let(
+      [ b: $o, k: $int ],
+      [ b := $true, k := 1 ],
+      ( b
+      & $greater(k,0) ) ) ).
--- a/checks/theory/let-bool-simultaneous.p
+++ b/checks/theory/let-bool-simultaneous.p
@@ -1,6 +1,5 @@
-% Two Boolean symbols bound in one $let, so endLet builds a definition for each.
-% One of them is Boolean and one is not, which used to give a sort error rather
-% than a crash.
+% Two symbols bound in one $let, one Boolean and one not, so endLet builds a
+% definition for each. On master this shape is a sort error, not a crash.
 tff(let_bool_simultaneous,conjecture,
     $let(
       [ b: $o, k: $int ],
--- a/Parse/TPTP.cpp
+++ b/Parse/TPTP.cpp
@@ -2736,7 +2736,16 @@ void TPTP::endLet()
       }
       vars = varList;
     }
-    auto binding = Formula::createDefinition(Term::create(symbol, args), body, vars);
+    Term* lhs;
+    if (isPredicate) {
+      // symbol is a predicate number, so it cannot go through Term::create.
+      // Wrap it as a formula to preserve the term-formula boundary, the same
+      // way SMTLIB2::parseLet does.
+      lhs = Term::createFormula(new AtomicFormula(Literal::create(symbol, args.size(), true, args.begin())));
+    } else {
+      lhs = Term::create(symbol, args);
+    }
+    auto binding = Formula::createDefinition(lhs, body, vars);
     let = TermList(Term::createLet(binding, let, sort));
   }
   _termLists.push(let);
--- a/checks/sanity
+++ b/checks/sanity
@@ -85,6 +85,16 @@ check_szs_status Theorem -newcnf on theory/let-tuple-bool.p
 check_szs_status Theorem -newcnf on -ile off theory/let-function.p
 check_szs_status Theorem -newcnf on -ile off theory/let-tuple.p
 check_szs_status Theorem -newcnf on -ile off theory/let-tuple-bool.p
+check_szs_status Theorem theory/let-bool.p
+check_szs_status Theorem -newcnf on theory/let-bool.p
+check_szs_status Theorem -ile off theory/let-bool.p
+check_szs_status Theorem -newcnf on -ile off theory/let-bool.p
+check_szs_status Theorem theory/let-bool-pred.p
+check_szs_status Theorem -newcnf on theory/let-bool-pred.p
+check_szs_status Theorem -ile off theory/let-bool-pred.p
+check_szs_status Theorem -newcnf on -ile off theory/let-bool-pred.p
+check_szs_status Theorem theory/let-bool-simultaneous.p
+check_szs_status Theorem -newcnf on theory/let-bool-simultaneous.p
 
 # Arrays
 check_szs_status Unsatisfiable theory/arrays.smt2
--- a/checks/theory/let-bool-pred.p
+++ b/checks/theory/let-bool-pred.p
@@ -1,0 +1,9 @@
+% As let-bool.p, but the bound Boolean symbol takes an argument, so the
+% definition is quantified over it. Also shadows a global of the same name.
+tff(r,type,
+    r: $int > $o ).
+
+tff(let_bool_pred,conjecture,
+    $let(r: $int > $o,
+      r(X) := ( r(X) | ~ r(X) ),
+      r(1) ) ).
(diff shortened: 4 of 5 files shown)


checks/theory/let-bool-simultaneous.p
```diff
@@ -0,0 +1,9 @@
+% Two Boolean symbols bound in one $let, so endLet builds a definition for each.
```
mezpusz: What do you mean two Boolean symbols? I see only one.

shalashaska117: Sorry, you're right, the first line is wrong because b is a boolean and k is $int, i wrote that both were boolean here but in the next line I correct myself. I was talking about a test that I did for this and placed the wrong line there.

mezpusz:
Sorry for not being able to look at this earlier. My tests suggests that this solves 5 TPTP crashes. Great work!

mezpusz (approved):

mezpusz:
@shalashaska117 Regarding the issue with `-newcnf on` in the section "A separate problem this uncovers", could you please file an issue?

shalashaska117:
> @shalashaska117 Regarding the issue with `-newcnf on` in the section "A separate problem this uncovers", could you please file an issue?

I will, being Vampire a project so big I thought that issues in non-main branches were not worth being published

mezpusz:
> I will, being Vampire a project so big I thought that issues in non-main branches were not worth being published

As far as I understand this already affects master when using SMTLIB, but merging this will also affect it when parsing TPTP, right?

shalashaska117:
> > I will, being Vampire a project so big I thought that issues in non-main branches were not worth being published
> 
> As far as I understand this already affects master when using SMTLIB, but merging this will also affect it when parsing TPTP, right?

Yes, just it was not exposed

mezpusz:
For full closure I ran some tests with `-newcnf on` too and there are 33 problems that previously crashed but now they don't. 🎉

PR #890, opened 2026-07-29
Fix eqModACRect renaming check and Injectivity success_3 expectation

### LINKED ISSUE #889: eqModACRect accepts clauses that are not equal up to renaming; Injectivity success_3 fails on gcc
`eqModACRect` accepts clause pairs that are not equal up to renaming, whenever the two clauses happen to share a subterm. That let a wrong expectation into `UnitTests/tInferences_HOL_Injectivity.cpp`, where it is green in CI under clang and red under gcc 15.2.

## The wrong expectation

```cpp
TEST_GENERATION(success_3,
    Generation::AsymmetricTest()
      .input( clause({ y.sort(srt) == z, ap(f, {x, z}) != ap(f, {x, y}) }))
      .EXPECTED(exactly(clause({ ap(inv(), {z, ap(f, {x, z})}) == x })))
    )
```

The input reads `f(x,z) = f(x,y) -> y = z`, so `f` is injective in its second argument and the left inverse is `inv(x, f(x,z)) = z`. The expectation says `inv(z, f(x,z)) = x`, which is `success_2`'s answer, where the differing argument is the first one instead. `Injectivity` never derives it, on any compiler.

## Why the compiler changes the derived clause

Equality arguments get normalised by term id in `TermSharing::argNormGt`, and ids are handed out in creation order. The two operands of `!=` are function call arguments, so C++ leaves their evaluation order unspecified. clang builds `f(x,z)` first, gcc builds `f(x,y)` first, so the two compilers put a different term on the lhs. `Injectivity::generateClauses` reads `differingArg` off the lhs, so it derives

* clang: `inv(X0, f(X0,X2)) = X2`
* gcc: `inv(X0, f(X0,X1)) = X1`

Both are correct and equal up to renaming. Neither is the expected clause.

## Why only gcc fails

`TestUtils::eqModAC_` opens with

```cpp
if (lhs == rhs) { return true; }
```

On shared terms that is pointer equality, and it returns without recording the variable renaming. Under clang the subterm `f(X0,X2)` is the same shared term in the derived and the expected clause, so the comparison short-circuits there. The only surviving constraints are `X0 -> X2` and `X2 -> X0`, and the permutation search satisfies both. Under gcc the derived subterm is `f(X0,X1)`, the short-circuit does not fire, the recursion demands `X0 -> X0` and `X1 -> X2`, and that contradicts `X0 -> X2`:

```
[     case ]: { X1 = X2 | ~(f @ X0 @ X1 = f @ X0 @ X2) }
[       is ]: [ { inv_f_0 @ X0 @ (f @ X0 @ X1) = X1 } ]
[ expected ]: exactly: [ { inv_f_0 @ X2 @ (f @ X0 @ X2) = X0 } ]
[ FAIL ] success_3
```

I checked the clang half rather than inferring it. Swapping only the written order of the two operands builds the same clause but reverses the creation order, and under gcc that makes the original expectation pass, which is the CI result.

The test expectation is a one line fix. The comparator is the part worth fixing: as it stands, any wrong expectation slips through as long as it shares a subterm with the derived clause, which is how this one got past review and CI.

## Also

The `TermList` overload of `eqModACRect` collects `vr` from `lhs`:

```cpp
auto vl = iterTraits(vi(new VariableIterator(lhs))).collect<Stack>();
...
auto vr = iterTraits(vi(new VariableIterator(lhs))).collect<Stack>();
```

so its variable count guard compares `lhs` against itself. The clause overload does not go through it, so it did not affect this test.

## Cost of dropping the fast path

Nothing measurable. With the fast path removed and the expectation corrected, the suite stays at 97/97 and around 20s, same as before.

PR to follow.

---

Disclosure: I used AI to generate this issue text. The bug was found out by myself testing for the PR #884

shalashaska117:
Fixes #889.

`TestUtils::eqModAC_` opened with

```cpp
if (lhs == rhs) { return true; }
```

On shared terms that is pointer equality, and it returned without recording the variable renaming, so `eqModACRect` accepted clause pairs that are not equal up to renaming whenever the two happened to share a subterm.

`success_3` in `tInferences_HOL_Injectivity` was riding on that. Its input makes `f` injective in its second argument, so the left inverse is `inv(x, f(x,z)) = z`, but the expectation was copied from `success_2` and says `inv(z, f(x,z)) = x`. Which clause `Injectivity` derives depends on the evaluation order of the two operands of `!=`, because equality arguments are normalised by term id. clang derives `inv(X0, f(X0,X2)) = X2`, which shares `f(X0,X2)` with the expectation and short-circuits, so CI stayed green. gcc derives `inv(X0, f(X0,X1)) = X1`, and the test fails.

Three changes:

* Drop the fast path in `eqModAC_`.
* Correct the `success_3` expectation.
* Fix the `TermList` overload of `eqModACRect`, which collected `vr` from `lhs` and so compared `lhs` against itself in its variable count guard. Nothing in the suite goes through that overload, so it is a separate bug I noticed while reading rather than something this PR needed.

Checked with gcc 15.2, Debug: 97/97 in about 20s, unchanged from before, so no other test was relying on the fast path.

With the fast path gone the old expectation also fails under clang's ordering. I reproduced that ordering under gcc by swapping the written order of the two operands, which builds the same clause but reverses the creation order:

```
[     case ]: { X1 = X2 | ~(f @ X0 @ X2 = f @ X0 @ X1) }
[       is ]: [ { inv_f_0 @ X0 @ (f @ X0 @ X2) = X2 } ]
[ expected ]: exactly: [ { inv_f_0 @ X2 @ (f @ X0 @ X2) = X0 } ]
[ FAIL ] success_3
```

---

Disclosure: I used AI to generate this PR text. The code has been wrote by myself.

### DIFF
--- a/Test/TestUtils.cpp
+++ b/Test/TestUtils.cpp
@@ -204,7 +204,9 @@ Stack<TermList> collect(unsigned functor, Term* t) {
 template<class Comparisons>
 bool TestUtils::eqModAC_(TermList lhs, TermList rhs, Comparisons comp)
 {
-  if (lhs == rhs) { return true; }
+  // NB: deliberately no `lhs == rhs` fast path. Two syntactically identical
+  // subterms still have to agree with the renaming `comp` is building up, so
+  // returning early here lets an inconsistent renaming through.
   if (lhs.isVar() && rhs.isVar()) {
     return comp.var(lhs.var(), rhs.var());
   } else if (lhs.isTerm() && rhs.isTerm()) {
@@ -270,7 +272,7 @@ bool TestUtils::eqModACRect(Kernel::TermList lhs, Kernel::TermList rhs)
   auto vl = iterTraits(vi(new VariableIterator(lhs))).collect<Stack>();
   vl.sort();
   vl.dedup();
-  auto vr = iterTraits(vi(new VariableIterator(lhs))).collect<Stack>();
+  auto vr = iterTraits(vi(new VariableIterator(rhs))).collect<Stack>();
   vr.sort();
   vr.dedup();
 
--- a/UnitTests/tInferences_HOL_Injectivity.cpp
+++ b/UnitTests/tInferences_HOL_Injectivity.cpp
@@ -110,8 +110,9 @@ TEST_GENERATION(success_2,
       .EXPECTED(exactly(clause({ ap(inv(), {z, ap(f, {x, z})}) == x })))
     )
 
+// f is injective in its second argument, so the inverse takes the first
 TEST_GENERATION(success_3,
     Generation::AsymmetricTest()
       .input( clause({ y.sort(srt) == z, ap(f, {x, z}) != ap(f, {x, y}) }))
-      .EXPECTED(exactly(clause({ ap(inv(), {z, ap(f, {x, z})}) == x })))
+      .EXPECTED(exactly(clause({ ap(inv(), {x, ap(f, {x, z})}) == z })))
     )


mezpusz (approved):
The tests are seemingly not slowed down by the removed check, so I approve. Thanks for the fix!

mezpusz:
This should also fix the coverage build, rerunning now 🤞 https://github.com/vprover/vampire/actions/runs/30460258799

PR #892, opened 2026-08-03
Add predicate elimination as a new preprocessing step

quickbeam123:
Implements the technique of Khasidashvili and Korovin (SAT 2016): a predicate P occurring at most once in every clause is eliminated by replacing S_P and S_~P with all pairwise resolvents on P. On problems without equality and theories, resolvents are computed with an mgu and non-unifiable pairs dropped; otherwise the P-literals are (virtually) flattened, introducing argument disequalities which are then simplified away by exhaustive equality substitution (this can introduce equality into a problem previously without it). FMB forces the equational mode, since its model reconstruction cannot rely on the Herbrand-interpretation argument justifying the mgu variant.

Elimination steps are gated SAT-VE-style by growth limits on the estimated clause count |S_P|*|S_~P| - |S_P| - |S_~P|: a  global cap relative to the original total (-peltl, default 2.0). Syntactic tautologies (complementary pair, s = s) and duplicate literals are removed from generated resolvents, and the actual surviving count feeds back into the budget.

The next predicate to eliminate is by default the one with the smallest estimated growth (pure predicates thus go first, their clauses being simply deleted)

With -pels, the clause set is kept forward-inter-subsumed throughout, using a standalone LiteralSubstitutionTree index (unit literal, or the least matchable one) plus SATSubsumptionAndResolution; backward subsumption is left as future work.

Every elimination records a model-repairing definition
  P(xs) <=> \/_{D \/ P(ts) in S_P} exists ys. (xs = ts /\ ~D)
via Problem::addEliminatedPredicate (addTrivialPredicate for pure ones), so both the textual model updates output and FMB model restoration stay correct (verified via --mode model_check).

Skipped for higher-order/polymorphic inputs (predicates could hide inside terms, breaking the occurrence counting) and for color-annotated problems.

### DIFF
--- a/Shell/PredicateElimination.cpp
+++ b/Shell/PredicateElimination.cpp
@@ -117,7 +117,7 @@ void PredicateElimination::apply(Problem &prb)
       }
     }
     _all.push(cl);
-    handleClause(cl, /*add=*/true);
+    handleClause</*add=*/true>(cl);
     if (_useSubsumption) {
       indexInsert(cl);
     }
@@ -134,7 +134,7 @@ void PredicateElimination::apply(Problem &prb)
 
   if (_modified) {
     UnitList *res = 0;
-    for (const auto& cl : _all) {
+    for (const auto& cl : iterTraits(ClauseStack::Iterator(_all))) {
       if (!_deleted.contains(cl)) {
         UnitList::push(cl, res);
       }
@@ -147,7 +147,8 @@ void PredicateElimination::apply(Problem &prb)
   }
 }
 
-void PredicateElimination::handleClause(Clause *cl, bool add)
+template<bool add>
+void PredicateElimination::handleClause(Clause *cl)
 {
   static DHMap<unsigned, int> occ;
   occ.reset();
@@ -167,9 +168,9 @@ void PredicateElimination::handleClause(Clause *cl, bool add)
   }
 
   for (const auto& [pred, val] : iterTraits(occ.items())) {
-    if (add) {
+    if constexpr (add) {
       if (val == 2) {
-      _preds[pred].blockers++;
+        _preds[pred].blockers++;
       }
       else if (val == 1) {
         _preds[pred].pos.insert(cl);
@@ -290,14 +291,14 @@ void PredicateElimination::eliminate(Problem &prb, unsigned pred)
   }
 
   for (Clause *cl : posCls) {
-    handleClause(cl, /*add=*/false);
+    handleClause</*add=*/false>(cl);
     if (_useSubsumption) {
       indexRemove(cl);
     }
     _deleted.insert(cl);
   }
   for (Clause *cl : negCls) {
-    handleClause(cl, /*add=*/false);
+    handleClause</*add=*/false>(cl);
     if (_useSubsumption) {
       indexRemove(cl);
     }
@@ -307,7 +308,7 @@ void PredicateElimination::eliminate(Problem &prb, unsigned pred)
 
   for (Clause *r : resolvents) {
     _all.push(r);
-    handleClause(r, /*add=*/true);
+    handleClause</*add=*/true>(r);
     if (_useSubsumption) {
       indexInsert(r);
     }
@@ -486,7 +487,7 @@ void PredicateElimination::recordElimination(Problem &prb, unsigned pred,
   VarShiftApplicator shift{ar}; // clause variables become >= ar, i.e. disjoint from the head's
 
   FormulaList *disjuncts = FormulaList::empty();
-  for (const auto& c : posCls) {
+  for (const auto& c : iterTraits(ClauseStack::ConstIterator(posCls))) {
     Literal *plit = findPredLiteral(c, pred, true);
 
     FormulaList *conjuncts = FormulaList::empty();
--- a/Shell/PredicateElimination.cpp
+++ b/Shell/PredicateElimination.cpp
@@ -25,7 +25,6 @@
 #include "Kernel/SortHelper.hpp"
 #include "Kernel/SubstHelper.hpp"
 #include "Kernel/Term.hpp"
-#include "Kernel/Unit.hpp"
 
 #include "Inferences/InferenceEngine.hpp"
 #include "Inferences/TautologyDeletionISE.hpp"
@@ -86,11 +85,8 @@ void PredicateElimination::apply(Problem &prb)
   Inferences::TautologyDeletionISE tautologyDeletion;
 
   ClauseStack input;
-  UnitList::Iterator uit(prb.units());
-  while (uit.hasNext()) {
-    Unit *u = uit.next();
-    ASS(u->isClause());
-    Clause *cl = static_cast<Clause *>(u);
+  for (const auto& u : iterTraits(UnitList::Iterator(prb.units()))) {
+    Clause *cl = u->asClause();
     Clause *simp = tautologyDeletion.simplify(duplicateLiteralRemoval.simplify(cl));
     if (simp != cl) {
       _modified = true;
@@ -121,7 +117,7 @@ void PredicateElimination::apply(Problem &prb)
       }
     }
     _all.push(cl);
-    registerClause(cl);
+    handleClause(cl, /*add=*/true);
     if (_useSubsumption) {
       indexInsert(cl);
     }
@@ -138,9 +134,7 @@ void PredicateElimination::apply(Problem &prb)
 
   if (_modified) {
     UnitList *res = 0;
-    ClauseStack::Iterator it(_all);
-    while (it.hasNext()) {
-      Clause *cl = it.next();
+    for (const auto& cl : _all) {
       if (!_deleted.contains(cl)) {
         UnitList::push(cl, res);
       }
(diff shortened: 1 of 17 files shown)


quickbeam123:
The original commit message is preserved in the commit but describes more than what I eventually decided to keep. (I used interactive rebasing and "fixup" to hide my update commits, not sure if for the better.)

A TPTP FOL discount comparison is now:
```
Sort by SAT
1025 ['problemsSTD_pel10964_dis10_i10K.pkl']
1073 ['problemsSTD_pel10964_dis10_pel_i10K.pkl']

Sort by UNS
8406 ['problemsSTD_pel10964_dis10_i10K.pkl']
8591 ['problemsSTD_pel10964_dis10_pel_i10K.pkl']

Total 1108 / 8826

Num only-by (UNS):
   problemsSTD_pel10964_dis10_i10K.pkl 235
   problemsSTD_pel10964_dis10_pel_i10K.pkl 420

Greedy cover (UNS):
# problemsSTD_pel10964_dis10_pel_i10K.pkl contributes 8591 total 8591
# problemsSTD_pel10964_dis10_i10K.pkl contributes 235 total 8406
Total 8826
```

Overall, I already reviewed this and as a separate module in preprocessing, this shouldn't cause any problems.

samplers/samplerFNT.smp
```diff
@@ -33,6 +33,15 @@ $nm=NZ > nm ~sgd 0.07,2
 $ins=Z > ins ~cat 0:1
 $ins=NZ > ins ~sgd 0.4,1
 
+# predicate_elimination
```
mezpusz, line 36: Looking at the duplicate changes in multiple sampler files, do you think some kind of inclusion mechanism could be added to the sampler syntax? We could then move out the common parts to separate files.

easychair:
This predicate elimination technique was implemented in very early versions
of Vampire, I think as early as in 1999. I am not claiming it is exactly
the same as when K&K published, but it was in the code. The elimination was
done for (1 x n) occurrences or (2 x 2) occurrences because in these two
cases the number of clauses does not increase (1 + n clauses are replaced
by n and 4 clauses replaced by 4.

There is a caveat though: you don't want to have definition introduction
during formula preprocessing following predicate definition elimination,
which effectively undoes the introduction.

A

samplers/samplerFNT.smp
MichaelRawson: That's a good idea. We now have a few different homebrew Vampire input files - do you know of some semi-standardised configuration language?

Shell/PredicateElimination.cpp
mezpusz, line 1: Would you mind adding these files to a directory `Transformations` (or `Preprocessing`) rather than `Shell`? I'd like to start separating preprocessing transformations into their own directory as I add unit tests for them in the future.

Shell/PredicateElimination.hpp
```diff
@@ -0,0 +1,132 @@
...
+ * Predicate elimination for preprocessing of clausified problems,
+ * after Khasidashvili and Korovin: "Predicate Elimination for Preprocessing
+ * in First-Order Theorem Proving" (SAT 2016).
+ *
+ * A predicate P which occurs at most once in every clause (is "non-self-referential")
+ * can be eliminated by replacing the clauses S_P and S_~P (those containing P positively,
+ * respectively, negatively) by all the pairwise resolvents on P. In the presence
+ * of equality (or theories), the resolvents need to be computed via (virtual) flattening
+ * of the P-literals, i.e. C \/ P(ts) and D \/ ~P(ss) yield C \/ D' \/ t1 != s1' \/ ... \/ tn != sn'
+ * (with D renamed apart), simplified by the equality substitution rule
```
mezpusz: Are the `si` renamed apart too?

Shell/PredicateElimination.hpp
```diff
@@ -0,0 +1,132 @@
...
+   * @param forceEquationally  force the flattening-based resolvent computation even on
+   *        problems without equality/theories (required under FMB, whose model
+   *        reconstruction cannot rely on the Herbrand-interpretation argument
+   *        that justifies the mgu mode)
+   * @param totalLimit  an elimination step is admissible only if the estimated
+   *        number of clauses afterwards does not exceed the initial number times this factor
+   * @param useSubsumption  keep the clause set forward-inter-subsumed and
+   *        subsumption-resolved
+   */
+  PredicateElimination(bool forceEquationally, float totalLimit, bool useSubsumption)
```
mezpusz, line 63: Could `totalLimit` be an integer?

Shell/PredicateElimination.hpp
```diff
@@ -0,0 +1,132 @@
...
+private:
+  struct PredInfo {
+    Lib::DHSet<Clause *> pos; // S_P: clauses in which P occurs exactly once, positively
+    Lib::DHSet<Clause *> neg; // S_~P: dtto, negatively
+    unsigned blockers = 0;    // number of clauses in which P occurs more than once
+    bool eliminated = false;
+  };
+
+  // options
+  bool _forceEquationally;
```
mezpusz: These could be `const`.

Shell/PredicateElimination.hpp
```diff
@@ -0,0 +1,132 @@
...
+  float _totalLimit;
+  bool _useSubsumption;
+
+  // clause set state
+  Lib::Stack<Clause *> _all;     // all clauses ever seen, in insertion order
```
mezpusz: This could be `ClauseStack`.

Shell/PredicateElimination.cpp
```diff
@@ -0,0 +1,694 @@
...
+using namespace Lib;
+using namespace Kernel;
+using namespace Indexing;
+
+/**
+ * Renames variables by adding a fixed offset.
+ */
+struct VarShiftApplicator {
+  unsigned off;
+  TermList apply(unsigned var) const { return TermList(var + off, false); }
```
mezpusz: Consider using `TermList::var(var + off)`.

Shell/PredicateElimination.cpp
```diff
@@ -0,0 +1,694 @@
...
+};
+
+/**
+ * Replaces a single variable by a term, leaving other variables intact.
+ */
+struct SingleVarApplicator {
+  unsigned var;
+  TermList term;
+  TermList apply(unsigned v) const { return v == var ? term : TermList(v, false); }
```
mezpusz: Consider using `TermList::var(v)`

Shell/PredicateElimination.hpp
quickbeam123: Not unless we make some convention, like that it is in percent of the original size. We are using this to express, stop eliminating if the estimated new size grows to, e.g., 1.02 multiple of the original size...

Shell/PredicateElimination.hpp
```diff
@@ -0,0 +1,132 @@
...
+  // resolvent construction; nullptr means no clause results (tautology / non-unifiable pair)
+  Clause *buildResolvent(Clause *c, Literal *plitC, Clause *d, Literal *plitD);
+  Clause *buildResolventMgu(Clause *c, Literal *plitC, Clause *d, Literal *plitD);
+  Clause *buildResolventEq(Clause *c, Literal *plitC, Clause *d, Literal *plitD);
+  Clause *assembleClause(Lib::Stack<Literal *> &lits, Clause *c, Clause *d);
+
+  // model reconstruction
+  void recordElimination(Problem &prb, unsigned pred, Lib::Stack<Clause *> const &posCls, Lib::Stack<Clause *> const &negCls);
+
+  // forward subsumption (and subsumption resolution) machinery, only used with _useSubsumption;
```
mezpusz, line 113: Do you think this could be changed to code tree subsumption? It could be a bit simpler (and even more efficient!).

Shell/PredicateElimination.hpp
mezpusz: You're right, I misunderstood the meaning.

Shell/PredicateElimination.hpp
quickbeam123: Yes, fixing the comment.

Shell/PredicateElimination.cpp
quickbeam123: Could we do this as a separate PR?

Not sure how the waiting PRs might cope with files they modify moving ...

Shell/PredicateElimination.hpp
quickbeam123: Should be there now, please check whether it couldn't be simplified further!

easychair:
This restriction can be relaxed:


> A predicate P which occurs at most once in every clause (is "non-self-referential")

For example, one can eliminate

  p(s) \/ p(t) \/ C

against

  p(r) \/ D

The restriction should be

1. Either all positive occurrences of p occur at most once in every
clause or all negative occurrences occur at most once in every clause

2. There is no clause with both positive and negative occurrences of p


Andrei

samplers/samplerFNT.smp
quickbeam123: We could also have global `if-then-else`s, so that half of the sampler files is not littered by `sa!=fmb >` and co.

A separate PR?

Shell/PredicateElimination.hpp
mezpusz: Looks good, thanks!

mezpusz (approved):

Shell/Preprocess.cpp
```diff
@@ -466,6 +467,24 @@ void Preprocess::preprocess(Problem& prb)
      bce.apply(prb);
    }
 
+   if (_options.predicateElimination()) {
+     if (prb.isHigherOrder() || prb.hasPolymorphicSym()) { // in both cases, predicates could hide inside terms, breaking the occurrence counting
```
MichaelRawson, line 471: It's not clear to me why *polymorphism* would be a problem. You'd have to be careful not to introduce equalities between sorts (although one could imagine supporting this usefully?!) but otherwise should be alright, no?

Shell/PredicateElimination.cpp
```diff
@@ -0,0 +1,606 @@
...
+};
+
+/**
+ * Replaces a single variable by a term, leaving other variables intact.
+ */
+struct SingleVarApplicator {
+  unsigned var;
+  TermList term;
+  TermList apply(unsigned v) const { return v == var ? term : TermList::var(v); }
+};
```
MichaelRawson, line 64: Maybe these can be lambdas or moved into e.g. `SubstHelper` if they're useful so everyone can benefit?

Shell/PredicateElimination.cpp
```diff
@@ -0,0 +1,606 @@
...
+  _equational = _forceEquationally || prb.hasEquality() || prb.hasInterpretedOperations() || prb.hasNumerals();
+
+  _preds.ensure(env.signature->predicates());
+
+  // clauses may still contain duplicate literals (or be tautologies) at this stage of
+  // preprocessing; besides making our occurrence counting needlessly conservative, they would,
+  // more importantly, violate an invariant of the multi-literal matching in ClauseCodeTree
+  // (which in saturation is maintained by running these very simplifications on every new clause)
+  Inferences::DuplicateLiteralRemovalISE duplicateLiteralRemoval;
+  Inferences::TautologyDeletionISE tautologyDeletion;
```
MichaelRawson, line 85: Do we perhaps want to do this at some point(s) during preprocessing? Saves handling this case here (and maybe elsewhere), some compute and maybe improves e.g. SInE? Similarly with subsumption.

MichaelRawson (approved):
Cool! I wondered if something similar could be achieved *during* saturation by literal selection - but it's not the same thing at all.

Shell/Preprocess.cpp
quickbeam123: You might be correct. But Claude decided to skip it and I was fine with that. The extra cognitive load for making sure polymorphism is handled correctly versus the expected impact on solving more problems did not seem to pay off... Could be a nice exercise for a motivated volunteer ;)

Shell/PredicateElimination.cpp
quickbeam123: I am not sure. Isn't this too simple to nudge into sharing? At the same time, with a class it's simpler to reuse a single instance at multiple locations as being done here (compared to a lambda).

Shell/PredicateElimination.cpp
quickbeam123: Saturation will do everything in the first pass anyway. So the question should be: is any other preprocessing step bothered by these not being done? Maybe BCE could benefit, but that's optional. And so is predicate elimination. So having them local is a manifestation of "don't impose on others something they might not need"...

(SInE is run on formulas, this waits for CNF.)

PR #894, opened 2026-08-04
fix wrong answer from -newcnf on with two or more Boolean lets

### LINKED ISSUE #893: Wrong answer from -newcnf on with two or more Boolean lets
Splitting this out of #885 as requested in https://github.com/vprover/vampire/pull/885#issuecomment-5164813279.

`-newcnf on` gives a wrong answer for a formula containing two or more Boolean lets. A single Boolean let is handled correctly.

It reproduces on current master with no parser change: the SMT-LIB parser has always built Boolean let definitions this way, so NewCNF sees the same shape there.

```smt2
(set-logic ALL)
(declare-fun q () Bool)
(assert (not (and (let ((b (or q (not q)))) b) (let ((d (or q (not q)))) d))))
(check-sat)
```

The assertion is the negation of a tautology, so the expected answer is Unsatisfiable, and that is what the default clausifier answers. `vampire --input_syntax smtlib2 -newcnf on` answers Satisfiable. Same result on master and on the #885 branch.

The TPTP form needs #885 first, because on master any Boolean $let crashes the TPTP parser (#863). With that fix applied, `-newcnf on` reports this Theorem as CounterSatisfiable:

```tptp
tff(q,type, q: $o ).
tff(c,conjecture,
    ( $let(b: $o, b := ( q | ~ q ), b )
    & $let(d: $o, d := ( q | ~ q ), d ) ) ).
```

The regression tests in #885 use one Boolean $let per file on purpose, so they pass in all four clausifier configurations and don't encode this bug as expected behaviour.

shalashaska117:
Closes #893.

NewCNF stores formulas in generalised clauses with any leading negation stripped and the sign of the generalised literal flipped instead (pushLiteral). `Occurrences::replaceBy` broke that invariant: it wrote the formula in unchanged, and `enqueue` compensated by dropping the negation from the queue key and flipping the occurrence signs. The slot was left holding `~A` with a sign that every later lookup reads as the sign of `A`.

processLet produces that state whenever ennf has pushed a negation into a let body. With one $let the slot is only ever rewritten in place, so nothing shows. With two, clausifying the first copies the second's slot into new generalised clauses through pushLiteral, which strips the negation again and flips the sign a second time. The second definition gets the wrong polarity, its clauses come out as tautologies, and tautologies are deleted, so the whole problem clausifies to nothing and saturation reports a model. On master (9f4717263, Release):

```
$ vampire --input_syntax smtlib2 -newcnf on checks/theory/let-bool-twice.smt2
% SZS status Satisfiable for let-bool-twice
$ vampire -newcnf on checks/theory/let-bool-twice.p
% SZS status CounterSatisfiable for let-bool-twice
$ vampire -newcnf on -ile off checks/theory/let-bool-twice.p
% SZS status Theorem for let-bool-twice
```

`-ile off` was already correct: the naming path never hands replaceBy a negated formula there.

The stale `~A` also breaks the occurrence count that pop maintains by looking the slot formula up in `_occurrences`. A Debug build dies on the same input without printing anything; under gdb:

```
#5  Debug::Assertion::violatedEquality<unsigned int, int> (val1Str="_size", val2Str="0", val1=1, val2=0)
#6  Shell::NewCNF::Occurrences::isNonEmpty at Shell/NewCNF.hpp:327
#7  Shell::NewCNF::process(JunctionFormula*, Occurrences&) at Shell/NewCNF.cpp:449
```

The fix handles the negation where the formula is written into the clause: replaceBy drops it and flips the sign of every occurrence, the same way it already normalises negative shared literals, and enqueue leaves the signs alone. Its invert() helper lost its only caller and is removed. process(QuantifiedFormula*) uses the same replaceBy/enqueue pair on the quantifier body, so it is covered by the same change.

With the fix, Debug build with assertions enabled:

```
PASS: twice.smt2 default
PASS: twice.smt2 newcnf
PASS: twice.smt2 newcnf ile off
PASS: twice.p default
PASS: twice.p newcnf
PASS: twice.p ile off
PASS: twice.p newcnf ile off
PASS: let-bool.p newcnf
PASS: let-bool.p newcnf ile off
PASS: let-bool-pred.p newcnf
PASS: let-bool-pred.p newcnf ile off
PASS: let-bool-simultaneous.p newcnf
PASS: let-function.p newcnf
PASS: let-tuple.p newcnf
PASS: let-tuple-bool.p newcnf
```

Also checked in default, `-newcnf on` and `-newcnf on -ile off`: three sibling lets and nested lets (Unsatisfiable), a satisfiable two-let problem (still Satisfiable), and two $lets under a universal quantifier (Theorem). checks/sanity passes with a Release+Z3 build and ctest is 97/97 with the Debug build. The new checks cover the TPTP and SMT-LIB shape of the problem in the same configurations as the other let-bool files.

### DIFF
--- a/Shell/NewCNF.hpp
+++ b/Shell/NewCNF.hpp
@@ -354,11 +354,23 @@ private:
       Occurrences::Iterator occit(*this);
 
       bool negateOccurrenceSign = false;
+      if (f->connective() == NOT) {
+        /**
+         * Generalised clauses store formulas without negations (cf. pushLiteral),
+         * so the negation is dropped here and the sign of every occurrence is
+         * flipped instead. Otherwise the stored formula would not match the
+         * formula the occurrences are registered under in _occurrences, and
+         * pushLiteral would flip the sign a second time when the generalised
+         * literal is copied to another generalised clause.
+         */
+        f = f->uarg();
+        negateOccurrenceSign = true;
+      }
       if (f->connective() == LITERAL) {
         Literal* l = f->literal();
         if (l->shared() && ((SIGN)l->polarity() != POSITIVE)) {
           f = new AtomicFormula(Literal::complementaryLiteral(l));
-          negateOccurrenceSign = true;
+          negateOccurrenceSign = !negateOccurrenceSign;
         }
       }
 
@@ -369,15 +381,6 @@ private:
         if (negateOccurrenceSign) {
           sign(gl) = OPPOSITE(sign(gl));
         }
-      }
-    }
-
-    void invert() {
-      Occurrences::Iterator occit(*this);
-      while (occit.hasNext()) {
-        Occurrence occ = occit.next();
-        GenLit& gl = occ.gc->_literals[occ.position];
-        sign(gl) = OPPOSITE(sign(gl));
       }
     }
 
@@ -568,19 +571,19 @@ private:
   void nameSubformula(Formula* g, Occurrences &occurrences);
 
   void enqueue(Formula* formula, Occurrences occurrences = Occurrences()) {
-    if ((formula->connective() == LITERAL) && formula->literal()->shared()) return;
-
     if (formula->connective() == NOT) {
       /**
        * Formulas are always stored without negations in genclauses,
-       * therefore it is safe to drop the negation before queueing,
-       * all the occurrences of the formula won't have it either
+       * therefore it is safe to drop the negation before queueing.
+       * The signs of the occurrences are not touched here: whoever puts
+       * a formula into a generalised clause is responsible for dropping
+       * its negation and flipping the sign of the generalised literal
+       * (cf. pushLiteral and Occurrences::replaceBy).
        */
       formula = formula->uarg();
-      ASS_REP(formula->connective() != LITERAL, formula->toString());
-
-      occurrences.invert();
     }
+
+    if ((formula->connective() == LITERAL) && formula->literal()->shared()) return;
 
     if (_occurrences.find(formula)) {
       Occurrences oldOccurrences;
--- a/checks/sanity
+++ b/checks/sanity
@@ -95,6 +95,12 @@ check_szs_status Theorem -ile off theory/let-bool-pred.p
 check_szs_status Theorem -newcnf on -ile off theory/let-bool-pred.p
 check_szs_status Theorem theory/let-bool-simultaneous.p
 check_szs_status Theorem -newcnf on theory/let-bool-simultaneous.p
+check_szs_status Theorem theory/let-bool-twice.p
+check_szs_status Theorem -newcnf on theory/let-bool-twice.p
+check_szs_status Theorem -ile off theory/let-bool-twice.p
+check_szs_status Theorem -newcnf on -ile off theory/let-bool-twice.p
+check_szs_status Unsatisfiable theory/let-bool-twice.smt2
+check_szs_status Unsatisfiable -newcnf on theory/let-bool-twice.smt2
 
 # Arrays
 check_szs_status Unsatisfiable theory/arrays.smt2
--- a/checks/theory/let-bool-twice.p
+++ b/checks/theory/let-bool-twice.p
@@ -1,0 +1,12 @@
+% Two Boolean $lets in one conjecture. NewCNF indexed each inlined definition
+% without its leading negation while the generalised clause kept it, so
+% clausifying one $let copied the other's literal and flipped its sign a
+% second time. Its clauses became tautologies and were dropped, and -newcnf on
+% answered CounterSatisfiable. Both definitions are tautologies, so the
+% conjecture is a theorem.
+tff(q,type,
+    q: $o ).
+
+tff(let_bool_twice,conjecture,
+    ( $let(b: $o, b := ( q | ~ q ), b )
+    & $let(d: $o, d := ( q | ~ q ), d ) ) ).
--- a/checks/theory/let-bool-twice.smt2
+++ b/checks/theory/let-bool-twice.smt2
@@ -1,0 +1,8 @@
+; The same shape as let-bool-twice.p in SMT-LIB. The SMT-LIB parser has
+; always built Boolean let definitions as formulas, so this wrong answer
+; predates Boolean $let support in the TPTP parser. The assertion negates a
+; conjunction of two tautologies, hence unsatisfiable.
+(set-logic ALL)
+(declare-fun q () Bool)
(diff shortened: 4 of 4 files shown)


shalashaska117:
@MichaelRawson @mezpusz @quickbeam123 could one of you take a look at this when you get a chance?

It is a soundness fix rather than a crash fix, which is why I would like eyes on it: with two or more Boolean `$let` bindings, `-newcnf on` could return the wrong result. `Occurrences::replaceBy` was storing negated formulas into generalised clauses, breaking the invariant that those hold none.

mezpusz:
I will try to run regression over TPTP and SMTLIB for this today. Since it touches a common path, it might potentially affect a lot of benchmarks with `-newcnf on`, so we have to be careful.

MichaelRawson:
Take your time, @mezpusz.

@shalashaska117 thanks for the fix. As you are a new contributor we (I) are trying to be prompt with review but generally Vampire moves quite slowly, so a few weeks is not uncommon. I try and avoid months turning into years...

There again, the kind of code-quality improvements you are doing helps to improve velocity, so there is hope for the future. Pepperidge farm remembers a time when things were *really* gunky.

shalashaska117:
> Take your time, @mezpusz.
> 
> @shalashaska117 thanks for the fix. As you are a new contributor we (I) are trying to be prompt with review but generally Vampire moves quite slowly, so a few weeks is not uncommon. I try and avoid months turning into years...
> 
> There again, the kind of code-quality improvements you are doing helps to improve velocity, so there is hope for the future. Pepperidge farm remembers a time when things were *really* gunky.

Being august I have plenty of time beside jobs and university to spend on this project, now that I have to the time it's also easier for me to study it deeper doing these PRs, but no problem I will make everything that I can as a draft and with things merging I will make them open!

mezpusz:
Regression over SMTLIB resulted in no change, while regression over TPTP showed some difference:

These problems were "lost" under `-sa discount`:
```
ITP/ITP211_4.p
ITP/ITP212_4.p
ITP/ITP216_4.p
ITP/ITP233_4.p
ITP/ITP261_4.p
ITP/ITP277_4.p
ITP/ITP278_4.p
SYN/SYN854-1.p
```
and these under `-sa otter` (mostly the same):
```
ITP/ITP211_4.p
ITP/ITP212_4.p
ITP/ITP213_4.p
ITP/ITP216_4.p
ITP/ITP233_4.p
ITP/ITP261_4.p
ITP/ITP277_4.p
ITP/ITP278_4.p
```

I've checked superficially that they result in slightly different clausifications (usually one polarity switch in a few clauses). @shalashaska117 could you please verify this?

shalashaska117:
Verified. Master's clausification is the wrong one of the two, and the eight ITP problems were being solved through the bug.

With `-newcnf on`, every one of master's answers on those eight is ContradictoryAxioms, reached in 0.3 to 6.3 s. None of them is a proof that uses the conjecture.

| problem | master, discount | branch, discount | master, otter | branch, otter |
| --- | --- | --- | --- | --- |
| ITP211_4 | ContradictoryAxioms 0.33 s | none in 60 s | ContradictoryAxioms 0.54 s | none in 60 s |
| ITP212_4 | ContradictoryAxioms 0.29 s | none in 60 s | ContradictoryAxioms 0.42 s | none in 60 s |
| ITP213_4 | ContradictoryAxioms 6.26 s | ContradictoryAxioms 5.77 s | ContradictoryAxioms 1.47 s | ContradictoryAxioms 1.81 s |
| ITP216_4 | ContradictoryAxioms 0.36 s | Theorem 5.99 s | ContradictoryAxioms 0.57 s | Theorem 13.46 s |
| ITP233_4 | ContradictoryAxioms 0.66 s | none in 60 s | ContradictoryAxioms 0.91 s | none in 60 s |
| ITP261_4 | ContradictoryAxioms 0.83 s | none in 60 s | ContradictoryAxioms 1.38 s | none in 60 s |
| ITP277_4 | ContradictoryAxioms 0.95 s | none in 60 s | ContradictoryAxioms 1.23 s | none in 60 s |
| ITP278_4 | ContradictoryAxioms 0.78 s | Theorem 2.93 s | ContradictoryAxioms 1.17 s | Theorem 4.79 s |
| SYN854-1 | Satisfiable 1.39 s | Satisfiable 1.39 s | Satisfiable 2.00 s | Satisfiable 2.02 s |

The polarity switch is where the contradiction comes from. In ITP211_4 the refutation uses

```
tff(f23844,plain,(
  ( ! [X1 : fun($o,$o)] : (~member($o,$true,aa(set($o),set($o),aa($o,fun(set($o),set($o)),insert2($o),$true),bot_bot(set($o)))) | ($true = aa($o,$o,X1,aa(product_unit,$o,product_Rep_unit,sK1417(X1)))) | ($true = aa($o,$o,X1,$true))) )),
  inference(cnf_transformation,[],[f14463])).
```

and f14463 is fact_4757_Rep__unit__induct after ennf:

```
! [X0 : $o,X1 : fun($o,$o)] : (aa($o,$o,X1,X0) | ? [X2 : product_unit] : ~aa($o,$o,X1,aa(product_unit,$o,product_Rep_unit,X2)) | ~member($o,X0,...))
```

The existential ranges over a negated atom, so the Skolemised literal has to be negative. Master emits it positive, which makes the clause stronger than the axiom it came from and lets the axiom set refute itself. The branch emits `$true != aa(...)`. ITP212_4 and ITP216_4 carry the same two clauses with the same flip.

Two of the eight are therefore not lost at all: ITP216_4 and ITP278_4 come out as Theorem on the branch, a bogus contradiction replaced by a real proof. ITP213_4 gives ContradictoryAxioms on both sides, so that one looks like genuinely inconsistent axioms rather than a clausification artefact. The other five are no longer solved inside 60 s, which is the shortcut disappearing.

SYN854-1 is unrelated. It is pure CNF, 3198 clauses with no `$let` or `$ite`, its clausification is byte-identical on both sides, and both answer Satisfiable in the same time.

Nothing outside `-newcnf on` moves: under default options the clausifications of all nine problems are byte-identical, and neither binary solves ITP211_4 in 60 s. Since the option is off by default, it is worth confirming that the regression ran newcnf-enabled strategies.

As a cross-check I generated 1008 small problems made of several Boolean lets under a connective and took `-newcnf off` as the oracle. Master with `-newcnf on` answers 4 of them wrongly, in both directions, including this satisfiable one reported Unsatisfiable:

```
(not (and (let ((v0 (and q (not q)))) v0) (let ((v1 (or q (not q)))) v1)))
```

The branch answers all 1008 correctly, and a further 200 randomly generated let/ite formulas produce no disagreement on either side.

mezpusz (approved):

PR #896, opened 2026-08-05
Randomized simplifications

quickbeam123:
The same spirit as randomized preprocessing.

This "leaks" the core action for `fd`, `fsr`, `bd`, `bsr` and avatar under some small probabilities.

### DIFF
--- a/Inferences/BackwardDemodulation.cpp
+++ b/Inferences/BackwardDemodulation.cpp
@@ -15,6 +15,7 @@
 #include "Lib/DHMultiset.hpp"
 #include "Lib/Environment.hpp"
 #include "Lib/Metaiterators.hpp"
+#include "Lib/Random.hpp"
 #include "Debug/TimeProfiling.hpp"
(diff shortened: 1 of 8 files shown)


MichaelRawson:
(now has conflicts somehow)

easychair:
What's the point in all these options?

Why not simply randomly deleting all clauses?

A

MichaelRawson (approved):
Alright! At some point in future I want to make inferences 'aspect-oriented', i.e. that this kind of logic would be done automatically outside of any specific inference. But this looks interesting in the meantime.

quickbeam123:
The point is to observe the middle ground between always performing and never performing. (Since we have option for the extremes and both values seem useful sometimes.) At the very least, this extends our toolkit for fuzzing, i.e., looking for bugs in randomized traces.

easychair:
I don't see enough justification for it to be in master. One can randomise
everything, thus adding 1,000 new options to an already overcrowded option
space.

A

Issue #898, opened 2026-08-12
make container hash functions explicit and remove DefaultHash


shalashaska117:
Lib/Hash.hpp picks a hash for each key type with a chain of enable_if overloads on DefaultHash and DefaultHash2: detection of defaultHash()/defaultHash2() methods, a special case for Unit-derived pointers, and fallbacks for other pointers, scalars, strings, containers, pairs and tuples. Map, Set, DHMap, DHSet, DHMultiset and BiMap all default their hash parameters to these two classes, so the hash a key ends up with is invisible at the use site. It was suggested off-list that users supply the hash explicitly and DefaultHash be removed; this issue tracks that work.

The tree has 594 container instantiations; 532 rely on the defaults: roughly 290 scalar keys, 65 non-Unit pointers, 50 TermList, 26 std::string, 25 pairs/tuples, 15 types with defaultHash() methods, 5 Unit-derived pointers, and 25 sites inside templates where the key type is itself a template parameter.

Every converted site keeps the exact hash values the current dispatch produces. Distribution does not change, so proof search is unchanged and every step can be checked against master for byte-identical output.

To keep review manageable this lands as a series of small PRs:

1. named single-purpose functors in Lib/Hash.hpp (FnvHash, UnitHash, UnitNumberHash, PtrIdentityHash, LengthHash, PairHash); the DefaultHash branches become one-line delegations to them; a unit test pins the FNV-1a reference vectors and the equivalence
2. pointer keys: Unit-derived to UnitHash/UnitNumberHash, the rest to FnvHash/PtrIdentityHash
3. scalar keys to FnvHash/IdentityHash
4. string, pair and tuple keys
5. per-type functors for the types with defaultHash() methods (TermList, SATLiteral, ...), removing the method detection
6. hash parameters threaded through the generic templates, default arguments removed from Forwards.hpp, DefaultHash and DefaultHash2 deleted

Each PR builds and tests green on its own. The defaults only disappear in the last one, so the tree stays consistent at every point in the series.

shalashaska117:
@MichaelRawson I'm working on the first task you gave me, feel free to comment its steps here, for now i have completed 1 and I will try to finish 2 for tomorrow. 3 and 4 should not be that long to code, I will try to finish everything in 1 week. I'm also studying how to do the Debian task too, never did something like that so for me it's an unknown territory.

MichaelRawson:
[...]

PR #899, opened 2026-08-12
add named hash functors and delegate DefaultHash to them

shalashaska117:
First PR of the series in #898.

Adds named single-purpose hash functors to Lib/Hash.hpp: FnvHash (owns hashBytes/hashNulTerminated/hashIter and hashes scalars, non-Unit pointers and strings), UnitHash and UnitNumberHash (Units by their number), PtrIdentityHash (pointer address cast to unsigned), LengthHash (container length) and PairHash<HashFst,HashSnd>. Every dispatch branch of DefaultHash/DefaultHash2 is now a one-line delegation to one of these, so hash values are unchanged by construction and no call site changes yet.

FnvHash and PtrIdentityHash static_assert against Unit-derived pointers, so later stages cannot hand a Unit key the pointer-bytes hash by mistake.

A new unit test pins the FNV-1a reference vectors and the named-vs-DefaultHash equality; the equality half leaves with DefaultHash at the end of the series.

Tested: debug ctest 99/99, including the new Hash test.

### DIFF
--- a/Lib/Hash.hpp
+++ b/Lib/Hash.hpp
@@ -69,6 +69,129 @@ struct IdentityHash
   { return static_cast<unsigned>(val); }
 };
 
+/**
+ * FNV-1a for types where hashing the underlying bytes is sensible:
+ * arithmetic and enumeration types, pointers (by address, no dereference)
+ * and strings (by character).
+ * Caveat: this implements the 32-bit variant of FNV-1a
+ * Therefore it assumes (incorrectly) that `unsigned` is always 32 bits in size
+ * Nothing terrible will happen, but it's not going to win any hashing competitions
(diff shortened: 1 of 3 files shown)


MichaelRawson (approved):

MichaelRawson:
Nice - even if we don't end up completely removing `DefaultHash` this is worthwhile so that people can use these hash functions explicitly if they wish.

shalashaska117:
> Nice - even if we don't end up completely removing `DefaultHash` this is worthwhile so that people can use these hash functions explicitly if they wish.

It will be the last step deleting everything legacy, so that every step is not the default path meanwhile I continue developing each phases

PR #902, opened 2026-08-13
Equality proxy: one implementation, chosen by the problem, with a deterministic output order

quickbeam123:
## Why

The equality proxy transformation existed twice. `Shell/EqualityProxyMono` was a copy-paste of
`Shell/EqualityProxy` — ~95 % identical, doc comments still naming the other class — and the two
copies had begun to drift, so every fix had to be made twice and every difference between them was
ambiguous between "deliberate" and "forgotten".

Worse, the difference that mattered was steered by an option rather than by the input. The
polymorphic variant introduces a proxy predicate with a type argument, i.e. it turns a monomorphic
problem polymorphic, and it never reported that to the `Problem`. Since `invalidateByRemoval()`
leaves an already known *false* alone, `hasPolymorphicSym()` kept answering "monomorphic" until
something happened to re-scan the problem. Steps that refuse polymorphic input — `TheoryFlattening`,
`PredicateElimination`, FMB, `TheoryFinder` — could then act on a problem that was no longer what
they took it for. The compensating hack was in `TermSharing::setPoly()`, which decided the
well-typedness test from *option values*.

And the polymorphic variant is no win on a monomorphic problem anyway, so `mono_ep`/`mep` was a knob
nobody should have had to set.

## What this PR does

Four self-contained commits; each commit message carries its own reproducer and the exact
verification.

**1. The two classes are merged into `Shell/EqualityProxy`.** One class, a `poly` constructor flag,
and only the handful of places that genuinely differ branch on it: proxy-predicate creation and
lookup, the arity of the proxy literal, premise collection, the local-axiom loop, and the
type-argument handling in congruence axioms. The mono variant's proxy maps, which were `static`,
became ordinary members. Behaviour is unchanged by this commit — the caller still derives the flag
from `mono_ep` — which is what makes it checkable by diffing.

**2. The variant is chosen from the problem, not from an option.** A polymorphic problem gets the
single polymorphic proxy predicate; a monomorphic problem gets the per-sort predicates and *stays
monomorphic*. `mono_ep`/`mep` is removed, along with the `mep=off` entries in the CASC schedules and
the `mep` lines in the samplers. `Problem` gains `reportPolymorphicSymAdded()`, which the
transformation now calls, so `TermSharing::setPoly()` no longer has to infer the answer from
options.

**3. The monomorphic variant emits its axioms in a deterministic order.** It enumerated a `DHMap`
keyed by sorts, and `DHMap` places entries by hashing `TermList::content()` — the *address* of the
sort term. The reflexivity/symmetry/transitivity axioms therefore came out in a different order in
every run, which is a pre-existing defect on master and not specific to this refactor. `Kernel/Term.hpp`
gains `SharedTermListHash`/`SharedTermListHash2`, the `TermList` counterpart of the existing
`SharedTermHash`, hashing `Term::getId()`; both are needed, since `DHMap` takes the bucket from
`Hash1` and the probing step from `Hash2`.

**4. No congruence axiom for a symbol with no term arguments to relate.** A type argument is not
related by the proxy — there is no equality between types, so it simply gets the same variable on
both sides — and `arity()` counts type arguments, so the `arity == 0` guard did not catch a
polymorphic constant like `nil : !>[X]: list(X)`. The axiom emitted for one was
`E(list(X0), nil(X0), nil(X0))`, an instance of the reflexivity axiom that is always added anyway;
the predicate analogue was the tautology `~p(X0) | p(X0)`. The monomorphic variant already skipped
predicates with nothing to relate; both loops in both variants do now.

## Effect

* One implementation instead of two: −626/+290 lines, `Shell/EqualityProxyMono.{cpp,hpp}` gone.
* **No change to the default behaviour.** `mep` defaulted to `on`, i.e. a monomorphic problem
  already got the monomorphic proxy — exactly what `hasPolymorphicSym()` now selects. Only the
  `-mep off` combination disappears, and with it two ways of breaking a monomorphic problem:
  `-ep R -sa fmb` on `PUZ001+1` used to die with "Finite model building is currently not compatible
  with polymorphism", and `-ep RSTC -p tptp` used to hit an assertion in `Kernel/Formula.cpp:202`
  while printing the proxy definition.
* Runs are reproducible again: the same binary twice on the same input now gives byte-identical
  `--mode clausify` output. The emitted axiom order changes once against master and then stays put.
* Fewer useless clauses on polymorphic problems: `ITP254_4` with `-ep RSTC` sheds 198 subsumed unit
  clauses, one per polymorphic constant. Monomorphic problems are unaffected by that commit.

## Testing

* clause sets of `--mode clausify` output compared against a build of the base commit for
  `NUM919_1`, `PUZ001+1`, `SET009-1`, `ITP294_3` (monomorphic TFX) and `ITP254_4` (polymorphic TFX)
  × all four values of `-ep`, with `-mep on` and `-mep off` for commit 1 and with the default for
  commit 2: identical throughout. Compared as multisets of formulas with normalised names, since
  clause numbering moves as soon as anything is emitted in a different order;
* `-ep RSTC -p tptp` proofs byte-identical on `PUZ001+1` and on a small FOF problem, including the
  `equality_proxy_definition` / `_replacement` / `_axiom` inference records;
* determinism: two runs of the same binary now agree on `ITP294_3` and `NUM919_1` × `-ep R`,
  `-ep RSTC`; they disagreed before commit 3;
* for commit 4, the clause set of `ITP254_4 -ep RSTC` loses exactly the 198 degenerate axioms and
  gains nothing, all other outputs in the matrix are byte-identical, and the general reflexivity
  axiom is of course still there;
* `ctest` passes (99/99), and the `-ep` lines of `checks/sanity` were run by hand.

A full regression over the problem lists has not been run locally; it is queued for the server.

## Notes for the reviewer

* The dispatch uses `prb.hasPolymorphicSym()`, which **over-approximates**: `Shell/Property` also
  sets it for an equality on a non-nullary *ground* sort such as `list(int)`, which the monomorphic
  variant would handle perfectly well. That matches the previous default exactly, so it is left as a
  `TODO` at the call site rather than changed here — a finer criterion would be a behaviour change
  worth measuring separately.
* One wart of the old code is preserved deliberately: `apply(Clause*)` records the same definition
  once per replaced literal, so a clause with several equalities of the same sort lists that premise
  several times in its inference record. Harmless, and deduplicating it would change the proof
  output of every equality-proxy run for no gain. (The other wart commit 1 preserved, the degenerate
  congruence axioms, is what commit 4 fixes.)
* `checks/sanity` has a **pre-existing** failure earlier in the script,
  `Problems/DAT/DAT023_1.p` with `-alasca on -alascai on -thf on`, which aborts the run before it
  reaches the equality-proxy lines. It fails on the base commit too.


quickbeam123:
I will come back here when I have run some tests on the server.

Kernel/Term.hpp
```diff
@@ -1316,6 +1316,22 @@ struct SharedTermHash {
...
+ */
+struct SharedTermListHash {
+  static bool equals(TermList t1, TermList t2) { return t1==t2; }
+  static unsigned hash(TermList t)
+  { ASS(t.isTerm() && t.term()->shared()); return DefaultHash::hash(t.term()->getId()); }
+};
+struct SharedTermListHash2 {
+  static unsigned hash(TermList t)
+  { ASS(t.isTerm() && t.term()->shared()); return DefaultHash2::hash(t.term()->getId()); }
+};
```
MichaelRawson, line 1333: Can just change the `TermList` hash if you prefer. Or we can call this `DeterministicTermListHash` or something and handle the variable case. Or nothing at all, nbd.

MichaelRawson (approved):
Nice, let's be having this finally.

Kernel/Term.hpp
quickbeam123: You mean the overload for `TermList`? Somewhere deep inside `Hash.hpp`? I was a bit afraid to touch it, given some ongoing plans to clean that up...

MichaelRawson: That one despatches to `defaultHash` in the class `TermList`, so it could be fixed 'locally' but it would be a very invasive change so I'd probably do one of the second two options (incl nothing). There are indeed plans to fix it, but you would be in time. :)

quickbeam123:
I ran all four "R" / "RS" / "RST" / "RSTC" on all of CNF/FOF/TF0/TX0/TX1 and found nothing suspicious.

There are a few problems on which master may succeed, but only sometimes, when the nondeterministic order makes the saturation god happy. New version is deterministic.

PR #903, opened 2026-08-13
ALASCA integer conversion: give the integrality axiom a sound Inference, and refresh the problem property

quickbeam123:
Two small commits, each with its reproducer in the commit message.

**1. Fix the Inference kind of the ALASCA integrality axiom.**
`AlascaPreprocessor::integerConversion` appends a premise-free clause `floor(f(X)) = f(X)` for every
fresh real-sorted symbol that replaced an int-valued one, but built it as
`NonspecificInferenceMany(ALASCA_INTEGER_TRANSFORMATION, nullptr)` — a rule from the
formula-clause-transformation interval, passed to a premise-carrying constructor. Debug builds now
abort on `isNonSpecificInferenceRule`; the `FormulaClauseTransformation(rule, nullptr)` spelling
segfaults outright. New `ALASCA_INTEGRALITY_AXIOM` in the non-specific interval, built with
`NonspecificInference0(UnitInputType::AXIOM, ...)`. Deliberately *not* a theory axiom: it is not
valid in the theory of reals, it only constrains the freshly introduced symbol.

**2. Invalidate the cached Problem property after the conversion.**
The conversion rewrote every unit and added clauses without telling `Problem`, so a later
`getProperty()` returned the pre-conversion `Property` and `hasEquality()` could stay `false` for a
clause set full of equalities.

New `checks/theory/alasca-integer-conversion.p` in `checks/sanity`: its refutation has to go through
the integrality axioms, so the check turns Satisfiable if they ever stop being added.
`checks/sanity` passes against a Release+Z3 build, `ctest` 99/99, and the previously failing
`Problems/DAT/DAT023_1.p -alasca on -alascai on -thf on` is now Theorem.

### DIFF
--- a/Kernel/ALASCA/Preprocessor.hpp
+++ b/Kernel/ALASCA/Preprocessor.hpp
@@ -258,6 +258,10 @@ public:
         }
       }
     }
+    // the conversion rewrites every unit and adds new ones (in particular equalities,
+    // which the original problem need not have contained), so nothing cached about the
+    // problem survives it
+    prb.invalidateEverything();
   }
 };
 
--- a/Kernel/ALASCA/Preprocessor.hpp
+++ b/Kernel/ALASCA/Preprocessor.hpp
@@ -24,7 +24,7 @@
 
 namespace Kernel {
 
-class AlascaPreprocessor 
+class AlascaPreprocessor
 {
   const InequalityNormalizer& _norm;
   Map<unsigned, unsigned> _preds;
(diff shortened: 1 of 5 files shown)


quickbeam123:
I also removed some trialing whitespace from `ALASCA::Preprocessor`, sorry...

Kernel/ALASCA/Preprocessor.hpp
```diff
@@ -243,25 +243,29 @@ class AlascaPreprocessor
...
+        if (!theory->isInterpretedFunction(func.value())
             && !R::isNumeral(func.value())
             && !R::isLinMul(func.value())
             ) {
           auto sym = env.signature->getFunction(func.value());
           if (orig_sym->fnType()->result() == Z::sort()) {
             auto t = TermList(Term::createFromIter(func.value(), range(0, sym->arity()).map([](auto x) { return TermList::var(x); })));
-            // TODO use something else than NonspecificInferenceMany
-            auto inf = Inference(NonspecificInferenceMany(INF_RULE, nullptr));
+            auto inf = Inference(NonspecificInference0(UnitInputType::AXIOM,
```
mezpusz, line 253: One second thought, maybe this should be a `TheoryAxiom`, what do you think? Then we have to move it in the inference rule enum too.

checks/theory/alasca-integer-conversion.p
```diff
@@ -0,0 +1,12 @@
+% Regression: the ALASCA int->real conversion (-alascai) appends an integrality
```
mezpusz, line 1: Note: these should be eventually converted to smaller-footprint preprocessing unit tests.

Kernel/ALASCA/Preprocessor.hpp
quickbeam123: TheoryAxiom is not appropriate, because the formula is about an introduced symbol. (This has been considered;))

mezpusz (approved):

PR #900, opened 2026-08-12
Predicate elimination 2

quickbeam123:
Implements (behind an option) an extension suggested by Andrei:

      "Also eliminate a predicate P occurring more than once in a clause, provided P never occurs"
      " both positively and negatively in a single clause and the multi-occurrence clauses all sit"
      " on one polarity side. Instead of pairwise resolvents, the replacement clauses are then all"
      " the hyper-resolvents, each occurrence of a multi-occurrence clause being resolved against"
      " its own (variable-disjoint) copy of a single-occurrence clause of the opposite polarity."
      
In addition, performance has been tuned and improved (also for the non-pelmo case).
(Sneakily, `cl->maxVar` no longer uses the expensive `getUniquePersistentIterator` which positively impacts other places too.)

PR #904, opened 2026-08-14
name the hash explicitly at pointer-keyed containers

shalashaska117:
Second PR of the series in #898, on top of #899.

Every container keyed on a pointer now names its hash functors instead of relying on DefaultHash/DefaultHash2. Pointers into the Kernel::Unit hierarchy (Clause*, Kernel::Unit*) take UnitHash and UnitNumberHash; every other pointer takes FnvHash and PtrIdentityHash. Map and Set have one hash slot and need equals() from it, so they take FnvHash alone. These are the functors the DefaultHash branches already delegate to since #899, so the hash values are unchanged and the edit is a spelling change at the use site. 76 sites across 36 files; every changed line is the old line plus one template argument.

The compiler checks the mapping: FnvHash and PtrIdentityHash static_assert that the key is not Unit-derived, and UnitHash::hash takes a const Kernel::Unit*, so a key routed to the wrong functor fails to compile in either direction.

Inferences/Induction.hpp:117 already named SharedTermHash as its primary hash. Only the empty secondary slot was filled, with the PtrIdentityHash that DefaultHash2 was resolving to.

Nine pointer-keyed sites stay on the defaults. Eight are commented-out code (FMB/ClauseFlattening.cpp:148-149 and Inferences/Instantiation.cpp:70-72). The ninth is Lib/InverseLookup.hpp:58, DHMap<T*, size_t> inside a class template: T can be Unit-derived, so naming FnvHash there would change values for those instantiations. Generic sites are stage 6.

Tested: debug and release builds clean, ctest 99/99, checks/sanity on the release binary exits 0. Output was also compared against master byte for byte under -stat full over 158 problems (MPTP2078 bushy, typed TPTP, and the checks/ corpus) and 34 strategy configurations. The cleanest pass bounds the saturation loop by -al 1000 rather than by the clock, so a run that hits the limit still stops in the same place twice: every comparable job is identical, and no job differs against the branch that a control run of master against itself could reproduce.

One caveat on that comparison, since it bounds what the sweep can claim. perf_event_open is unavailable on this kernel, so -i is silently ignored and wall-clock-bounded runs stop in a different place each time. Every sweep therefore ran twice, master against branch and master against itself, and only jobs the self-comparison reproduced count as evidence.

### DIFF
--- a/DP/SimpleCongruenceClosure.hpp
+++ b/DP/SimpleCongruenceClosure.hpp
@@ -272,7 +272,7 @@ private:
   /** Constants corresponding to terms */
   DHMap<TermList,unsigned> _termNames;
   /** Constants corresponding to literals */
-  DHMap<Literal*,unsigned> _litNames;
+  DHMap<Literal*,unsigned, FnvHash, PtrIdentityHash> _litNames;
 
   /**
    * Equality that caused unsatisfiability; if CEq::isInvalid(), there isn't such.
--- a/FMB/DefinitionIntroduction.hpp
+++ b/FMB/DefinitionIntroduction.hpp
@@ -222,8 +222,8 @@ namespace FMB {
     ClauseIterator _cit;
     Stack<Clause*> _processed;
 
-    DHMap<Term*,Term*> _introduced;
-    DHMap<Term*,unsigned> _introducedNG;
+    DHMap<Term*,Term*, FnvHash, PtrIdentityHash> _introduced;
+    DHMap<Term*,unsigned, FnvHash, PtrIdentityHash> _introducedNG;
 
   };
 
--- a/FMB/FiniteModelMultiSorted.hpp
+++ b/FMB/FiniteModelMultiSorted.hpp
@@ -120,7 +120,7 @@ private:
 
   // the pairs of <constant number, sort>
   DHMap<std::pair<unsigned,unsigned>,Term*> _domainConstants;
-  DHMap<Term*,std::pair<unsigned,unsigned>> _domainConstantsRev;
+  DHMap<Term*,std::pair<unsigned,unsigned>, FnvHash, PtrIdentityHash> _domainConstantsRev;
 public:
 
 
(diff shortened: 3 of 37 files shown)


shalashaska117:
@MichaelRawson @mezpusz @quickbeam123 this one is ready whenever you have time.

It is the second of the series in #898, after #899. #905 (scalar keys) and #906 (string and pair keys) follow it and are both drafts, so they need nothing from you until this merges. GitHub will not take a fork branch as a pull request base, so each of them carries the commits below it and shows an isolated diff through a compare link in its description.

MichaelRawson:
> Pointers into the Kernel::Unit hierarchy (Clause*, Kernel::Unit*) take UnitHash and UnitNumberHash; every other pointer takes FnvHash and PtrIdentityHash. 

@quickbeam123 - we could take this as an opportunity to `static_assert` that we don't create iterators over hash tables with pointer keys, if that sounds appealing? I'll put it on the list if you agree.

MichaelRawson (approved):
Looks good. Please rebase to fix merge conflicts, I think it's just that `PredicateElimination` got changed a lot this week.

quickbeam123:
> create iterators over hash tables with pointer keys, if that sounds appealing?

It sounds appealing indeed! Thank you!

shalashaska117:
Rebased onto master.

The conflict was `PredInfo::pos` and `neg` in `Shell/PredicateElimination.hpp`, which #900 changed from `DHSet<Clause*>` to `DHMap<Clause*, unsigned>`; both now carry `UnitHash` and `UnitNumberHash` in the map's hash slots.

Master grew three pointer-keyed sites since the branch was cut, so they are converted here too: the `updateSide` parameter and `premiseSeen` in `Shell/PredicateElimination.cpp`, and `seen` in `Shell/EqualityProxy.cpp`. `updateSide` is not optional, since its parameter type has to follow `PredInfo::pos`. The stage is now 79 sites across 37 files.

Debug and release build clean, ctest 99/99, checks/sanity passes.

PR #905, opened 2026-08-14
name the hash explicitly at scalar-keyed containers

shalashaska117:
Third PR of the series in #898, rebased onto master now that #904 has merged, so the diff here is this stage alone.

Every container keyed on an arithmetic or enumeration type now names its hash functors: FnvHash primary and IdentityHash secondary, or FnvHash alone where Map and Set have a single slot. As in #904 these are the functors DefaultHash and DefaultHash2 already delegate to, so the values are unchanged and the edit is a spelling change at the use site. 301 sites across 112 files. The keys are unsigned, int, pid_t, the enum Theory::Interpretation, and the typedefs SymId, SplitLevel, StoredType and VampireVar.

Thirteen of those sites already named IdentityHash as their primary and DefaultHash as their secondary, swapping the usual arrangement because variable numbers distribute well under identity. Their DefaultHash becomes the FnvHash it was resolving to, which puts FnvHash in the secondary slot, the reverse of every other site in this diff. They are in Kernel/Matcher.cpp, MLMatcher.hpp, MLVariant.cpp, Renaming.hpp and KBO.hpp, Indexing/SubstitutionTree.hpp, Saturation/Splitter.cpp, and Shell/FunctionDefinition.cpp, FunctionDefinition.hpp and EqResWithDeletion.hpp.

Forwards.hpp gains forward declarations of FnvHash and IdentityHash next to the existing ones for DefaultHash and DefaultHash2. Kernel/SortHelper.hpp and Kernel/InductionTemplate.hpp name these containers in declarations while including only Forwards.hpp, and including Lib/Hash.hpp there would pull Kernel/Unit.hpp in with it, which is why Forwards.hpp forward-declares hashes at all.

Master moved two of these sites since the branch was cut. EqualityProxyMono was merged into EqualityProxy, which turns its s_proxyPredicateSorts into the member _proxyPredicateSorts in Shell/EqualityProxy.hpp; that site is converted here and the two in the old files are gone. PredicateElimination's occ map changed value type upstream and carries the functors in its new form.

Three scalar-keyed sites stay on the defaults, all commented-out code: FMB/CliqueFinder.hpp:40 and :100, and Parse/TPTP.cpp:5167, the last of which is stale text naming a member that has since changed type. Three explicit DefaultHash uses also remain, all of them applying it to a type with a defaultHash() method: DerefPtrHash<DefaultHash> in Kernel/TermOrderingDiagram.cpp:496 and Saturation/Splitter.hpp:302, and the SATLiteral BiMap in SAT/Z3Interfacing.hpp:347. Those belong to stage 5.

SATSubsumption/subsat/SubstitutionTheory.hpp:367 now reads DHMap<VampireVar, VampireVarPos, Kernel::IdentityHash, Lib::IdentityHash>, the same functor in both slots, which Lib/Hash.hpp:380 advises against. DefaultHash2 was resolving to IdentityHash there already, so the pair was degenerate before this PR and fixing it would change the values. Worth a follow-up issue.

Tested: debug and release builds clean, ctest 99/99, checks/sanity on the release binary exits 0. The thirteen sites above swap DefaultHash for FnvHash in a secondary slot, and UnitTests/tHash.cpp already asserts that FnvHash::hash equals DefaultHash::hash for unsigned and for int, so that substitution is checked by the test suite rather than by sampling. Output was also compared against master byte for byte over 161 problems in 10 strategy configurations, each bounded by -al 1000 rather than by the clock: 1607 of 1610 jobs identical. Two of the other three are wall-clock noise that a control run of master against itself reproduces, and the third is a pre-existing crash on hol/hol1.p under -qa plain that both binaries hit in the same place, where the only difference left is the commit hash in the version banner.

### DIFF
--- a/CASC/PortfolioMode.cpp
+++ b/CASC/PortfolioMode.cpp
@@ -392,7 +392,7 @@ std::optional<fs::path> PortfolioMode::runSchedule(Schedule schedule) {
   pid_t me = getpid();
   Schedule::BottomFirstIterator it(schedule);
   pid_t successful = 0; // PID 0 should be invalid
-  Set<pid_t> processes;
+  Set<pid_t, FnvHash> processes;
   int remainingTime;
   bool scheduleRepeat = false;
   while(remainingTime = env.remainingTime() / 100, remainingTime > 0) {
--- a/DP/SimpleCongruenceClosure.hpp
+++ b/DP/SimpleCongruenceClosure.hpp
@@ -239,7 +239,7 @@ private:
   };
 
   struct ConstOrderingComparator;  
-  typedef DHMap<unsigned,TermList> NFMap;
+  typedef DHMap<unsigned,TermList, FnvHash, IdentityHash> NFMap;
   void computeConstsNormalForm(unsigned c, NFMap& normalForms);
   
 #if VDEBUG
--- a/FMB/CliqueFinder.hpp
+++ b/FMB/CliqueFinder.hpp
@@ -21,7 +21,7 @@ namespace FMB {
 
   class CliqueFinder {
   public:
-    static unsigned findMaxCliqueSize(DHMap<unsigned,DHSet<unsigned>*>* Ngraph)
+    static unsigned findMaxCliqueSize(DHMap<unsigned,DHSet<unsigned, FnvHash, IdentityHash>*, FnvHash, IdentityHash>* Ngraph)
     {
       //std::cout << "findMaxCliqueSize with " << Ngraph->size() << std::endl;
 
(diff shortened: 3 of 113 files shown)


shalashaska117:
@MichaelRawson rebased onto master now that #904 is in. The only conflict was Shell/PredicateElimination.cpp, where the occ map changed value type upstream. The EqualityProxyMono merge also moved one scalar-keyed site into Shell/EqualityProxy.hpp, which this stage now covers. Debug and release builds, ctest 99/99 and checks/sanity are clean; the output sweep is in the description.

MichaelRawson (approved):

PR #906, opened 2026-08-14
name the hash explicitly at string-, pair- and tuple-keyed containers

shalashaska117:
Stage 4 of #898. Containers keyed on std::string, std::pair or std::tuple now name their hash functors instead of relying on DefaultHash/DefaultHash2. std::string takes FnvHash/LengthHash; a pair takes PairHash<HA,HB> built from what each element resolves to on its own, so ULit (pair<Literal*,Clause*>) takes PairHash<FnvHash,UnitHash>/PairHash<PtrIdentityHash,UnitNumberHash>. Map and Set take the primary alone, which also supplies equals(). Every hash value is unchanged, since these are the functors DefaultHash already delegates to. 53 sites across 29 files.

Three pair keys are spelled as a typedef: BindingSpec is a pair<unsigned,unsigned>, and InductionFormulaIndex's Key is a pair<Stack<LiteralStack>,pair<Literal*,Literal*>>, where DefaultHash walks the stacks element by element and DefaultHash2 takes the outer length.

TupleHash takes one functor per element now and has an equals(). DefaultHash and DefaultHash2 were its only callers and wanted one functor throughout, so they expand the tuple themselves. That moves Kernel/PartialOrdering.cpp:174, whose std::tuple<const PartialOrdering*, size_t, size_t, PoComp> key had no secondary spelling before.

Forwards.hpp gains a declaration of LengthHash, since StringUtils.hpp includes only Forwards.hpp and Lib/Output.hpp.

The six pair and tuple keys left on the defaults all carry a TermList or a TermSpec, and move in stage 5 with those types.

tHash.cpp pins three shapes against both defaults: a mixed three-element tuple, PartialOrdering's four-element one, and Key's nested pair. Beyond that: ctest 99/99, checks/sanity clean on release, and a byte-identical output sweep against master at 1608 of 1608 comparable jobs, plus 483 of 483 each under -fgj on and -ind both -indmd 1, the configurations that reach the PartialOrdering and InductionFormulaIndex caches. The sweep runs a master-against-master control pass and only counts jobs it reproduces; the two it does not are wall-clock-limited and differ in both passes.

### DIFF
--- a/DP/SimpleCongruenceClosure.hpp
+++ b/DP/SimpleCongruenceClosure.hpp
@@ -263,9 +263,9 @@ private:
   /**
    * Map from signature symbols to the local constant numbers.
    */
-  DHMap<std::pair<unsigned,SignatureKind>,unsigned> _sigConsts;
+  DHMap<std::pair<unsigned,SignatureKind>,unsigned, PairHash<FnvHash,FnvHash>, PairHash<IdentityHash,IdentityHash>> _sigConsts;
 
-  typedef DHMap<CPair,unsigned> PairMap;
+  typedef DHMap<CPair,unsigned, PairHash<FnvHash,FnvHash>, PairHash<IdentityHash,IdentityHash>> PairMap;
   /** Names of constant pairs (modulo the congruence!)*/
   PairMap _pairNames;
 
--- a/FMB/FiniteModelBuilder.cpp
+++ b/FMB/FiniteModelBuilder.cpp
@@ -390,8 +390,8 @@ void FiniteModelBuilder::init()
 
   env.statistics->phase = ExecutionPhase::FMB_PREPROCESSING;
 
-  DHSet<std::pair<unsigned,unsigned>> vampire_sort_constraints_nonstrict;
-  DHSet<std::pair<unsigned,unsigned>> vampire_sort_constraints_strict;
+  DHSet<std::pair<unsigned,unsigned>, PairHash<FnvHash,FnvHash>, PairHash<IdentityHash,IdentityHash>> vampire_sort_constraints_nonstrict;
+  DHSet<std::pair<unsigned,unsigned>, PairHash<FnvHash,FnvHash>, PairHash<IdentityHash,IdentityHash>> vampire_sort_constraints_strict;
   if(env.options->fmbDetectSortBounds()){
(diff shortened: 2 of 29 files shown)


MichaelRawson:
Hi @shalashaska117 - thanks for the last PR! Please fix merge conflicts and I'll review.

shalashaska117:
> Hi @shalashaska117 - thanks for the last PR! Please fix merge conflicts and I'll review.

Can you just review the paragraph 5 of the PR? It's needed for future changes

MichaelRawson:
Ah - yes, sorry! Let's have a per-element `TupleHash<...>`?

As a side note I hope eventually to unify `Map` and `DHMap`, and this PR series will definitely help with that.

shalashaska117:
> As a side note I hope eventually to unify `Map` and `DHMap`, and this PR series will definitely help with that.

For what I have seen in the code, it's long but double change, if you have any ideas about that let me know!

shalashaska117:
@MichaelRawson Rebased; with #905 merged this is one commit again.

TupleHash<Hs...> is per-element now, as you proposed, and has an equals() so it can fill a Map or Set slot. DefaultHash and DefaultHash2 were the only callers that wanted one functor throughout, so they expand the tuple themselves. Kernel/PartialOrdering.cpp:174 moves with it, and tHash.cpp pins its four-element shape against both defaults. The remaining pair and tuple keys carry a TermList or a TermSpec and go to stage 5.

ctest 99/99, sanity clean, sweep 1608/1608 comparable jobs, plus 483/483 under -fgj on, which is what actually reaches that cache.

shalashaska117:
Pushed three more pair keys I had missed, all spelled as a typedef: BindingSpec in Shell/FunctionDefinition.cpp is a pair<unsigned,unsigned>, and InductionFormulaIndex's Key is a pair<Stack<LiteralStack>,pair<Literal*,Literal*>>. 53 sites now.

Same evidence as before, re-run: ctest 99/99, sanity clean, sweep 1608/1608 comparable jobs, and 483/483 each under -fgj on and -ind both -indmd 1, which are what actually reach the PartialOrdering and InductionFormulaIndex caches.

MichaelRawson (approved):

PR #910, opened 2026-08-17
fix newcnf crash on a Boolean let bound to another let

### LINKED ISSUE #908: newcnf segfaults on a let bound to another let
NewCNF segfaults when a let is bound to another let.

```smt2
(set-logic ALL)
(declare-fun p () Bool)
(declare-fun q () Bool)
(declare-fun r () Bool)
(assert (let ((b (let ((c (and r p))) c))) (or (=> q b) b)))
(check-sat)
```

```
$ vampire -newcnf on crash.smt2
% Aborted by signal SIGSEGV on crash.smt2
% Termination reason: Unknown
% Termination phase: NewCNF
```

Release build, commit 484427078.

Three things are load-bearing. The outer let has to be bound to another let: `(let ((b (and r p))) (or (=> q b) b))` is answered normally. The inner body has to be a conjunction: `(or r p)` in its place does not trigger it. And the bound variable has to occur both under an implication and on its own.

Found while testing #894, and it reproduces with and without that branch, so it is independent of it.

shalashaska117:
NewCNF segfaults on a Boolean `$let` bound to another `$let` when the bound symbol occurs twice in one clause (#908).

Both occurrences are inlined separately and, with a nested binding, come out as the same formula, so it lands in two generalised literals of one generalised clause: `Occurrences::replaceBy` writes into the clauses directly and so bypasses `pushLiteral`. Only the one the popped occurrence points at was expanded, and the copy left behind sat in no `_occurrences` entry, so nothing ever expanded it and it reached `toClause` as a non-literal, where a release build reads a formula as a literal.

`introduceExtendedGenClause` now drops the other generalised literals holding the formula it expands, and discards the clause as a tautology when one of them has the opposite sign. `pop()` accounts for them. Checks added in `checks/theory/let-nested-bool.{p,smt2}`.

Closes #908.

### DIFF
--- a/Shell/NewCNF.hpp
+++ b/Shell/NewCNF.hpp
@@ -468,6 +468,20 @@ private:
     SPGenClause gc = occ.gc;
     unsigned position = occ.position;
 
+    /**
+     * The formula being expanded can occupy more than one generalised literal
+     * of gc: Occurrences::replaceBy writes into generalised clauses directly
+     * and so bypasses pushLiteral, which normally keeps one formula from being
+     * stored twice. Two distinct subformulas replaced by the same formula are
+     * enough, as happens when a let-bound symbol occurring twice is inlined.
+     * The other generalised literals say the same thing as the one occ points
+     * at, so they are dropped instead of copied over. A copy would stay in the
+     * new clause unprocessed forever, because a formula is out of _occurrences
+     * while it is being processed, so registerGenClause registers nothing
+     * for it.
+     */
+    GenLit expanded = gc->_literals[position];
+
     unsigned size = gc->size() + List<GenLit>::length(gls) - 1;
     SPGenClause newGc = SPGenClause(new GenClause(size, gc->bindings, gc->foolBindings));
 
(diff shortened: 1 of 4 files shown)


shalashaska117:
Validation of this branch, gcc 13, Z3 4.14.0.0 at the pinned submodule commit.

ctest, Debug: 99/99.

checks/sanity, Release: 12 of the 90 checks fail here, and every one of them reaches the status the script expects once I give it `-t 60` instead of the 0.1-1s the check allows (ANA135_1, ARI700_1, ARI724_1, DAT023_1, HWV087_2, ITP319_1, ITP320_1, ITP412_1, PLA046_1, PLA050_1, SWC478_1, SWW605_2), so this machine is slower than the CI runners rather than the code being wrong. A binary built from master's Shell/NewCNF.hpp fails the same set apart from the two new let-nested-bool checks, plus ITP319_1, ITP320_1 and ITP412_1, which sit on the time limit and flip between runs on either binary: three A/B runs each gave one Theorem for master on ITP320_1, one for this branch on ITP412_1, and timeouts otherwise.

Clausification diff against master over the whole checks corpus, 68 problems in four configurations each (default, `-newcnf on`, `-newcnf on -ile off`, `-newcnf on -nm 2`): 272 pairs of `--mode clausify` output, byte-identical everywhere except the two new tests, where master crashes.

Shapes I tried but did not add as checks: the bound symbol three times in one disjunction, the bound symbol with both signs in one disjunction, which is the tautology branch of the fix, three levels of nesting, a nested let under a quantifier binding a symbol with an argument, an inner `ite`, and two nested lets side by side. Five of the six crash on master, and with the fix all six agree with the default clausifier.

quickbeam123 (approved):

shalashaska117:
@quickbeam123 is this resolved or does it need other things?

PR #913, opened 2026-08-27
support the TPTP tcf dialect, on input and on output

quickbeam123:
A tcf() unit is parsed exactly as a tff() one; only at the end, in endFof(), do we strip its universal prefix and check that what remains is a clause. Unlike cnf(), a tcf() formula must be closed, and a conjecture role is refused, since negating a clause does not give a clause. Type declarations inside tcf() come for free, and so does everything our parser is habitually more liberal about: chained quantifiers, mixed dialects in one file, polymorphic prefixes.

--mode tclausify now prints tcf() rather than a tff() formula faking one. This is what the mode always meant, and it lets clausifyMode drop the Formula::fromClause wrapper along with the overwriteNumber and CONJECTURE-rewriting hacks that went with it. Both TPTPPrinter::toString and UIHelper::outputSymbolDeclarations take the new flag with a default, so every other output mode is unchanged.

checks/parse/SYN000-3.p is the TPTP sample exercising every tcf construct; checks/parse/tcf.p adds a two-sorted problem in which "X = Y" needs the prefix to make sense, and its tclausify output is checked to be re-readable.

### DIFF
--- a/Parse/TPTP.hpp
+++ b/Parse/TPTP.hpp
@@ -226,6 +226,8 @@ public:
     TFF,
     /** THF declaration */
     THF,
+    /** tcf declaration */
+    TCF,
     /** read type declaration */
     TYPE,
     /** after a top-level type declaration */
@@ -551,6 +553,10 @@ private:
   Stack<State> _states;
   /** input type of the last read unit */ // it must be int since -1 can be used as a value
   UnitInputType _lastInputType;
+  /** the top-level dialect the unit currently being read came from;
+   *  CNF and TCF units must end up being clauses, FOF and TCF units must be closed */
+  enum class Dialect { CNF, FOF, TCF };
+  Dialect _lastDialect;
   /** true if the last read unit is a question */
   bool _isQuestion = false;
   /** */
@@ -724,7 +730,7 @@ private:
   static Formula* makeJunction(Connective c,Formula* lhs,Formula* rhs);
   void unitList();
   void fof(bool fo);
-  void tff();
+  void tff(bool tcf);
   void vampire();
   void consumeToken(Tag);
   std::string name();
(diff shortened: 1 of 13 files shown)


Parse/TPTP.hpp
```diff
@@ -551,6 +553,10 @@ class TPTP
   Stack<State> _states;
   /** input type of the last read unit */ // it must be int since -1 can be used as a value
   UnitInputType _lastInputType;
+  /** the top-level dialect the unit currently being read came from;
+   *  CNF and TCF units must end up being clauses, FOF and TCF units must be closed */
+  enum class Dialect { CNF, FOF, TCF };
```
MichaelRawson, line 558: No `HOL` dialect?

MichaelRawson (approved):

Parse/TPTP.hpp
MichaelRawson: Ah, because it's handled separately. Weird, but alright.

Shell/TPTPPrinter.cpp
```diff
@@ -455,6 +456,42 @@ std::string TPTPPrinter::toString(const Formula* formula)
...
+  if (varSorts.isEmpty()) {
+    return "";
+  }
+
+  Stack<unsigned> vars;
+  vars.loadFromIterator(varSorts.domain());
+  vars.sort([&varSorts](unsigned v1, unsigned v2) {
+    bool t1 = varSorts.get(v1).isTerm() && varSorts.get(v1).term()->isSuper();
+    bool t2 = varSorts.get(v2).isTerm() && varSorts.get(v2).term()->isSuper();
+    return t1 != t2 ? t1 : v1 < v2;
```
mezpusz: I don't know the operator precedences from the top of my head so this is a bit hard to read. What does the condition say?

quickbeam123: Good point, adding parens.

mezpusz (approved):

PR #916, opened 2026-08-28
remove `ELEMENT_TYPE` and `ITERATOR_TYPE` macros, templates

MichaelRawson:
While poking around in `Reflection.hpp` I noticed the whole apparatus is essentially unused, and quite confusing. Remove it.


Kernel/InferenceStore.cpp
```diff
@@ -113,9 +113,8 @@ void InferenceStore::recordIntroducedSplitName(Unit* u, std::string name)
  *
  * It is caller's responsibility to ensure that variables in @b vars are unique.
  */
-template<typename VarContainer>
-std::string getQuantifiedStr(const VarContainer& vars, std::string inner, DHMap<unsigned,TermList, FnvHash, IdentityHash>& t_map, bool innerParentheses=true){
-  VirtualIterator<unsigned> vit=pvi( getContentIterator(vars) );
```
MichaelRawson, line 119: This was the only place where `getContentIterator` was used. Pass the iterator rather than the container into the function.

Lib/Reflection.hpp
```diff
@@ -70,188 +67,6 @@
...
-template<typename T>
-struct ElementTypeInfo<T*>
-{
-  typedef T Type;
-};
-
-template<class C> 
-struct ElementTypeInfo<std::initializer_list<C>> {
-  using Type = C;
-};
```
MichaelRawson, line 218: These seem like they would be important, but they are never used. Removing them should give a compiler error (as otherwise `T` would need to have a member `_ElementType`) but it does not.

Lib/VirtualIterator.hpp
```diff
@@ -47,13 +44,7 @@ template<typename T>
 template<typename T>
 class IteratorCore {
 public:
-  IteratorCore(const IteratorCore&) = delete;
-  IteratorCore& operator=(const IteratorCore&) = delete;
-  IteratorCore(IteratorCore&&) = default;
-  IteratorCore& operator=(IteratorCore&&) = default;
-
-  DECL_ELEMENT_TYPE(T);
-  IteratorCore() = default;
```
MichaelRawson, line 56: These (non) constructors are implied.

Lib/VirtualIterator.hpp
```diff
@@ -102,33 +92,21 @@ class EmptyIterator
  * of element retrieval through the polymorphism of @b IteratorCore objects
  *
  * @tparam T type returned by the iterator
- *
- * The @b VirtualIterator object performs reference counting on @b IteratorCore
- * objects and deletes them when the reference count drops to zero. The reference
- * count is kept in the @b IteratorCore::_refCnt field.
- *
- * @see IteratorCore
```
MichaelRawson, line 110: Comment no longer accurate.

Lib/VirtualIterator.hpp
mezpusz: What implies that copying is disallowed here?

MichaelRawson: Good point. I don't actually know what I was thinking, let me get back to you.

Lib/VirtualIterator.hpp
```diff
@@ -102,33 +92,21 @@ class EmptyIterator
...
    * Create an object with @b core as its core.
    */
   inline
   explicit VirtualIterator(IteratorCore<T>* core) : _core(core) {}
 
-  VirtualIterator(const VirtualIterator &) = delete;
-  VirtualIterator &operator=(const VirtualIterator &) = delete;
-
-  VirtualIterator(VirtualIterator &&) noexcept = default;
-  VirtualIterator &operator=(VirtualIterator &&other) noexcept = default;
```
MichaelRawson, line 130: Oh, right - @mezpusz I meant *these* constructors, sorry - these are implied-deleted because `_core` is a `std::unique_ptr<IteratorCore<T>>`.

Lib/VirtualIterator.hpp
MichaelRawson: See other comment - these are *not* implied-deleted but there's actually no reason to forbid copying a (safe) subclass of `IteratorCore<T>` if you want to, and unsafe subclasses should forbid it themselves. You should not copy a `std::unique_ptr<IteratorCore<T>>` - because you don't know whether the boxed iterator minds being copied - but the compiler won't let you anyway.

MichaelRawson: (but I don't mind putting these back!)

mezpusz: No need to put back, thanks for the analysis!

mezpusz (approved):
IMO macros should be removed where possible, so this is a nice change (and lots of removals), thanks!

quickbeam123 (approved):

PR #917, opened 2026-08-29
internal iteration for clause generation

MichaelRawson:
Something a little controversial to think about.

Vampire builds massive iterator chains in order to carry out inference. However, all of them end up being consumed and passed through `SaturationAlgorithm` more-or-less unconditionally.

This PR adds a parallel method to e.g. `GeneratingInferenceEngine` that implements [internal iteration](https://en.wikipedia.org/wiki/Iterator#Internal_iterator) rather than constructing a `ClauseIterator`, i.e.

```c++
  virtual ClauseIterator generateClauses(Clause* premise) = 0;
  virtual void generateClauses(Clause* premise, ClauseReceiver receive) {
    for(Clause *cl : iterTraits(generateClauses(premise)))
      receive(cl);
  }
```

I've re-plumbed `SaturationAlgorithm` to use internal iteration at clause activation.

The pay-off for all this is that writing inferences is often much more straightforward, easier to debug and possibly more performant. I've changed `Factoring` as a demonstration.

(also renamed `SimplifyingGeneratingInference` `SimplifyingGeneratingInferenceEngine` because it was confusing me)

This is just a proof of concept and would need more attention if we want to go ahead with it.

### DIFF
--- a/Inferences/ALASCA/Coherence.hpp
+++ b/Inferences/ALASCA/Coherence.hpp
@@ -223,7 +223,7 @@ struct Coherence : public BinInf<CoherenceConf<NumTraits>> {
 
 
 template<class NumTraits>
-struct CoherenceNormalization : SimplifyingGeneratingInference {
+struct CoherenceNormalization : SimplifyingGeneratingInferenceEngine {
   const AlascaState& _shared;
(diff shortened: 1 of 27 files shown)


MichaelRawson:
CI failure is weird, it works locally. :shrug:

MichaelRawson:
I for one greatly prefer #920. Let's keep the discussion going there.

PR #918, opened 2026-08-29
re-shuffle includes to reduce unnecessarily large translation units

MichaelRawson:
I am trying to improve compile times. Many files end up pulling in all of ALASCA or other heavy machinery by an unfortunate series of includes, but this is not necessary.

I've re-structured things by splitting header files up a bit. Nothing too dramatic, but it makes a difference.

There should be no performance hit here as the final code is the same.


cmake/sources.cmake
```diff
@@ -178,6 +178,8 @@ set(SOURCES
     Indexing/CodeTree.hpp
     Indexing/CodeTreeInterfaces.cpp
     Indexing/CodeTreeInterfaces.hpp
+    Indexing/DemodulationIndex.hpp
```
mezpusz, line 181: I'm just wondering, how did you come across this split? Was `Index.hpp` included in too many places because of this?

MichaelRawson: I was chasing the include diagram of random translation units to see why certain things end up there. `TermOrderingDiagram` ends up almost everywhere (TODs are awesome but maybe we don't want this much death), which is because of of the include in `Index.hpp`. But you need it because of `DemodulatorData`, so split out `DemodulatorData`.

mezpusz (approved):
Great if this reduces compiler stress! Do you have any suggestions how to keep this up in the future?

Also, only marginally related but I've seen `SharedSet.hpp` as an unused include in many places, but it can't be removed for some reason. Did you come across this?

cmake/sources.cmake
mezpusz: Death to all! 💀 What tool did you use? Otherwise it makes sense.

MichaelRawson:
I think if we're reasonably hawkish about includes in header files we should be OK. I don't mind repeating the exercise whenever I notice things getting sluggish.

Ah, the `SharedSet` thing! This is annoying, my LSP also suggests this incorrectly. It's because of `SplitSet` (used in `Clause`), which is just `SharedSet<unsigned>`. I might try and fix this in a future PR.

cmake/sources.cmake
MichaelRawson: Some combination of `g++ -E` and Doxygen (!). Doxygen spits out a GraphViz-generated graph of includes and included-by for each file. To some extent at least, it gets confused sometimes.

PR #919, opened 2026-08-29
use named hash functors for types with defaultHash()

shalashaska117:
Stage 5 of #898, following #906. Containers keyed on `TermList` (including `SortId`), `VarSpec`, `TermSpec`, `SATLiteral`, `SATClause`, `Coproduct`, and term ordering polynomials now name their hash functors. This includes the pair and tuple keys deferred from stage 4 and the three container sites that explicitly named `DefaultHash`.

The per-type functors hold the hash formulas, and `defaultHash()`/`defaultHash2()` delegate to them. `CoproductHash` and `CoproductHash2` call the existing methods. The method protocol stays until stage 6 threads hashes through generic code and removes `DefaultHash`.

`SATClause` is the exception to identical hash values. Moving its hash loop exposed different results in `-O3` builds when the trailing array declared as `_literals[1]` was indexed directly. `SATClauseHash` uses pointer traversal and computes the XOR of the literal hashes, which can differ from the old optimized loop's result. The regression test derives its expected value from the input literals and checks that permuting them preserves the hash; the output sweeps below check the effect on proof search.

## Checks

- Debug unit tests: 100/100 passed.
- Release build and `checks/sanity` passed.
- `checks/headers` and `git diff --check` passed.
- Output sweeps against `master` (`719d180e2`): 3,768 matching comparisons across 19 configurations, including `-fgj on`, `-ind both -indmd 1`, and `-alasca on`.

Comparisons include full statistics and ignore the version banner, elapsed time, and peak memory. The controls comparing `master` with itself exclude 69 jobs whose output varies between identical builds.

One further timeout mismatch, `int_sum_y_geq_0.smt2` with ALASCA and otter, matched at 300 activations with a time limit of 180 seconds. `NUM919_1` also matched with the longer time limit and its original limit of 1,000 activations. Both reruns included controls.

### DIFF
--- a/SAT/SATClause.hpp
+++ b/SAT/SATClause.hpp
@@ -49,12 +49,7 @@ public:
   void* operator new(size_t,unsigned length);
   void operator delete(void *, size_t);
 
-  unsigned defaultHash() const {
-    unsigned hash = 0;
-    for(unsigned i = 0; i < length(); i++)
-      hash ^= DefaultHash::hash(_literals[i]);
-    return hash;
-  }
+  unsigned defaultHash() const;
 
   bool operator==(const SATClause &other) const {
     if(length() != other.length())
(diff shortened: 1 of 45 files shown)


shalashaska117:
@MichaelRawson ready for review!

SAT/SATClause.hpp
```diff
@@ -114,6 +109,23 @@ class SATClause
...
+// different hash values in optimized builds when this loop was moved.
+struct SATClauseHash {
+  static bool equals(SATClause const& c1, SATClause const& c2) { return c1 == c2; }
+
+  static unsigned hash(SATClause const& c) {
+    const SATLiteral* lits = &c[0];
+    unsigned hash = 0;
+    for(unsigned i = 0; i < c.length(); i++)
+      hash ^= SATLiteralHash::hash(lits[i]);
+    return hash;
```
MichaelRawson, line 123: Aaaahhhh flexible array members are scary. Probably this is UB and will bite us eventually. I've made a note to audit our use of these.

MichaelRawson (approved):

SAT/SATClause.hpp
shalashaska117: Do you want me to change it?

MichaelRawson: This code's fine! But how we do flexibly-sized data structures is broken, I'll fix it separately.

PR #920, opened 2026-08-31
coroutines for clause generation

quickbeam123:
Another idea for how to address what #917 is proposing to fix. 

This is just cosmetics and readability. Perf check found no measurable difference on TPTP.

### DIFF
--- a/Lib/Generator.hpp
+++ b/Lib/Generator.hpp
@@ -1,0 +1,173 @@
+/*
+ * This file is part of the source code of the software program
+ * Vampire. It is protected by applicable
+ * copyright laws.
+ *
+ * This source code is distributed under the licence found here
+ * https://vprover.github.io/license.html
+ * and in the source directory
+ */
+/**
+ * @file Generator.hpp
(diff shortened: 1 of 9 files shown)


MichaelRawson:
:exploding_head:

Lib/Generator.hpp
```diff
@@ -0,0 +1,173 @@
...
+    friend class Generator;
+
+    Option<T> _value;
+    std::exception_ptr _exception;
+
+  public:
+    // route the coroutine frame through Vampire's allocator; the compiler prefers the
+    // sized deallocation function, which is the one Lib::free needs
+    void *operator new(size_t size) { return Lib::alloc(size); }
+    void operator delete(void *ptr, size_t size) { Lib::free(ptr, size); }
```
MichaelRawson, line 79: This should not be necessary.

MichaelRawson:
This is awesome. I think I'd much prefer this to #917 - although we should be very careful to understand coroutine semantics before committing to this too hard.

PR #921, opened 2026-08-31
`Shell::Option` demolition

MichaelRawson:
`Shell::Options` had grown over a decade or so and become quite convoluted. Straighten it out a bit:
- Move most of the implementation details to `Options.cpp` so it doesn't bother every other translation unit.
- Thereby remove some extraneous `#include`s in the header.
- Use `const char *` rather than `std::string` for option names (since they are all statically known), and various other de-heapifications.
- Re-implement the `OptionValueConstraint<T>` system - you can still write exactly the same expressions, but without the evil `AbstractOptionValueConstraint`/`WrappedConstraint` headache.
- `Options::init()` is now just `Options()`
- `OptionValue<T>` no longer has a scary default constructor that does...something. This was only necessary because options were not initialised in `Options()`.
- Remove some completely-unused fields and options, and wire some in properly that were used but never initialised (eek!).

 I'm not sure what the best way to review this is, as it amounts to a rewrite. Maybe I'll just grab @quickbeam123 at some point and inform him of its various virtues.


MichaelRawson:
At least on my compiler this hacking seems to have allowed a reasonable release compile time for `Shell/Options.cpp` again, so I've removed the no-optimise compiler pragma.

quickbeam123 (approved):

mezpusz (approved):

PR #922, opened 2026-08-31
try bumping actions version to fix cache?

MichaelRawson:
CI seems to fail to recover an extant Z3 build from cache, saying (cryptically)

```
Cache hit for: z3-build-3c47fd96cf5645d0c42b2c819d9e9a84380aa721
Warning: Failed to restore: 
Cache not found for input keys: z3-build-3c47fd96cf5645d0c42b2c819d9e9a84380aa721
```

The key does exist. There's a new version of the GitHub `cache` script so let's try that as a first step.


MichaelRawson:
Doesn't seem to work. :shrug: I'm all out of ideas.

MichaelRawson:
Clearing the cache seemed to fix this.

PR #924, opened 2026-09-02
Make code tree default for generalizations

mezpusz:
The main goal of the PR was to completely remove the matching part of substitution trees, that have a big intersection of functionality with the allegedly more efficient code trees.

The only missing feature from the code tree side was a "literal code tree", which is very similar in spirit to "term code trees" that can index single terms, but they have to also handle commutativity of equalities.

The PR therefore contains
(1) some refactoring to the code tree infrastructure to allow adding the literal code tree implementation with too much duplication,
(2) lots of deletion and refactoring on the high-level literal index side, where the interface essentially broke into two parts, a generalizing and non-generalizing (i.e. unifying and instantiating) part, and
(3) to avoid having to add the usual higher-order templating to all the inferences that now freshly use code trees, I removed the higher-order templating completely from code trees.

The results with default options are surprisingly good, with TPTP FOL Discount winning 7 new problems, Otter winning 3 new problems, no change on TPTP HOL. I will run some experiments with `-cts off` to see how the SAT subsumption has changed.


Indexing/CodeTree.cpp
```diff
@@ -725,7 +725,7 @@ inline bool CodeTree::Matcher<removing, checkRange, higherOrder>::doAssignVar()
     fte++;
     ASS_EQ(fte->_tag(), FlatTerm::FUN_TERM_PTR);
     ASS(fte->_term());
-    if constexpr (higherOrder) {
+    if (env.higherOrder()) {
```
mezpusz, line 728: Here's the infamous change for HOL that previously needed so much templating. It looks like the path is not too hot after all..

mezpusz:
Results with `-cts off` (this should actually heavily use the literal code tree, run 0 is master, run 1 is this branch):

FOL Discount:

run 0 unsat: 8996 (14) sat: 1040 (1) cputime: 54415.21 s instructions: 228530921 Mi memory: 1115017.82 MB
run 1 unsat: 9006 (24) sat: 1041 (2) cputime: 53247.08 s instructions: 228461094 Mi memory: 1231806.50 MB

FOL Otter:

run 0 unsat: 8900 (31) sat: 1028 (3) cputime: 55226.70 s instructions: 230644257 Mi memory: 1023557.69 MB
run 1 unsat: 8938 (69) sat: 1027 (2) cputime: 53314.88 s instructions: 230305318 Mi memory: 1129932.88 MB

HOL Discount:

run 0 unsat: 2061 (2) sat: 1 (0) cputime: 9645.78 s instructions: 50693535 Mi memory: 119887.27 MB
run 1 unsat: 2061 (2) sat: 1 (0) cputime: 9572.82 s instructions: 50745078 Mi memory: 121430.79 MB

HOL Otter:

run 0 unsat: 2007 (7) sat: 1 (0) cputime: 9435.65 s instructions: 51411090 Mi memory: 101481.53 MB
run 1 unsat: 2003 (3) sat: 1 (0) cputime: 9377.15 s instructions: 51435440 Mi memory: 109068.85 MB

easychair:
SAT subsumption should be the slowest of all techniques. I can imagine it
to be better on a few problems due to a different pre-filtering, but
otherwise code trees should be considerably faster.

On clauses with multiple literals, we should probably add a couple of
instructions to code trees to avoid effects like a "last literal fail" with
exponential search before it. I am happy to discuss it in Vienna.

Equality subsumption can be optimized for multi-literal search when the all
of the indexed equalities in this node are pre-ordered. It can result in an
exponential improvement too in some cases.

Andrei

mezpusz:
I agree that we should focus our efforts on code tree subsumption, and I like the preordered equality trick, we can talk about it in Vienna. Yet, we should also maintain SAT subsumption as it remains the more stable and complete approach  that can be used as a baseline and debug checks (there are some known cases where code trees cannot detect a subsumption resolution, we can also discuss these).

easychair:
> I agree that we should focus our efforts on code tree subsumption, and I like the preordered equality trick, we can talk about it in Vienna. Yet, we should also maintain SAT subsumption as it remains the more stable and complete approach that can be used as a baseline and debug checks (there are some known cases where code trees cannot detect a subsumption resolution, we can also discuss these).

This means you simply use a wrong version/implementation of code
trees. There is nothing in FSR which makes it immune from code trees.

Andrei

Indexing/AcyclicityIndex.cpp
```diff
@@ -56,9 +57,7 @@ namespace Indexing
   }
 
   AcyclicityIndex::AcyclicityIndex(SaturationAlgorithm&) :
-    _sIndexes(),
-    _tis(new TermSubstitutionTree<TermLiteralClause>())
-  {}
+    _sIndexes(), _tis() {}
```
MichaelRawson: Constructor is now pointless, I think.

Indexing/CodeTree.hpp
```diff
@@ -492,7 +492,8 @@ class CodeTree
 
   //////// member variables //////////
 
-  bool _clauseCodeTree;
+  bool _clauseCodeTree = false;
+  bool _containsLiterals = false;
```
MichaelRawson: Could you explain what these variables do? When is `_clauseCodeTree` true but not `_containsLiterals`?

Indexing/LiteralCodeTree.hpp
```diff
@@ -0,0 +1,80 @@
...
+ * @file LiteralCodeTree.hpp
+ * Defines class LiteralCodeTree.
+ */
+
+#ifndef __LiteralCodeTree__
+#define __LiteralCodeTree__
+
+#include "Forwards.hpp"
+
+#include "Lib/Vector.hpp"
```
MichaelRawson: Is this needed?

MichaelRawson (approved):

Indexing/AcyclicityIndex.cpp
mezpusz: Due to the `SaturationAlgorithm&` argument I have to retain it but I will add a trivial one in the header.

Indexing/LiteralCodeTree.hpp
mezpusz: Nope, good catch!

Indexing/CodeTree.hpp
mezpusz: Great question. I added `_containsLiterals` to be able to print predicates properly, but the logic disappeared in a way I don't understand. Anyways, since printing is not used too much and we want to merge the symbol domains in #925, this won't be needed, so I just removed it for now.

PR #927, opened 2026-09-03
Our server's gcc does not like the missing semicolon

quickbeam123:
Plus some whitespace edit


MichaelRawson:
Whoops. I don't actually know how this compiled on my machine, that's weird.

MichaelRawson (approved):

PR #928, opened 2026-09-03
Fix VTIME_PROFILING-only part of the code so that TimeTrace::printPretty() does not crash in the timer thread when something else is happening in the main thread

quickbeam123:
More details in the commit message ...

### DIFF
--- a/Debug/TimeProfiling.cpp
+++ b/Debug/TimeProfiling.cpp
@@ -33,45 +33,57 @@ TimeTrace::ScopedTimer::ScopedTimer(const char* name)
 
 TimeTrace::ScopedTimer::ScopedTimer(TimeTrace& trace, const char* name)
   : _trace(trace)
+  , _active(trace._enabled.load(std::memory_order_relaxed))
 #if VDEBUG
   , _start()
   , _name(name)
 #endif
 {
-  if (_trace._enabled) {
-    auto& children = std::get<0>(trace._stack.top())->children;
-    auto node = iterTraits(children.iter())
-      .map([](auto& x) { return &*x; })
-      .find([&](Node* n) { return n->name == name; })
-      .unwrapOrElse([&]() { 
-          children.push(std::make_unique<Node>(name));
-          return &*children.top();
-      });
+  if (_active) {
+    auto& children = std::get<0>(trace._stack.back())->children;
+    Node* node = nullptr;
+    for (auto& c : children) {
+      if (c->name == name) {
+        node = &*c;
+        break;
+      }
+    }
+    if (!node) {
+      children.push_back(std::make_unique<Node>(name));
+      node = &*children.back();
+    }
     auto start = Clock::now();
 #if VDEBUG
     _start = start;
-#endif 
+#endif
 
-    _trace._stack.push(std::make_pair(node, start));
+    _trace._stack.push_back(std::make_pair(node, start));
   }
 }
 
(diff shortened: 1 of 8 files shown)


Lib/Timer.cpp
```diff
@@ -109,6 +121,10 @@ static std::recursive_mutex EXIT_LOCK;
     }
   }
 
+  // terminateImmediately is std::_Exit, which does not flush: without this the tail of
+  // the report is lost whenever stdout is a (fully buffered) file rather than a tty.
+  std::cout.flush();
```
MichaelRawson: We could consider putting this into `terminateImmediately`.

Debug/TimeProfiling.cpp
```diff
@@ -193,22 +207,34 @@ void TimeTrace::Node::printPrettyRec(std::ostream& out, NodeFormatOpts& opts)
   }
   out << ", cnt: "  << msetw(6) << cnt
       << ")" << std::endl;
-  std::sort(children.begin(), children.end(), [](auto& l, auto& r) { return l->totalDuration() > r->totalDuration(); });
-  indent.push(indentBeforeLast);
+
+  // Order a local copy rather than sorting `children` in place: this is called on the
+  // timer thread while the main thread may still be scanning and appending to
+  // `children` (Lib/Timer.cpp, limitReached()), and an in-place sort would move
+  // elements under it. Also makes printing idempotent.
```
MichaelRawson, line 214: The whole point of acquiring the lock is that this part is not multi-threaded, no?

MichaelRawson (commented):
The "set a flag and wait for a bit" approach feels like a mutex done badly. Did you try a mutex? Otherwise no objection.

quickbeam123:
> The "set a flag and wait for a bit" approach feels like a mutex done badly. Did you try a mutex? Otherwise no objection.

Claude-Suggested reply ;) : "Considered it. A mutex would cover the sort and the _stack iteration but not the third problem — flatten() allocating from the unsynchronised global allocator while the main thread allocates clauses from the same free lists — so the container/allocator change is needed regardless. And the lock would sit on a path entered ~6 M times per run (61 M in the worst case), costing ~3% and, worse, distorting exactly the small-per-call nodes the profiler exists to measure. What's there is a one-shot latch rather than a lock, deliberately: the main thread must never resume writing after the report, which a mutex would allow. You're right that the 1 ms settle is a probabilistic argument rather than a guarantee; a rigorous handshake needs seq_cst on the hot path, which costs what the mutex costs."

Martin's own: This really did fix the practical issue of having some pretty-prints destroyed in the middle or never delivered due to various SIGnalls killing Vampire instead.

Lib/Timer.cpp
quickbeam123: Excellent idea. See the new version!

Debug/TimeProfiling.cpp
quickbeam123: The idea of the lock is only to quickly stop adding new records on the stack by any new ScopedTimers.

Many ScopedTimers will still have their records there, but they have promised not to touch them anymore.

The 1ms wait is only for the single ScopedTimer, who might have read `_active == true` in its constructor, upon which the timer thread switched to "want to print, stop messing around with records", and the child is still finishing the `if (_active) {` part of its constructor.

MichaelRawson (approved):

PR #932, opened 2026-09-07
Tuning the LRS code a bit

quickbeam123:
Mainly:
- LRS was found to spend large fractions of Vampire's time budget on repeated updates of its estimate
In a nutshell, it would fire too often (looking like it was once per 50 activations, it was actually once per 50 pops from unprocessed) This PR proposes a solution where updates are not performed if they would happen too often. It tries to keep the time/instructions spent performing LRS updates under 5% of the overall budget.

Because the LRS estimate is essentially a linear scan of the passive set, it is memory intensive and often takes a lot of time (if not so many instructions). Therefore, the fix is more a win for the time-limited than for instruction limited runs. Also, the longer the runs, the larger the win. Here TPTP (v9.3.1 ALL) results for `-t 60`:

```
Sort by SAT
1114 ['problemsALLlocal_tstat11142_tstat-on_t60s.pkl']
1117 ['problemsALLlocal_tstat11156_tstat-on_t60s.pkl']

Sort by UNS
13343 ['problemsALLlocal_tstat11142_tstat-on_t60s.pkl']
13469 ['problemsALLlocal_tstat11156_tstat-on_t60s.pkl']

Total 1120 / 13509

Num only-by (UNS):
   problemsALLlocal_tstat11142_tstat-on_t60s.pkl 40
   problemsALLlocal_tstat11156_tstat-on_t60s.pkl 166

Greedy cover (UNS):
# problemsALLlocal_tstat11156_tstat-on_t60s.pkl contributes 13469 total 13469
# problemsALLlocal_tstat11142_tstat-on_t60s.pkl contributes 40 total 13343
Total 13509
```

In addition:
- no ugly goto in `SaturationAlgorithm::doUnprocessedLoop`
- options to save/load LRS traces for 100% reproducibility (if needed; off by default)

### DIFF
--- a/Saturation/LRS.hpp
+++ b/Saturation/LRS.hpp
@@ -31,9 +31,9 @@ public:
   using Otter::Otter;
 
 protected:
-  void poppedFromUnprocessed(Clause* c) override;
+  void afterUnprocessedLoop(unsigned popsElapsed) override;
 
-  bool shouldUpdateLimits();
+  bool shouldUpdateLimits(unsigned popsElapsed);
 
   long long estimatedReachableCount();
 };
--- a/Saturation/LRS.hpp
+++ b/Saturation/LRS.hpp
@@ -46,12 +46,35 @@ private:
    * solved in one process. */
   unsigned _leftoverPops = 0;
 
+  /** Running cost of limit maintenance, against which the -lmb budget is checked.
+   * Kept in both units because which one matters depends on which limit binds;
+   * see bindingResourceIsInstructions(). */
+  long long _maintenanceMicros = 0;   //< microseconds spent in updates
+  long _maintenanceInstrs = 0;        //< mega-instructions spent in updates
+
+  bool withinMaintenanceBudget();
+  bool bindingResourceIsInstructions();
+
   /** Trace of limit-update decisions, for reproducing an LRS run exactly.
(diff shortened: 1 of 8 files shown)


Saturation/LRS.hpp
```diff
@@ -20,6 +20,9 @@
 
 #include "Otter.hpp"
 
+#include <fstream>
```
MichaelRawson: Maybe `<iosfwd>` suffices here?

checks/.gitignore
```diff
@@ -0,0 +1,6 @@
+# scratch files written by the sanity script
+vampire-stdout
+vampire-stderr
+lrs-trace
+lrs-recorded
+lrs-replayed
```
MichaelRawson: Could the sanity check script either remove these on exit or put them in `/tmp`?

MichaelRawson (approved):
Good find! But perhaps this is telling us that the linear scan eventually bites, not that LRS needs throttling inherently.

I'm not coming up with any alternative quickly, so let's merge this for now but remember to revisit this? I'm imagining some kind of index to do the scan fast, or doing the deletion lazily somehow.

Saturation/LRS.hpp
quickbeam123: OK!

checks/.gitignore
quickbeam123: Improved!

Issue #925, opened 2026-09-02
Refactoring signature


mezpusz:
`Shell::Signature` is painful and error-prone to use, make it better. We could do the following things to make it better:

- [ ] Remove `setType` and pass `OperatorType` directly when creating a symbol. This removes the need for the weird `fnType`, `predType`, `typeConType` triplet. Currently being implemented in `require-symbol-operatortype`.
- [ ] Have a unified domain for all symbol types. This should be straightforward to do, we just have some edge cases where we have to enumerate functions/predicates/typecons, so we have to additionally maintain some structures storing these. The benefit of this is that a tremendous amount of branching can be removed from the code, having `getSymbol` instead of `getFunction`, `getPredicate` and `getTypeCon`, and of course, that debugging becomes much easier because symbols map one-to-one to `unsigned`s, so no more "not sure what this symbol is".
- [ ] Have some kind of nicer, builder interface for all the `makeIntroduced`, `makeSkolem`, etc. and maybe also for getting the functor itself, so the `addFunction` and immediate `getFunction` patterns could disappear. (suggested by @MichaelRawson)

shalashaska117:
Can I solve this one?

mezpusz:
Yes, feel free to do so, just note that I have an almost working first item on `require-symbol-operatortype`. The following still doesn't work:
```
./vampire --ignore_missing warn --decode fmb+10_1_si=on:fmbes=contour:fmbswr=3:random_seed=12345:fmbsr=1.29942:i=5000:nm=0:rtra=on_0 $TPTP/Problems/GRP/GRP655+1.p
...
Condition n < _size at location /Users/mezpusz/vampire/Lib/DArray.hpp:100 was violated, as:
n == 0
_size == 0
```
Probably the changes in `FMB/SortInference.cpp` are wrong.

quickbeam123:
I would prefer if someone with a longer experience with the codebase orchestrates this. I asked Michael to look into this...

MichaelRawson:
One way we could do this if you're keen @shalashaska117 is that I work out what the new interface should be and you help me adapt the existing code.

mezpusz:
@MichaelRawson @quickbeam123 feel free to add more items, or elaborate on them, I just added what was on the top of my mind.

shalashaska117:
> One way we could do this if you're keen @shalashaska117 is that I work out what the new interface should be and you help me adapt the existing code.

For me it's ok, let me know what I have to do

PR #934, opened 2026-09-07
Require symbol operatortype

mezpusz:
First item in #925. Mostly just straightforward changes from adding a symbol and then setting its type to directly supplying the type through the symbol addition call, and changing `fnType`/`predType`/`typeConType` to `type`. The `OperatorType` and `Signature` interface has been also cleaned up a bit accordingly, but I'm open to ideas how to improve it.

A few more interesting cases:
- `FMB/SortInference` had some non-trivial logic where symbols were added and then later their types were set, this has been fixed by @MichaelRawson.
- TPTP allows symbols to be declared on the fly in certain logics, these get `($i * ... * $i) > $i` types. I think this was the reason for the `fnType`/`predType`/`typeConType` setting the `OperatorType` of a symbol on the first call, but maybe there are other historical reasons too. This has been cleaned up, as this type setting can be done unambiguosly inside the TPTP parser too.
- `Splitter` declared a function constant without any type only to get a fresh function name prefix, which is weird, maybe this should be looked at more closely.
- The interface was slightly misused in one case for a tuple symbol (a bigger stack of args was supplied than the arity).
- A weird bug with `$tType` surfaced, as its creation required a type constructor to be already present in the signature, due to checking `ASS_EQ(env.signature->typeConArity(typeCon), arity);`. This was not an issue before as first we created the `$i` symbol, then created `$tType`. Now the order changed, so I just replaced the non-shared creation of `$tType` with a direct call `new AtomicSort`.

FOL regression and debug sampling are okay, the rest I'm running now (including ALASCA to test theory symbol creation).


mezpusz:
HOL regression and debug sampling are okay, ALASCA regression also came back without any change.

mezpusz:
Same with induction and synthesis regressions, no change. So the PR is ready to go!

Kernel/Signature.hpp
```diff
@@ -81,15 +81,14 @@ class Signature
     /** print name */
     std::string _name;
 
+    OperatorType* _type;
     // both _arity and _typeArgsArity could be recovered from _type. Storing directly here as well for convenience
 
     /** arity */
     unsigned _arity;
     /** arity of type arguments */
     unsigned _typeArgsArity;
```
MichaelRawson: Now that `_type` is mandatory, are these constants simply mirrors of the `_type`'s total arity/type argument arity? I suspect there are some edge cases (arithmetic?) but if not it would be good to fold them in.

Shell/TermAlgebra.cpp
```diff
@@ -216,17 +216,8 @@ std::string TermAlgebra::getSubtermPredicateName() {
 
 unsigned TermAlgebra::getSubtermPredicate() {
   bool added;
-  unsigned s = env.signature->addPredicate(getSubtermPredicateName(), nTypeArgs()+2, added);
-
-  if (added) {
```
MichaelRawson, line 221: What happened to this logic, sorry?

MichaelRawson (approved):

Shell/TermAlgebra.cpp
mezpusz: The `OperatorType` creation just collapses to calling it with `{ _sort, _sort }` and then we just supply it to `addPredicate` no matter what, so potentially creating a `OperatorType` unnecessarily but avoiding a branch. 🤷

Kernel/Signature.hpp
mezpusz: You're right, these can be deleted. Interpreted symbols also use `Symbol` with all its functionality, they just have some extra special functions on top.

MichaelRawson: Great! OK, perhaps in a separate PR then, I guess there could be some performance impact.

PR #935, opened 2026-09-09
require explicit hashes and remove DefaultHash

### LINKED ISSUE #898: make container hash functions explicit and remove DefaultHash

shalashaska117:
Closes #898 (6/6).

Make hash arguments mandatory on the container templates and pass them through the remaining generic helpers. Add explicit hashes for coproduct alternatives and polynomial keys, then delete `DefaultHash`, `DefaultHash2`, the forwarding methods, and the tuple hash macro.

Keep the hash formulas from #919, including the distinction between FNV and identity hashes in polynomial caches. The replacement tests use fixed hash vectors and cover mixed coproduct alternatives, polynomial keys, and unique iterators.

Checks:

- After rebase onto master: GCC Debug, all 101 test suites pass; GCC Release build, copyright headers, and `git diff --check` pass.
- Release sanity: `SWC478_1` times out at the standard 0.1-second limit on both current master and this branch. All remaining checks pass with that case excluded.
- Before rebase, output sweeps against #919: 3,777 comparable results match across 202 inputs and 19 configurations; 61 cases are excluded by repeated-baseline controls.
- In those sweeps, five timeout cases match on rerun with `-t 180`: four at `-al 1000`, and `checks/parse/thf-not-eq.p` under Otter with ALASCA at `-al 100`. The latter still times out on the baseline at 1,000 activations.


MichaelRawson:
Hey @shalashaska117 - can you e.g. rebase this so I only get the second commit as a diff? Some very minor merge conflict too.

Thanks for all the work on this series. If you're up for it, I think the next step is to factor out the `equals()` method that `Map` (but not `DHMap`) uses from the `Hash` parameter so that it's obvious when we're using something weird and not `operator==`. After that we can look at aligning the interface of `Map` and `DHMap` for an eventual merger.

shalashaska117:
Rebased onto current master, so the diff now contains only stage 6. Resolved the conflict and added explicit hashes to the new `_distinctPredicates` map.

Both builds and all 101 unit test suites pass. The release sanity check for `SWC478_1` times out on both current master and this branch; all remaining sanity checks pass with that case excluded.

I'll open an issue for the next target: separating `equals()` from the hash parameter in `Map`.

MichaelRawson (approved):

PR #936, opened 2026-09-09
All the stuff Geoff asked for regarding the implicitly distinct typed "constants" and $distinct handling in the TPTP version

quickbeam123:


checks/parse/distinct-objects-poly.p
```diff
@@ -0,0 +1,7 @@
+% a distinct object of a polymorphic (non-ground) sort is not supported
```
MichaelRawson, line 1: Good. This is arguably the same case as functions, as a polymorphic constant is simply a kind of dependent function from sorts to a constant of that sort.

Kernel/Signature.cpp
```diff
@@ -869,6 +870,68 @@ void Signature::addToDistinctGroup(unsigned constantSymbol, unsigned groupId)
...
 
+/**
+ * Return the distinct group collecting the string constants ("distinct objects") of
+ * @c sort, creating it if this is the first one of that sort.
+ *
+ * $i uses STRING_DISTINCT_GROUP, which is reserved in the constructor; this cannot be
+ * folded into the map, since the constructor must not call AtomicSort::defaultSort().
+ */
+unsigned Signature::getStringDistinctGroup(TermList sort)
+{
```
MichaelRawson, line 881: `ASS(sort.ground())`?

Kernel/Signature.cpp
```diff
@@ -869,6 +870,68 @@ void Signature::addToDistinctGroup(unsigned constantSymbol, unsigned groupId)
...
+  unsigned result;
+  if (_distinctPredicates.find(key,result)) {
+    return result;
+  }
+
+  result = _preds.length();
+  Symbol* sym = new Symbol("$distinct", OperatorType::getPredicateTypeUniformRange(arity,sort),
+        /*       interpreted */ false,
+        /*    preventQuoting */ true);
+  sym->markDistinctPred();
```
MichaelRawson, line 924: `sym->markSkipForCongruence` would prevent any equality-proxy shenanigans.

Shell/DistinctGroupExpansion.cpp
```diff
@@ -45,13 +102,143 @@ void DistinctGroupExpansion::apply(Problem& prb)
...
+
+  return changed;
+}
+
 /**
- * Attempts to expand each recorded distinct group
+ * Phase 2: attempts to expand each recorded distinct group
  * (this includes those for builtin sorts i.e. ints, strings...)
  * If all groups are expanded we indicate there are no distinct groups left, which will
  * prevent the distinct simplifier being added later
```
MichaelRawson, line 239: Probably this could also be done by having `Parse::TPTP` emit some final axiom like `$distinct("apple", "banana")` before it finishes parsing. I have no opinion on what is better, just so that you consider the option.

MichaelRawson:
OK to go when you want, I only have non-critical comments!

Kernel/Signature.cpp
quickbeam123: It's meant to disappear as the first thing during preprocessing.

Shell/DistinctGroupExpansion.cpp
quickbeam123: This is a good idea to remember! 

Especially with the `DistinctEqualitySimplifier`. Because the expansion at least says `NonspecificInference0(UnitInputType::AXIOM,InferenceRule::DISTINCTNESS_AXIOM))` for the quadratic thing it produces for the "distinct strings". But the inference rule does not have anything to point to in that case.

But I think it's low prio to work on right now.

PR #941, opened 2026-09-14
AVATAR calls addFreshFunction when it should be addFreshPredicate

MichaelRawson:
@shalashaska117 says in #939 

> Proof checks use -av off because baseline master also crashes while printing an AVATAR symbol declaration: Splitter::getFormulaStringFromName() passes a predicate type to addFreshFunction(). The unchanged Debug binary asserts in UIHelper::outputSymbolTypeDeclarationIfNeeded() before printing the issue's original $ite formula.

Vampire currently crashes if you ask for TPTP proof output on an TFF input, e.g. `PUZ012_1`.

Whoops, this was probably my contributions to @mezpusz 's otherwise-successful #934 changes.


mezpusz (approved):
I wouldn't have expected that this makes a difference, but then let's do it the right way.

quickbeam123 (approved):

PR #943, opened 2026-09-14
Do not run DistinctGroupExpansion on higher-order problems

quickbeam123:
Short story from Martin:

The last, `$distinct` PR did not think of HOL, who might parse "strings" too (we disallow), but worse, we can't run `DistinctGroupExpansion` the HOL presentation. The fix: skip `DistinctGroupExpansion` for HOL but complain if is DistinctGroup has been parsed (for "strings").

Long story from Claude:

Vampire currently aborts on essentially every higher-order input, printing only

    Not implemented at Kernel/FormulaTransformer.cpp:113

Reproducer, four lines and no $distinct anywhere:

    thf(p_decl, type, p: ($i > $o) > $o).
    thf(q_decl, type, q: $i > $o).
    thf(ax, axiom, (p @ (^[X: $i]: (q @ X)))).
    thf(co, conjecture, (?[F: $i > $o]: (p @ F))).

    vampire --input_syntax tptp that-file.p

DistinctExpander::applyLiteral descends into the arguments of any unshared literal, to find a $distinct marker hiding inside a FOOL term. In higher-order logic a literal holding a lambda is unshared as well, so HOL terms enter that descent, reach FormulaTransformer::apply(TermList) and hit its NOT_IMPLEMENTED for SpecialFunctor::LAMBDA. Two earlier changes combine: removing the hasDistinctGroups() guard from Preprocess (right in itself -- a $distinct marker can exist with no group yet) made the pass run on every problem, and the FOOL descent gave it something to crash on. A sweep of TPTP has 3 316 runs dying this way; TH0 refutations fall from 1 951 to 824.

Skip the pass for higher-order problems. Sound for the markers, which thf rejects at parse time. Not sound for the other source of distinctness -- a distinct object still parses in thf, both in an equality and in a type declaration, so a higher-order problem can carry a distinct group, and silently dropping the disequalities it stands for would lose what the input asserted. No TPTP problem does this, so rather than leave a trap, reject it explicitly.

checks/sanity had no test that could catch any of this: hol1.p is its only end-to-end higher-order entry and has a higher-order *variable* but no lambda, so nothing unshared carrying a special term ever reaches preprocessing; every other thf entry stops at --mode output, before preprocessing runs. Add hol2.p, which is the reproducer above and fails before this commit, and hol-distinct-object.p for the rejected case.


shalashaska117:
We could make make a Project Dashboards with every step that needs to be done for HOL implementation, because there is not a discussion topic here for vprover/vampire

mezpusz (approved):

PR #944, opened 2026-09-14
Unify signature symbol IDs and compact category storage

### LINKED ISSUE #925: Refactoring signature

shalashaska117:
Functions, predicates, and type constructors share one symbol table and global ID domain; equality stays at ID 0. Category-specific caches use stable, dense indices, while precedence maps and FMB tables share storage indexed by global ID.

The fluent construction API and its caller migration are now in #960, as requested in review. This PR keeps the symbol-ID, storage, and iteration changes, including the predicate-registration fixes from #941.

Category iteration preserves snapshot ranges because FMB registers functions while iterating existing ones. Tests cover registration during iteration, interleaved IDs, dense indices, late-symbol ordering, usage counting, and ground-model evaluation.

Checks:
- GCC 15.2 Debug and Release builds; all 102 unit-test groups pass. Header and whitespace checks pass.
- After the split, 590 problem/option cases match the previous PR head in status and exit code. Successful clausification output is identical after removing build/runtime metadata. Four crashes on two parser fixtures reproduce before the split.
- Release sanity passes with the known `SWC478_1` timeout omitted in a temporary harness. The unchanged suite hits the same timeout on master; repository limits are unchanged.

Earlier measurements of the dense-index changes versus the rebased draft (`5474ab043`), using identical GCC 15.2 Release builds: five alternating runs with `-sa discount -al 1` reduced median peak RSS from 215 to 164 MiB for 100,000 functions/100 predicates, and from 32.0 to 26.9 MiB for 10,000/100. Median runtimes were 0.397 vs 0.374 s and 0.0387 vs 0.0371 s. The inverse 100/10,000 case stayed near 32.5 MiB and 0.036 s. These short runs do not establish a general speedup.

Part of #925; the construction API is handled separately in #960.


shalashaska117:
@mezpusz @MichaelRawson @quickbeam123 let me know what you think about the implementation, I cleaned some code too

Kernel/Signature.hpp
```diff
@@ -661,29 +682,45 @@ class Signature
...
-  {
-    ASS_L(n, _preds.length());
-    return _preds[n];
-  } // getPredicate
-  inline Symbol* getTypeCon(unsigned n)
-  {
-    ASS_L(n, _typeCons.length());
-    return _typeCons[n];
+  /** A snapshot of a category's IDs; remains valid if registration grows its storage. */
+  class SymbolRange {
```
mezpusz, line 689: Where is this needed vs just plain `const Stack<unsigned>&`?

mezpusz:
The changes look okay although I haven't checked in detail. But we have to be really careful with this as lots of code relies on the assumption that a symbol category is in `[0, categorySize)` range (and this part I haven't anticipated in #925). So either we're allocating much bigger arrays for counting symbols (for example in predicate elimination AFAIK), or we might even get incorrect results (in precedence orderings there are precedence maps for predicates and functions separately, these are probably broken and should be merged).

Kernel/Signature.hpp
shalashaska117: the advantage is that callers can write a normal range-for safely, even when something deeper in the call chain registers new symbols. They don't have to remember to save the size and iterate by index. It also avoid copying the ID list. Maybe it's harder to maintain but it makes everything more secure and easier to read, Michael said where I could to improve the code readability.

shalashaska117: maybe if we just need this behavior for one or two loops it would help make it easier but for the long term my abstraction would help

shalashaska117:
> The changes look okay although I haven't checked in detail. But we have to be really careful with this as lots of code relies on the assumption that a symbol category is in `[0, categorySize)` range (and this part I haven't anticipated in #925). So either we're allocating much bigger arrays for counting symbols (for example in predicate elimination AFAIK), or we might even get incorrect results (in precedence orderings there are precedence maps for predicates and functions separately, these are probably broken and should be merged).

Uh, yes, with 100 predicates and 10.000 functions, we would need 10100 entries and not 100. We could merge their storage, so avoiding the unused entries.

mezpusz:
Perhaps a reverse map that stores which symbol has what index within the function/predicate/type-con ranges. The question is whether this extra indirection is faster than allocating excessive arrays.

shalashaska117:
> Perhaps a reverse map that stores which symbol has what index within the function/predicate/type-con ranges. The question is whether this extra indirection is faster than allocating excessive arrays.

I could run some benchmark for speed and memory if it's useful

Kernel/Signature.hpp
```diff
@@ -442,6 +455,46 @@ class Signature
   // Uninterpreted symbol declarations
   //
 
+  /** A registered symbol and its ID. Flag setters can be chained at creation. */
```
mezpusz: Would it be possible to have this as a separate PR, including changing all places to use this construction?

shalashaska117: Yes, give me 1 hour and it will be done. Is this amount of comments good? Or do you need more/less?

mezpusz: Thanks! I prefer comments only to indicate something non-trivial or very important, the more concise the better.

MichaelRawson:
I like the fluent API, that's cool. And I like the *idea* of merging the function/predicate/type constructor tables, but as @mezpusz points out it is risky and Vampire's code does rely on subtle invariants at times. How about we do the fluent API first (I see you have a PR, thanks) and then have a very nervous follow-up to exhaustively (exhaustingly) test that we didn't break anything with the merged symbol tables?

shalashaska117:
> I like the fluent API, that's cool. And I like the _idea_ of merging the function/predicate/type constructor tables, but as @mezpusz points out it is risky and Vampire's code does rely on subtle invariants at times. How about we do the fluent API first (I see you have a PR, thanks) and then have a very nervous follow-up to exhaustively (exhaustingly) test that we didn't break anything with the merged symbol tables?

Working on that right now!

shalashaska117:
Follow up thing, we need to improve the testing part of Vampire, we need way more test, with all edge cases and so on, probably it could help us spotting things easier and more precisely

shalashaska117:
Check flexibleTail::flexibleTail(), it calculates from the base-subobject address, assuming that address equals the derived-object address. This is valid where than can be a misaligned pointer and an out-of-bonds write.

shalashaska117:
also, we should add tests for alignment, empty tails and allocation reuse. We should also track SkipList::Node, which still uses the one-element trailing-array pattern

MichaelRawson:
> Follow up thing, we need to improve the testing part of Vampire, we need way more test, with all edge cases and so on, probably it could help us spotting things easier and more precisely

If you have any ideas I think we're open to it! There are a few ways we test Vampire now:
- Unit tests. These work reasonably well but don't seem to run particularly fast (?!) for some reason I don't yet understand.
- The sanity-check script, `checks/sanity`. This does some kind of "integration test".
- Manual "crater runs" on TPTP for performance and/or crash detection.

The requirements are:
1. Catches bugs.
2. Doesn't require excess tooling, Vampire is already hard to get started with/distribute.
3. Isn't otherwise a burden.

Kernel/Signature.cpp
```diff
@@ -31,10 +31,9 @@ const unsigned Signature::STRING_DISTINCT_GROUP = 0;
  * Standard constructor.
  * @author Andrei Voronkov
  */
-Signature::Symbol::Symbol(unsigned number, const std::string& nm, OperatorType* type, bool interpreted, bool preventQuoting)
+Signature::Symbol::Symbol(const std::string& nm, OperatorType* type, bool interpreted, bool preventQuoting)
```
mezpusz, line 34: Why remove adding the number here?

shalashaska117: I moved ID assignment into `Signature::addSymbol()`, which inserts the symbol into the shared table and assigns its category index. It uses the next global ID, except for equality, which gets the reserved slot 0.

The symbol still stores its number: `_number` starts as `UINT_MAX` and is set during registration, before the creation API returns it. The intent was to keep ID allocation in one place; passing the number through the constructors is still possible.

PR #946, opened 2026-09-16
abstraction for safe (?) flexible array members

MichaelRawson:
Anti-UB patches from the train, volume N of many.

The problem here is that some classes (`Clause`, `SATClause`, `FlatTerm`, `Term`) use a flexible array member, which is standard C but not C++. We run into UB like in #919 where I think the problem is that the compiler assumes that e.g. `SATLiteral _literals[1];` means that there is *exactly one* literal in the struct and optimises accordingly.

Instead, skip the member and simply allocate some memory, then compute where the pointer should start.

I've done this with a "mixin" style `FlexibleTail` class that can be applied to anything that needs a flexibly-sized tail. Hopefully this abstracts the evil away sufficiently. I've demonstrated with `SATClause` - if you like this I can extend it to other similar classes.

@shalashaska117 - could you take a look and see if this fixes the flakiness you saw in #919 ?


shalashaska117:
Tested b4ddfb0e4 with GCC 15.2.0 on x86-64 WSL. The hash mismatch still happens at `-O3`, but I can now reproduce it with an ordinary `int[6]`, without any Vampire code. `-O0`, `-O2`, and `-O3 -fno-tree-vectorize` all give the expected result. This points to a GCC vectorization issue, so my earlier explanation blaming the one-element array was too strong.

I also found two things left over from the old layout in `SATClause.cpp`:

- The constructor loop still starts at `i = 1`. There is no longer an array member to construct literal 0 automatically, so the loop needs to start at 0. Filling the allocation with `0xA5` before construction leaves the first literal unchanged on this branch; master initializes it to zero. The destruction loop and its comment need the same update.
- `destroy()` still uses the old allocation-size formula. It should use `allocationRequired(_length)`, as the updated `operator delete` does. A unit clause now requests 20 bytes, but `destroy()` frees it as 16, putting it into the wrong allocator size class. I confirmed that a subsequent `Lib::alloc(16)` reuses that address.

The Release build and all 101 Debug unit test suites pass. The hash mismatch needs an optimized test to catch it.

<details>
<summary>Standalone hash reproducer</summary>

```cpp
#include <cstdio>

unsigned fnv(int value) {
  const auto* bytes = reinterpret_cast<const unsigned char*>(&value);
  unsigned hash = 2166136261u;
  for (unsigned i = 0; i < sizeof(value); ++i)
    hash = (hash ^ bytes[i]) * 16777619u;
  return hash;
}

__attribute__((noinline)) unsigned hash(const int* values, unsigned length) {
  unsigned result = 0;
  for (unsigned i = 0; i < length; ++i)
    result ^= fnv(values[i]);
  return result;
}

int main() {
  const int values[] = {104730, -112649, 120568, -128487, 136406, -144325};
  std::printf("%u\n", hash(values, 6));
}
```

With `g++ -std=c++20 -O3 repro.cpp`, this prints `92697581`. The expected result is `4018666989`, which I get with `-O0`, `-O2`, or `-O3 -fno-tree-vectorize`.

</details>

shalashaska117:
Let me know if you want me to look it further, I honestly don't know right now how to solve it right now

MichaelRawson:
Thanks for catching those! Fixed - and weird with the compiler bug, but no need to figure it out for our sake. My GCC (14.2)  also does this, which is a bit alarming. Clang is well-behaved.

shalashaska117:
> Thanks for catching those! Fixed - and weird with the compiler bug, but no need to figure it out for our sake. My GCC (14.2) also does this, which is a bit alarming. Clang is well-behaved.

C++ is the most beautiful language in the world also thanks to this funny things

shalashaska117:
I can't commit on this PR but I wrote the fix here: https://github.com/shalashaska117/vampire/commit/c3407ad35b735589f60c7d7065422a591cdcfed4

It works with `-03`  flag on

MichaelRawson:
Thanks for investigating, but I think we'll just tolerate this kind of thing - I imagine that the GCC folks have a fix, and as long as it hashes consistently it's not a problem. If on the other hand it hashed inconsistently then we'd probably have to do something.

shalashaska117:
> Thanks for investigating, but I think we'll just tolerate this kind of thing - I imagine that the GCC folks have a fix, and as long as it hashes consistently it's not a problem. If on the other hand it hashed inconsistently then we'd probably have to do something.

I tried with 16.2 and it's not fixed, if it matters

Lib/FlexibleTail.hpp
```diff
@@ -0,0 +1,85 @@
...
+ * Utility class for derived classes that need a flexible array member at the end, like
+ *
+ * struct Term {
+ *   unsigned arity;
+ *   TermList args[]; // zero or more TermLists
+ * };
+ *
+ * which are intended as one allocation:
+ *
+ * arity | <maybe padding> | arg[0] | ... | arg[arity]
```
quickbeam123, line 22: I think we want this until `arg[arity-1]`

Also a bit futher below in another comment

Lib/FlexibleTail.hpp
```diff
@@ -0,0 +1,85 @@
...
+  constexpr static size_t tailOffset() {
+    size_t derived = sizeof(Derived);
+    // may need some padding
+    if(derived % alignof(T))
+      derived += alignof(T) - (derived % alignof(T));
+    return derived;
+  }
+
+  // number of bytes required for Derived with a tail of `length`
+  constexpr static size_t allocationRequired(unsigned length) {
```
quickbeam123, line 66: `allocationRequired` sounds a bit like a boolean. Could we rename it to something like `allocateHowMany` or even better?

quickbeam123 (approved):

quickbeam123:
Looking cool. Looking forward to see it extended!

shalashaska117:
> Looking cool. Looking forward to see it extended!

Would it be possible to get permission to commit to PRs? I would love to add my commit, if it's a good solution

MichaelRawson:
Extended to everything I could find (with a regex!), except:

1.  `FlatTerm`, which would require lifting `FlatTerm::Entry` out in order to forward-declare it.
2. `Vector<T>`, which is generally fine but then the `GET_CONTAINING_OBJECT` macro breaks and I don't understand the code that uses it well enough to touch it.

If @mezpusz feels inspired he can fix these too, but it may be more trouble than it is worth.

MichaelRawson:
Testing: shuffle runs on FOF didn't find any assertion violations or crashes. There should not be a performance regression, but if you are suspicious I can run something.

MichaelRawson:
> Would it be possible to get permission to commit to PRs? I would love to add my commit, if it's a good solution

@shalashaska117 - I hope you won't be offended, but not yet. We tend to be cautious with external contributors and the current AI climate only makes us more so.

However, we're happy that you want to contribute to Vampire and are grateful for your patches. If you ping the author of a PR with a commit that they're happy to take, they should be able to cherry-pick from your remote (which gives you credit).

In this particular case I don't think the GCC miscompile (if that's what it is) is worth modifying our FNV implementation for. If it's still broken on 16.2 and you are absolutely sure your MWE is not UB* you should file it upstream against GCC! Finding a real miscompile is rare and a feather in anyone's cap.

*I don't see anything wrong personally. But C++ semantics can be very surprising. Let me know, either way - I'd be curious.

shalashaska117:
> @shalashaska117 - I hope you won't be offended, but not yet. We tend to be cautious with external contributors and the current AI climate only makes us more so.
> 
> However, we're happy that you want to contribute to Vampire and are grateful for your patches. If you ping the author of a PR with a commit that they're happy to take, they should be able to cherry-pick from your remote (which gives you credit).
> 

Don't worry! I understand it perfectly, cautious is more important than ever today, I completely understand. Security, especially in a C++ source code, is probably the most important thing.

MichaelRawson:
@shalashaska117 - probably you meant to comment here?

> Check flexibleTail::flexibleTail(), it calculates from the base-subobject address, assuming that address equals the derived-object address. This is valid where than can be a misaligned pointer and an out-of-bonds write.

Not sure I understand this. Could you spell it out for me? `this` should be the same pointer in both base and derived classes (I hope), otherwise a pointer downcast couldn't work.

> SkipList::Node still uses the one-element trailing-array pattern

So it does. Looking into it...

shalashaska117:
sorry, wrong page! deleting the comment over there. 

> flexibleTail::flexibleTail()

A base-to-derived `static_cast` can adjust the address when the base sub-object is at a nonzero offset. That adjustment is what makes the downcast work.

Since `FlexibleTail` is empty, empty-base optimization normally gives it the same address as the derived object. My reproducer uses this layout:

```cpp
struct TailOwner;
using Tail = Lib::FlexibleTail<TailOwner, int>;

struct Prefix { Tail member; };

struct TailOwner : Prefix, Tail {
  int length = 1;
};
```

On my GCC/x86-64 build, the `Tail` base is one byte after the start of `TailOwner`. `tailOffset()` is 8, so the current implementation returns an address 9 bytes from the start of the object instead of 8. Allocating `bytesRequiredFor(1)` and writing the tail element produces UBSan reports for misalignment and insufficient space.

Converting to `Derived*` before doing the byte arithmetic fixes the reproducer:

```cpp
T* flexibleTail() {
  auto* derived = static_cast<Derived*>(this);
  return reinterpret_cast<T*>(
      reinterpret_cast<char*>(derived) + tailOffset());
}
```

The const overload would need the equivalent change.

I haven't found a current Vampire caller that triggers this layout. I should have been clearer that this is an edge case in the reusable helper.

shalashaska117:
When I code something, I test all edge cases, so that even if the program is working, I can check for things that could break later, this helps understand the program better, its limit and how I use c++.

MichaelRawson:
Eek! That's a scary corner of the language that I didn't know. Do you need multiple inheritance to get there? Applying your fix, anyhow.

Ideally we just shouldn't be doing this kind of thing. But since we have to, let's put all the evil in one place.

> When I code something, I test all edge cases

That is the ideal, yes. :)

quickbeam123:
Nice! I was especially surprised how smoothly things went with `Term`. 

No performance regression detected. Good to go?

MichaelRawson:
I was also surprised tbh, I expected at least one nightmare bug. Maybe I will still be right. :) Good to go!

PR #950, opened 2026-09-17
Fix issue in `ClauseCodeTree::optimizeLiteralOrder`

Synrom:
This PR improves `ClauseCodeTree::optimizeLiteralOrder` 
Previously, literals were not reordered when the tree was empty.

It also fixes the handling of `LIT END`s. 
Each literal was compiled independently, so the global variable numbers assigned did not match the numbers produced by the actual compilation.
This matters because whether two LIT ENDs can be merged depends on their global variable numbers.

The changes are evaluated on TPTP using `-i 20000`:
[ordering.otter.stat.txt](https://github.com/user-attachments/files/32344465/ordering.otter.stat.txt)
[ordering.discount.stat.txt](https://github.com/user-attachments/files/32344466/ordering.discount.stat.txt)


quickbeam123:
Are the evaluation txt files pointing in the direction of a mild improvement? 

It's not clear to me what `run 0` and `run 1` are...

mezpusz:
`run 0` is this branch, `run 1` is master. I think it's clearly a bugfix, so we should merge it, but it also improves performance a bit.

mezpusz (approved):

PR #954, opened 2026-09-18
Let EqualityProxy find the used symbols itself

quickbeam123:
Congruence axioms are added only for symbols that occur in the problem, and that test was Signature::Symbol::usageCnt() != 0 -- a count maintained as a side effect of Property::scan, used here as a boolean. Preprocess had to force a whole extra scan of the problem right before the equality proxy just to refresh it ("refresh symbol usage counts, can skip unused symbols for equality proxy"), and the exactness of that count is part of why Property::scan cannot skip a subterm it has already visited.

Compute it locally instead. EqualityProxy::collectUsedSymbols walks the clauses it has been handed and fills two DArray<bool>, which addCongruenceAxioms then reads; nothing else pays for it, and nothing pays for it at all unless -ep rstc really asks for congruence axioms. Because only presence is recorded and never a count, the walk can stop at a subterm it has already seen -- whether a symbol occurs below a shared term does not depend on how many times that term occurs -- so it is a walk over the term DAG rather than over the tree it unfolds to. The forced rescan in Preprocess goes away with it; prb.hasPolymorphicSym() on the next line still refreshes the property if it needs to.

This turns out to fix a real defect. PredicateDefinition:: removeUnusedDefinitionsAndPurePredicates (-updr, on by default) mutates the unit list without calling any invalidate*, so prb.getProperty() found _propertyValid still set and returned the cached Property without rescanning -- leaving usageCnt describing a unit list from before -updr ran. Congruence axioms were then emitted for predicates -updr had already eliminated: isolated clauses whose predicate occurs nowhere else in the problem, so they can only ever resolve against themselves.

Pre-fix reproducer (sQ0_eqProxy congruence axioms for barrel, down, in, white, dirty, lonely and partof, none of which occurs in any other clause):

  ./vampire --mode clausify -ep rstc Problems/NLP/NLP002-1.p

Verified over 18 problems x 9 configurations (-ep {rstc,rst,rs,r}, plus rstc with -newcnf on, -updr off, -updr off -newcnf on, -bce on, -ins 3), comparing --mode clausify output: 150 of 162 byte-identical. All 12 differences are the three NLP problems under the six configurations that leave -updr on; the two -updr off configurations agree for all three, which is what pins the cause on the stale cache. Every difference is a removal, never an addition (NLP014-1: 101 clauses -> 89), and each removed axiom is for a predicate with zero occurrences elsewhere in the clause set. ctest 101/101, checks/sanity green against a release build.

The behaviour change is sweep-worthy: -ep rstc on a problem where -updr removes predicates now produces a smaller clause set.


Shell/EqualityProxy.hpp
```diff
@@ -17,6 +17,7 @@
 
 #include "Forwards.hpp"
 
+#include "Lib/DArray.hpp"
```
MichaelRawson, line 20: I think you only need the declaration from `Forwards.hpp` here, then `#include "DArray.hpp"` in `EqualityProxy.cpp`.

Shell/EqualityProxy.cpp
```diff
@@ -210,13 +210,70 @@ bool EqualityProxy::getArgumentEqualityLiterals(unsigned cnt, LiteralStack& lits
...
+        // all of its arguments, so the whole subtree is of no interest here
+        if (t->isSort() || !seen.insert(t)) {
+          continue;
+        }
+        usedFunctions[t->functor()] = true;
+        for (TermList* ts = t->args(); ts->isNonEmpty(); ts = ts->next()) {
+          if (ts->isTerm()) {
+            todo.push(ts->term());
+          }
+        }
```
MichaelRawson, line 255: Isn't this whole loop just `NonVariableNonTypeIterator`?

MichaelRawson (changes requested):

Shell/EqualityProxy.hpp
quickbeam123: The new code will actually be moved out of `EqualityProxy.cpp` higher up in out PR tower, but thanks for spotting this, `#include "Lib/DArray.hpp"` would have otherwise stayed...

Shell/EqualityProxy.cpp
quickbeam123: Later on, this is called `collectUsedSymbols` in `SymbolUsage`.

The main difference to an iterator is the `seen`-skipping which makes this DAG-aware. 

The could be a different kind of iterator or a templated version, but I don't think you can simply iterate in a plain way and `seen`-skipping in the loop and claim you have reaped the benefit, because you will at the very least do the expensive iteration, if not the expensive "in loop" work.

Maybe the `seen`-skipping should be added to `NonVariableNonTypeIterator` by calling its `right()` method on `seen`. What do you think, should we have it instead? 

(I would prefer doing it on top of the PR tower though, there is definitely some auto-rebase going on behind the scene, which is a bit disorienting...)

quickbeam123: More learned: `NonVariableNonTypeIterator` tries some limited support of special terms, which this iteration pattern boldly declares unreachable and skips. Whether to support those as well would be another decision to make.

MichaelRawson: Right - sorry, I got myself confused with the DAG + `seen` dance. Maybe `BottomUpEvaluation` could help here? But it's certainly not simple, let's merge and move on.

MichaelRawson (approved):

PR #956, opened 2026-09-18
Ask the clauses which functions FMB's sort predicates must cover

quickbeam123:
Under -fmbas predicate, FiniteModelBuilder::init builds a deleted_functions array and hands it to Monotonicity::addSortPredicates, which uses it to decide for which f it need not assert "!args: p(f(args))". The array was filled from Signature::Symbol::usageCnt(), i.e. from whatever the last Property::scan of the run happened to leave on the signature -- while the clauses it is supposed to describe are pushed from _prb.clauseIterator() on the very next line.

Move the presence walk introduced for EqualityProxy in the previous commit out to Kernel/SymbolUsage, where both callers can reach it, and take a ClauseIterator rather than a UnitList so it fits FMB too. deleted_functions is then the complement of a walk over exactly the clause list being transformed, which is right by construction rather than by luck. EqualityProxy keeps its behaviour; it adapts its unit list the way Problem::clauseIterator does.

The stale read was observable. On Problems/SYN/SYN000-1.p with -fmbas predicate, usageCnt claimed 1 for three function symbols -- 'A constant', 'A function' and 'A \'quoted \\ escape\'' -- that occur in no clause reaching finite model building, so sort predicate axioms and a non-emptiness constraint were emitted for each, inflating the inferred sort. The problem now needs domain sizes up to 3 where it needed 5, reaching the same SZS status Unsatisfiable by the same proof (46 clauses created rather than 52).

The cause is the same one behind the previous commit: -updr (PredicateDefinition::removeUnusedDefinitionsAndPurePredicates) mutates the unit list without calling any invalidate*, so the property stays "valid" and the usage counts keep describing the pre-updr problem. Accordingly the difference disappears under -updr off, which is what identifies it.

Verified over 12 problems x 8 configurations (-fmbas predicate alone and with -fmbdsb on / -fmbsso usage / -updr off / -bce on, plus -fmbas function, group and expand), comparing full output: 93 of 96 identical. The three differences are SYN000-1 under -fmbas predicate, -fmbas predicate -bce on and -fmbas predicate -fmbsso usage; SYN000-1 under -fmbas predicate -updr off is among the identical ones. The EqualityProxy move is a pure refactor: the 162-output clausify comparison from the previous commit is 162/162 unchanged. ctest 101/101, checks/sanity green against a release build.


Kernel/SymbolUsage.cpp
```diff
@@ -0,0 +1,71 @@
...
+        }
+      }
+      while (todo.isNonEmpty()) {
+        Term* t = todo.pop();
+        ASS(t->shared());
+        // a sort is built from type constructors, not from function symbols, and so are
+        // all of its arguments, so the whole subtree is of no interest here
+        if (t->isSort() || !seen.insert(t)) {
+          continue;
+        }
```
MichaelRawson, line 59: Could argue another time that this utility should be extended to include type constructors (and thereby encourage people thinking about them), but another time.

Kernel/SymbolUsage.hpp
```diff
@@ -0,0 +1,40 @@
...
+ * @file SymbolUsage.hpp
+ * Answers questions about how the symbols of the signature are used by a set of clauses.
+ */
+
+#ifndef __SymbolUsage__
+#define __SymbolUsage__
+
+#include "Forwards.hpp"
+
+#include "Lib/DArray.hpp"
```
MichaelRawson, line 20: Same deal with fwd-decl here, but 🤷  I'll do another sweep for include performance separately.

MichaelRawson (approved):

Kernel/SymbolUsage.hpp
quickbeam123: Something seems to be different here:
```
./Kernel/SymbolUsage.hpp:41:20: error: implicit instantiation of undefined template 'Lib::DArray<unsigned int>'
   41 |   DArray<unsigned> functions;
      |                    ^
./Forwards.hpp:29:25: note: template is declared here
   29 | template<class C> class DArray;
      |                         ^
```

MichaelRawson: Whoops. Then forget it, sorry. Rule is that pointers and reference only need a forward declaration, but this is the structure itself.

PR #958, opened 2026-09-19
Build the ordering's frequencies from the clauses, not from the signa…

quickbeam123:
…ture

Symbol precedence under sp=frequency and its four relatives, and KBO weights under kws=frequency / inv_frequency, asked Signature::Symbol::usageCnt() how often a symbol occurs. To make sure that counter said something, the PrecedenceOrdering constructor called prb.getProperty() with a comment claiming it would "(re-)compute usageCnt's for all the symbols". It would not: getProperty() rescans only when the cached Property has been invalidated, and returns the old object otherwise, leaving whatever counts the last real scan left behind.

Count the clauses instead. SymbolCounts (the three arrays from the previous commit, now a struct with a countIn method) becomes a protected member of PrecedenceOrdering, filled once from prb.clauseIterator() at construction and shared by all three precedences and by KBO::weightsFromOpts. The comparators take it along the existing template recursion convention, so FreqComparator reads the array rather than the symbol.

It is filled only if needsSymbolCounts(opt) says an option in force wants it, so sp=arity, occurrence, const_max, const_min and reverse_arity, together with a non-frequency kws, now build an ordering without counting anything -- and without the Property scan the old code could trigger. The condition is deliberately a little generous, since counting when nobody asks wastes one pass whereas the converse is a bug: both places that read the counts assert they were demanded. That assertion is the point of the arrangement. unary_first reaches FreqComparator two template levels down, inside
UnaryFirstComparator<false,ArityComparator<false,FreqComparator<>>>, which is exactly the kind of dependency a later edit adds without updating the condition; today that would silently degrade the precedence to occurrence, and now it cannot.

The counts change on problems where the cached Property was stale, which is the point. Measured by keeping both alongside each other temporarily and comparing symbol by symbol over 14 problems x 6 configurations: in all 72 cells the new counts agree exactly with what a forced rescan produces, and in 10 of them they disagree with what the old code actually read. Two independent causes:

 - NLP002-1, 23 symbols, and the disagreement goes away under -updr off: the familiar PredicateDefinition::removeUnusedDefinitionsAndPurePredicates, which mutates the unit list without invalidating;
 - SET010^5, 2 symbols -- $$true and $$false -- and it survives -updr off. TheoryAxioms::applyFOOL adds two clauses mentioning them through addAndOutputTheoryUnit, which pushes onto _prb.units() directly and reports only formulas added, so the cached Property never saw them. Both constants read as unused where the clauses use them twice.

Verified over 13 problems x 17 configurations, comparing the --show_ordering dump: 208 of 221 identical, the 13 differences being exactly NLP002-1 and SET010^5 under the configurations that consult the counts. FMB is untouched (96/96 on the -fmbas grid). The assertion was checked to fire, with its intended message, by temporarily making needsSymbolCounts return false for the frequency-based precedences. ctest 101/101, checks/sanity green against a release build.

Sweep-worthy: sp=frequency is the default, so any problem whose property went stale during preprocessing gets a different precedence than before.


Kernel/SymbolUsage.hpp
```diff
@@ -36,22 +36,36 @@ using namespace Lib;
...
+   * Unlike collectUsedSymbols this must NOT skip a subterm it has already seen: the answer
+   * is a multiplicity, so every occurrence of a shared term has to be counted again, once
+   * for each way the term DAG reaches it. Adding a visited set here would silently change
+   * every count on a problem with any sharing.
+   *
+   * The counting rule is the one Property::scan maintained as a side effect on
+   * Signature::Symbol::usageCnt, so that callers moving off that counter keep their
+   * behaviour.
+   */
+  void countIn(ClauseIterator clauses);
```
MichaelRawson, line 61: For maximum flexibility I would expect a family of functions `countIn(T)` for `T` in `Term *`, `Literal *`, `Clause *`, and then *maybe* `ClauseIterator`. But not a big deal for now, just an idea.

MichaelRawson (approved):

PR #959, opened 2026-09-19
Remove Signature::Symbol::usageCnt

quickbeam123:
Nothing reads it any more. Equality proxy, finite model building, symbol precedence and KBO weight generation each count the clauses they care about, so the counter Property::scan maintained on the side has no consumers left.

Drop _usageCount and its three accessors, and the increments and the reset loop from Property::scan. What the scan still writes to the signature is inGoal and inUnit -- two per-unit booleans, independent of how often a symbol occurs. A scan that skipped subterms it had already visited would set exactly the same two marks, which is what makes a DAG-aware Property::scan worth attempting.


MichaelRawson (approved):
Slam-dunk. Nice work, anything we can move out of `Property`/`Signature` is a win.

PR #960, opened 2026-09-19
Make symbol construction chainable and symbol lookups const

shalashaska117:
Symbol creation now returns `Symbol*`, and marking methods return the same pointer for chaining. Symbols store their signature number, so callers can write `addSkolemFunction(type)->markSkipCongruence()->number()`. Keeps the `add...` names and the specialized Skolem/naming helpers; removes `SymbolBuilder`.

Flag fields, getters, and markers share a macro. Symbol getters are const and return `const Symbol*`; readers and caches use const pointers. Required updates after registration use explicit `Signature` methods. Tests cover registration semantics, flags, stable pointers, stored numbers for all symbol kinds, and const getter types.

Split out of #944. Part of #925.

## Checks
- GCC 15.2 Debug and Release builds with Z3; all 103 Debug CTest suites pass. Copyright headers and whitespace checks pass.
- 985 output comparisons match the merged baseline, with baseline self-comparisons. Covers Discount, Otter, NewCNF, ALASCA, Z3, both clausifiers, and finite model building. Four cases time out on both versions and are excluded.
- `checks/sanity` hits the known `SWC478_1.p` timeout on both versions. A temporary script omitting that check passes on both; repository limits are unchanged.


Parse/TPTP.cpp
```diff
@@ -4104,7 +4105,7 @@ Unit* TPTP::processClaimFormula(Unit* unit, Formula * f, const std::string& nm)
     // only clauses can have free variables at this point!
     ASS_EQ(freeVariables(f),VList::empty())
   }
-  env.signature->getPredicate(pred)->markLabel();
+  symbol.label();
```
mezpusz: This could be chained right after construction.

mezpusz:
Nice change, thanks! Would it be possible to make everything on the `Signature` and `Symbol` interfaces now called through the new construction private? I was also thinking that `addFunction` and the like could simply return the `Symbol*` to make things cleaner.

FMB/SortInference.cpp
```diff
@@ -646,8 +646,7 @@ void SortInference::doInference()
     // we avoid actually creating new symbols until `type` can be computed reasonably,
     // but firstFreshConstant...fresh should be a new contiguous block in the signature
     // we kind of pretend these functions already exist above
-    DEBUG_CODE(unsigned inserted =) env.signature->addFreshFunction(type, "fmbFreshConstant");
-    env.signature->getFunction(f)->markIntroduced();
+    DEBUG_CODE(unsigned inserted =) env.signature->freshFunction(type, "fmbFreshConstant").number();
```
mezpusz: Why not introduced anymore?

Kernel/Signature.hpp
```diff
@@ -440,46 +447,89 @@ class Signature
...
+  class SymbolBuilder {
+    Symbol* _symbol;
+    unsigned _number;
+    friend class Signature;
+    SymbolBuilder(Symbol* symbol, unsigned number) : _symbol(symbol), _number(number) {}
+  public:
+    unsigned number() const { return _number; }
+    Symbol* operator->() const { return _symbol; }
+    Symbol& symbol() const { return *_symbol; }
+    SymbolBuilder& introduced() { _symbol->markIntroduced(); return *this; }
```
mezpusz: I was thinking... most of these could be factored out by some macroing, e.g. introducing the field and a marker/getter, but the `SymbolBuilder` is kind of in the way of this. Is there any reason we use this builder apart from going for the design pattern?

shalashaska117: I thought it was easier to read, I looked at other part of the code and it was done similarly

mezpusz:
As a general comment, would it be possible to leave the function names `addSomething` instead of just `something`? I think this could confuse readers later. Also, removing the special functions for e.g. Skolems seems a bit error-prone to me, I would leave all such functions in too.

How about just changing the `Symbol` such that upon marking it with something it returns itself, and it also stores its number? Then, we could:
1. make macros for the marked fields
2. use the old add functions that return `Symbol*` and make symbol getter functions const and return `const Symbol*`, so there's no funny business after creation (not sure this can be enforced).

Sorry for asking for a lot of changes, but if we get this right and simple once, then we benefit long term.

shalashaska117:
@mezpusz Reworked this in b3d6bb47b:

- Restored the `add...` names and the specialized Skolem/naming helpers.
- Removed `SymbolBuilder`. `Symbol` stores its number, and its marking methods return `this` for chaining.
- Added a macro for the flag fields, getters, and markers, preserving the protection side effects of labels and answer predicates.
- Made the symbol getters const and return `const Symbol*`, and updated readers/caches accordingly. Required later updates, such as parser coloring/protection and polarity flipping, go through explicit `Signature` methods. A retained creation pointer is still mutable, so this doesn't enforce a strict freeze after creation.

On the FMB comment: the symbol is still marked introduced. `addFreshFunction()` already calls `markIntroduced()`, so the removed call was redundant.

Also merged current master and resolved the conflicts. Debug and Release builds pass, along with all 103 unit-test suites and 985 output comparisons against the merged baseline. Four comparison cases time out on both versions. The known `SWC478_1.p` sanity timeout also reproduces on both; the remaining sanity checks pass.

quickbeam123:
There is no such thing as the "known SWC478_1.p sanity timeout". I have told my Claude to remember that `sanity` must be run in Release mode.

shalashaska117:
> There is no such thing as the "known SWC478_1.p sanity timeout". I have told my Claude to remember that `sanity` must be run in Release mode.

Oh i tought that being also on the version on master it was acknowledged, need to recheck my python script for testing, was my error not an AI error

Kernel/NumTraits.hpp
```diff
@@ -321,7 +321,7 @@ struct NumTraits;
     static TermList constantTl(ConstantType const& i) { return TermList(constantT(i)); }  \
     static Option<ConstantType const&> tryNumeral(unsigned functor)                       \
     {                                                                                     \
-      Signature::Symbol* sym = env.signature->getFunction(functor);                       \
+      const Signature::Symbol* sym = env.signature->getFunction(functor);                       \
```
mezpusz: Could you please align this?

Inferences/TermAlgebraReasoning.cpp
```diff
@@ -79,8 +79,8 @@ namespace Inferences {
     if (!lit->isEquality())
       return false;
 
-    Signature::Symbol *s = termAlgebraConstructor(lit->nthArgument(0));
-    Signature::Symbol *t = termAlgebraConstructor(lit->nthArgument(1));
+    const Signature::Symbol*s = termAlgebraConstructor(lit->nthArgument(0));
```
mezpusz: Could you please put a space before or after `*`?

Kernel/Signature.cpp
```diff
@@ -34,6 +34,7 @@ const unsigned Signature::STRING_DISTINCT_GROUP = 0;
 Signature::Symbol::Symbol(const std::string& nm, OperatorType* type, bool interpreted, bool preventQuoting)
   : _name(nm),
     _type(type),
+    _number(UINT_MAX),
```
mezpusz: What prevents us from initializing this in the ctor?

Kernel/Signature.hpp
```diff
@@ -419,46 +377,50 @@ class Signature
...
+  Symbol* addSkolemPredicate(OperatorType* type, const char* suffix = nullptr) {
+    return addFreshPredicate(type, "sK", suffix)->markSkolem();
+  }
+  Symbol* addSkolemTypeCon(unsigned arity) {
+    return addFreshTypeCon(arity, "sK")->markSkolem();
+  }
+  Symbol* addNameFunction(OperatorType* type) { return addFreshFunction(type, "sP"); }
+  Symbol* addNamePredicate(OperatorType* type) { return addFreshPredicate(type, "sP"); }
+
+  // Updates that are needed after registration go through the signature.
```
mezpusz, line 414: Could you please add a TODO to eventually remove these?

Shell/SMTCheck.cpp
```diff
@@ -81,9 +81,9 @@ static std::ostream &operator<<(std::ostream &out, Escaped escaped) {
 }
 
 struct FunctionName {
-  FunctionName(Signature::Symbol *symbol) : symbol(symbol) {}
+  FunctionName(const Signature::Symbol*symbol) : symbol(symbol) {}
```
mezpusz: Add space before/after `*` please.

mezpusz: Same in other places in the file.

UnitTests/tRebalance.cpp
```diff
@@ -81,8 +81,8 @@ void test_rebalance(Literal* lit, initializer_list<expected_t> expected);
         auto xL = Trm<UninterpretedTraits>(TermList::var(0));                                                 \
         auto yL = Trm<UninterpretedTraits>(TermList::var(1));                                                 \
         auto zL = Trm<UninterpretedTraits>(TermList::var(2));                                                 \
-        env.signature->getFunction(nil .functor())->markTermAlgebraCons();                                    \
-        env.signature->getFunction(cons.functor())->markTermAlgebraCons();                                    \
+        env.signature->markTermAlgebraConstructor(nil .functor());                                    \
```
mezpusz: Realign spaces please

mezpusz (approved):
Thanks for the changes, see some minor comments inline.

shalashaska117:
Fixed in 1cd5b05d6:

- Initialize symbol numbers in constructors, keep them const, and assert they match the registration slot.
- Add the TODO to remove post-registration update methods.
- Fix pointer spacing in TermAlgebraReasoning and SMTCheck, and macro alignment in NumTraits and tRebalance.

Merged master in 97478cddd, keeping its deletion of the commented-out FMB code.

Debug/Release builds and all 103 unit tests pass after the merge. Unchanged Release sanity still times out on SWC478_1.p at 0.1s, also reproduced on the saved baseline. Corrected the PR description: calling this a "known timeout" was inaccurate.

PR #962, opened 2026-09-19
options X-macro

MichaelRawson:
`Shell::Options` contains a goodly number of options. However, their details are spread over several locations.

Add an [X-macro](https://en.wikipedia.org/wiki/X_macro) to write a table once and then instantiate it to the kind of activity we need at various call sites.

Somewhat Claude-assisted. It verified its work by checking that the output of `--show_output on --show_experimental on` is the same.


Kernel/MainLoop.cpp
```diff
@@ -112,12 +101,9 @@ MainLoop* MainLoop::createFromOptions(Problem& prb, const Options& opt)
...
   case Options::SaturationAlgorithm::Z3:
-    if(!isComplete || !prb.getProperty()->allNonTheoryClausesGround()){
-      reportSpiderStatus('u');
-      USER_ERROR("Z3 saturation algorithm is only appropriate where preprocessing produces a ground problem");
-      //TODO should return inappropriate result instead of error
-    }
-    res = new SAT::Z3MainLoop(prb,opt);
+    //TODO should return inappropriate result instead of error
+    //TODO so `sa=z3` useless?
+    USER_ERROR("Z3 saturation algorithm is only appropriate where preprocessing produces a ground problem");
```
quickbeam123, line 106: I am in favor of removing `Z3MainLoop` completely, 

(It's use was to simply pass the whole problem (if ground) to Z3 and thus have a chance to sometimes report SAT under theories (iff Z3 would).) 

but was this what you set out to do under the heading "options X-macro"?

easychair:
If the intension is also to remove z3 from the sa option values, then I am
against.

Let users happily run z3 via Vampire.

One motivation is that we have / will have many input and output syntaxes,
so with this option one can run Z3 on inputs Z3 itself does not understand.

Andrei

quickbeam123:
Overall, my gut feeling is a bit against macro-fications as they make code a notch harder to understand.

Happy to discuss more though and look at some of the details together.

The option constraints are anyway left dangling in `init` as the last aspect that's hard to X-ify, right?

Kernel/MainLoop.cpp
MichaelRawson: > but was this what you set out to do under the heading "options X-macro"?

Not exactly! :) But fiddling around exposed some options that were never wired in and therefore had fixed values. Which led me to this - selecting `sa=z3` goes in a straight line to `USER_ERROR` currently, I think.

MichaelRawson:
> Let users happily run z3 via Vampire.
> 
> One motivation is that we have / will have many input and output syntaxes,
so with this option one can run Z3 on inputs Z3 itself does not understand.

Agree! I have actually used `sa=z3` at some point. I wouldn't be heartbroken if it went away, but as some minimal hack it's alright.

> Overall, my gut feeling is a bit against macro-fications as they make code a notch harder to understand.

I think you're right - looking at this with fresh eyes it's not clearly a win for readability/maintainability, so I'm inclined to change my mind and close.

> The option constraints are anyway left dangling in init as the last aspect that's hard to X-ify, right?

Yep. Perhaps these are a good argument against 'X' - I already had to do some macro-spaghetti to conditionally disable option (values) based on `VZ3`, `VDEBUG` and so on.

easychair:
There was a year when Alexander Riazanov left and a new version of GCC
could not compile Vampire anymore. Thanks to his excessive use of spaghetti
macros, it was useless to try repair the code, it had to be rewritten and
what exactly had to be rewritten was impossible to read, interpret and
understand.

I would avoid using macros as plague. They can seriously hurt as compilers
change.

A

MichaelRawson:
Thanks for feedback, I was too excitable. :) In discussion with @quickbeam123 I'll do another version without the macro  but moving more things into `OptionValue`'s constructor to achieve a similar effect.

PR #963, opened 2026-09-21
separate Map and Set equality from their hash parameters

### LINKED ISSUE #947: separate Map equality from its hash parameter
Follow-up to #898 and Michael's [comment on #935](https://github.com/vprover/vampire/pull/935#issuecomment-5699409042).

`Map` currently gets both hashing and equality from its `Hash` parameter, while `DHMap` compares keys with `operator==`. A custom equality operation is therefore hidden inside what looks like a hash choice.

Separate the equality operation from `Map`'s hash parameter, with `operator==` as the default. Make callers that need different equality name it explicitly, and preserve their current hashing and comparison behavior.

This issue covers separating equality from hashing. Bringing the `Map` and `DHMap` interfaces together for a possible merger can follow separately.

shalashaska117:
`Map` and `Set` now take separate equality predicates, defaulting to `std::equal_to<Key>` and `std::equal_to<Val>`. Structural sharing and heterogeneous lookups use explicit predicates, and the Z3 maps use `z3::eq`. `BiMap`, memoization helpers, and `IterTraits::unique` pass equality predicates through.

Removes the unused hash `equals()` methods. Hash calculations are unchanged. Ten new regression tests cover default and custom equality, collisions and resizing, bidirectional lookup, heterogeneous keys, and structural sharing. The test hashes declare `equals()` as deleted to catch accidental use.

Closes #947.

## Checks
- GCC 15.2 Debug and Release builds with Z3; all 102 Debug CTest suites passed. Copyright headers and whitespace checks passed.
- 797 output comparisons matched the previous PR head, with baseline self-comparisons, across Discount, Otter, NewCNF, ALASCA, Z3, and forward ground joinability. `thf-binder-eq.p` under Otter timed out on both versions and was excluded.
- `checks/sanity` times out on `SWC478_1.p` at its 0.1-second limit on both versions. A temporary script omitting that check passes on both; the repository's sanity script is unchanged.


MichaelRawson (changes requested):
Really nice! Thanks. I assume that all the `Hash` classes now have a useless `equals()` member, could you go through and delete those as well? Happy to merge after that, this worked out much nicer than I thought. :-)

shalashaska117:
> Really nice! Thanks. I assume that all the `Hash` classes now have a useless `equals()` member, could you go through and delete those as well? Happy to merge after that, this worked out much nicer than I thought. :-)

It was smaller than I thought the change, it took some hours. I have to check if there aren't any legacy code using that member.

shalashaska117:
@MichaelRawson One question, Set still uses Hash::equals(), would you prefer that I removed the unused equals() in this PR or also give Set a separate equality paramenter so we can finish separating hashing from equality?

MichaelRawson:
Ohh, I forgot that `Set` is not just a `Map` with `Value = EmptyStruct`. `DHSet<K> = DHMap<K, EmptyStruct>`, for instance. :facepalm: Ugh, something else to straighten out in this series.

If you have the energy, yes, please do the required templating for `Set` also. Then we should be able to verify that all the `equals()` methods are unreachable.

shalashaska117:
@MichaelRawson done, let me know what's your idea for merging dhmap and map, and which name it should keep

MichaelRawson (approved):

MichaelRawson:
Nice. Alright, context then a vague plan.

We have `Map`, `DHMap`, and friends (`*Set`, `*MultiSet`) as some kind of historical accident. At some point Vampire used *no* standard containers because it handled its own memory allocation. System allocators have moved on (and memory allocation has become more hazardous) so I've been trying to purge Vampire's allocator and use standard containers where possible. e.g. `std::vector<T>` is a fairly good drop-in replacement for `Stack<T>`, so I'd use `std::vector` in new code. By the way, `DH` stands for "double hashing" if you hadn't already clocked that.

However, it's not as easy as "just use `std::unordered_map`" here. `std::unordered_map` is slow essentially [because the standard says it has to be](https://www.youtube.com/watch?v=fHNmRkzxHWs&t=2702s). We will get `std::flat_map` in C++ '23, but it's quite minimal. I have tried swapping out `Map`/`DHMap` for `std::unordered_map<K, V>` in the past and unfortunately it *is* slower.

--------

So, what's the plan? Well, I think eventually we want one `Map` implementation (called `Map`, ideally). `Set` should be `Map<K, ()>` and `MultiSet` should be `Map<K, unsigned>`, possibly with some extra methods. To get there:

1. Align the interfaces of both `Map` and `DHMap` so that they are identical, with the same semantics. This should allow swapping `DHMap` for `Map` carefully.
2. Very carefully swap `DHMap` for `Map`. This may require improving the underlying implementation for `Map` and/or the primary hash function on a case-by-case basis - it's possible that some poor hashtable performance when using `Map` was fixed by using double hashing. I'd prefer to use a single hash and something a bit smarter from the modern era, like Robin Hood hashing. We can take some inspiration from e.g. swisstable here, but there's probably no need to go crazy, neither container is incredibly well-implemented.
3. Cleanup: `Set`, `Multiset`, and so on. Remove all the old `Hash2` implementations.

MichaelRawson:
Oh, and once this is done we can restore *some* of the functionality of `DefaultHash`. Probably just by defaulting to `std::hash`, like you defaulted to `std::equal_to` here - this is the sane way to do things.

shalashaska117:
> Nice. Alright, context then a vague plan.
> 
> We have `Map`, `DHMap`, and friends (`*Set`, `*MultiSet`) as some kind of historical accident. At some point Vampire used *no* standard containers because it handled its own memory allocation. System allocators have moved on (and memory allocation has become more hazardous) so I've been trying to purge Vampire's allocator and use standard containers where possible. e.g. `std::vector<T>` is a fairly good drop-in replacement for `Stack<T>`, so I'd use `std::vector` in new code. By the way, `DH` stands for "double hashing" if you hadn't already clocked that.
> 
> However, it's not as easy as "just use `std::unordered_map`" here. `std::unordered_map` is slow essentially [because the standard says it has to be](https://www.youtube.com/watch?v=fHNmRkzxHWs&t=2702s). We will get `std::flat_map` in C++ '23, but it's quite minimal. I have tried swapping out `Map`/`DHMap` for `std::unordered_map<K, V>` in the past and unfortunately it *is* slower.
> 
> --------
> 
> So, what's the plan? Well, I think eventually we want one `Map` implementation (called `Map`, ideally). `Set` should be `Map<K, ()>` and `MultiSet` should be `Map<K, unsigned>`, possibly with some extra methods. To get there:
> 
> 1. Align the interfaces of both `Map` and `DHMap` so that they are identical, with the same semantics. This should allow swapping `DHMap` for `Map` carefully.
> 2. Very carefully swap `DHMap` for `Map`. This may require improving the underlying implementation for `Map` and/or the primary hash function on a case-by-case basis - it's possible that some poor hashtable performance when using `Map` was fixed by using double hashing. I'd prefer to use a single hash and something a bit smarter from the modern era, like Robin Hood hashing. We can take some inspiration from e.g. swisstable here, but there's probably no need to go crazy, neither container is incredibly well-implemented.
> 3. Cleanup: `Set`, `Multiset`, and so on. Remove all the old `Hash2` implementations.

I will study everything needed, then I will write an issue for the plan like I did for the hashing

PR #965, opened 2026-09-21
First: don't do expensive things in `Property::scan`, second: do `Property::scan` cheaply on large trees

quickbeam123:
Driven by the evil input `NUM795^4` as target example, Claude came up with some simplifications (in SortHelper, others will benefit from) until we closed this with making Property::scan DAG-aware (repeated-subterms skipping).

`Property::scan` is no longer the bottleneck in `NUM795^4`.

The benefit for everyone:
```
Sort by UNS
10617 ['problemsALL_master11267_otter_i5K.pkl']
10668 ['problemsALL_cheaper11272_otter_i5K.pkl']

Num only-by (UNS):
   problemsALL_master11267_otter_i5K.pkl 0
   problemsALL_cheaper11272_otter_i5K.pkl 51
```

A pure subset, no butterfly-wings.


Shell/Property.cpp
```diff
@@ -563,7 +595,14 @@ void Property::scan(Literal* lit, int polarity)
 
 
     for (int i=0; i<arity; i++) {
-      scanSort(SortHelper::getArgSort(lit, i));
+      TermList arg = *lit->nthArgument(i);
+      if (!scanWillFindItsOwnSort(arg)) {
+        scanSort(SortHelper::getArgSort(lit, i));
```
MichaelRawson, line 594: ~I don't mind particularly, but `getArgSort` will build a whole polymorphic substitution, then apply it to argument `i`. On the next iteration it builds the same substitution, then applies it to argument `i + 1`. There's not really a way around this with the current `SortHelper` API, we could consider something like `Substitution buildSortSubstitutionForTerm(Term *)`.~

Seems you are thinking along the same lines. :)

Shell/Property.cpp
```diff
@@ -263,9 +263,23 @@ void Property::scan(Clause* clause)
...
+    // already been walked by the time a later occurrence comes up (a term's arguments are
+    // pushed above its right sibling, so they drain first). The one thing that is not
+    // idempotent is the count of variable occurrences below the term -- and the term
+    // carries exactly that, since TermSharing maintains numVarOccs as the tree count.
     SubtermIterator stit(literal);
     while (stit.hasNext()) {
-      scan(stit.next());
+      TermList ts = stit.next();
+      if (ts.isTerm() && ts.term()->shared() && !_scannedTerms.insert(ts.term())) {
+        _variablesInThisClause += ts.term()->numVarOccs();
```
MichaelRawson: What if we kept a hashtable of `TermList` rather than `Term *`? Then we would also visit variables only once. Or is `variablesInThisClause` counting occurrences rather than variables?

easychair:
If we have dags, we eventually convert them to clauses which will have a
tree size. Generally, whatever we do in preprocessing is linear in the size
of the input.

Why would we gain anything by complicating linear-time procedures that take
0 time?

AV

MichaelRawson (approved):

quickbeam123:
The issue (manifested greatly by this `NUM795^4`) is that a reasonably-sized tree grows through extensive "function definition elimination" into a reasonably-sized DAG representing an enormous tree. To have a chance of finding a simple refutation (which could still be there), we must not spend long seconds traversing this tree to just learn some statics.

Shell/Property.cpp
quickbeam123: It is counting occurrences.

PR #966, opened 2026-09-21
move more into the initalisation of OptionValues

### LINKED ISSUE #393: A bug with fmbswo was fixed by disabling the option
We used to have
```
./vampire_z3_rel_master_6300 --decode fmb+10_1:1_fmbas=expand:fmbes=smt:fmbswo=diagonal_0 temp.p
Detected maximum model sizes of [1,max,max,max,1,max,max,max,max,max,1,max,max,max,max,max,1,1,1]
TRYING [1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1]
TRYING [1,1,1,1,1,2,1,1,1,1,1,1,1,1,1,1,1,1,1]
TRYING [1,1,1,1,1,3,1,1,1,1,1,1,1,1,1,1,1,1,1]
```
which is clearly wrong for the `temp.p` here
```
cnf(cls_knows__Spy__Notes_1,axiom,
    ( c_Event_Oknows(c_Message_Oagent_OSpy,c_List_Olist_OCons(c_Event_Oevent_ONotes(V_A,V_X),V_evs,tc_Event_Oevent)) = c_Event_Oknows(c_Message_Oagent_OSpy,V_evs)
    | hBOOL(hAPP(hAPP(c_in(tc_Message_Oagent),V_A),c_Event_Obad)) )).
cnf(cls_used__nil__subset_0,axiom,
    ( c_lessequals(c_Event_Oused(c_List_Olist_ONil(tc_Event_Oevent)),c_Event_Oused(V_evs),tc_fun(tc_Message_Omsg,tc_bool)) )).
cnf(cls_not__Cons__self2_0,axiom,
    ( c_List_Olist_OCons(V_x,V_t,T_a) != V_t )).
cnf(cls_neq__Nil__conv_1,axiom,
    ( c_List_Olist_OCons(V_x,V_xa,T_a) != c_List_Olist_ONil(T_a) )).
cnf(cls_subset__trans_0,axiom,
    ( c_lessequals(V_A,V_C,tc_fun(T_a,tc_bool))
    | ~ c_lessequals(V_B,V_C,tc_fun(T_a,tc_bool))
    | ~ c_lessequals(V_A,V_B,tc_fun(T_a,tc_bool)) )).
```
Preliminary investigations lead to suspecting `fmbswo=diagonal` under `fmbas=expand`. Since non-default `fmbswo` does not seem to be essential for us (and does not appear in schedules), a preliminary fix was to disconnect the option until more time can be spent on fixing this properly.

MichaelRawson:
Take 2 on localising option data a bit, now using fancy-pants C++ '20 initialisation. Partially achieved via LLM, but the agent wrote a throwaway script to do it and it seemed to do a relatively systematic job.

Also some fixes I for bugs introduced in the last refactor. I'm not actually sure how, maybe some rebasing error. Or I sneezed or something. Anyway, things happen and it's fixed here.


Shell/Options.cpp
```diff
@@ -461,7 +461,7 @@ void AbstractOptionValue::output(std::ostream &out, bool linewrap) const
     out << "\t";
     int count = 0;
     for (const char *p = description; *p; p++) {
-      out << p;
+      out << *p;
```
MichaelRawson, line 471: Bug 1: failed to actually dereference this pointer. This broke `--show_options on` hilariously.

Shell/Options.cpp
```diff
@@ -4375,13 +4420,16 @@ Stack<const char *> Options::getSimilarOptionNames(const char *name, bool is_sho
...
   auto options = _lookup.values();
+  size_t len = strlen(name);
+  if(!len)
+    return similar_names;
+
   while(options.hasNext()){
     AbstractOptionValue* opt = options.next();
     auto opt_name = is_short ? opt->shortName : opt->longName;
     size_t dif = 2;
-    size_t len = strlen(name);
```
MichaelRawson, line 4382: Don't compute `strlen(name)` over and over in a loop.

Shell/Options.cpp
```diff
@@ -4375,13 +4420,16 @@ Stack<const char *> Options::getSimilarOptionNames(const char *name, bool is_sho
...
     if(!is_short) dif += len/4;
-    if(len!=0 && StringUtils::distance(name,opt_name) < dif)
+    if(opt_name && StringUtils::distance(name,opt_name) < dif)
```
MichaelRawson, line 4079: When `opt_name == nullptr` this crashed. I don't think I broke this so I'm not actually sure how this worked before. 🤷

quickbeam123: Wait, `len` was/is about `name`. But your new check is about `opt_name`.

MichaelRawson: Yep - this is two changes in one diff line.
1. Move the non-zero `len` check out of the loop.
2. Introduce the check for non-null `opt_name`.

mezpusz (approved):

Shell/Options.cpp
mezpusz: Isn't it perhaps because now you have a default `nullptr` for `shortName` and `longName` whereas it was always set as `""` in the ctor before?

MichaelRawson: Possibly - but there was always some logic in `AbstractOptionValue` to interpret `""` as `nullptr` anyway. So I am not completely seeing it, but unless you want a full explanation (which would be reasonable) I'll just say that your explanation/this change makes sense and move on.

